PR #2, opened 2018-11-08
stronger types on route params

TomWoodward:
this depends on #1 merge that one first and then change the base on this.

the guard `navigation.utils.matchForRoute` now narrows the type of `match` properly so that the match.route and match.params have the correct type

### DIFF
--- a/src/app/errors/routes.ts
+++ b/src/app/errors/routes.ts
@@ -2,7 +2,7 @@ import * as components from './components';
 
 const CATCH_ALL = '/(.*)';
 
-export const notFound: Route = {
+export const notFound: Route<undefined> = {
   component: components.PageNotFound,
   getUrl: () => '/404',
   name: 'NotFound',
--- a/src/app/navigation/types.ts
+++ b/src/app/navigation/types.ts
@@ -1,14 +1,28 @@
 import { Location } from 'history';
+import { routes } from '../';
 
 export type State = Location;
 
-export interface Match<Params> {
-  route: Route<Params>;
-  params: Params;
+type RouteParams<R> = R extends Route<infer P> ? P : never;
+type UnionRouteMatches<R> = R extends AnyRoute ? Match<R> : never;
+
+interface MatchWithParams<R extends AnyRoute> {
+  route: R;
+  params: RouteParams<R>;
 }
+interface MatchWithoutParams<R extends AnyRoute> {
+  route: R;
+}
+
+export type GenericMatch = MatchWithParams<AnyRoute> | MatchWithoutParams<AnyRoute>;
+
+export type Match<R extends AnyRoute> = RouteParams<R> extends undefined ? MatchWithoutParams<R> | MatchWithParams<R> : MatchWithParams<R>;
 
 export type historyActions =
   {method: 'push', url: string} |
   {method: 'replace', url: string};
 
 export type reducer = (state: State, action: AnyAction) => State;
+
+export type AnyRoute = typeof routes[number];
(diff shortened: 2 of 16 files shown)


src/app/errors/routes.ts
```diff
@@ -2,7 +2,7 @@ import * as components from './components';
 
 const CATCH_ALL = '/(.*)';
 
-export const notFound: Route = {
+export const notFound: Route<undefined> = {
```
philschatz: nice!

src/app/navigation/types.ts
```diff
@@ -2,9 +2,9 @@ import { Location } from 'history';
 
 export type State = Location;
 
-export interface Match {
-  route: Route;
-  params: any;
+export interface Match<Params> {
+  route: Route<Params>;
+  params: Params;
```
philschatz: Oh, I didn't realize `Params` was a generic type. I was looking to see where `Params` was imported from. I realize we said no `I` in front of interfaces so maybe there's no need for generic types to begin with `T`? But I did notice `P` was used elsewhere in this diff... 


Any reason for choosing one (`Params`) over the other (`P`)?

(I think `P` helps clarify that it's a generic

philschatz: Also, what if the Match was enumerated with the different possible params used in the system? That would help prevent the `Match<any>` in `src/app/navigation/actions.ts`

Something like....

```ts
interface RouteAndParams<P> {
  route: Route<P>;
  params: P;
}

export type Match = RouteAndParams<undefined> | RouteAndParams<Params>
```

TomWoodward: you mean create a union type of the possible params? i can work on that, it seems possible.

TomWoodward: @philschatz i updated so the match typing is a bit more strict. it think its good now.

re the generic template names, i think convention is to use single letter names, but i had started using more verbose ones to make the hinting more useful when querying types from other files. the example you called out `Match<Params>` is kind of moot now because it became `Match<R extends AnyRoute>` which is pretty self explanatory. in general i'm going to try hard to make my generic parameters extend something so that you're limited in what you can pass in and you get a tip for what type of thing it should be. if the names are still too generic in some case i'll add a docblock, my editor at least will show me the docblock when i query the types.

src/app/navigation/types.ts
```diff
@@ -1,14 +1,28 @@
...
 
 export type State = Location;
 
-export interface Match<Params> {
-  route: Route<Params>;
-  params: Params;
+type RouteParams<R> = R extends Route<infer P> ? P : never;
+type UnionRouteMatches<R> = R extends AnyRoute ? Match<R> : never;
+
+interface MatchWithParams<R extends AnyRoute> {
```
philschatz: Ah nice! At first glance I was wondering why anyone would extend `AnyRoute`... but this way `route` and `params` have the *same* `AnyRoute`.

src/app/navigation/types.ts
```diff
@@ -1,14 +1,28 @@
 import { Location } from 'history';
+import { routes } from '../';
 
 export type State = Location;
 
-export interface Match<Params> {
-  route: Route<Params>;
-  params: Params;
+type RouteParams<R> = R extends Route<infer P> ? P : never;
```
philschatz: I'm having trouble parsing this... oh, it looks like it's something like the following (which doesn't work):

```ts
type RouteParams<_ extends Route<P>> = P
```

philschatz (approved):

PR #4, opened 2018-11-08
chore: verify source maps are in the build

philschatz:
This verifies that sourcemaps are included in the build (`yarn build`). 

To verify it works, one can:

- change `GENERATE_SOURCEMAP=true`  in `package.json` to be `GENERATE_SOURCEMAP=false`
- run `yarn test` and check that the exit status is non-zero

It is currently pretty verbose because I noticed that the sourcemap JSON does not actually have the correct relative path. It does however have the source code of the file in the `.map` file so it's probably ok. If it really is ok, then we can remove the `console.warn(...)` statement.

# Example outputs

### Warnings (expected)

```
WARN: In 'build/static/js/1.ba200448.chunk.js.map': Unable to find source file but found sourceContent for '../node_modules/webpack/buildin/harmony-module.js'
WARN: In 'build/static/js/main.398f5adf.chunk.js.map': Unable to find source file but found sourceContent for 'helpers/createStore.ts'
```

### Error (missing sourcemap file)

```
UnhandledPromiseRejectionWarning: Error: BUG: Sourcemap not found! file='build/static/js/1.ba200448.chunk.js'
```

### Error (no JS files found)

```
UnhandledPromiseRejectionWarning: Error: BUG: Could not find js files
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "typescript": "^3.1.6"
   },
   "scripts": {
-    "lint": "tslint --fix -p ./tsconfig.json",
+    "lint": "tslint -p ./tsconfig.json",
     "start": "react-scripts start",
     "build": "GENERATE_SOURCEMAP=true react-scripts build",
     "codecov": "codecov",
--- a/package.json
+++ b/package.json
@@ -15,11 +15,12 @@
     "typescript": "^3.1.6"
   },
   "scripts": {
-    "lint": "tslint -p ./tsconfig.json",
+    "lint": "tslint --fix -p ./tsconfig.json",
     "start": "react-scripts start",
     "build": "GENERATE_SOURCEMAP=true react-scripts build",
     "codecov": "codecov",
     "test": "react-scripts test --coverage",
+    "posttest": "npm run-script build && node ./script/verifySourcemaps.js",
     "eject": "react-scripts eject"
   },
   "eslintConfig": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -15,11 +15,12 @@
     "typescript": "^3.1.6"
   },
   "scripts": {
-    "lint": "tslint -p ./tsconfig.json",
+    "lint": "tslint --fix -p ./tsconfig.json",
```
TomWoodward: i'd rather not have `--fix` turned on for CI, can we make this a different command? or specify --fix manually?

TomWoodward (changes requested):
can we deal with the warnings before merging this? seems like we should either fix the issue or remove them.

package.json
philschatz: ok, fixed the `--fix` 😄

TomWoodward:
@philschatz did you see my comment https://github.com/openstax/books-web/pull/4#pullrequestreview-173374935 ?

philschatz:
oops, I forgot to remove the warnings. Fixed.

Any idea why the sources do not have the correct relative path? I have a feeling that will become a problem when running browser tests (& extracting code coverage)

TomWoodward:
i'm not sure, that seems like something cra should be handling for us

TomWoodward (approved):

TomWoodward (approved):

PR #5, opened 2018-11-08
feat: add dev link to source

philschatz:
This adds a little message to the dev console if someone opens up the developer tools in their browser.

Just a little 🎨 for people that are exploring our code. 

_Fortunately_ this line of code is not tested yet so it does not appear in code coverage.

### Testing

The browser test runs puppeteer and verifies that the console message is printed out.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1,0 +1,129 @@
+import * as http from 'http';
+import * as path from 'path';
+import portfinder from 'portfinder';
+import puppeteer from 'puppeteer';
+import staticServer from 'serve-handler';
+
+class SinglePage {
+    public readonly page: puppeteer.Page;
+    public messages: {
+        warning: string[]
+        log: string[]
+        info: string[]
+        debug: string[]
+        error: string[]
+    };
+    private readonly browser: puppeteer.Browser;
+    constructor(browser: puppeteer.Browser, page: puppeteer.Page) {
+        this.browser = browser;
+        this.page = page;
+        this.messages = {
+            debug: [],
+            error: [],
+            info: [],
+            log: [],
+            warning: [],
+        };
+        // collect browser console messages to be checked later
+        this.page.on('console', (consoleMessage) => {
+            const type = consoleMessage.type();
+            const text = consoleMessage.text();
+
+            switch (type) {
+                case 'debug':
+                case 'error':
+                case 'info':
+                case 'log':
+                case 'warning':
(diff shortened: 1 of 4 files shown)


TomWoodward (approved):

src/index.tsx
```diff
@@ -8,6 +8,11 @@ if (typeof(document) === 'undefined') {
   throw new Error('Browser entrypoint must be used in the browser');
 }
 
+if (window && window.top === window.self) {
```
TomWoodward: should do 

```
typeof(window) !== 'undefined'
```

not that it matters because this file will always be in the browser, but i'd like to get in the habit of doing the checks correctly

TomWoodward (changes requested):

src/index.tsx
philschatz: I wonder, what is the advantage of checking the type of window vs just checking for the existence of it?

TomWoodward: `if (window)` will throw an error if window is not set, which is the case in node. `typeof(window) !== 'undefined'` will not throw an error if window is unset

TomWoodward (approved):

src/index.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+    let devServer: http.Server | null = null;
+    let wrapper: SinglePage | null = null;
+
+    // beforeEach does not seem to work well with async functions
+    const before = async() => {
+        jest.setTimeout(5 * 60 * 1000); // dev browser takes a while to spin up
+
+        // pick a random port
+        // const port1 = (2048 + Math.round(10000 * Math.random()))
```
TomWoodward: we can delete this line right?

src/index.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        const url = `http://localhost:${port}/`;
+        wrapper = await startBrowser(url);
+        return wrapper;
+    };
+
+    afterEach(async() => {
+        if (wrapper) {
+            await wrapper.close();
+            wrapper = null;
```
TomWoodward: this seems probably excessive, we can re-use the dev server and browser right? can we move that stuff to a `beforeAll` and `afterAll`? or if `before` functions don't play nicely with async we can have a helper `getBrowser` or something that re-uses that stuff internally?

TomWoodward (changes requested):

TomWoodward (approved):

PR #10, opened 2018-11-13
chore: add docker config

TomWoodward:
@philschatz i moved your dev server to a new home so i could share it with the docker config. in the future we'll add proxy middleware that can be shared between docker, tests, and dev.

also jq was just straight not working so i replaced with a custom postinstall script

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:10.9-alpine
+
+COPY . /code
+WORKDIR /code
+RUN yarn && yarn build
+
+EXPOSE 8000
+
+CMD yarn server
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,11 @@ FROM node:10.9-alpine
 
 COPY . /code
 WORKDIR /code
-RUN yarn && yarn build
+
+RUN yarn install
+RUN yarn run build
 
 EXPOSE 8000
 
-CMD yarn server
+ENTRYPOINT ["yarn", "run"]
+CMD ["server"]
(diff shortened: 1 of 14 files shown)


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:10.9-alpine
+
+COPY . /code
+WORKDIR /code
+RUN yarn && yarn build
+
+EXPOSE 8000
+
+CMD yarn server
```
philschatz:
```suggestion
CMD yarn run server
```

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:10.9-alpine
+
+COPY . /code
+WORKDIR /code
+RUN yarn && yarn build
```
philschatz: Just to clarify to people not necessarily familiar with yarn's commandline syntax. What do you think?

```suggestion
RUN yarn install
RUN yarn run build
```

philschatz (changes requested):
Yay! It works for me. Should docker instructions be included in the README?

TomWoodward:
@philschatz added readme and rejiggered the structure a little bit.

i made `yarn run` the entrypoint so that in the future we could do `docker run openstax/books-web pre-render`. `server` is still the default command.

i almost added `docker run openstax/books-web test` as an example command but it doesn't work. i'll need to extend the dockerfile a bit to install chrome and stuff for puppeteer to work. i'll work on that tomorrow.

TomWoodward:
ui tests work in the docker container now

philschatz (approved):

PR #12, opened 2018-11-14
rearrange scripts

TomWoodward:
I enjoy running `yarn test --watchAll` while developing, which wasn't working with the combined test commands so i made `yarn test:unit` which i can use for that purpose and reference it from `yarn test`.

running the build from `pretest` was wasted effort from the docker container because the build is known to be up to date, i switched travis to run it explicitly before testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,7 @@ node_js: "10"
 
 cache: yarn
 
+before_script:
+  - yarn build
+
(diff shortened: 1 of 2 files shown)


philschatz (dismissed):

TomWoodward:
@philschatz i touched this so it needs review again

philschatz (approved):

PR #13, opened 2018-11-14
combine browser and unit tests

TomWoodward:
@philschatz this is just a suggestion for how to combine the browser and unit tests into one suite. we move the main jest config into the package.json and reference the bits of the react-scripts config as necessary, then we just tag browser test files with `@jest-environment puppeteer`

does this seem like a good thing to you?


TomWoodward:
i'm not sure what we're losing by bypassing the `react-scripts test` command, we can reference things like the `babelTransform` directly and still benefit from the configs that are there. but there is a lot of code in that `test` script that we'll no longer be using.

the tests pass though and i like being able to have the suites combined so /shrug? maybe we can re-visit if something becomes a problem

PR #15, opened 2018-11-15
chore: Screenshot tests

TomWoodward:


src/test/utils.ts
```diff
@@ -15,4 +15,11 @@ export {
 
 const DEV_SERVER_PORT = 8000;
 
-export const url = (path: string) => `http://localhost:${DEV_SERVER_PORT}/${path}`;
+export const url = (path: string) => `http://localhost:${DEV_SERVER_PORT}/${path.replace(/^\/+/, '')}`;
+
+export const navigate = async(target: puppeteer.Page, path: string) => {
+  await target.goto(url(path));
+  await target.evaluate(async() => {
+    await (window as any).hooks.calm();
```
philschatz: nice! Since we redefine the `window` in `global.d.ts` to optionally be undefined, could it be extended to include the `.hooks`? (or maybe something easier to grep for & less likely to collide like `_RENDERING_HOOKS`)

src/index.tsx
```diff
@@ -18,6 +18,9 @@ if (typeof(window) !== 'undefined' && window.top === window.self) {
 
 const app = createApp();
 
+// bind this to the window so profiling tools can access it
+(window as any).hooks = app.hooks;
```
philschatz: nice!

philschatz (dismissed):

philschatz:
Yay! Running the tests locally worked for me (the macos screenshots match) but it seems the Travis screenshot is still different.

src/test/utils.ts
philschatz: something like what is in commit 72173a85ae90124016470bf9189aed37316133ec

src/app/content/hooks/locationChange.ts
```diff
@@ -1,22 +1,33 @@
...
+const fonts = fontMatches ? fontMatches.slice(1) : [];
+
+export default routeHook(content, ({dispatch, getState, fontCollector}) => async({match}) => {
   const state = getState();
   const {bookId, pageId} = match.params;
   const book = select.book(state);
   const page = select.page(state);
+  const promises: Array<Promise<any>> = [];
+
+  fonts.forEach((font) => fontCollector.add(font));
```
TomWoodward: invented font collector. dual purpose:
- pre-render script will hoist these fonts into the head so we aren't relying on styled-components `@include` polyfill to fetch them from google.
- can hook into when the fonts are loaded

src/app/content/hooks/locationChange.ts
```diff
@@ -1,22 +1,33 @@
...
 
   if ((!page || page.id !== pageId) && pageId !== select.loadingPage(state)) {
     dispatch(requestPage(pageId));
-    archiveLoader(`${bookId}:${pageId}`).then((pageData) => dispatch(receivePage(pageData)));
+    promises.push(archiveLoader(`${bookId}:${pageId}`)
+      .then((pageData) => dispatch(receivePage(pageData)))
+    );
   }
+
+  await Promise.all(promises);
```
TomWoodward: this needs to return a master promise to make hooks work (otherwise async function promise resolves before data fetching promises)

src/app/index.tsx
```diff
@@ -39,9 +41,14 @@ export default (options: Options = {}) => {
     navigation: navigation.createReducer(history.location),
   });
 
+  const services = {
```
TomWoodward: so far these are always the same and are created here, in the future we may want to add some to the options passed into this function which creates a dependency injection pattern to be used between browser and pre-renderer

src/test/matchers/toMatchImageSnapshot.ts
```diff
@@ -0,0 +1,13 @@
+import { MatchImageSnapshotOptions, toMatchImageSnapshot } from 'jest-image-snapshot';
+
+type ToMatchImageSnapshot = (image: Buffer, config: MatchImageSnapshotOptions) => {pass: boolean, message(): string};
+
+export default function(this: jest.MatcherUtils, image: Buffer, config: {[key: string]: MatchImageSnapshotOptions}) {
+  const options = process.env.CI
+    ? config.CI
+    : config.DEV;
```
TomWoodward: here i'm allowing different image comparison configs in dev vs ci. so far i'm considering running tests locally to be the source of truth and i'm passing looser configs that allow the tests to pass in jenkins CI

src/app/content/hooks/locationChange.ts
```diff
@@ -1,22 +1,33 @@
...
 import { receiveBook, receivePage, requestBook, requestPage } from '../actions';
 import { content } from '../routes';
 import * as select from '../selectors';
 import { archiveLoader } from '../utils';
 
-export default routeHook(content, ({dispatch, getState}) => async({match}) => {
+const fontMatches = css.match(/"(https:\/\/fonts\.googleapis\.com\/css\?family=.*?)"/);
+const fonts = fontMatches ? fontMatches.slice(1) : [];
+
+export default routeHook(content, ({dispatch, getState, fontCollector}) => async({match}) => {
```
TomWoodward: i'm going to make a separate change that makes these hook files export just the hook implementation and move the call to the `routeHook` creator into a more centralized place. it doesn't really matter but it will make the tests more digestible to not maintain knowledge of what `routeHook` is doing

src/app/content/components/Header.tsx
```diff
@@ -0,0 +1,7 @@
+import styled from 'styled-components';
+
+export default styled.h1`
+  font-family: "Roboto Condensed",serif;
+  font-size: 19px;
+  line-height: 29px;
```
TomWoodward: should the h1 actually be styled by the cnx-recipes styles? i guess it depends on the page design.

src/test/browserutils.ts
```diff
@@ -0,0 +1,60 @@
...
+  await target.evaluate(async() => {
+    if (window) {
+      await window.__APP_ASYNC_HOOKS.calm();
+    }
+  });
+};
+
+export const finishRender = async(_: puppeteer.Page) => {
+  // HACK - there is no convenient way to tell if chrome is finished rendering,
+  // we should investigate inconvenient possibilities.
```
TomWoodward: @philschatz the problem is:

i've invented ways to be sure that the relevant data and network requests for fonts have been loaded, but i can't tell when chrome is finished drawing the page. this is irrelevant for some browser tests because the interactions will already work, but for screenshots there are problems like:

- [image](https://jenkins-dev.cnx.org:8080/job/openstax-continuous-integration/job/rex-web/job/PR-15/49/artifact/content-browserspec-ts-content-looks-right-1-diff.png) the page just isn't drawn at the bottom, adding the 1s delay fixes this issue so i'm assuming its just draw time.
- fonts rendering is different from the network request being done, there is no great way to tell when fonts are done loading. `document.fonts.ready` is a thing, but doesn't seem reliable yet (very experimental).

i don't have any really good ideas, maybe keep taking screenshots every 50ms or so and see when they stop changing? i feel like that might give false positives if the rendering happens in chunks. do you have any thoughts? 

we can merge this as is and circle back if necessary, it seems to be passing find on jenkins

src/helpers/FontCollector.ts
```diff
@@ -0,0 +1,20 @@
+type Observer = (font: string) => void;
+
+export default class FontCollector {
+  private _fonts: string[] = [];
```
philschatz:
```suggestion
  public readonly fonts: string[] = [];
```

That would remove the getter and having to name the variable `_`

src/helpers/PromiseCollector.ts
```diff
@@ -0,0 +1,25 @@
+export default class PromiseCollector {
+  private _promises: Array<Promise<any>> = [];
```
philschatz:
```suggestion
  public final promises: Array<Promise<any>> = [];
```

Similarly, it seems like this would obviate the need for a helper. However, I _do_ like that the current method exposes a line that can be covered in codecov (to see if the public field is ever read)

philschatz (commented):
Just a question about getters. Otherwise it looks great!

src/helpers/PromiseCollector.ts
TomWoodward: i did the `get` helper to convert the type to `ReadonlyArray` otherwise the readonly var just means you can't overwrite it but you can still call like `array.push` or whatever to modify it.

philschatz: oooh, interesting. okies. I'm still not used to using getters in JS.

TomWoodward: it would be nice if typescript had a way to designate the type as like `PrivateArray` or something so i didn't have to jigger the type to be different internally vs externally using that getter. i wonder if they have that...

philschatz (approved):

PR #18, opened 2018-11-20
chore: use port specified in .env file

philschatz:
I frequently run different services and `:8000` seems to be a popular port number.

This moves the port number into the `.env.test` file.


.env.production
```diff
@@ -1 +1,2 @@
 REACT_APP_ARCHIVE_URL=https://archive.cnx.org/contents/
+SERVER_PORT=8000
```
TomWoodward: we probably don't even need to specify this here

jest-puppeteer.config.js
```diff
@@ -7,8 +18,8 @@ module.exports = {
     ],
   },
   server: {
+    command: `yarn run server ${SERVER_PORT}`, // This takes longer & times out: `PORT=${SERVER_PORT} BROWSER=none yarn start`
```
TomWoodward: starting the dev server does take longer, but i feel like its worth it to not have to worry about building before testing. the suite passes for me and jenkins and travis, what issue were you having?

jest-puppeteer.config.js
```diff
@@ -1,3 +1,14 @@
+const path = require('path');
+const dotenv = require('dotenv');
+
+dotenv.config({path: path.join(__dirname, '.env.development')});
```
TomWoodward: in `development-proxy` i added a script that loads the environment files in the same way as react-scripts so this will work in other environments like `test`

server/index.js
```diff
@@ -1,12 +1,12 @@
...
 const staticServer = require('serve-handler');
+require('dotenv').config();
 
 module.exports = (options = {}) => new Promise(resolve => {
   const fallback404 = !!options.fallback404;
 
-  portfinder.getPortPromise().then(startServer);
+  const port = process.env.SERVER_PORT
 
   function startServer(port) {
```
TomWoodward: if we're passing in port theres no reason for any of this function to exist

src/test/browserutils.ts
```diff
@@ -22,9 +23,7 @@ page.on('console', (consoleMessage) => {
   }
 });
 
-const DEV_SERVER_PORT = 8000;
-
-export const url = (path: string) => `http://localhost:${DEV_SERVER_PORT}/${path.replace(/^\/+/, '')}`;
+export const url = (path: string) => `http://localhost:${puppeteerConfig.server.port}/${path.replace(/^\/+/, '')}`;
```
TomWoodward: neato

TomWoodward (changes requested):

jest-puppeteer.config.js
philschatz: Yep, I would also rather use the dev server at this point... when I use the dev server though, The 2 browser tests fail, not from timeouts but from being blank (could still be because of a HTTP request for the JS file though):

- `Browser sanity tests › displays the "Hello developer" console text`
- `Browser sanity tests › loads the 404 page`

Maybe I can add a line after `await page.goto(...)` that checks to make sure the JS file loaded

.env.production
philschatz: oops, yep. it's not needed.

jest-puppeteer.config.js
philschatz: oh, Jenkins fails in the same way as my local machine: https://jenkins-dev.cnx.org:8080/blue/organizations/jenkins/openstax-continuous-integration%2Frex-web/detail/PR-18/1/pipeline/#step-20-log-155

package.json
```diff
@@ -58,7 +58,6 @@
     "jest-image-snapshot": "^2.7.0",
     "jest-puppeteer": "^3.5.1",
     "pify": "^4.0.1",
-    "portfinder": "^1.0.19",
```
TomWoodward: please leave this in i'm using it on the pre-rendering branch

server/index.js
```diff
@@ -2,31 +2,28 @@ const connect = require('connect');
 const http = require('http');
 const serveStatic = require('serve-static')
 const path = require('path');
-const portfinder = require('portfinder');
 const setupProxy = require('../src/setupProxy');
+require('dotenv').config();
```
TomWoodward: should use `require('./src/env');` here as well

TomWoodward (changes requested):
needs master merged, but looks generally 👍

package.json
philschatz: oh, why not use the value in `.env.test`?

server/index.js
philschatz: Updated!

package.json
TomWoodward: yea pre-rendering doesn't need to be interacted with i'm just starting the server to use the proxy config so i figured using a made up port would be preferable there.

TomWoodward (approved):

PR #19, opened 2018-11-21
feat: add Skip to Content link

philschatz:
Fixes https://github.com/openstax/unified-issues/issues/82

which is hidden until it is tabbed to. When the link is clicked it scrolls down to the beginning of the content.

skip-to-content

# Notes

- ~~Travis is failing because the browser tests are not starting up. Hopefully #18 will help with that~~
- Is it 🆗 if the URL contains a hash in it when someone clicks the link? It's probably fine since the user is actively moving to another part of the page but I thought it is worth mentioning
  - **Decision:** No, JS should handle the scrolling and focus when JS is enabled.

# TODO

- [x] verify that the "Skip to Content" link is the first element that is tabbed to
- [x] verify that the link worked and the focus changed


TomWoodward:
github doesn't add the hash to the url, do they just scroll with js?

src/app/content/components/SkipToContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  private isLoading() {
+    return this.props.loading;
+  }
+}
+
+export default connect(
+  (state: AppState) => ({
+    loading: !!select.loadingBook(state) || !!select.loadingPage(state),
```
TomWoodward: you don't need to connect this component, `src/app/content/components/Content.tsx` already short circuits if the content is loading

src/app/content/components/SkipToContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+interface PropTypes {
+  targetId: string;
+  loading: boolean;
+}
+
+class SkipToContent extends Component<PropTypes, never> {
+
+  public render() {
+    const { targetId } = this.props;
+    return !this.isLoading() && <HiddenLink href={`#${targetId}`} tabIndex={1}>Skip to Content</HiddenLink>;
```
TomWoodward: if we scroll with js instead, can we pass a ref to the target element into `SkipToContent` instead of relying on the id and hash?

TomWoodward (changes requested):

src/app/content/components/SkipToContent.tsx
philschatz: That would be cleaner... I can try although I'm not sure how to pass a ref to the target element since it is constructed in a child component(my React-fu is fuzzy).

Also, I think we do need to support the href & `id="..."` as fallback if JS did not load

TomWoodward: do we support not js not being loaded? that will affect many other things

TomWoodward: i think the reacty way to do it would be to have a `<SkipToContentWrapper>` component that rendered the link before its children, and a `<SkipToContentTarget>` HOC that wrapped your actual target, then use [context](https://reactjs.org/docs/context.html) to communicate between the two and use [find dom node](https://reactjs.org/docs/react-dom.html#finddomnode) on `SkipToContentTarget`'s child to get the dom reference to scroll too.

i know it sounds more complex, but it'll enable using the SkipToContent pattern on more pages with less id wrangling

philschatz: One concern with that is that the SkipToContent link needs to appear at the beginning of the DOM (before the navigation, the ToC, etc) while the actual content element lives after the navigation, the ToC. Would the pattern you mentioned still work? If so, it seems like the `<SkipToContentWrapper>` would need to surround the entire app, right?


RE Not having JS enabled: What else would break? (I am assuming that only non-logged-in users would not have JS enabled)

TomWoodward: if we do want it to work without js, `<SkipToContentTarget>` could inject a anchor with a randomly generated id and communicate that back to `<SkipToContentWrapper>` via context on `componentWillMount`. all that will get baked into the pre-rendered html so then the link will work without js, but i still think we should have a js handler so we don't get the url change when we can avoid it

philschatz: I added a JS handler but was wondering if it is worth adding the `react-scroll` package or just using the part of it that we need (seen in b277764) befoe I try writing tests.

src/app/components/SkipToContentWrapper.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return <Provider value={{registerMainContent: this.registerMainContent}}>
+      <HiddenLink onClick={this.scrollToTarget} href={`#${MAIN_CONTENT_ID}`} tabIndex={1}>Skip to Content</HiddenLink>
+      {this.props.children}
+    </Provider>;
+  }
+
+  private scrollToTarget = (event: React.MouseEvent<HTMLAnchorElement>) => {
+    if (this.mainContent) {
+      event.preventDefault();
+      scrollTo(this.mainContent);
```
TomWoodward: this doesn't seem to match src/helpers/scrollTo.ts

TomWoodward (changes requested):

src/app/components/SkipToContentWrapper.tsx
TomWoodward: oh never mind the second and third params are optional. do we really need those to be there?

src/app/components/SkipToContentWrapper.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+    }
+    const node = ReactDOM.findDOMNode(c)
+    if (!node) {
+        throw new Error(`BUG: Could not find DOM node`)
+    }
+    return {component: c, node};
+}
+
+describe('SkipToContentWrapper', () => {
+    it('errors when no main content is provided', () => {
```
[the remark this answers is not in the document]
philschatz: I am worried that hiding the link will silently cause us to stop adding the SkipToContent link.

src/app/components/SkipToContentWrapper.tsx
philschatz: I removed them. Mostly because I wanted to keep code coverage @ 💯

src/app/components/SkipToContentWrapper.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return {component: c, node};
+}
+
+describe('SkipToContentWrapper', () => {
+
+    it('errors when no main content is provided', () => {
+        const {node} = renderToDom(<SkipToContentWrapper/>);
+        const breaks = () => ReactTestUtils.Simulate.click(node);
+        expect(breaks).toThrowErrorMatchingSnapshot();
```
philschatz: I still get a bunch of red text when a test expects an `Error` to be raised. I noticed another test had suppressed the text... is that really the only way?

TomWoodward: i didn't realize `.toThrowErrorMatchingSnapshot();` was a thing, but if it shows the error in the output i'd rather not use it... i like the clean output

src/app/components/SkipToContentWrapper.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+            const spyScroll = jest.spyOn(window, 'scrollTo');
+            const spyFocus = jest.spyOn(mainContent, 'focus');
+
+            ReactTestUtils.Simulate.click(node);
+
+            expect(spyScroll).toHaveBeenCalledTimes(1);
+            expect(spyScroll).toHaveBeenCalledWith(0, 0); // the vertical offset of the element
+
+            expect(spyFocus).toHaveBeenCalledTimes(1);
+            expect(spyFocus).toHaveBeenCalled();
```
TomWoodward: do you really need both of these spyFocus assertions?

TomWoodward (changes requested):

src/app/components/SkipToContentWrapper.spec.tsx
philschatz: oops, removed 😄

src/app/components/SkipToContentWrapper.spec.tsx
philschatz: The cause of the console.error messages is from JSDOM trying to be helpful. I suppressed them by adding a `expectError(msg, fn)` which mocks `console.error` when the function runs and then resets `console.error` back.

src/app/components/SkipToContentWrapper.spec.tsx
```diff
@@ -18,17 +18,41 @@ function renderToDom<P>(component: ReactElement<P>) {
     return {component: c, node};
 }
 
+// JSDom logs to console.error when an Error is thrown.
+// Disable the console just in this instance, and re-enable after.
+// See https://github.com/facebook/jest/pull/5267#issuecomment-356605468
+function expectError(message: string, fn: () => void) {
```
TomWoodward: can we move this to `src/test/reactutils.ts` or maybe invent a `src/test/utils.ts`?

TomWoodward (changes requested):

TomWoodward (approved):

PR #20, opened 2018-11-21
chore: create data fixtures for archive to use in test and ci

TomWoodward:


src/test/fixtures/archive/contents/testbook1-shortid
```diff
@@ -0,0 +1,128 @@
+{
```
philschatz: Are these manually updated? If so, would it make sense to use something like https://github.com/vcr/vcr to record the fixtures?

Or drop in a `fetch` implementation that records/plays-back the fixtures? like https://www.npmjs.com/package/fetch-vcr

philschatz: Ok, if it is fake data then could it be lorem-ipsum-style ... whatever is the minimum used by the tests? That would help make it clear whether a particular combination of elements is being tested rather than the content just being in the original copy/paste

TomWoodward: is that not what it is?

philschatz: Do we use (or plan to use) fields like `submitlog`, `legacy_id`, `stateid`, `parentId` (bingo!, got all 3 ways to combine word+id)?

and others like `parentVersion`, `legacy_version`, `history`, `buyLink`, `doctype`, `licensors`, `publishers`, `authors`?

jest-puppeteer.config.js
```diff
@@ -8,7 +8,9 @@ module.exports = {
   },
   server: {
     launchTimeout: 60000,
-    command: 'PORT=8000 BROWSER=none yarn start',
+    // react-scripts start unconditionally sets the NODE_ENV to development,
+    // so we're setting CI here even if that isn't necessarily true
+    command: 'CI=true PORT=8000 BROWSER=none yarn start',
```
philschatz: oh, `NODE_ENV=production PORT=8000 BROWSER=none yarn start` does not work?

src/setupProxy.js
```diff
@@ -11,12 +13,23 @@ const archivePaths = [
   'specials',
 ];
 
-const ARCHIVE_URL = process.env.ARCHIVE_URL;
 
 module.exports = function(app) {
+  process.env.NODE_ENV === 'test' || process.env.CI
```
philschatz: hmm, I'm only familiar with `production` and `development` as being valid values for `NODE_ENV` ... is `test` another commonly-used value?

philschatz (commented):
Thanks for updating the mocks. I noticed 1 other thing before merging

src/setupProxy.js
TomWoodward: its what create-react-app sets it to in testing https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/test.js#L13

jest-puppeteer.config.js
TomWoodward: i was trying to set it to `test` but cra overrides it on boot with `yarn start` so CI is just a way to tell you're in a testing mode so that it'll use the fixtures

philschatz:
Drat. is there a link to the failed travis build? I clicked around on the commits but on most of them travis passed.

TomWoodward:
@philschatz starting with https://github.com/openstax/rex-web/pull/20/commits/6cb8ffd7533f78fda5f0756c101b814455ebcdaf and the few commits after it. i thought it was just unstable at first and re-ran that commit several times and it consistently failed

PR #23, opened 2018-11-26
feat: implement a way to specify meta tag values

TomWoodward:


philschatz (approved):
Looks good! From reading it seems like the meta tags are visible only during prerender, not when running `yarn start`... is that correct?

I just added one small change (unused checks). If that looks fine then merge away

TomWoodward:
@philschatz yea the meta tags aren't actually rendered into the head when running live on webpack-dev-server, you can see the values in the redux state though.

its the pre-render script that takes the values out of the state and adds them to the html

PR #25, opened 2018-11-28
feat: Prerender

TomWoodward:


jest-puppeteer.config.js
```diff
@@ -20,7 +20,9 @@ module.exports = {
     launchTimeout: 60000,
     // react-scripts start unconditionally sets the NODE_ENV to development,
     // so we're setting CI here even if that isn't necessarily true
-    command: `CI=true PORT=${SERVER_PORT} BROWSER=none yarn run start`,
+    command: process.env.SERVER_MODE === 'built'
```
philschatz: ❤️

script/entry.js
```diff
@@ -0,0 +1,39 @@
...
+  .map(extension => fs.existsSync(path.resolve(__dirname, scriptPath + extension)))
+  .filter(identity)
+  .length > 0;
+
+if (!exists) {
+  console.error(`script "${script}" does not exist`);
+  process.exit(1);
+}
+
+require('@babel/register')({
```
philschatz: Oh nice! So now all scripts can be TypeScript!

src/app/content/components/Content.tsx
```diff
@@ -17,25 +17,39 @@ interface PropTypes {
...
+  }
+
   public loadBook(props: PropTypes) {
-    if (!props.book) {
+    const {book, services} = props;
+    if (!book) {
       return;
     }
-    this.props.services.archiveLoader(props.book.shortId).then((book) => this.setState({book}));
+    this.setState({book: services.archiveLoader.cachedBook(book.shortId)});
```
philschatz: Why does the `cachedBook` method now need to be called rather than just the `archiveLoader` function?

src/helpers/createArchiveLoader.ts
```diff
@@ -1,15 +1,28 @@
...
+    })
+    .then((response) => response.json() as Promise<ArchiveContent>)
+    .then((response) => {
+      cache.set(id, response);
+      return response;
+    })
+  );
+
+  return {
+    book: (bookId: string) => loader(bookId) as Promise<ArchiveBook>,
```
philschatz: Could this be `=> cache.get(bookId) || loader(bookId)` ? and therefore avoid the need for an explicit `cachedBook`?

src/app/content/components/Content.tsx
TomWoodward: this is done in the react component to avoid extraneous promises that mess with the server side render

src/helpers/createArchiveLoader.ts
TomWoodward: no the loader returns promises and the cache returns json

philschatz (approved):

PR #27, opened 2018-11-29
feat: Jenkins build

TomWoodward:
this obviously doesn't actually work without the changes from the prerender branch

proof: https://thu1.sandbox.openstax.org/books/Ax2o07Ul/pages/05gOEZHv


philschatz:
Did jenkins not run because there is no Jenkins file on `master`? I was expecting to see that it ran

philschatz (approved):

PR #28, opened 2018-11-30
feat: basic toc functionality

TomWoodward:
i still have some more functionality from the issue to work on, but this covers the basics.


src/app/content/components/Sidebar.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  ${(props) => !props.open && css`
+    span {
+      margin-top: 15%;
+    }
+  `}
+
+  ${(props) => props.open && css`
+    span:nth-child(1) {
+      transform: rotateZ(45deg);
```
philschatz: Ooh, nifty transition!

src/app/content/components/Sidebar.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  public render() {
+    const {open, book, closeToc, openToc} = this.props;
+    return <SidebarPlaceholder open={open}>
+      <SidebarBody open={open}>
+        <SidebarControl
+          open={open}
+          onClick={() => open ? closeToc() : openToc()}
+          role='button'
+          aria-label={`Click to ${open ? 'close' : 'open'} the Table of Contents`}
```
philschatz: Hmm, maybe it's time to open an i18n PR...

src/app/content/guards.ts
```diff
@@ -0,0 +1,4 @@
+import { ArchiveTree, ArchiveTreeSection } from './types';
+
```
philschatz: Ooh, good idea. One place for all the type-guards

philschatz (dismissed):

src/app/content/components/Sidebar.tsx
TomWoodward: yea i would prefer to do that before we dig our hole much bigger

philschatz (approved):

PR #29, opened 2018-12-03
style: lint indent

TomWoodward:


tslint.json
```diff
@@ -6,11 +6,11 @@
   "jsRules": {},
   "rules": {
     "interface-name": [true, "never-prefix"],
-    "max-line-length": [true, 200],
     "prefer-object-spread": true,
     "quotemark": [true, "single"],
     "space-before-function-paren": [true, "never"],
     "variable-name": [true, "check-format", "allow-leading-underscore"],
+    "indent": [true, "spaces", 2],
```
philschatz: Hmm, time for me to update my editor to use 2 space-indents. Or... maybe this is a good point to do a bit of bike-shedding.... I used to use 2 spaces but found that with horizontally-wide screens (& python conventions) that 4 spaces was easier to see scoping & it discourages nesting things too deeply. Thoughts?

TomWoodward: with our line length restrictions the size of the screen doesn't really matter, and even a couple levels of nesting would be like 30% of your available line length. 

i've found that with larger indents you spent more time coming up with creative ways to break code onto multiple lines that doesn't look like 🗑 .

i prefer two spaces but in general i use whatever the convention for the language is, in php i use 4 spaces, in js two seems to be standard.

philschatz (approved):

PR #32, opened 2018-12-03
perf: Navigation ids

TomWoodward:
when navigating only the content short ids are available in the url, this adds additional metadata to `location.state` so that we can avoid fetching it repeatedly.


src/app/content/components/Content.spec.tsx
```diff
@@ -77,14 +78,11 @@ describe('content', () => {
       </Services.Provider>
     </Provider>);
 
-    process.nextTick(() => {
-      const tree = component.toJSON();
-      expect(tree).toMatchSnapshot();
-      done();
-    });
+    const tree = component.toJSON();
```
philschatz: Yay, no more `nextTick`

script/prerender.tsx
```diff
@@ -47,22 +51,31 @@ async function render() {
     }, null, 2));
   }
 
-  // book version is currently ignored, and will be until we rejigger the book content
-  // routing to preserve long id and version data on navigation
-  async function renderContentPage(bookId: string, _bookVersion: string, pageId: string) {
-    const book = await archiveLoader.book(bookId);
-    const page = await archiveLoader.page(bookId, pageId);
+  async function renderContentPage(bookId: string, bookVersion: string, pageId: string) {
+    const book = await archiveLoader.book(`${bookId}@${bookVersion}`);
```
philschatz: Why concatenate the string here rather than in `archiveLoader` ? It seems archiveLoader is already doing string concatenation for page-in-a-book.

By stripping the version out in multiple places (& keeping the same type for `yqV9q0HH` and `yqV9q0HH@2`) it is difficult to see when a version is being used/needed and when it is not.

src/app/content/components/Content.tsx
```diff
@@ -14,56 +14,10 @@ import Wrapper from './Wrapper';
 interface PropTypes {
   page: State['page'];
   book: State['book'];
-  loading: boolean;
```
philschatz: Why remove the loading state? I think it would be good to have so that Selenium tests have a way to wait until the new page has loaded. Also, people on slower connections might wonder if clicking did anything

src/app/content/hooks/locationChange.spec.ts
```diff
@@ -102,7 +102,7 @@ describe('locationChange', () => {
 
   it('doesn\'t load page if its already loaded', () => {
     localState.page = {
-      id: 'pageId',
+      shortId: 'pageId',
```
philschatz: Since we use many identifiers in the system (uuid, shortId, uuid+ver, shortId+ver, slug) and they all have the same type (`string`) would it make sense to rename the field to help show which identifier is being used?

src/app/navigation/types.ts
```diff
@@ -6,32 +6,44 @@ import { AnyAction } from '../types';
...
 
-export type historyActions =
-  {method: 'push', url: string} |
-  {method: 'replace', url: string};
+export type AnyHistoryAction = UnionHistoryActions<AnyRoute>;
 
 export type reducer = (state: State, action: AnyAction) => State;
 
-export interface Route<Params> {
+// @ts-ignore: 'S' is declared but its value is never read.
```
philschatz: Why is `S` never used?

philschatz (changes requested):
Since we use many identifiers in the system (uuid, shortId, uuid+ver, shortId+ver, slug) and they all have the same type (`string`) it might help to rename the fields to help show which identifier is being used.

I saw that this is being done with `shortId` but not others.

src/app/content/components/Content.tsx
TomWoodward: there is still a loading state in the redux state, this just wasn't doing anything here anymore so i removed it. when we have design for a loading state we can decide where to put it

src/app/navigation/types.ts
TomWoodward: the shape of `state` isn't used in defining the members of the route object, but it is a defining characteristic of a route, its used in other types to make sure the state you pass with the route matches the state the route was defined with

TomWoodward: eg: https://github.com/openstax/rex-web/pull/32/files#diff-558360260e96a66c6b633c30d170dd6bR21

src/app/content/hooks/locationChange.spec.ts
TomWoodward: the trouble with shortId is that in the full payload it is only the id, but in the tree it is shortId + version. currently the only place i care about this is in ContentLink but i could probably push it down to ArchiveLoader if you want

script/prerender.tsx
TomWoodward: i'll make a change so that the book version gets passed into archiveloader explicitly, i think that'll help

philschatz (dismissed):
Thanks! This has a merge conflict but otherwise, its ready to merge

philschatz (approved):

PR #33, opened 2018-12-04
chore: add tsc to test command

TomWoodward:
until now we've been catching type errors as a function of the build commands run as part of the tests, but some type error have leaked through in the tests themselves and in the scripts which aren't compiled with the app code.

adding tsc to the test command before was annoying because of this bug https://github.com/Microsoft/TypeScript/issues/28202

so i've upgraded typescript to the new release with that bug fixed but now there is a different bug
https://github.com/Microsoft/TypeScript/issues/28748

but that one is easier to ignore and document so i've pressed on anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "reselect": "^4.0.0",
     "styled-components": "^4.0.3",
     "typesafe-actions": "^2.0.4",
-    "typescript": "^3.1.6"
+    "typescript": "^3.2"
   },
   "scripts": {
     "lint": "tslint -p ./tsconfig.json",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "server": "NODE_ENV=development node script/entry server/cli",
     "codecov": "codecov",
     "postinstall": "node ./script/postinstall",
-    "test": "npm run-script test:unit && npm run-script lint && npm run-script test:build",
+    "test": "npm run-script test:unit && npm run-script lint && npm run-script test:build && ./node_modules/.bin/tsc",
```
philschatz:
```suggestion
    "test": "npm run-script test:unit && npm run-script lint && npm run-script test:build && tsc",
```

philschatz (dismissed):
Other than the 1 suggestion I think it's fine to merge

package.json
TomWoodward: hey look at that

philschatz (approved):

PR #34, opened 2018-12-04
chore: add i18n

### LINKED ISSUE #14: Internationalize strings
### Description
Before we start adding in copy text, it is probably worthwhile to add string internationalization.

These 2 were brought up:
- https://github.com/yahoo/react-intl
- https://github.com/i18next/react-i18next

Both seem to support Typescript. yahoo's seems to support attributes, and yahoo's seems to be built on top of the browser i18n APIs. Not sure about the `i18next` one.

#### Branches

Both packages were checked by replacing the `page not found` with internationalized text:

- https://github.com/openstax/rex-web/compare/i18n-react-intl
- https://github.com/openstax/books-web/compare/i18n-react-i18next

#### Notes

The `react-i18next` package seems to offer these features & drawbacks:

- has option to load the translations via xhr but they _can_ be explicitly included (not sure if they can be nested)
- can translate components that have children and can move them to different positions in the translated text (`Hello [Phil], how are you? <avatar/>` vs `Hola! <avatar/> [Phil]`)
- does not seem to error when a **default is not provided**
- does not seem to do the translations for unit tests
- displays the token when something is not found in the translation file

The `react-intl` package seems to offer these features/drawbacks:

- simpler API (no need for an `i18n.js` for configuring
- might be the only one that uses browser i18n APIs
- _might_ support text with elements in it (& reordering those elements). e.g. `Hello [Phil], how are you? <avatar/>` vs `Hola! <avatar/> [Phil]`
- uses `id="i18ntokenhere"` which might be confusing
- unsure what happens when a token is missing
Context: https://openstax.slack.com/archives/C8V1XET1R/p1542216834346800

#### Katalyst notes:

> I was using `http://kazupon.github.io/vue-i18n/` for Vue.js app. It seems like your proposition of `i18next` is very similiar so I can say that I had good experiences with this kind of internationalization. However I wasn't using all of the functions, only basic ones.

> With current internationalization library for webview there was problem, because everything had to be in the same order and if someone left additional `space` or `enter` then it was breaking. With `i18n` this shouldn't be a problem.

Source: https://openstax.slack.com/archives/CBW9DFKGF/p1542314135166300

### Checklist for Done
1A-PRE DESIGN
- [ ] Description is complete 
- [ ] High-level natural language acceptance tests are complete
- [ ] If UX design is needed then the UX label is added
- [ ] If UX design is needed then the UX LoFi Design is completed, approved by the Product Manager, and added to the issue.
- [ ] A high-level solution (development approach for the code change) is added to the issue
- [ ] If the issue is a story, then Epic is selected

1B-DECOMP
- [ ] The work is broken into reasonable and consistent work items, and work item issues created. The work item issues are moved below the Done marker in the Decomp column.
- [ ] Estimate is updated for the work item issues.
- [ ] If there is a parent story epic issue, it is linked with the child work item issues and moved to the Epic column.
- [ ] If an issue is included in the quarterly plan, then Milestone is selected
- [ ] If an issue is included in a product phase, then Release is selected.

1C-DESIGN
- [ ] The UX HiFi Design is complete, approved by the Product Manager, and added to the issue.
- [ ] Distilled acceptance tests are added to the issue.
- [ ] Automated testing criteria are added to issue.
- [ ] The solution is updated.

2A-CODE
  - [ ] A pull request is opened, the automated pull request checks pass, and the change has been approved by other developers.
  - [ ] The acceptance tests are finalized in "Given / When / Then" format.
  - [ ] Regression test categories are identified on the issue. Categories should be referenced if any changes might affect them, even if the intended functionality is unchanged.

2B-PROD ACCEPT
- [ ] The pull request is merged into master by the Product Manager, and the change branch is deleted.

3A-PEND RELEASE
- [ ] All the changes in the release are deployed to the Release Candidate environment.
- [ ] The release is declared frozen by the Product Manager.

3B-RELEASE REVIEW
- [ ] Acceptance tests pass for all issues in the release
- [ ] Blessing from Product Manager that the release is good to go.

3C-RELEASED
- [ ] Production is updated and verified.
- [ ] All issues in the release are closed.

philschatz:
This adds Yahoo's `react-intl` library. Fixes #14

The `react-intl` package seems to offer these features/drawbacks:

- simpler API (no need for an `i18n.js` for configuring
- uses browser i18n APIs
- supports text with elements in it (& reordering those elements). e.g. `Hello [Phil], how are you? <avatar/>` vs `Hola! <avatar/> [Phil]`
- **Note:** uses `id="i18ntokenhere"` which might be confusing
- unsure what happens when a token is missing
Context: https://openstax.slack.com/archives/C8V1XET1R/p1542216834346800


src/app/errors/components/PageNotFound.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
+import { FormattedMessage } from 'react-intl';
 
 export default () => {
-  return <h1>page not found</h1>;
+  return <h1><FormattedMessage id='i18n:404'/></h1>;
```
TomWoodward: can we use this syntax to avoid the extraneous spans?

```
<FormattedMessage id="title">
    {(txt) => (
        <H1>
            {txt}
        </H1>
    )}
</FormattedMessage>
```

from https://github.com/yahoo/react-intl/wiki/Components#formattedmessage

src/app/index.tsx
```diff
@@ -70,10 +73,24 @@ export default (options: Options) => {
     reducer,
   });
 
+  const language = (typeof window === 'undefined') ? 'en-us' : window.navigator.language;
+  addLocaleData([...en, ...cs]);
+
+  const messages = new Map()
+  messages.set('en', { 'i18n:404': 'page not found' });
+  messages.set('cs', { 'i18n:404': 'požadovaná stránka neexistuje'});
```
TomWoodward: we may want to hold off on the second language file, that will definitely slow us down

src/app/index.tsx
```diff
@@ -70,10 +73,24 @@ export default (options: Options) => {
...
+
+  function getMessages(language: string) {
+      const lang = language.split('-')[0]
+      return messages.get(lang) || messages.get('en')
```
TomWoodward: you're defaulting to english in two places (here and on 76) we should consolidate this

src/app/index.tsx
```diff
@@ -70,10 +73,24 @@ export default (options: Options) => {
...
+  }
+
   const container = () => <Provider store={store}>
+  <IntlProvider locale={language} messages={getMessages(language)}>
```
TomWoodward: indentatatoin

src/app/index.tsx
```diff
@@ -70,10 +73,24 @@ export default (options: Options) => {
     reducer,
   });
 
+  const language = (typeof window === 'undefined') ? 'en-us' : window.navigator.language;
```
TomWoodward: i think it would be beneficial to pass the language as an option, browser can get it out of window.navigator and pre-render will either hardcode to english or potentially render different content in different languages

TomWoodward (changes requested):

philschatz:
Since we are only supporting 1 language so far, I just hardcoded the language to `en` and moved the `<IntlProvider>` into a `<MessageProvider>`.

That should delay the need for determining the current language until we start supporting a second language but allows us to have a place to start putting strings; thoughts?

philschatz:
Also, all of the tests now require `<MessageProvider><Provider>`. Would it make sense to have a Provider component that just combines both? I don't know of a case where we would not need to have the MessageProvider available. Plus, it would be harder to forget to include it.

src/app/MessageProvider.tsx
```diff
@@ -0,0 +1,22 @@
+
```
TomWoodward: extra whitespace

src/app/MessageProvider.tsx
```diff
@@ -0,0 +1,22 @@
...
+addLocaleData(en);
+
+const messages = {
+  'i18n:404': 'page not found',
+  'i18n:a11y:skipToContent': 'Skip to Content',
+  'i18n:toc:bookDetails': 'book details',
+  'i18n:toc:title': 'Table of Contents',
+  'i18n:toc:toggle:closed': 'Click to open the Table of Contents',
+  'i18n:toc:toggle:opened': 'Click to close the Table of Contents',
```
TomWoodward: is there some value to the `i18n:` prefix on everything?

src/app/content/components/Sidebar.tsx
```diff
@@ -87,6 +95,8 @@ const SidebarControl = styled(({isOpen, ...props}: React.HTMLProps<HTMLButtonEle
 
 `;
 
+const SidebarControl = injectIntl(StyledSidebarControl);
```
TomWoodward: is this `injectIntl` necessary to programmatically access the messages with `intl.formatMessage` rather than the `FormattedMessage` component?

src/app/content/components/Sidebar.tsx
```diff
@@ -31,8 +32,15 @@ const SidebarPlaceholder = styled.div<{isOpen: boolean}>`
...
-const SidebarControl = styled(({isOpen, ...props}: React.HTMLProps<HTMLButtonElement> & {isOpen: boolean}) =>
-  <button {...props}>
+type StyledSidebarControlProps = React.HTMLProps<HTMLButtonElement> & {
+  isOpen: boolean,
+  intl: IntlShape & InjectedIntl
+};
+
+const StyledSidebarControl = styled(({isOpen, intl, ...props}: StyledSidebarControlProps) =>
+  <button {...props}
+    aria-label={intl.formatMessage({id: isOpen ? 'i18n:toc:toggle:opened' : 'i18n:toc:toggle:closed'})}
```
TomWoodward: is there any reason to use formatMessage here instead of the FormattedMessage component?

src/app/index.tsx
```diff
@@ -83,9 +84,11 @@ export default (options: Options) => {
   });
 
   const container = () => <Provider store={store}>
-    <Services.Provider value={services} >
-      <navigation.components.NavigationProvider routes={routes} />
-    </Services.Provider>
+    <MessageProvider>
```
TomWoodward: if MessageProvider is here it doesn't need to also be in SkipToContentWrapper right?

TomWoodward (changes requested):

src/app/content/components/Sidebar.tsx
philschatz: Yes, I used it in order to set attributes.

TomWoodward: if you use the `FormattedMessage` with the render function you could accomplish the same thing. i suppose its useful to follow both patterns though because if you had multiple messages on different attributes nested `FormattedMessage` components would be super gross

src/app/content/components/Sidebar.tsx
philschatz: ooh, that might work. I'll try. I used the example from https://github.com/yahoo/react-intl/wiki#formatting-data but consistency with `FormattedMessage` would be nicer

src/app/index.tsx
philschatz: Yes, I can change that. I was debating whether it made more sense to wrap only components that need the MessageProvider with it or to have a global Provider.

src/app/MessageProvider.tsx
philschatz: I added the prefix initially to distinguish the i18n strings from other strings in the code like `opened` and to easily search-and-replace later. I figured it's easier to remove later than to add later. 

Can we keep it until we've explored ways for CI to verify that all strings are i18n-ized ? I'm not sure how to do that but one way is to just not add the lookup table in tests... then the tests could verify that `body.innerHTML` only contains strings that begin with `i18n:` or something.

TomWoodward: yea we can leave it for now

src/app/index.tsx
TomWoodward: i think providing it once globally is a little better, i could see the argument to applying it only to individual components though

src/test/reactutils.tsx
```diff
@@ -25,19 +26,28 @@ export function expectError(message: string, fn: () => void) {
...
-        throw new Error(`BUG: Component was not rendered`);
-    }
-    const node = ReactDOM.findDOMNode(c) as Element;
-    if (!node || !node.parentNode) {
-        throw new Error(`BUG: Could not find DOM node`);
+// Utility to handle nulls and SFCs
+export function renderToDom<C extends ComponentType<{}>>(subject: ReactElement<C>) {
+
+  // tslint:disable-next-line:variable-name
+  const Wrapper = class extends Component {
```
philschatz: Ah, so a SFC can't be rendered directly... it _always_ needs a `<Component>` around it. Thanks!

TomWoodward: yea that was a bummer, but i think this will work to just always wrap it

philschatz:
@TomWoodward can you review again? I think all of the issues are fixed, no?

TomWoodward:
we were leaving this open for a reason, but now i forget what....

TomWoodward:
oh no i was thinking of the other one, will review this

src/app/content/components/Sidebar.tsx
```diff
@@ -32,12 +33,22 @@ const SidebarPlaceholder = styled.div<{isOpen: boolean}>`
...
-    <span></span>
-  </button>
+type StyledSidebarControlProps = React.HTMLProps<HTMLButtonElement> & {
+  isOpen: boolean,
+};
+
+const SidebarControl = styled(({isOpen, ...props}: StyledSidebarControlProps) =>
+  <FormattedMessage id={isOpen ? 'i18n:toc:toggle:opened' : 'i18n:toc:toggle:closed'}>
+    {(txt) => {
+      return typeof txt === 'string' && <button {...props}
```
TomWoodward: i see that you're doing this because `aria-label` can't be an `Element` but i find this formatting to be a little gross *and* if for some reason it was an Element you'd want something more dramatic than the button silently disappearing to happen right?

i'd suggest making an `assertString` in `src/app/utils.ts` where i made an `assertDefined` helper for a similar purpose. that way you can do `aria-label={assertString(txt)}` and typescript will be ok because it either returns the thing as a string type or throws an exception.

src/app/index.tsx
```diff
@@ -86,9 +87,11 @@ export default (options: Options) => {
   });
 
   const container = () => <Provider store={store}>
-    <Services.Provider value={services}>
-      <navigation.components.NavigationProvider routes={routes} />
-    </Services.Provider>
+    <MessageProvider>
```
TomWoodward: this is still here

TomWoodward (changes requested):

src/app/index.tsx
philschatz: I think we ended up on including the `<MessageProvider>` only once on the app root, no? Or was it that we'd wrap every component that uses messages with a `<MessageProvider>`?

The former has fewer `<MessageProvider>` sprinkled in the code while the latter has fewer `<MessageProvider>` sprinkled through the tests

philschatz:
Updated. I wish there was a way for the tests to error when code contains non-18n strings... Any ideas?

TomWoodward:
for enforcing internationalization... it seems like probably a linter thing rather than a unit testing thing right? seems easier to statically analyze the code for string literals that don't start with `i18n` rather than add individual tests for components

TomWoodward:
i'm realizing that static analysis becomes even harder because text in jsx doesn't look like a string literal, and i looked around and there are a surprising number of string literals that aren't copy even in components (like attribute values)

testing for copy outside of the messaging system seems possible but hard, and i think you'd need some kind of a snapshot of things to ignore

TomWoodward (approved):

philschatz:
Hmm, good points. I'm not familiar with the `.tsx` parser but maybe there's a way to write custom lint rules?

In either case (linter or tests) we _could_ have a heuristic that checks to make sure there are no spaces in attributes or element text when rendered using a non-english language (I think `react-intl` will just output the `i18n:...` string if there is no default language).

PR #35, opened 2018-12-04
fix: some unmet peer dependency warnings

TomWoodward:
the only one left is 
```
warning "cnx-recipes > sass-extract@2.1.0" has unmet peer dependency "node-sass@^3.8.0 || 4".
```

which should probably be removed from cnx-recipes


philschatz (approved):
It's odd that these 2 need to be explicitly set even though they are in react-scripts. Oh, it's probably because we pull in `jest-puppeteer` or something that requires these packages to be explicitly set (rather than implicitly in other packages)

PR #37, opened 2018-12-05
fix: skip to content so that it actually skips to content

TomWoodward:


src/app/components/MainContent.tsx
```diff
@@ -1,11 +1,22 @@
 import React, { SFC } from 'react';
+import styled from 'styled-components';
 import { Consumer, MAIN_CONTENT_ID } from '../context/SkipToContent';
 
+// tslint:disable-next-line:variable-name
+const MainContentDispaly = styled.div`
+  outline: none;
```
philschatz: Is the outline being removed only for aesthetic reasons? If so, then I seem to recall a11y websites indicating that outline should not be suppressed because it helps sighted people see where their focus currently is.

```suggestion
```

philschatz (changes requested):

src/app/components/MainContent.tsx
TomWoodward: if we want to keep the outline i should update the other styles so the outline appears in a more useful place, currently its mostly hidden outside the window and one line shows up around the sidebar scrollbar, it doesn't really indicate the content at all

philschatz (approved):

PR #38, opened 2018-12-05
feat: Lighthouse a11y tests

philschatz:
Performs an audit of `/` and `TEST_PAGE` using https://github.com/GoogleChrome/lighthouse.

- `/` has a score of 100% for accessibility
- `TEST_PAGE` has a score of 100% for accessibility


src/lighthouse.browserspec.ts
```diff
@@ -0,0 +1,50 @@
...
+  afterAll(async() => {
+    return chrome && await chrome.kill();
+  });
+
+  it('reports about the root page', async() => {
+    await checkLighthouse('/');
+  });
+
+  it('reports about the test content page', async() => {
+    await checkLighthouse(TEST_PAGE, 0.95);
```
philschatz: I would like to see what the report generated & would like to try to get an HTML report out (similar to [tutor/scripts/accessibility-audit/index.js](https://github.com/openstax/tutor-js/blob/b7b9cc4a69d547ce7773906dde8279ad82743466/tutor/scripts/accessibility-audit/index.js) ) but maybe running the audit in the browser is sufficient for now. Thoughts?

TomWoodward: yea i don't think we have to worry about the html reports for now, we might generate those separately later but it wouldn't be part of the unit test suite

TomWoodward (dismissed):

TomWoodward (approved):

PR #39, opened 2018-12-06
fix: clarify errors and add env defaults

philschatz:
... because I was having trouble debugging the following:

- why the browser test was printing a console message saying that `business.json` was missing (there was zombied `node` process)
- why `yarn prerender` did not work (it needed `NODE_ENV` to be set)
- why `yarn server` did not start up (it needed `PORT` to be set but was silent)
- why I could not run `NODE_ENV=test yarn server` (no way to change the default; I wanted the fixtures to load)
- whether an argument is a file or not (I use `./foo` in `package.json` to clarify that the argument is a file rather than the command `foo`)

Goal is to be able to run either of these commands and have a webserver running on `:3000` (by default):

- `yarn start`
- `yarn build && yarn prerender && yarn server` or just `yarn start:static`

I would still like to be able to add `debugger` lines anywhere (in app code, tests, and server) and have node stop but that might be weekend work. It might involve sprinkling `--inspect-brk` inside `package.json` which is not ideal.


script/server/cli.ts
```diff
@@ -2,8 +2,15 @@
 import { argv } from 'yargs';
 import server from '.';
 
-server({port: parseInt(process.env.PORT || '', 10), ...argv})
-  .then(({port}) => console.log(`running on port: ${port}`))
+const port = parseInt(process.env.PORT || '', 10);
+
+if (!process.env.PORT || Number.isNaN(port)) {
```
TomWoodward: you shouldn't need to make these changes, it should get caught by https://github.com/openstax/rex-web/blob/master/script/server/index.ts#L22

i would prefer not to add the check here because when i run it manually i like to specify the port with `--port` rather than the env

package.json
```diff
@@ -21,18 +20,18 @@
...
-    "server": "NODE_ENV=development node script/entry server/cli",
+    "server": "NODE_ENV=${NODE_ENV:-development} node ./script/entry server/cli",
     "codecov": "codecov",
     "postinstall": "node ./script/postinstall",
     "test": "npm run-script test:unit && npm run-script lint && npm run-script test:build && tsc",
     "test:unit": "jest --coverage --testPathPattern=\"(\\.|/)(browser)?spec\\.tsx?\"",
     "test:build": "npm run-script build:clean && npm run-script test:sourcemap && npm run-script test:prerender",
     "test:sourcemap": "node ./script/verifySourcemaps.js",
-    "test:prerender": "NODE_ENV=test npm run-script prerender && SERVER_MODE=built jest --testPathPattern=\"(\\.|/)(prerender|browser)spec\\.tsx?\""
+    "test:prerender": "NODE_ENV=${NODE_ENV:-test} npm run-script prerender && SERVER_MODE=built jest --testPathPattern=\"(\\.|/)(prerender|browser)spec\\.tsx?\""
```
TomWoodward: `test:prerender` is never going to work outside `NODE_ENV=test`, we shouldn't allow changing it here

TomWoodward (changes requested):

script/server/index.ts
```diff
@@ -18,7 +18,7 @@ export default (options: Options): Promise<{server: http.Server, port: number}>
   const baseDir = path.join(__dirname, '../../build');
   const {port, onlyProxy} = {...defaultOptions, ...options};
 
-  if (!port) {
+  if (!port || !(port > 0)) {
```
TomWoodward: `port < 1` ?

TomWoodward (changes requested):

script/server/index.ts
philschatz: It was easier than writing `!port || !Number.isNAN(port) || port <= 0` . I ran into this area silently failing because `const port = parseInt(process.env.PORT || '', 10);` was set to `NAN`.

TomWoodward: the `not greater than 0` confused me. if your goal is to actually check for NaN i think we should do that explicitly

package.json
```diff
@@ -21,18 +20,19 @@
...
     "start": "react-scripts start",
-    "clean": "rm -rf build",
-    "build": "GENERATE_SOURCEMAP=true react-scripts build",
+    "start:static": "npm run-script build && npm run-script prerender && npm run-script server",
+    "clean": "rm -rf ./build",
+    "build": "GENERATE_SOURCEMAP=${GENERATE_SOURCEMAP:-true} react-scripts build",
     "build:clean": "npm run-script clean && npm run-script build",
-    "prerender": "node ./script/entry prerender",
-    "server": "NODE_ENV=development node script/entry server/cli",
+    "prerender": "NODE_ENV=test node ./script/entry prerender",
```
TomWoodward: definitely don't want to hardcode this env, i'd default it to `production` if anything

package.json
```diff
@@ -21,18 +20,19 @@
...
+    "server": "NODE_ENV=${NODE_ENV:-development} node ./script/entry server/cli",
     "codecov": "codecov",
     "postinstall": "node ./script/postinstall",
     "test": "npm run-script test:unit && npm run-script lint && npm run-script test:build && tsc",
     "test:unit": "jest --coverage --testPathPattern=\"(\\.|/)(browser)?spec\\.tsx?\"",
     "test:build": "npm run-script build:clean && npm run-script test:sourcemap && npm run-script test:prerender",
     "test:sourcemap": "node ./script/verifySourcemaps.js",
-    "test:prerender": "NODE_ENV=test npm run-script prerender && SERVER_MODE=built jest --testPathPattern=\"(\\.|/)(prerender|browser)spec\\.tsx?\""
+    "test:prerender": "npm run-script prerender && SERVER_MODE=built jest --testPathPattern=\"(\\.|/)(prerender|browser)spec\\.tsx?\""
```
TomWoodward: we want this test env set here

script/server/cli.ts
```diff
@@ -2,8 +2,10 @@
 import { argv } from 'yargs';
 import server from '.';
 
-server({port: parseInt(process.env.PORT || '', 10), ...argv})
-  .then(({port}) => console.log(`running on port: ${port}`))
+const port = parseInt(process.env.PORT || '0', 10);
```
TomWoodward: it seems like maybe we don't even need this here? because https://github.com/openstax/rex-web/pull/39/files#diff-3f7ce2d17b676da9527903e740774c69R16 is reading out of the env for the defaultOption

TomWoodward (changes requested):
you have a diff file content-browserspec-ts-content-looks-right-1-diff.png committed there

philschatz:
RE `.diff` file: I consistently get that diff but it seems I missed it this time 😞 . I think the other concerns were addressed. I'm happy the code seems a bit cleaner without checking multiple times for the `PORT`.

package.json
```diff
@@ -21,11 +20,12 @@
   "scripts": {
     "lint": "tslint -p ./tsconfig.json",
     "start": "react-scripts start",
-    "clean": "rm -rf build",
-    "build": "GENERATE_SOURCEMAP=true react-scripts build",
+    "start:static": "npm run-script build && NODE_ENV=${NODE_ENV:-test} npm run-script prerender && npm run-script server",
```
TomWoodward: doest the NODE_ENV default get applied to the `npm run-script server` as well? it needs to be for the proxies to work properly

TomWoodward (dismissed):

package.json
philschatz: oh, it does not. What if `"server": ...` also contains the `NODE_ENV=${NODE_ENV:-test}` default?

philschatz: ok, how about this (see commit below)?

.env.development
```diff
@@ -1,3 +1,2 @@
 ARCHIVE_URL=https://archive.cnx.org/
 REACT_APP_ARCHIVE_URL=/contents/
-PORT=3000
```
TomWoodward: you can leave the port in development, its possible to run the pre-renderer in development it'll just render all of college physics instead of just the fixtures

TomWoodward (dismissed):

TomWoodward (approved):

PR #46, opened 2018-12-10
feat: Render Mathjax in the browser

TomWoodward:
mathjax-render

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -10,6 +10,8 @@
       homescreen on Android. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
     <!--
       Notice the use of %PUBLIC_URL% in the tags above.
       It will be replaced with the URL of the `public` folder during the build.
--- a/src/helpers/mathjax.ts
+++ b/src/helpers/mathjax.ts
@@ -94,7 +94,7 @@ function startMathJax() {
     },
   };
 
-  if (!window) {
+  if (typeof(window) === 'undefined') {
     throw new Error('BUG: Window is undefined');
   }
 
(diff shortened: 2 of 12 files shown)


public/index.html
```diff
@@ -10,6 +10,8 @@
       homescreen on Android. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
```
philschatz: Oh nice, there is some page jitter but maybe MathJax corrects itself after rendering? For example, http://localhost:3000/books/Ax2o07Ul/pages/M_qlK4M9#63848_copy_1 points to "10.4 Rotational Kinetic Energy: Work and Energy Revisited" and after some jitter, continues to point to the same place.

Hmm, what happens if we have a floating (absolute positioned) bar at the top? Will MathJax's `scrollTo` cause content to no longer be visible?

Also, any thoughts on reducing the jitter in the future? Some options: 

- wrap each `<math>` with a `<span/div style="display: block; width: ###; height: ###">`
- prerender using MathJax and then re-render in the browser (would need to squirrel away the MathML)

src/helpers/mathjax.ts
```diff
@@ -0,0 +1,123 @@
+import { Element } from '@openstax/types/lib.dom';
+import debounce from 'lodash/debounce';
+import isEmpty from 'lodash/isEmpty';
+import memoize from 'lodash/memoize';
+import WeakMap from 'weak-map';
+
+const MATH_MARKER_BLOCK  = '\u200c\u200c\u200c'; // zero-width non-joiner
```
philschatz: Was this copied from tutor? Just curious.

TomWoodward: yea this whole file was converted from https://github.com/openstax/tutor-js/blob/master/shared/src/helpers/mathjax.js

public/index.html
TomWoodward: yea we have openstax/unified-issues#53 to try out pre-rendering math, i'm not sure how its going to work out but i think that would be ideal

philschatz (dismissed):

philschatz (approved):
Wow, MathJax adds ~30MB of memory even if the page does not contain any math.

The test I did was: Moving from Preface to 3.5 to Preface.

Without MathJax the heap is: 10.3MB, 11.8MB, 12.1MB
With MathJax the heap is: 41.6MB, 47.1MB, 44.8MB

PR #47, opened 2018-12-11
chore: add custom Lighthouse audits

philschatz:
Fixes https://github.com/openstax/unified-issues/issues/92

@TomWoodward I left the audit code in `.js` so that lighthouse can run from the terminal so we can generate an HTML report. But if it's worth generating a report whenever the tests run (maybe behind a `CI=true` flag) then I can translate them to TypeScript.


src/test/lighthouse-config.js
```diff
@@ -0,0 +1,92 @@
...
+      return [...elementsWithId].map(el => el.getAttribute('id'))
+    };
+    return driver.evaluateAsync(`((${fn.toString()}) ())`);
+  }
+}
+
+
+const ALL_AUDITS = [ SingleH1, UniqueIds, ];
+
+module.exports = {
```
TomWoodward: can we move the custom audit classes into separate files?

TomWoodward (changes requested):

TomWoodward (approved):

PR #48, opened 2018-12-13
feat: Content references

TomWoodward:


src/app/content/utils.ts
```diff
@@ -1,8 +1,37 @@
 import { HTMLElement } from '@openstax/types/lib.dom';
+import flatten from 'lodash/fp/flatten';
+import { isArchiveTree } from './guards';
+import { ArchiveTree, ArchiveTreeSection } from './types';
 
 export const stripIdVersion = (id: string): string => id.split('@')[0];
 export const getIdVersion = (id: string): string | undefined => id.split('@')[1];
 
+export const getContentPageReferences = (content: string) =>
+  (content.match(/\/contents\/([a-z0-9\-]+(@[\d\.]+)?)(:([a-z0-9\-]+(@[\d\.]+)?))?/g) || [])
```
philschatz: I am worried about extracting links by executing a regular expressions over the whole HTML page. Why are these resolved before injecting the HTML into the DOM rather than after?

philschatz (commented):
Just 1 question on using a regexp on the whole document to find content links.

Also, I pushed a commit updating the fixture file because I had trouble finding a piece of content that contained a link to another piece of content.

README.md
```diff
@@ -39,6 +39,9 @@ Run `PORT=8000 yarn start:static` to change the webserver port.
 
 Run `NODE_ENV=development yarn start:static` to build the books (defined in [./.env.development](./.env.development)) and serve them.
 
+
+To prerender the test fixture book and serve it, run `NODE_ENV=test npm run-script prerender && NODE_ENV=test PORT=3000 node ./script/entry server/cli`
```
TomWoodward: can you not just `NODE_ENV=test yarn serve:static` ?

src/app/content/utils.ts
TomWoodward: its being applied to the content string before being loaded into the dom (differentiating from `HTML page` which could mean the whole page, not just the content). its happening this way so that we can load content references as part of the initial data load before anything is rendered. this has some benefits:

- modified links are pre-rendered
- there is no interstitial time where a user could click a link before it gets replaced

philschatz: Okies, I just hope no one writes something like:

`<p>To find out more, you can visit <a href="/contents/{uuid}">/contents/foobar</a> or just type /contents/foobar into your browser`

TomWoodward: the regex isn't looking for a link, its looking for the paths, so that would actually work 😄

README.md
philschatz: oops, yep! fixed

src/app/content/utils.ts
philschatz: that was my original concern 😄.

RE interstitial time: I see what you mean... it _could_ be alleviated by loading the HTML into a detached DOM node, finding the links, rewriting them, and then attaching it to the DOM but maybe that's overkill?

philschatz (dismissed):

philschatz (approved):

PR #49, opened 2018-12-13
chore: add snapshot helper for styled-components

TomWoodward:


philschatz (approved):
🎉 now we have style diffs instead of cryptic class names!

PR #50, opened 2018-12-13
perf: move around some pre-render logic

TomWoodward:
added some interesting numbers to the output and parallelize some stuff to make it faster, speeds up locally rendering college physics by almost 10x, we'll see what happens on jenkins

### DIFF
--- a/script/prerender.tsx
+++ b/script/prerender.tsx
@@ -1,4 +1,5 @@
 import fs from 'fs';
+import chunk from 'lodash/fp/chunk';
 import flatten from 'lodash/fp/flatten';
 import fetch from 'node-fetch';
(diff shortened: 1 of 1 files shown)


script/prerender.tsx
```diff
@@ -116,20 +119,47 @@ async function render() {
...
+        );
+      }
+
+      await Promise.all(promises);
+    }
+  }
+
+  console.info('rendering...'); // tslint:disable-line:no-console
+
+  for (const pageChunk of chunk(50, pages)) {
```
philschatz: Wow! Makes sense since the network is the bottleneck

philschatz (approved):

PR #54, opened 2018-12-16
chore: refactor to emotion for css sourcemaps

philschatz:
This is mostly for feedback on whether this is a promising tangent to continue down.

This swaps out `styled-components` with `@emotion/styled/macro` (the `/macro` is for [babel-plugin-macros](https://github.com/kentcdodds/babel-plugin-macros) which CRA uses).

From Travis, it seems like some of the styles that use `:nth-child` might be buggy and might work better if replaced with `:nth-of-type` (it generated warnings)

Also, from Travis it seems like the prerender styles do not quite work yet.


script/prerender.tsx
```diff
@@ -5,7 +5,6 @@ import path from 'path';
 import portfinder from 'portfinder';
 import React from 'react';
 import { renderToString } from 'react-dom/server';
-import { ServerStyleSheet, StyleSheetManager } from 'styled-components';
```
TomWoodward: without this stuff in pre-render, how do the styles get into the pre-rendered html?

src/emotion-styled-macro.d.ts
```diff
@@ -0,0 +1,180 @@
+declare module '@emotion/styled/macro' {
```
TomWoodward: is there a reason this shorter workaround didn't work? https://github.com/emotion-js/emotion/issues/933#issuecomment-447831850

src/app/content/components/__snapshots__/Content.spec.tsx.snap
```diff
@@ -3,21 +3,21 @@
 exports[`content matches snapshot 1`] = `
 Array [
   <a
-    className="sc-bdVaJa RfHPw"
+    className="css-1axmsgr e1wgaj6u0"
```
TomWoodward: i wonder is these class names suffer from the same issue as styled where they will all update anytime any styles changes? 

also it looks like you branched off before i added the jest-snapshot thing, is there a comparable thing for emotion that will show the styles in the snapshot?

src/emotion-styled-macro.d.ts
philschatz: that comment was made 4 hours ago! I'll try it 😄

script/prerender.tsx
philschatz: Ahh, maybe I need to include the `<CacheProvider>` (from https://emotion.sh/docs/ssr ) . I will try later

TomWoodward: oh apparently you don't need anything special for ssr in emotion because it injects the style tags adjacent to the elements

philschatz:
It seems like the sourcemaps all point to `Header.tsx:3`. I'm not sure why that is the case.

image

Earlier commits did not seem to work either (commit `@233fcd5` was `inline:1`)

TomWoodward:
@philschatz it seems to be working for me in dev 
untitled

but after reading https://github.com/emotion-js/emotion/blob/master/docs/source-maps.md i checked the pre-rendered version and it indeed does not include the sourcemaps. looking at https://github.com/emotion-js/emotion/issues/395 it doesn't seem like production sourcemaps are supported at all by design.

TomWoodward:
adding `hold` label until we figure out either:
- sourcemaps in production
- if we want to use emotion instead of styled in general

PR #56, opened 2018-12-18
feat: highlight the targeted element

philschatz:
(this was part of investigating what needs to be added to make book styles more useful)
Refs https://github.com/openstax/unified-issues/issues/60

When the link contains a `#` and targets a specific element, highlight the element.

This is particularly helpful for things like linking to a solution in the back of the book, discussing just a table, or seeing a specific footnote.

```
  *:target {
    background-color: #ffffdd;
  }
```

# Examples

## A Term (from the index)

1. Click a link in the index: For example, https://rex-web-pr-56.herokuapp.com/books/college-physics/pages/26-6-aberrations#term676
  - You might need to refresh the page to see it (because of SinglePageApp quirkiness)

image

## A Table

Example: https://rex-web-pr-56.herokuapp.com/books/college-physics/pages/11-6-gauge-pressure-absolute-pressure-and-pressure-measurement#eip-286

image


src/app/content/components/BookStyles.tsx
```diff
@@ -8,4 +8,8 @@ const Hoc = ({children, className}: {className?: string, children: (className?:
 
 export default styled(Hoc)`
   ${bookStyles}
+  /* Occur later so the bookstyles do not bleed below (e.g. @page) */
+  *:target {
```
TomWoodward: i have experienced problems with adding additional styles to the same component as the `bookStyles`, thats why i added some patchup styles in `ContentPane`

philschatz: Thanks! I moved them over & left a comment in here

src/app/content/components/BookStyles.tsx
```diff
@@ -8,4 +8,5 @@ const Hoc = ({children, className}: {className?: string, children: (className?:
 
 export default styled(Hoc)`
   ${bookStyles}
+  // Insert custom styles into ContentPane
```
TomWoodward: not sure if this will be correctly ignored inside the css, might want to move it up to line 8

TomWoodward:
adding `hold` label until:
highlighting elements need to look a little nicer, images in particular. we may figure this out as part of search and can apply it here as well.

PR #57, opened 2018-12-21
feat: Change URL structure

TomWoodward:
i started wanting to be able to access environment variables in the bundled code in order to get at the default book version in the books config, previously there were no variables used in the bundled code that were different between environments so this is the first time i ran into the problems described [here](https://github.com/openstax/rex-web/compare/master...url-structure?diff=unified&expand=1#diff-9d9a6cd82f41984872a66a3ab0d440c4R2) and ended up ripping out all the dotenv stuff and replacing with js configs.

its important to note that the way webpack handles environment variables the configs for other environments do not get bundled, so we don't have to worry about having dead development code in production.

[link to useful place in the review environment](https://rex-web-pr-57.herokuapp.com/books/college-physics/pages/preface)


TomWoodward:
[link to useful place in the review environment](https://rex-web-pr-57.herokuapp.com/books/college-physics/pages/Preface)

TomWoodward:
concourse pr-test

TomWoodward:
concourse pr-test

TomWoodward:
concourse pr-test

TomWoodward:
concourse pr-test

TomWoodward:
concourse pr-test

TomWoodward:
concourse pr-test

philschatz:
Oh no... the heroku deployment no longer works... maybe it is because [cp build/books/Ax2o07Ul/pages/HR_VN3f7 build/index.html](https://github.com/openstax/rex-web/pull/57/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L38) was removed from `package.json`?

TomWoodward:
it probably just expired over vacation

TomWoodward:
works again, the link to the useful place may not be necessary in the future, lets talk offline

src/app/content/hooks/locationChange.ts
```diff
@@ -1,60 +1,113 @@
...
 const fontMatches = css.match(/"(https:\/\/fonts\.googleapis\.com\/css\?family=.*?)"/);
 const fonts = fontMatches ? fontMatches.slice(1) : [];
 
 const hookBody: RouteHookBody<typeof content> = (services) => async({match}) => {
-  const {dispatch, getState, fontCollector, archiveLoader} = services;
+  const {fontCollector} = services;
+
+  fonts.forEach((font) => fontCollector.add(font));
+
+  await resolveBook(services, match).then(resolvePage(services, match));
```
philschatz: Form a cursory skim, it seems like `.then` does not do anything... but `resolvePage(...)` returns a function rather than evaluating which was a bit unexpected. What about something like:

```suggestion
  const [book: Book, bookLoader: ReturnType<AppServices['archiveLoader']['book']>] = await resolveBook(services, match);
  await resolvePage(services, match, book, bookLoader);
```

src/app/content/hooks/locationChange.ts
```diff
@@ -1,60 +1,113 @@
...
+    };
+    return [newBook, archiveLoader.book(newBook.id, newBook.version)];
+  };
 
-  const [bookRefId, bookRefVersion, pageRefId] = match.state
-    ? [match.state.bookUid, match.state.bookVersion, match.state.pageUid]
-    : [bookId, undefined, pageId];
+  if (bookSlug !== select.loadingBook(state)) {
+    dispatch(requestBook(bookSlug));
+    return await response().then((params) => {
```
philschatz: Why combine both `await` and Promise syntax?

```suggestion
    const params = await response();
    dispatch(receiveBook(params[0]));
    return params;
```

src/app/content/utils.ts
```diff
@@ -59,3 +61,77 @@ export const scrollTocSectionIntoView = (sidebar: HTMLElement | undefined, activ
...
+    throw new Error(`BUG: could not URL encode page title: "${section.title}"`);
+  }
+
+  const cleanNumber = sectionNumber || splitTitleParts(section.parent.title)[0];
+
+  const cleanTitle = replaceAccentedCharacters(sectionTitle)
+    // handle space delimiters
+    .replace(/[-_]+/g, ' ')
+    // remove special characters
+    .replace(/[^a-z0-9 ]/gi, '')
```
philschatz: Should we worry about UTF-8 characters? And is there a package we could use rather than rolling our own? I am concerned about having to maintain the URL-computing code indefinitely and it would be nice if there were some language-agnostic "best practices" methods of computing this string

src/app/content/utils.ts
```diff
@@ -59,3 +61,77 @@ export const scrollTocSectionIntoView = (sidebar: HTMLElement | undefined, activ
...
+  flattenArchiveTree(book.tree).find((section) =>
+    stripIdVersion(section.shortId) === stripIdVersion(pageId)
+    || stripIdVersion(section.id) === stripIdVersion(pageId)
+  );
+
+const getUrlParamForPageTitle = (section: LinkedArchiveTreeSection): string => {
+  const splitTitleParts = (str: string) => {
+    const match = str
+      // remove html tags from tree title
+      .replace(/<[^>]+>/g, '')
```
philschatz: Would injecting the text into a detached `<div>` be less error-prone? What if the title contains an HTML entity tag (like `&nbsp;` or `&lt;`)

Example:

```js
const div = document.createElement('div');
const strip = (xml) => {
  div.innerHTML = xml
  const stripped = div.textContent
  div.innerHTML = ''
  return stripped
}
```

src/app/content/utils.ts
```diff
@@ -59,3 +61,77 @@ export const scrollTocSectionIntoView = (sidebar: HTMLElement | undefined, activ
...
+      // split out section number from title
```
philschatz: I wonder if having `.` in the URL will cause people problems... It _might_ look like the file has a _really_ long file extension

src/app/utils.ts
```diff
@@ -31,3 +31,17 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
       return result;
     };
   };
+
+/*
+ * util for dealing with array and object index signatures
+ * don't include undefined
+ *
+ * ref: https://github.com/Microsoft/TypeScript/issues/13778
```
philschatz: 😄

src/config.js
```diff
@@ -0,0 +1,29 @@
+/*
+ * `react-scripts start` hardcodes NODE_ENV='development' but we run it in
```
philschatz:
```suggestion
 * This file is written in JS so the unused configs are optimized out of the built file.
 *
 * `react-scripts start` hardcodes NODE_ENV='development' but we run it in
```

philschatz (changes requested):
Just a couple of concerns RE the URL-escaping logic.

Also, I did not check but does this mean that there 2 JS files will be loaded (the app and the config) on each HTML page?

src/app/content/utils.ts
TomWoodward: at first i was removing things like &amp; or &lt; but ended up preferring to leave them in the url, i thought it looked better that way.

&nbsp; should probably be replaced and treated like a space though.

TomWoodward:
re: Also, I did not check but does this mean that there 2 JS files will be loaded (the app and the config) on each HTML page?

only in dev, in prod the extra data is always preloaded

src/app/content/utils.ts
TomWoodward: are you referring to my `replaceAccentedCharacters` function? it looks like this package provides basically the same function https://github.com/tyxla/remove-accents

the stackoverflow i copied handles more characters than the package, but i'm not sure how much it matters.

src/app/content/utils.ts
philschatz: What about others like the matching double-quotes, math symbols, or Chinese characters? They could be removed... `webview` did that initially but then had to add them back in for SEO around the time the Polish versions of books came out.

TomWoodward: the replaceAccentedCharacters function replaces many wonky characters with their non-wonky equivalent for the url, anything not in that map would be stripped out by the regex

src/app/content/utils.ts
TomWoodward: how about this one? https://www.npmjs.com/package/slugify

philschatz: It seems like slugify performs some ad-hoc character replacements... I am not sure how much that helps the situation.

It looks like the landscape is less-than-ideal:

- django uses at least [3 libraries depending on which is installed](https://github.com/justinmayer/django-autoslug/blob/master/autoslug/settings.py#L30-L50).
  - https://pypi.org/project/translitcodec/ seems to be the most in-depth, providing 3 different strategies depending on if you want long, short, or single-character replacements (see [package link](https://pypi.org/project/translitcodec/) for examples)
- jekyll has [several regular expressions](https://github.com/jekyll/jekyll/blob/4271495fcaad4073b2375b76089041acd16ffb49/lib/jekyll/utils.rb#L13-L18) that seem to replace some characters with a `-`
- wordpress : I had trouble finding the definition of the `wp_unique_post_slug` function (see code search: https://github.com/search?p=1&q=org%3AWordPress+wp_unique_post_slug&type=Code) so threw my hands up


So, it's fine as-is unless you wanted to do something like replace (jekyll's method): 

```
source:  some_1_and_._and_,_and_:_and_&_and_ä_and_á_and_€_and_☺_and_(_and_)
slugged: some_1_and_._and_,_and_-_and_&_and_ä_and_á_and_-_and_-_and_(_and_)
```

src/app/content/utils.ts
philschatz: ok, I don't think it's a big issue... periods are probably fine. I assume/hope that LMS' and things handle periods in URLs (like autodetecting what is a URL from a big paragraph of text). That might be something we can test before release? Thoughts @kjdav ?

philschatz (dismissed):

philschatz (dismissed):
Yay for codeclimate

src/setupProxy.js
```diff
@@ -24,14 +24,27 @@ function setupTestProxy(app) {
...
+    const queryFilePath = path.join(filePath, parts.search ? encodeURIComponent(parts.search) : '');
+    const indexFilePath = path.join(filePath, 'index.html');
+
+    const sendFile = path => fs.readFile(path, (err, contents) => {
+      res.end(contents);
+    });
+
+    const isFile = path =>
+      fs.existsSync(path)
+      && fs.existsSync(fs.realpathSync(path))
```
philschatz: Why is `realPathSync` needed here? And why is it not necessary for checking below in `isDirectory`?

src/setupProxy.js
```diff
@@ -24,14 +24,27 @@ function setupTestProxy(app) {
...
+      && fs.lstatSync(fs.realpathSync(path)).isFile();
+
+    const isDirectory = path => fs.existsSync(path) && fs.lstatSync(path).isDirectory();
+
+    if (isFile(queryFilePath)) {
+      sendFile(queryFilePath);
+    } else if (isFile(filePath)) {
+      sendFile(filePath);
+    } else if (isDirectory(filePath) && isFile(indexFilePath)) {
+      sendFile(indexFilePath);
```
philschatz: When is the index file served?

src/setupProxy.js
```diff
@@ -24,14 +24,27 @@ function setupTestProxy(app) {
...
+
+    const isFile = path =>
+      fs.existsSync(path)
+      && fs.existsSync(fs.realpathSync(path))
+      && fs.lstatSync(fs.realpathSync(path)).isFile();
+
+    const isDirectory = path => fs.existsSync(path) && fs.lstatSync(path).isDirectory();
+
+    if (isFile(queryFilePath)) {
+      sendFile(queryFilePath);
```
philschatz: When do we serve filenames with the query string in them? I did not see any of them in the diff

src/setupProxy.js
TomWoodward: because the fixtures use symlinks to allow loading the archive files with either short or uid, so the second check makes sure the symlink is valid. the fixtures don't use any symlinks on directories so i didn't bother.

src/setupProxy.js
TomWoodward: the simulations

src/setupProxy.js
TomWoodward: osweb queries, this logic was already here i just moved it

philschatz (approved):

PR #58, opened 2019-01-03
chore: stabilize testing environment

TomWoodward:


jest-puppeteer.config.js
```diff
@@ -16,6 +16,8 @@ module.exports = {
       height: 400,
     },
     args: [
+      // so that tests can be run as root, if you're into that sort of thing.
+      '--no-sandbox',
```
philschatz: Could this be something like `process.env.CI === 'true' ? '--no-sanbox' : undefined` ? That way puppeteer doesn't reuse cookies when running tests locally?

philschatz (changes requested):

jest-puppeteer.config.js
TomWoodward: yea definitely, i'll change

philschatz (approved):

PR #60, opened 2019-01-09
chore: screenshot test sims

TomWoodward:
i moved the lighthouse audit to another page so it wouldn't complain about the issues in the sim, we should capture work to update sims somewhere...

this copies some of the fixture changes from #57, will conflict when merged. the proxy in this pr is better.


philschatz (approved):

philschatz:
It's a little weird to rely on s3 for CI tests... is there a reason we need a real sim for the tests?

TomWoodward:
@philschatz i dont know... asserting that iframes were iframy with some page with a red background or whatever didn't seem like it was testing much

PR #63, opened 2019-01-15
chore: bundle analyzer

TomWoodward:
we should run this periodically to make sure nothing crazy shows up, so far i don't see any low hanging fruit to optimize.

it generates a report.html in the build folder when you run it. we could add something to CI that posts an image of the report when deps are added. i think that would be more neat than useful though. 

image


philschatz (dismissed):
Nice! 

html5-entities is **huge**... the input is XML so we won't have named entities like `&nbsp;` (except for the 2-ish XML entities `&amp;` and `&quot;`) so I'm not sure if it is actually necessary. We will likely have numbered entities like `&1234;`

philschatz:
Do you know why the `index.ts + 56 files` is collapsed?

philschatz (commented):
Nice! Do you know why the `index.ts + 56 files` is collapsed?


I've been using the following steps and it looks like this combines them into 1:

1. `webpack --profile --json > ./webpack-stats.json`
1. Upload the `webpack-stats.json` to https://webpack.github.io/analyse/#modules

script/webpack-bundle-analyzer.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * hack to run the analyzer as a webpack plugin taken from:
+ * https://github.com/facebook/create-react-app/issues/3518
+ */
+
+process.env.NODE_ENV = "production"
```
philschatz: Ah, concatenating files (& causing the single entry of `index.ts + 54 files (concatenated)` is a default config option for `production`: https://webpack.js.org/plugins/module-concatenation-plugin/ ... so we can still analyze the individual file sizes if dev mode was turned on.

Would it be useful to default to `production` but allow a `NODE_ENV=development yarn analyze-build`?

TomWoodward: that would be cool, except https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/scripts/build.js#L13

philschatz (dismissed):

philschatz (approved):

PR #65, opened 2019-01-15
chore: get review apps to spin down

philschatz:
after 30 minutes of idle time. Maybe this works?

See https://devcenter.heroku.com/articles/dyno-types


TomWoodward (approved):
/shrug lets give it a shot!

PR #68, opened 2019-01-16
chore: Add pytest and selenium test framework to the project

m1yag1:
Added pytest and selenium test framework to the project


.gitignore
```diff
@@ -27,3 +27,100 @@ yarn-error.log*
 
 # lighthouse report
 /localhost_*.report.html
+
+# Jetbrains files
+.idea/
```
TomWoodward: i'm not morally opposed to ignoring editor files in the project gitignore, but since your editor is a personal decision, you could add them to your [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore)

.gitignore
```diff
@@ -27,3 +27,100 @@ yarn-error.log*
...
+.python-version
+
+# celery beat schedule file
+celerybeat-schedule
+
+# SageMath parsed files
+*.sage.py
+
+# Environments
+.env
```
TomWoodward: `.env` is intentionally allowed, use `.env.local` for stuff outside version control. that said we don't use the environment files for the JS on this project anymore so if you want to leave this you can. if you do leave it maybe combine it with all the env related stuff above

.gitignore
```diff
@@ -27,3 +27,100 @@ yarn-error.log*
...
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.hypothesis/
```
TomWoodward: i feel as if some of this stuff isn't necessary

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
```
TomWoodward: @philschatz maybe we should add the faster checks like `tsc` and `tslint` as commit hooks?

Makefile-py
```diff
@@ -0,0 +1,37 @@
+
```
TomWoodward: can this be in the folder with all the python stuff? or rename it to just `Makefile` and i'll start adding rando scripts here too

pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[tool.black]
```
TomWoodward: if we could move as many of the python project files into a subdirectory as possible that would make me happy

testrail.example.cfg
```diff
@@ -0,0 +1,10 @@
+[API]
+url = https://instance.testrail.net/
```
TomWoodward: i know this is just an example, but in reality we're going to have to eventually pull all these configs from the environment, is that possible?

tests/ui/README.md
```diff
@@ -0,0 +1,153 @@
...
+### Additional Pytest Options
+
+The pytest plugin that we use for running tests has a number of advanced
+command line options available. To see the options available, run
+`pytest --help`. The full documentation for the plugin can be found
+[here][pytest-selenium].
+
+## Uploading results to TestRail
+
+The TestRail integration is currently intended to be used during a local test run of the cnx-automation suite when the uploading of results to TestRail is desired.
```
TomWoodward: rex-web

TomWoodward (commented):
what is there so far of the page structure and framework looks really good. i think all my concerns were about merging the py and ts project configs at the top level of the repo. it would make me feel better to move the python specific configs into a subdirectory

.gitignore
m1yag1: I can put it in my global gitignore. no biggie.

.gitignore
m1yag1: Yea, again, it's the cause of some copy pasta.. I'll remove it.

pyproject.toml
m1yag1: Sure, I explored this at first but there were a few things that looked to the root directory for configs (the big one is black which is the code formatter for python) and the work arounds were like, "create a bash script that automatically adds the --config-file /some/path/to/config". My main idea was to go for easy to run w/o having someone remember a bunch of command line flags for config files.

Additionally, I've worked in a number of projects that use python and js. I don't think I've ever been like, "let me shove the node config stuff into another folder b/c it's not python." The root allows one to see the files necessary to run the project. However, I do understand the need to keep the root clean and only have files necessary for the project's runtime. An example of python and js living happily together is here: https://github.com/openstax/research-eGrader

I'll try to get things working where python is hidden. That may require more time investment than I have this week so I may not finish this soon as next week I have other things I need to work on.

testrail.example.cfg
m1yag1: Hmm.. I don't think the pytest-testrail plugin uses env vars. I'll have to double check.

pyproject.toml
m1yag1: oh hypothesis is another example that has js stuff in the root along with the python: https://github.com/hypothesis/h

.gitignore
m1yag1: Also, .DS_Store is in there.. i had added it again on accident, but that would also apply to your comment too. That's a OSX specific thing and operating systems to develop are a personal choice too ;).

.gitignore
m1yag1: Ah, that's a good practice. I'll have to start doing that with the other projects we have. 👍

.pre-commit-config.yaml
m1yag1: I really like the use of pre-commit hooks. I think they are a great way to run the checks as you commit instead of waiting for some other time. Even though I've had linters run based on the test command I always seem to break CI for some dumb linting error. Now that I use pre-commit hooks it doesn't happen anymore.

Makefile-py
m1yag1: My thought was that if other Make commands were introduced you could create one Make file that combines them all. It can be useful to separate out the docker specific Make stuff using something like `make -f Makefile-docker` but also combine them with a core Makefile and use `include Makefile-docker`. I thought the separation might help but maybe not. 🤷‍♂️

testrail.example.cfg
m1yag1: Yea, pytest-rail out of the box doesn't support environment variables. I'll need to add those.

Makefile-py
TomWoodward: having sub-makefiles and composing them sounds like a cool pattern, in that case could we move the Makefile-py into a python subdirectory? or would that mess with the cwd of the scripts and be confusing? i'm not _opposed_ to having stuff in the top level, i'd just like to put things in subdirectories that can be.

pyproject.toml
TomWoodward: yea, again, not opposed. the idea is that we would be working in the py tests regularly (for almost every feature developed probably) and i'm not trying to make the py code second class, even for the ts files i push them down the tree as much as i can

testrail.example.cfg
TomWoodward: ok, in CI we can specify them via command line, that'll be easier

m1yag1: 👍

Makefile-py
m1yag1: yes we could move the py Makefile to a subdirectory.

tests/ui/test_toc.py
```diff
@@ -0,0 +1,35 @@
+from tests.ui.pages.content import Content
+from tests.ui import markers
+
+
+@markers.parametrize("book_slug,page_slug", [("college-physics", "preface")])
+@markers.nondestructive
+def test_toc_toggle_button_opens_and_closes(selenium, base_url, book_slug, page_slug):
```
m1yag1: @TomWoodward Here I added a basic test of clicking the TOC toggle button. This test currently passes for both desktop and mobile resolutions. However, in the mobile view it really covers up the screen so I would suggest that the mobile table of contents be hidden by default.

Makefile-py
m1yag1: I moved it. This is resolved.

pyproject.toml
m1yag1: I was able to move most of the files into the tests directory. The only one I wasn't able to is the config for pre-commit.

testrail.example.cfg
m1yag1: I removed the file and added instructions in the readme to create a file if someone wanted to run it locally and create their own run.

tests/ui/README.md
m1yag1: Fixed!

tests/ui/test_toc.py
TomWoodward: i believe this is the plan you can see the upcoming new designs here openstax/unified-issues#104

tests/ui/pages/base.py
```diff
@@ -0,0 +1,84 @@
...
+    def get_window_size(self, dimension=None):
+        """Return the current window dimensions."""
+        get_size = self.driver.get_window_size()
+        if not dimension:
+            return get_size
+        if dimension not in get_size:
+            raise IndexError("Unknown dimension: {}".format(dimension))
+        return get_size[dimension]
+
+    def set_window_size(self, width=0, height=0):
```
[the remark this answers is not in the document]
TomWoodward: in javascript i would create a util decorator function that did the size guard and sleep that you just pass a function reference for `set_window_size` or `set_window_position` into.

i'm not sure if there is a comparably easy thing to do in python, but these two blocks of code are nearly identical.

tests/ui/pages/base.py
```diff
@@ -0,0 +1,84 @@
...
+        """Attempt to change the browser window size."""
+        if width >= 1 and height >= 1:
+            self.driver.set_window_size(width, height)
+            sleep(1.0)
```
TomWoodward: is there no way to check the window size to verify this worked instead of arbitrarily sleeping?

tests/ui/pages/content.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Content(Page):
+    URL_TEMPLATE = "/books/{book_slug}/pages/{page_slug}"
+
+    _main_content_locator = (
+        By.CSS_SELECTOR,
+        '#main-content > div > div > div h1[data-type="document-title"]',
+    )
+    _table_of_contents_button_locator = (By.CSS_SELECTOR, "#root > div > div > button")
+    _table_of_contents_nav_locator = (By.CSS_SELECTOR, "#root > div > div > div > div > nav")
```
TomWoodward: @philschatz should weigh in but i think we should just sprinkle some `aria-label`s on these to select them by

tests/ui/pages/content.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def loaded(self):
+        return self.find_element(*self._main_content_locator)
```
TomWoodward: we should really use the js hook that the app provides, that is the canonical way that we're identifying when the page is done doing _stuff_

tests/ui/regions/base.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        return self.page.scroll_down()
+
+    def scroll_up(self):
+        """Scrolls using page up once. Returns the active element.
+
+        """
+        return self.page.scroll_up()
+
+    def scroll_to(self, element=None):
```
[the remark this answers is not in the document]
TomWoodward: similar to the other situation, i'd consider changing this to `return self.page.scroll_to(element or self.root)` or `return self.page.scroll_to(defaultElement(element))`

TomWoodward (changes requested):

tests/ui/pages/content.py
m1yag1: I agree. They didn't exist so this is what i did to get the tests to identify the right elements.

tests/ui/pages/content.py
TomWoodward: when we run these tests against a built environment the page will be complete before the js is even loaded, so then this code would thought the page was done loading but all interactions would fail

tests/ui/pages/base.py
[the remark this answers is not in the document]
m1yag1: I removed this as it's probably better to implement when we need it.

tests/ui/pages/base.py
m1yag1: I removed this code like the previous code. This was done for some reason like it b/c of something with cnx-automation. we probably don't need to bring that code over unless the specifics of the test are similar.

tests/ui/pages/content.py
m1yag1: Could you provide me with some js code that would allow me to resolve the js hook. It would look something like this.

https://github.com/openstax/cnx-automation/blob/try-applitools/tests/webview/ui/test_applitools.py#L25-L27

I have no idea what the return on something like that is.. i would need to experiment.

TomWoodward:
@m1yag1 when i try to run `pytest --base-url http://localhost:3000 tests/ui` i get 

```bash
============================= test session starts ==============================
platform darwin -- Python 3.7.0, pytest-4.1.1, py-1.7.0, pluggy-0.8.1 -- /Users/tomwoodward/code/openstax/books-web/.venv/bin/python3
cachedir: .pytest_cache
sensitiveurl: ^(?:https?\:\/\/)?openstax\.org
metadata: {'Python': '3.7.0', 'Platform': 'Darwin-17.7.0-x86_64-i386-64bit', 'Packages': {'pytest': '4.1.1', 'py': '1.7.0', 'pluggy': '0.8.1'}, 'Plugins': {'variables': '1.7.1', 'selenium': '1.15.1', 'metadata': '1.8.0', 'html': '1.20.0', 'base-url': '1.4.1'}, 'Base URL': '', 'Driver': None, 'Capabilities': {}}
rootdir: /Users/tomwoodward/code/openstax/books-web/tests/ui, inifile: setup.cfg
plugins: variables-1.7.1, selenium-1.15.1, metadata-1.8.0, html-1.20.0, base-url-1.4.1
collecting ... collected 0 items / 1 errors

==================================== ERRORS ====================================
_________________________ ERROR collecting test_toc.py _________________________
ImportError while importing test module '/Users/tomwoodward/code/openstax/books-web/tests/ui/test_toc.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/ui/test_toc.py:1: in <module>
    from tests.ui.pages.content import Content
E   ModuleNotFoundError: No module named 'tests'
- generated html file: /Users/tomwoodward/code/openstax/books-web/ui-test.html -
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 0.17 seconds ============================
```

i haven't installed the webdriver stuff yet, but this doesn't look related i don't think...

README.md
```diff
@@ -97,14 +97,13 @@ this suite:
 
 ### Selenium Tests
 
-this suite has not been implemented yet.
+run these with `pytest --base-url http://localhost:3000 tests/ui`
```
TomWoodward: all of our other tests are distributed in the js code next to the stuff that they're testing. we have a /src/test place for making testing utilities, mocks, etc.

because of this, i don't think we're ever going to want to put anything else in the top level /tests folder, and having it wrap the single ui folder seems a little weird. can we rename the whole thing to /selenium or /ui-tests or something like that?

TomWoodward (changes requested):

README.md
m1yag1: I figured this would come up and just picked tests as that was standard for python. I also wasn't sure if we would employ python to do some other kind of testing like integration tests... so i split it up like that... thinking it may be `tests(or something else)/ui`,`tests/integration`, `tests/visual`

maybe just name the root folder to `pytest`?

m1yag1: I based it off what mozilla had. https://github.com/mozilla/addons-server/tree/master/tests

TomWoodward: 👍 for root folder `pytest`, if you think its prudent to make a subfolder below that go for it

Makefile
```diff
@@ -0,0 +1,14 @@
+.PHONY: help pre-commit clean clean-build clean-pyc clean-test venv
```
TomWoodward: this line is also in the py makefile, does it need to have the same commands in both places?

pytest-selenium/README.md
```diff
@@ -0,0 +1,151 @@
...
+The pytest plugin that we use for running tests has a number of advanced
+command line options available. To see the options available, run
+`pytest --help`. The full documentation for the plugin can be found
+[here][pytest-selenium].
+
+## Uploading results to TestRail
+
+The TestRail integration is currently intended to be used during a local test run of the rex-web pytest suite when the uploading of results to TestRail is desired.
+
+Make a copy of of the testrail.example.cfg:
```
TomWoodward: this doesn't exist

pytest-selenium/tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+
+import pytest
+
+# Window resolutions
+DESKTOP = (1080, 1920)
+MOBILE = (414, 738)
+
+
+@pytest.fixture(
+    scope="function", params=[DESKTOP, MOBILE], ids=["Resolution: 1080x1920", "Resolution: 414x738"]
```
TomWoodward: resolutions duplicated here and above?

TomWoodward (changes requested):

Makefile
m1yag1: i believe so. if someone were to do `make -f pytest-selenium/Makefile-py`

m1yag1: However, that is not likely so i'll remove it ;)

pytest-selenium/tests/conftest.py
m1yag1: The ids are mainly for display when the tests are running. It's not actually duplicating them but rather adding a label.

m1yag1: Example:
image

TomWoodward (changes requested):

pytest-selenium/tests/conftest.py
TomWoodward: but why not use the constants in these strings?

README.md
```diff
@@ -97,14 +97,13 @@ this suite:
 
 ### Selenium Tests
 
-this suite has not been implemented yet.
+run these with `pytest --base-url http://localhost:3000 --driver pytest-selenium/tests`
```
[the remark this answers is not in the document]
m1yag1: Haha, yea you're right

pytest-selenium/README.md
```diff
@@ -0,0 +1,151 @@
...
+#### Activate the virtualenv
+
+    $ source .venv/bin/activate
+
+## Run the tests using Pytest
+
+To run a specific test, pass in a value for `-k`:
+
+```bash
+$ pytest -k test_my_feature tests/ui
```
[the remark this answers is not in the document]
m1yag1: Yes, a lot of these have changed with the new changes. I'll correct them.

pytest-selenium/tests/conftest.py
m1yag1: I did

README.md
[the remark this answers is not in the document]
m1yag1: yea sorry.. I typically pass in Chrome as a default. Since we have a focus on cross browser it's something more explicit. I'm not used to it having to be passed in manually.

TomWoodward (dismissed):
this is working both locally and through saucelabs and can post the results back to testrail. i'll make some additional changes when i try to automate things more (the `make test` command is mostly just note taking for me) but i think we should get this merged down! :shipit:

philschatz (changes requested):
This is just a review of the precommit hooks, `Makefile`, and `README.md` (awesome to read through!) but will submit another one for the actual code

pytest-selenium/pages/content.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class Content(Page):
+    URL_TEMPLATE = "/books/{book_slug}/pages/{page_slug}"
+
+    _main_content_locator = (
+        By.CSS_SELECTOR,
+        '#main-content > div > div > div h1[data-type="document-title"]',
+    )
+    _table_of_contents_button_locator = (By.CSS_SELECTOR, "#root > div > div > button")
```
philschatz:
```suggestion
    _table_of_contents_button_locator = (By.CSS_SELECTOR, "[aria-label='Click to close the Table of Contents']")
```

Relying on the `[aria-label]` rather than a fragile selector helps test accessibility _and_ is less fragile than a `>` selector

pytest-selenium/pages/content.py
```diff
@@ -0,0 +1,45 @@
...
+from pages.base import Page
+from regions.base import Region
+
+
+class Content(Page):
+    URL_TEMPLATE = "/books/{book_slug}/pages/{page_slug}"
+
+    _main_content_locator = (
+        By.CSS_SELECTOR,
+        '#main-content > div > div > div h1[data-type="document-title"]',
```
philschatz: What are some reasons for using very-specific selectors in the Selenium tests?

```suggestion
        'h1',
```

If the WCAG guidelines are followed (I think we do so far) there should _only_ be one `h1` in the whole document. And if we don't follow the guidelines then something should blow up anyway 😄

pytest-selenium/pages/content.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    _table_of_contents_button_locator = (By.CSS_SELECTOR, "#root > div > div > button")
+    _table_of_contents_nav_locator = (By.CSS_SELECTOR, "#root > div > div > div > div > nav")
+
+    @property
+    def loaded(self):
+        return self.find_element(*self._main_content_locator)
```
philschatz: Does `self.find_element` throw an error when the locator matches `0` or `> 1` elements? If not, would it make sense to add a help that does that?

pytest-selenium/pages/content.py
```diff
@@ -0,0 +1,45 @@
...
+
+class Content(Page):
+    URL_TEMPLATE = "/books/{book_slug}/pages/{page_slug}"
+
+    _main_content_locator = (
+        By.CSS_SELECTOR,
+        '#main-content > div > div > div h1[data-type="document-title"]',
+    )
+    _table_of_contents_button_locator = (By.CSS_SELECTOR, "#root > div > div > button")
+    _table_of_contents_nav_locator = (By.CSS_SELECTOR, "#root > div > div > div > div > nav")
```
philschatz:
```suggestion
    _table_of_contents_nav_locator = (By.CSS_SELECTOR, "nav")
```

.pre-commit-config.yaml
TomWoodward: i would definitely not want a pre-commit hook to run tests because i don't want to wait for them every time i commit. the linter is pretty fast though. i like to have control over when my code gets re-formatted and would not love to have it happen automatically on commit

pytest-selenium/pages/content.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            self.table_of_contents_toggle_button.click()
+        return self.table_of_contents.wait_for_region_to_display()
+
+    @property
+    def table_of_contents(self):
+        return self.TableOfContents(self)
+
+    class TableOfContents(Region):
+        _root_locator = (By.CSS_SELECTOR, "#root > div > div > div > div > nav")
```
philschatz: Is this different than the `_table_of_contents_nav_locator` ? If not, can they be combined?

pytest-selenium/setup.cfg
```diff
@@ -0,0 +1,3 @@
+[tool:pytest]
+addopts=-vs --tb=long --showlocals --html ui-test.html --self-contained-html
+sensitive_url=^(?:https?\:\/\/)?openstax\.org
```
philschatz: What is this field used for? I don't see it used anywhere directly... maybe it is used by an included python package?

pytest-selenium/pages/base.py
```diff
@@ -0,0 +1,24 @@
+from time import sleep
+
+import pypom
+from selenium.webdriver.common.keys import Keys
+
+
+class Page(pypom.Page):
```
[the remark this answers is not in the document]
TomWoodward: the reason we're using python in the first place is because mike has already solved this problem using python and doesn't have to re-invent everything to set up rex in the same way. down the line once we get things working i'd consider porting the suite over to typescript. in the meantime i don't think its worth going too far out of our way to modify the base pytest framework (like adding types)

pytest-selenium/tests/conftest.py
```diff
@@ -0,0 +1,79 @@
+import os
+
+import pytest
+
+# Window resolutions
+DESKTOP = (1080, 1920)
```
philschatz: Is there a reason these are defined as `height,width` here but they are used in both the code and the docs as `width,height`?

pytest-selenium/tests/conftest.py
```diff
@@ -0,0 +1,79 @@
...
+
+    This fixture is also important to add options to the command line that
+    support running the Chrome WebDriver in Travis or other languages
+
+    """
+    if pytestconfig.getoption("--headless"):
+        chrome_options.headless = True
+
+    # Required to run in Travis containers
+    if pytestconfig.getoption("--no-sandbox"):
```
philschatz: Could these be conditional on running in a container (environment `CI=true` or `TRAVIS=true`)? Once we start storing cookies (logged-in, whether popups were dismissed, etc) it might cause tests to fail unexpectedly.

pytest-selenium/tests/test_toc.py
```diff
@@ -0,0 +1,23 @@
+from pages.content import Content
+from . import markers
+
+@markers.test_case('C250849')
+@markers.parametrize("book_slug,page_slug", [("college-physics", "preface")])
+@markers.nondestructive
+def test_toc_toggle_button_opens_and_closes(selenium, base_url, book_slug, page_slug):
```
philschatz: ❤️ The whole test is so sooo concise and readable 😄

README.md
```diff
@@ -97,14 +97,13 @@ this suite:
 
 ### Selenium Tests
 
-this suite has not been implemented yet.
+run these with `pytest --base-url http://localhost:3000 --driver Chrome pytest-selenium/tests`
```
philschatz:
```suggestion
make venv
source ./.venv/bin/activate
run these with `pytest --base-url http://localhost:3000 --driver Chrome pytest-selenium/tests`
```

This _partially_ worked for me but I got the following error (do we need separate instructions for installing chromedriver?):

```
selenium.common.exceptions.WebDriverException: Message: 'chromedriver' executable needs to be in PATH.
```

Maybe it would make sense to add a Makefile target instead? Like `make test-local` or something?

philschatz (changes requested):

.pre-commit-config.yaml
m1yag1: yea, i don't run tests.. i don't like that. only linting checks on the code i'm committing. We can just use this for the python stuff. I can move it to the python Makefile.

pytest-selenium/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+The TestRail integration is currently intended to be used during a local test run of the rex-web pytest suite when the uploading of results to TestRail is desired.
+
+Create a testrail.cfg file:
+
+    $ touch testrail.cfg
+
+Add these configs with the appropriate values:
+
+    [API]
```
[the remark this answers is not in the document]
m1yag1: Most of the time i believe this will be ran by automation. These instructions are here in the event if someone wants to do that locally which may not be that often.

pytest-selenium/pages/base.py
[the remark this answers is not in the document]
m1yag1: it will be some time before I can go through and add all the type hints.

pytest-selenium/pages/content.py
m1yag1: Did you test this change?

pytest-selenium/tests/conftest.py
m1yag1: Selenium uses these backwards from the way we typically see them. I've adjusted them in the documentation so that they use the typical order.

pytest-selenium/setup.cfg
m1yag1: These are defaults passed into the pytest command. You run pytest --driver Chrome pytest-selenium/tests and these are inserted in there so you don't have to type them out.

pytest-selenium/tests/conftest.py
m1yag1: Sure, it might but I'd rather not make changes on what might happen. This is basically copied from our other automation framework that does run in travis.

pytest-selenium/pages/content.py
philschatz: eep, there are 36 `<nav>` elements. I think that's a bug... there should only be one.

pytest-selenium/setup.cfg
philschatz: ah ok. I was just curious about the purpose of `sensitive_url` but thanks for enumerating the others

pytest-selenium/pages/content.py
philschatz: This will probably be fixed to be just one `<nav>` as part of the "Add ToC Toggle" story in "Reading Experience": https://docs.google.com/document/d/1-KX5NMSeJTP5309L7mlXw85YHG_7XRFm-C7ofIdWtCE/edit#heading=h.apfasqkrfa3g

TomWoodward: @philschatz the nav element is used for navs and subnavs, so each chapter has its own nav. i don't know of any reason to limit to just 1 nav per page

pytest-selenium/tests/conftest.py
m1yag1: Actually it might be that pytest injects them in the opposite order.

pytest-selenium/pages/content.py
philschatz: From the [MDN Description of nav](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) it seems like they are used to denote separate navigations on a page (e.g. ToC, site navigation) and are not nested. In the epub spec the entire ToC is a `<nav>` which contains nested `<ul>`, `<li>`, and `<a href="">` elements.

TomWoodward: ok, we can definitely change it to only be one nav for the ToC, we may end up having another one for site nav though so we should assume there are at least a couple

philschatz: Yeppers, and when a 2nd `<nav>` is introduced they can be differentiated via `aria-label` attributes so the selenium tests _and_ screenreaders can differentiate 😄

philschatz (dismissed):

philschatz:
@m1yag1 Do you want to make any comments before this gets merged? I pushed a few commits addressing my review comments yesterday but forgot to ask you to review

philschatz (dismissed):

README.md
```diff
@@ -97,14 +97,13 @@ this suite:
 
 ### Selenium Tests
 
-this suite has not been implemented yet.
+run these with `make test-local`
```
TomWoodward: this command doesn't seem to exist

TomWoodward (dismissed):

m1yag1:
I did a merge of Phil’s code but had a conflict. I thought i put that in there. ugh. i think i messed something up.

TomWoodward (dismissed):

pytest-selenium/pages/content.py
philschatz: Even when the locator matches `> 1` elements? I'm hoping that selenium can:

- help ensure there is exactly 1 `h1` on the page
- explicitly fail when more than 1 element matches a selector

TomWoodward (dismissed):

philschatz (dismissed):
I just fixed a link and clarified that `make test` does not work by printing out the help & erroring so people know to run `make test-local` or `make test-rail`

pytest-selenium/pages/content.py
m1yag1: We would need to write a test for that. Selenium doesn't enforce any standards like that.

m1yag1: What could be automated is some sort of HTML validation. I don't know any tools that do that off the top of my head but i'm sure something could be integrating. It would be something very similar to an aXe report. We would have it fail when there were any failures or some threshold.

m1yag1: Something like this seems like it would do what you want. https://validator.w3.org/docs/api.html

TomWoodward (approved):

philschatz (approved):

philschatz:
oooh, pyyaml has a security vulnerability & should be upgraded to `pyyaml>=4.2b1`. https://github.com/openstax/rex-web/network/alerts

pytest-selenium/pages/content.py
philschatz: (just bike-shedding)

Oh, I guess that I was trying to say that if the test-writer expects the selector to match one element (by using `find_element` rather than `find_elements`) then it seems like that assumption should throw an error if more than one element is matched. Otherwise, `find_element` really means `find_first_element`.

Once I contribute a test that uses `find_element` I might refactor all of the `find_element` references to be something like `util.find_element(self, ...)` or something.

m1yag1:
Find element returns the first one it finds. If you want more than one you have to use find_elements. I do see where it would be useful.
  

  
>

PR #69, opened 2019-01-18
feat: prompt to use releases

TomWoodward:
i didn't want to get too lost in creating a styling framework, i did add a place to put colors, and implemented that in the button components, but we'll have to clean up the rest as we establish more patterns.

also:
- added parsed query params to location state
- add a way to initialize redux state via query param

https://rex-web-pr-69.herokuapp.com/books/college-physics/pages/preface?initialState={%22notifications%22:[{%22type%22:%22Notification/updateAvailable%22}]}

# Desktop Screen

image

---

# Mobile Screen

image


src/helpers/pollUpdates.ts
```diff
@@ -0,0 +1,37 @@
...
+  document.addEventListener('visibilitychange', visibilityListener);
+
+  const cancel = () => {
+    document.removeEventListener('visibilitychange', visibilityListener);
+    clearInterval(interval);
+  };
+
+  async function poll() {
+    const release = await fetch('/rex/release.json')
+      .then((response) => response.json());
```
philschatz: Will/Should this quietly fail if the computer is not connected to the internet? Or if, for some reason the file is not available?

src/helpers/pollUpdates.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    if (release.id !== RELEASE_ID) {
+      cancel();
+      store.dispatch(updateAvailable());
+    }
+  }
+
+  activateInterval();
```
philschatz: Should polling stop once the browser knows it is out-of-date?

philschatz (commented):
Nice use of the query string to provide the initial state 😄 . Just two questions

src/helpers/pollUpdates.ts
TomWoodward: on line 29 it cancels the polling after recognizing the new id

src/helpers/pollUpdates.ts
TomWoodward: yea i'm ashamed to admit that i actually had that in there before but removed it when it was totally impossible to test... i'll try again

src/helpers/pollUpdates.ts
philschatz: D'oh, just saw that. thanks!

src/helpers/pollUpdates.ts
```diff
@@ -2,36 +2,40 @@ import { updateAvailable } from '../app/notifications/actions';
 import { Store } from '../app/types';
 import { APP_ENV, RELEASE_ID } from '../config';
 
+export type Cancel = () => void;
+
+export const poll = (store: Store, cancel: Cancel) => async() => {
+  const release = await fetch('/rex/release.json')
+    .then((response) => response.json())
```
philschatz: What happens when a non-200 response is returned? Oh, `response.json()` will throw an error & be caught by `.catch(...)`. ok 😄

philschatz (dismissed):

src/app/theme.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const color = {
+  neutral: {
+    base: '#fff',
+    darker: '#f1f1f1',
+    darkest: '#e5e5e5',
+    foreground: textColors.default,
+  },
+  primary: {
```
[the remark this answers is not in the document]
TomWoodward: i'm leaving these because the theme is still incomplete and i'm not sure what the best way to dry it up will be yet

snm8:
I just checked it. Looks good on desktop. But it doesn't change design when I made the window as mobile (and when I opened it on my mobile phone). Not sure if it is because it is on herokuapp...it needs to be more horizontal on mobile as in the mock-up

TomWoodward:
@snm8 is right, mobile is broken, i will revise

TomWoodward:
@snm8 i added the mobile view, i had less visual guidance on this than the larger view so i don't think it looks as nice, but its ok. let me know if you have any feedback.

snm8:
@TomWoodward looks good on mobile too. thank you!

src/app/notifications/components/Notifications.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { getType } from 'typesafe-actions';
+import theme from '../../theme';
+import { AppState } from '../../types';
+import * as actions from '../actions';
+import * as select from '../selectors';
+import { AnyNotification } from '../types';
+import UpdatesAvailable from './UpdatesAvailable';
+
+// tslint:disable-next-line:variable-name
+const Container = styled.div`
```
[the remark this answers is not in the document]
TomWoodward: i think this is only being flagged because of the stupid comment that i have to put on every component. really need to figure out how to not do that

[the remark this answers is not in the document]
philschatz: I'm fine with removing the linter rule completely 😄  I haven't noticed us break this rule before. Or we could add that the following are allowed:

- `lowerCamel`
- `UpperCamel`
- `UPPER_SNAKE` (not sure if we use this one)

philschatz (approved):

kjdav:
Looks good.

PR #71, opened 2019-02-01
fix: links to be relative

philschatz:
This addresses an issue brought up in release testing. links used to be `/books/{book_name}/pages/{page_name}`.

Now they are:

- `{page_name}` when in the same book
- `../{book_name}/pages/{page_name}` when in a different book
- see snapshot file for how other (unused) cases are handled


src/app/content/components/ContentLink.tsx
```diff
@@ -14,17 +16,18 @@ interface Props extends React.HTMLProps<HTMLAnchorElement> {
...
 // tslint:disable-next-line:variable-name
-export const ContentLink: SFC<Props> = ({book, page, navigate, ...props}) => {
+export const ContentLink: SFC<Props> = ({book, page, currentPath, navigate, ...props}) => {
   const params = {
     book: book.slug,
     page: getUrlParamForPageId(book, page.shortId),
   };
 
-  const url = content.getUrl(params);
+  const url = relative(dirname(currentPath), content.getUrl(params)) || `../${basename(params.page)}`;
```
TomWoodward: i assume this `||` is to catch the case where `If from and to each resolve to the same path (after calling path.resolve() on each), a zero-length string is returned.` but i don't see any test cases for the requested and current path being the same

TomWoodward: also i think it might be technically correct to leave such a case as an empty href....

TomWoodward (changes requested):

src/app/content/components/ContentLink.tsx
philschatz: The case occurs when the current url is something like `/books/A/pages/B/C`.

In this case we could maybe get away with `""` but:

- I'm not sure if browser handle that case properly
- this would mean that the target URL would be `/books/A/pages/B/` rather than `/books/A/pages/B` (no trailing slash) and I don't know if amazon handles that

TomWoodward: interesting... does it also happen when the requested and current path are the same? i don't see a test for that case.

also this is inconsistent with the "deeply nested" one that just returns ".."

philschatz: I think I addressed the first concern in the recent commit. The 2nd concern is unused... fixing it would likely add more code... should we handle it?

src/app/content/__snapshots__/utils.spec.ts.snap
```diff
@@ -0,0 +1,21 @@
...
+
+exports[`toRelativeUrl when in a different book 1`] = `"../../book1/pages/page1"`;
+
+exports[`toRelativeUrl when in the same book 1`] = `"page1"`;
+
+exports[`toRelativeUrl when not in a book and not at the root (unused) 1`] = `"../books/book1/pages/page1"`;
+
+exports[`toRelativeUrl when not in a book and not at the root (unused) 2`] = `"../books/book1/pages/page1"`;
+
+exports[`toRelativeUrl when the current path ends in a / (unused) 1`] = `"../page1"`;
```
TomWoodward: this doesn't even work on cloudfront so we can remove it

src/app/content/__snapshots__/utils.spec.ts.snap
```diff
@@ -0,0 +1,21 @@
...
+
+exports[`toRelativeUrl when the current path ends in a / but is not the current page (unused) 1`] = `"../page1"`;
```
TomWoodward: and this

src/app/content/__snapshots__/utils.spec.ts.snap
```diff
@@ -0,0 +1,21 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`toRelativeUrl when at the root (unused) 1`] = `"books/book1/pages/page1"`;
+
+exports[`toRelativeUrl when deeply under the same Page (unused) 1`] = `".."`;
```
TomWoodward: i still don't like that this is just `..` when "toRelativeUrl when under the same Page" is `../page1`

TomWoodward (changes requested):

src/app/content/__snapshots__/utils.spec.ts.snap
philschatz: Since we don't use it yet can this one be removed for now? I added a `throw new Error(...)` for the cases when `path.relative(...)` returns `''`

TomWoodward: i mean, i also don't love introducing a constraint about being not able to link to places from certain places. 

can we check to see if the basename of the result path is  `..` and append another `..` plus the basename of the target path?

TomWoodward (approved):

PR #73, opened 2019-02-06
chore: automate RC environments and regression tests

TomWoodward:
what this does is:

* create unified environment (via unified-deployment scripts) for milestones created in openstax/unified
* when rex-web master is changed:
  * bundle and pre-render the release using books defined in the codebase, name the release `master/<sha>`
  * push it to s3
  * update any environments for milestones that aren't closed with this release
  * find test suites in testrail for environments that were updated and run their test cases in a variety of browsers through saucelabs against the release environment

the pipeline:
image

the overall pipeline is configured in unified-deployment, and deploymenty scripts are there too. the changes here are just for the build/testing scripts that are rex specific

links to cool places:
the pipeline: https://concourse.openstax.org/teams/Dev/pipelines/unified
current testrail plan: https://openstax.testrail.com/index.php?/plans/view/750
unified-deployment pr: https://github.com/openstax/unified-deployment/pull/9


m1yag1 (dismissed):

concourse/create-master-build-configs/script.js
```diff
@@ -0,0 +1,37 @@
...
+  console.log(JSON.stringify(args, null, 2));
+  fs.writeFile(argFile, JSON.stringify(args), handleErr);
+
+  console.log(`Generating commit file with: ${commit}`);
+  fs.writeFile(commitFile, commit, handleErr);
+
+  console.log(`Generating release id file with: ${releaseId}`);
+  fs.writeFile(releaseFile, releaseId, handleErr);
+
+  console.log(`Generating tag file with: ${tag}`);
```
philschatz: Why is the tag file necessary? It seems like a tag is generated _after_ a commit is created and a tag might never even end up on `master` (e.g. a hotfix but merging it into master might squash the actual commit sha that was on production)

concourse/create-master-build-configs/task.yml
```diff
@@ -0,0 +1,15 @@
...
+  source:
+    repository: node
+
+inputs:
+- name: rex-web 
+outputs:
+- name: build-configs
+
+run:
+  path: rex-web/concourse/create-master-build-configs/script.js
```
philschatz: Is it possible to just use `./script.js` (in case any parent directories are renamed)

Hmm, concourse probably does not though. But maybe worth checking?

concourse/execute-test-runs/script.bash
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -e -u
```
philschatz: Would it be a good time to introduce a shell script linter?

[cnx-recipes](https://github.com/openstax/cnx-recipes) uses [shellcheck](https://github.com/koalaman/shellcheck) which is installed on travis containers

concourse/create-master-build-configs/script.js
TomWoodward: this is the docker tag

concourse/create-master-build-configs/task.yml
TomWoodward: no, the working directory will be the build folder, this is just the config file it loads

concourse/execute-test-runs/script.bash
TomWoodward: sure, i'm not super familiar with shell scripting best practices, any tools would be helpful.

in general i've been choosing whatever is the easiest language for these scripts, i have some bash, some js, and some ruby. i like that concourse lets you easily choose the best tool for the job but it does make code convention a little messy... open to thoughts...

concourse/execute-test-runs/script.bash
```diff
@@ -5,20 +5,22 @@ BASE_DIR=$(pwd)
...
 
   echo "checking for runs using $BASE_URL"
 
-  for TR_RUN_ID in $(ls $BASE_DIR/test-plans)
+  for TR_RUN_ID in "$BASE_DIR"/test-plans/*
   do
-    if [ "$BASE_URL" == $(< $BASE_DIR/test-plans/$TR_RUN_ID/base_url.txt) ]; then
+    browser=$(< "$BASE_DIR/test-plans/$TR_RUN_ID/browser.txt")
+    
+    if [[ "$BASE_URL" == "$browser" ]]; then
```
TomWoodward: this isn't the same conditional

script/test.bash
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+yarn test || exit 111
```
TomWoodward: why run `yarn test` from here instead of running this from `yarn test`?

philschatz: It felt a little weird to run shell scripts from `yarn test` but we already run `build-simulations.bash` from it so I can move it there

philschatz (approved):

concourse/execute-test-runs/script.bash
```diff
@@ -5,20 +5,22 @@ BASE_DIR=$(pwd)
 
 cd rex-web
 
-for MILESTONE_NUMBER in $(ls $BASE_DIR/unified-issues-milestones)
+for MILESTONE_NUMBER in "$BASE_DIR"/unified-issues-milestones/*
```
TomWoodward: @philschatz this internal variable is different after this change. previously it was just the directory name, with the glob now its getting the full path.

why did you change this?

philschatz: Thanks for testing (or checking the concourse build)! It is probably wrong but I changed it because of the shellcheck message [SC2045: Iterating over ls output is fragile. Use globs](https://github.com/koalaman/shellcheck/wiki/SC2045)

I can add a `# shellcheck disable=SC2045` if there is no good way to write it properly.

https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-one-specific-instance-in-a-file

TomWoodward: is there a way to get the glob to only return the directory name instead of the whole path?

philschatz: hmm, my shell-fu is not great but `basename ${MILESTONE_NUMBER}` might work. How can I test it?

TomWoodward: https://stackoverflow.com/questions/2664740/extract-file-basename-without-path-and-extension-in-bash might also be relevant to our interests. its hard to test, if you push to this branch concourse will run and "run test plans" should turn green here https://concourse.openstax.org/teams/Dev/pipelines/unified

there is a way to test the job before pushing with `fly execute` but i haven't totally figured out the workflow especially when you have lots of inputs

philschatz: Thanks! I'll try to make the fixes now that I have a way to verify they work (I forgot that concourse was already running 😊 ).

Also, I got feedback from https://stackoverflow.com/users/26428/dennis-williamson ( @dtwilliamson ) a while back that local variables should be `under_snake` case so they are visually distinct from `ENVIRONMENT_VARS`

TomWoodward: that makes sense on variable naming

concourse/execute-test-runs/script.bash
```diff
@@ -5,15 +5,17 @@ BASE_DIR=$(pwd)
...
   ENV_NAME=release-$MILESTONE_NUMBER
   export BASE_URL=https://$ENV_NAME.sandbox.openstax.org
 
   echo "checking for runs using $BASE_URL"
 
-  for TR_RUN_ID in "$BASE_DIR"/test-plans/*
+  for TR_RUN_PATH in "$BASE_DIR"/test-plans/*
   do
+    TR_RUN_ID=$(basename "$TR_RUN_PATH")
     base_url=$(< "$BASE_DIR/test-plans/$TR_RUN_ID/base_url.txt")
```
TomWoodward: you could use the TR_RUN_PATH here for less string replacing

TomWoodward:
@philschatz your changes are working on concourse.openstax.org, should we wait for @jpslav to review before merging?

philschatz:
👍 for merging now and we can do a concourse discussion when @jpslav gets back from vacation if he wants

PR #74, opened 2019-02-07
chore: Clean up clean up everybody do your share

TomWoodward:
upgrade typescript and fix some stuff in high curn + low maintainability sections of the code 

https://codeclimate.com/github/openstax/rex-web/trends/churn


src/app/content/components/Page.tsx
```diff
@@ -100,20 +101,18 @@ export class PageComponent extends Component<PropTypes> {
...
+    this.mapLinks((a) => a.addEventListener('click', this.clickListener(a)));
+  }
+
   private linksOff() {
-    if (this.container) {
-      Array.from(this.container.querySelectorAll('a')).forEach((a) =>
-        a.removeEventListener('click', this.clickListener(a))
-      );
-    }
+    this.mapLinks((a) => a.removeEventListener('click', this.clickListener(a)));
```
philschatz: Does this work? It seems like `this.clickListener(a)` always returns a new function. How does `a.removeListener(` know to remove it?

TomWoodward: yea i mean, this is the same way it was before, but i see what you mean. the listeners will be garbage collected with the element though so maybe we don't even need to care about the `linksOff` stuff

philschatz: I think that explicitly turning them off is useful I just wished that I had noticed this problem earlier. I think that in order to do it properly we would need a private array of all the `clickListeners` and then remove them when `linksOff` is called, right?

TomWoodward: the problem is getting the anchor in scope, getting the anchor from the event is annoying so i was trying to provide it in a parent scope, but its looking like that may not be possible and maintain a reference to the handler function

TomWoodward: updated to store the anchor => handler mapping

philschatz (approved):

PR #75, opened 2019-02-07
chore: add body attr when page is loaded

TomWoodward:
only works on initial load, adds attribute when everything is done loading

image


m1yag1 (approved):

TomWoodward:
looks good to me. @philschatz can merge if he approces

m1yag1:
pinging @philschatz

src/index.tsx
```diff
@@ -51,6 +51,12 @@ app.services.fontCollector.handle((font) => {
   app.services.promiseCollector.add(loadFont(font));
 });
 
+app.services.promiseCollector.calm().then(() => {
+  if (typeof(document) !== 'undefined') {
+    document.body.setAttribute('data-rex-loaded', 'true');
```
philschatz: I've run into a similar problem for browser-testing a side-project of mine. It seems like a boolean flag is not sufficient information for selenium; a counter would be better. For example:

let's say Selenium does the following:

1. start by going to Section `1.1`
1. click the ToC to move to Section `3.1`

There is no way to know if step 2 has finished yet. Since the attribute is a boolean and even if we switched it to be `false` when loading and then `true` when done loading there is no guarantee that selenium will ever see the `false`.

However, running this pseudocode would work 

```js
lastLoaded = page.getAttribute('body', 'data-rex-loaded') || 0
page.waitUntil(`body[data-rex-loaded=${lastLoaded + 1}]`)
```

philschatz (changes requested):

src/index.tsx
TomWoodward: i agree with you, but its currently only used to tell when the initial load is done (right @m1yag1 ?) and for that purpose it is only checking for the presence of the attribute, not its value, so i think we can merge this as is and revise later. thoughts @philschatz @m1yag1 ?

m1yag1: That's true. We only really care that the attribute is there.

m1yag1: Also from my testing it works really well! I wish something like this existed for cnx.org (webview).

philschatz (approved):

PR #76, opened 2019-02-13
spike highlighting search results

TomWoodward:
findings:
- types for NodeFilter need to be updated to differentiate between the variable and the interface
- using openstax/highlighter as typescript module is broken, need to update that repo to include type definitions 
- searching for a single letter like "a" results in all the math being highlighted, probably need skip trying to highlight any text node that is inside math
- there is no real identifying information in the elasticsearch results, so this is doing a totally separate search of the current page.
  - this example parses the user input, if elasticsearch is doing some fancy logic, we might want to parse out the matched terms from the search result. (if elasticsearch is stemming words or something)
  - some effort will need to be made to line up the matching logic, for instance elasticsearch by default only matches on whole words, but my implementation would match a subword, leading to many more matches on the page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@openstax/highlighter": "^1.1.9",
     "cnx-recipes": "github:connexions/cnx-recipes#master",
     "history": "^4.7.2",
     "html-entities": "^1.2.1",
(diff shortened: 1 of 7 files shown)


TomWoodward:
@philschatz can staxly be updated to look for the WIP label in addition to the name prefix?

kjdav:
Great work, Tom. @snm8 things to note: Mathjax does weird stuff to the highlights, so we need to consider not highlighting any text node that is inside math... Also, Tom's current implementation returns results in a case sensitive way but we probably don't want that in the final release.

TomWoodward:
in general search matching stuff in math is something that won't work with our current approaches. if we decide we want to try to make it work that is a whole conversation starting back at how even you type math in the search field

PR #78, opened 2019-02-26
chore: add architecture overview

TomWoodward:
this could probably have more details, we can add more stuff as people are confused. i think this will help people get started at least a little better.


KarinaMendez2 (commented):
Looks great! Definitely helpful for starting the project, maybe you could add this part: go to "src/setupProxy.js"  and remove lines 64 - 67, that will stop it from forwarding your requests to osweb and it'll be easier to tell what is in rex

TomWoodward:
@KarinaMendez2 i had considered disabling the proxy by default in development, but as we add more navigation elements, it really might seem pretty broken without it. i'll add a configuration to turn off the osweb proxy and a note in the readme about enabling it if you're confused

philschatz (approved):
❤️

philschatz (commented):
I find the proxy-by-default confusing (not knowing what is in rex and what is not). It seems like it would be more useful once we start working on integration points between the two services.

I like the routes listing. Since rex does not have a root route, what if `/` showed the routes list _and_ links to all the configured books? Then, the `[View Deployment]` links in the PR would work for the most part and instead of adding links to the Pull Request body (like in #69) we could just add those links in `/`.

TomWoodward:
@philschatz the proxy will be necessary for any of the navigation between osweb and rex to work, of which we currently have none, but we will be adding some in the upcoming cards

TomWoodward:
@philschatz i moved the developer page to `/` and added books but they aren't linked yet because i want some of the helpers that are currently on other branches. i think we should merge this as is and revise later.

philschatz (approved):
yay, works for me!

KarinaMendez2 (approved):
Super helpful README!

PR #80, opened 2019-02-27
chore: make 1rem = 10px

TomWoodward:
make 1rem = 10px, update places we were already using rems, convert px to rem in the update notification, update tests to match, make a change to hopefully help test stability


src/test/browserutils.ts
```diff
@@ -48,14 +48,13 @@ export const navigate = async(target: puppeteer.Page, path: string) => {
 };
 
 export const finishRender = async(target: puppeteer.Page) => {
-  // wait for any new async hooks to register...
-  await new Promise((resolve) => setTimeout(resolve, 1000));
+  // HACK
+  // - there is no convenient way to tell if chrome is finished rendering fonts and things.
+  // - its hard to tell when simulations are done loading.
```
philschatz: can the `readyState` be polled? (Not sure it is worth adding in the complexity into production code but maybe tests can use it? Something like https://stackoverflow.com/a/36155560

Similarly for fonts, I think MathJax creates a `<span style="font-face: FancyFont">W</span>` and polls to check the width to see if the font has loaded. Not sure it is worth adding

philschatz (approved):

PR #81, opened 2019-02-27
feat: attribution

TomWoodward:
https://rex-web-pr-81.herokuapp.com/books/college-physics/pages/preface

PR #82, opened 2019-03-04
feat: reading experience headers

KarinaMendez2:
link to review app: https://rex-web-pr-82.herokuapp.com/books/college-physics/pages/preface


src/app/components/Typography.ts
```diff
@@ -0,0 +1,13 @@
+const fontsize = {
+  h3: '2.4rem',
+};
```
TomWoodward: in here you can 

```typescript
export const h3Style = css`
  font-size: ...;
  font-weight: bold;
  color: ...;
  line-height: ...;
`;
export const H3 = styled.h3`
  ${h3Style}
`;
```

and then in the navbar component you can extend those styles with whatever overrides (like the color) like this

```typescript
const CoolHeading = styled(H3)`
  color: ...;
  padding: ...;
`;
```

and in the main nav, where you have the login text that uses the h4 styles but is not actually an h4 because it needs to be an `a`, you can do

```typescript
const LoginLink = styled.a`
  ${h4Style}
  
  // hover state
`;

File: src/app/content/components/BookBanner.tsx [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+interface PropTypes {
+  page?: Page;
+  book?: Book;
+  services: AppServices;
+}
+
+const TopBar = styled.div`
+  height: 13rem;
+  width: 100%;
+  max-width: 1170px;
```
TomWoodward: rems

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  font-size: 1.25rem;
+  font-weight: bold;
```
TomWoodward: do you need these font styles on the whole bar? don't you just override them in the text elements?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  background: transparent;
+  margin: 0 auto;
+  text-align: left;
+  padding-top: 1.4rem;
+`;
+
+const BookTitle = styled.h4`
```
TomWoodward: extend h4 component from typography here

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  color: #FFFFFF;
+  text-align: left;
+  display: inline-block;
+  width: 100%;
+  max-width: 87rem;
+  margin: 1.25rem 0;
+  line-height: 1.5rem;
+`;
+
+const BookChapter = styled.h3`
```
TomWoodward: same

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  letter-spacing: -0.04rem;
+  color: #FFFFFF;
```
TomWoodward: this color is going to depend on the background color, some of them, like maybe the yellow ones, will want darker text. in the `theme` i put a `foreground` on some of the color sets to keep track of what text color things should use, we should add one to blue and use it here

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  line-height: ${typography.lineheight.h3};
+  font-size: ${typography.fontsize.h3};
+`;
+
+const blue = `${theme.color.primary.blue.base}`;
+const color = Color(`${blue}`).lighten(0.7); 
+
+const BarWrapper = styled.div`
+  width: 100%;
+  text-align: center;
```
TomWoodward: pretty sure you don't need this

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+  padding: 0 135px;
+  box-shadow: 0px 2px 2px 0px rgba(0, 0, 0, 0.1);
```
TomWoodward: rems?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    const treeSection = findArchiveTreeSection(book, page.id);
+
+    if(!treeSection) {
+      return null;
+    }
+
+    return <BarWrapper>
+      <TopBar>
+        <BookTitle><LeftArrow/>{book.tree.title}</BookTitle>
```
TomWoodward: this whole book title situation needs to be linked to the book details page

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+`;
+
+const InputWrapper = styled.div`
+  position: relative;
+  display: flex;
+  align-items: center;
+  margin-right: 4rem;
+`;
+
+const SearchTxt = styled.input`
```
TomWoodward: `SearchInput` probably more descriptive name

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  height: 2.5rem;
+`;
+
+const PrintTxt = styled.h3`
+  color: #818181;
+  font-size: 1.6rem;
+  cursor: pointer;
+`;
+
+const TableTitleWrapper = styled.div`
```
TomWoodward: took me a while to figure out what this meant, maybe `ToCButtonWrapper` would be better name

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  text-align: left;
+  float: left;
+`;
+
+const SearchPrintWrapper = styled.div`
+  text-align: right;
+  float: right;
```
TomWoodward: does this float do anything?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+const BarWrapper = styled.div` 
+  width: 100%;
+  text-align: center;
```
TomWoodward: pretty sure you don't need this

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  padding: 0 155px;
```
TomWoodward: you have this padding in a bunch of places, you should define it as a constant and export it from somewhere so that it can be changed in one place

also it should be in rems

maybe `/src/app/theme.ts` `contentBuffer` or something?

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+`;
+
+interface PropTypes {
+  page?: Page;
+  book?: Book;
+  services: AppServices;
+}
+
+const TopBar = styled.div`
+  width: 100%;
```
TomWoodward: default for divs

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  height: 2rem;
+  width: 2rem;
+  color: #FFFFFF;
+  margin-right: 0.7rem;
+`;
+
+interface PropTypes {
+  page?: Page;
+  book?: Book;
+  services: AppServices;
```
TomWoodward: you don't need services

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+}
+
+const TopBar = styled.div`
+  width: 100%;
+  max-width: 117rem;
+  background: transparent;
```
TomWoodward: i don't think you need this

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  margin: 0 auto;
+  text-align: left;
+`;
+
+const BookTitle = styled.a`
+  ${typography.h4Style}
+  letter-spacing: -0.04rem;
```
TomWoodward: shouldn't this be part of the h4 style?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  color: ${theme.color.primary.blue.foreground};
+  text-align: left;
```
TomWoodward: isn't this default?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const TopBar = styled.div`
+  width: 100%;
+  max-width: 117rem;
+  background: transparent;
+  margin: 0 auto;
+  text-align: left;
+`;
+
+const BookTitle = styled.a`
```
TomWoodward: this needs ellipsis text overflow

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  color: ${theme.color.primary.blue.foreground};
+  text-align: left;
+  display: flex;
+  align-items: center;
+  width: 100%;
+  max-width: 87rem;
+  margin: 0;
+  text-decoration: none;
+
+  @media (max-width: 700px) {
```
TomWoodward: this should be part of the h4 style

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+    font-size: 1.6rem;
+    line-height: 2.5rem;
+  }
+`;
+
+const BookChapter = styled.h3`
+  ${typography.h3Style}
+  letter-spacing: -0.04rem;
```
TomWoodward: part of the h3 style?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  color: ${theme.color.primary.blue.foreground};
+  font-weight: bold;
+  text-align: left;
```
TomWoodward: default?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  margin: 0;
+  text-decoration: none;
+
+  @media (max-width: 700px) {
+    font-size: 1.6rem;
+    line-height: 2.5rem;
+  }
+`;
+
+const BookChapter = styled.h3`
```
TomWoodward: this needs ellipsis text overflow, but its actually kind of complicated for the book chapter because Devika wants it to ellipsis normally for desktop, but allow wrapping onto two lines before ellipsis on mobile, which is not actually possible using the text-overflow property. 

i haven't looked into it a lot, but it seems _possible_ if hacky http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+  font-weight: bold;
+  text-align: left;
+  display: inline-block;
+  width: 100%;
+  max-width: 87rem;
+  margin-top: 1rem;
+  margin-bottom: 0;
+
+  @media (max-width: 700px) {
+    font-size: 1.6rem;
```
TomWoodward: this font stuff should be part of the h3 style

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+    line-height: 2.2rem;
+    margin-top: 0.3rem;
+  }
+`;
+
+const blue = `${theme.color.primary.blue.base}`;
+const color = Color(`${blue}`).lighten(0.7);
```
TomWoodward: can just pass blue here  `Color(blue)`

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+    line-height: 2.2rem;
+    margin-top: 0.3rem;
+  }
+`;
+
+const blue = `${theme.color.primary.blue.base}`;
+const color = Color(`${blue}`).lighten(0.7); 
+
+const BarWrapper = styled.div`
+  width: 100%;
```
TomWoodward: default

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,120 @@
...
+    </BarWrapper>;
+  }
+}
+
+export default connect(
+  (state: AppState) => ({
+    book: select.book(state),
+    page: select.page(state),
+  })
+)(withServices(TitleComponent));
```
TomWoodward: can remove the `withServices` here and make it just 
```
export default connect(
  (state: AppState) => ({
    book: select.book(state),
    page: select.page(state),
  })
)(TitleComponent);
```

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+import styled from 'styled-components';
+import {ListOl} from 'styled-icons/boxicons-regular/ListOl';
+import {Search} from 'styled-icons/fa-solid/Search';
+import {Print} from 'styled-icons/fa-solid/Print';
+import theme from '../../theme';
+
+const ListIcon = styled(ListOl)`
+  height: 2rem;
+  width: 2rem;
+  color: #027EB5;
```
TomWoodward: we should get all these colors from variables somewhere

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const SearchIcon = styled(Search)`
+  height: 2rem;
+  width: 2rem;
+  color: #818181;
+  margin-right: 0.7rem;
+  position: absolute;
+
+  @media (max-width: 700px) {
+    height: ${theme.iconStyles.mobile.height};
```
TomWoodward: this height i'm not sure belongs in the theme. would it be used anywhere outside of this file? maybe just declare it as a const at the top of this file?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+    width: ${theme.iconStyles.mobile.width};
+  }
+`;
+
+const TopBar = styled.div`
+  height: 5rem;
+  width: 100%;
```
TomWoodward: default

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  max-width: 117rem;
+  font-size: 1.25rem;
```
TomWoodward: move fonty things to the text elements please

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  font-weight: bold;
+  background: transparent;
```
TomWoodward: don't think you need this

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  margin: 0 auto;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  @media (max-width: 700px) {
```
TomWoodward: i haven't been commenting on all of them, but all these media queries need to be updated

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+    position: initial;
+  }
+`;
+
+const SearchInput = styled.input`
+  font-size: 1.6rem;
+  font-weight: normal;
+  font-style: normal;
+  font-stretch: normal;
```
TomWoodward: do you need all these `font: normal`s?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  line-height: 2.5rem;
+  letter-spacing: normal;
+  color: #818181;
```
TomWoodward: we probably don't want this color for typing, just for the placeholder

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+`;
+
+const PrintOptWrapper = styled.div`
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+
+  @media (max-width: 700px) {
+    display: initial;
```
TomWoodward: is it necessary to override this?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  }
+`;
+
+const PrintOptions = styled.h3`
+  color: #818181;
+  font-size: 1.6rem;
```
TomWoodward: don't you want all the h3 style here?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const SearchPrintWrapper = styled.div`
+  text-align: right;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+const BarWrapper = styled.div` 
+  width: 100%;
```
TomWoodward: default

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+  padding: ${theme.contentBuffer.toolbar.padding};
+  box-shadow: 0 0.2rem 0.2rem 0 rgba(0,0,0,0.14);
+  display: inline-block;
```
TomWoodward: do you really want this to be inline-block? i think block is better for this

TomWoodward (changes requested):

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
...
+    color: #0064A0;
+  }
+
+`;
+
+const SearchIcon = styled(Search)`
+  height: 2rem;
+  width: 2rem;
+  color: #818181;
+  margin-right: 0.7rem;
```
TomWoodward: this is the right margin, but because the element is `position: absolute;` its not doing anything. is there a reason for the absolute positioning? the print icon isn't absolute and has good looking margin

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,167 @@
+import React, { SFC } from 'react';
+import styled from 'styled-components';
+import {ListOl} from 'styled-icons/boxicons-regular/ListOl';
+import {Search} from 'styled-icons/fa-solid/Search';
+import {Print} from 'styled-icons/fa-solid/Print';
+import theme from '../../theme';
+
+const ListIcon = styled(ListOl)`
+  height: 2rem;
+  width: 2rem;
```
TomWoodward: according to the design this should be `width: 16px; height: 14px;` but i think maybe we should just set the height to `1.6rem` to match the font size and let the width figure itself out. it looks a little weird currently because you're still using the `boxicons` version instead of fa

TomWoodward (changes requested):

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+    color: ${theme.color.secondary.darkBlue.foreground};
+  }
+
+`;
+
+const SearchIcon = styled(Search)`
+  height: ${iconStyles.mobile.height};
+  width: ${iconStyles.mobile.width};
+  color: ${theme.color.secondary.darkGray.base};
+  margin-right: 0.7rem;
```
TomWoodward: lets put `outline: none` to hide that box when you select the input

src/app/components/NavBar.tsx
```diff
@@ -0,0 +1,75 @@
...
+const HeaderImage = styled.img`
+  width: auto;
+  height: 3rem;
+
+  @media (max-width: 48em) {
+    height: 2rem;
+  }
+`;
+
+const LoginTxt = styled.a`
```
TomWoodward: this button kinda twitches up when you hover, maybe you need some negative margin on the hover state to compensate for the border?

src/app/components/NavBar.tsx
```diff
@@ -0,0 +1,75 @@
...
+  ${typography.h4Style}
+  font-family: Helvetica Neue;
+  text-decoration: none;
+  font-weight: bold;
+  color: #5E6062;
+  padding: 1rem 0;
+  overflow: visible;
+
+  :hover {
```
TomWoodward: you can combine these `:hover, :active, :focus`

src/app/components/Typography.ts
```diff
@@ -0,0 +1,39 @@
...
+  line-height: 2.5rem;
+`;  
+
+export const h4Style = css`
+  font-size: 1.8rem;
+  line-height: 2.5rem;
+  letter-spacing: -0.04rem;
+  text-decoration: none;
+
+  @media (max-width: 48em) {
```
TomWoodward: this is the main mobile breakpoint right? lets put this in the theme somewhere, lots of stuff is going to need it

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,152 @@
...
+    border: none;
+    margin-right: 1rem;
+  }
+`;
+
+const SearchInput = styled.input`
+  ${Typography.bodyCopyRegularStyle};
+  color: ${theme.textColors.default};
+  border: none;
+  height: 2.5rem;
```
TomWoodward: you might not need this, since you get the line height from the bodyCopyRegularStyle

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,152 @@
+import React, { SFC } from 'react';
+import styled from 'styled-components';
+import {ListOl} from 'styled-icons/fa-solid/ListOl';
+import {Search} from 'styled-icons/fa-solid/Search';
+import {Print} from 'styled-icons/fa-solid/Print';
+import theme from '../../theme';
+import Typography from '../../components/Typography';
```
TomWoodward: this is why you need the default export, but it would probably be better to import it like `import {bodyCopyRegularStyle} from '../../components/Typography';`

src/app/theme.ts
```diff
@@ -16,6 +16,21 @@ const textColors = {
   white: '#fff',
 };
 
+const contentBuffer = {
+  default: {
+    padding: '0 13.5rem',
+  },
+  toolbar: {
+    padding: '0 15.5rem',
```
TomWoodward: i didn't realize that the toolbar had a different width than everything else. lets just add the additional 2rem padding in the toolbar rather than having this here

TomWoodward (changes requested):

src/app/theme.ts
```diff
@@ -45,6 +65,11 @@ const color = {
       darkest: '#818181',
       foreground: textColors.white,
     },
+    darkGray: {base: '#818181'},
+    darkBlue: {
```
TomWoodward: this isn't one of the theme colors from the pattern library so we shouldn't put it here, it looks like what you want is link styles, maybe copy [linkStyle](https://github.com/openstax/rex-web/pull/81/files#diff-2dbbc782d04e2cc39b4f4bcad6fe5049R4) from the other PR

TomWoodward (changes requested):

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,153 @@
...
+`;
+
+const ToCButton = styled.h3`
+  ${bodyCopyRegularStyle};
+  color: ${theme.color.secondary.darkBlue.base};
+  cursor: pointer;
+  margin: 0;
+
+  :hover {
+    color: ${theme.color.secondary.darkBlue.foreground};
```
TomWoodward: can we organize these so all the ToCButton, ListIcon and ToCButtonWrapper are next to each other? also maybe the `color` and hover state should be defined in ToCButtonWrapper? currently when you hover over the text the icon doesn't change, and the reverse, they should change together

TomWoodward (changes requested):

src/app/content/components/Toolbar.tsx
```diff
@@ -1,43 +1,42 @@
 import React, { SFC } from 'react';
-import styled from 'styled-components';
+import styled, {css} from 'styled-components';
 import {ListOl} from 'styled-icons/fa-solid/ListOl';
 import {Search} from 'styled-icons/fa-solid/Search';
 import {Print} from 'styled-icons/fa-solid/Print';
-import theme from '../../theme';
+import theme, { linkStyle } from '../../theme';
```
TomWoodward: linkStyle should be in typography

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,123 @@
...
+  ${h3Style}
+  color: ${theme.color.primary.blue.foreground};
+  font-weight: bold;
+  display: inline-block;
+  width: 100%;
+  max-width: 87rem;
+  margin: 1rem 0 0 0;
+  padding: 0;
+  overflow: hidden;
+  display: -webkit-box;
```
TomWoodward: @KarinaMendez you don't need this right?

TomWoodward (changes requested):

TomWoodward (approved):

src/app/components/NavBar.tsx
```diff
@@ -0,0 +1,89 @@
...
+  `)}
+`;
+
+// tslint:disable-next-line:variable-name
+const NavigationBar: SFC = ({}) =>
+  <BarWrapper>
+    <TopBar>
+      <FormattedMessage id='i18n:nav:logo:alt'>
+        {(msg: Element | string) => <LogoLink href='/'>
+          <HeaderImage src={openstaxLogo} alt={assertString(msg, 'alt text must be a string')} />
```
philschatz: Is there an advantage to using `` rather than `<svg .../>` directly? Other icons in this PR are injected directly.

If using `` to reduce file size, I think `role="img"` needs to be added (see http://web-accessibility.carnegiemuseums.org/code/svg/ )

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { maxNavWidth } from '../../components/NavBar';
+import { h3MobileLineHeight, h3Style, h4Style } from '../../components/Typography';
+import theme from '../../theme';
+import { AppState } from '../../types';
+import * as select from '../selectors';
+import { Book, Page } from '../types';
+import { bookDetailsUrl, findArchiveTreeSection } from '../utils';
+
+export const bookBannerDesktopHeight = 13;
+export const bookBannerMobileHeight = 10.4;
```
philschatz: How were these determined? Can these numbers be computed based on other numbers? If not, would a comment in the code help with maintenance?

src/app/content/components/BookBanner.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+// tslint:disable-next-line:variable-name
+const TopBar = styled.div`
+  width: 100%;
+  max-width: ${maxNavWidth}rem;
+  margin: 0 auto;
+`;
+
+const bookBannerTextStyle = css`
+  max-width: 87rem;
```
philschatz: Is this value based on another number (like the `@media (width)` breakpoint)? If so, could they be linked somehow in the code as a comment or by basing this value on the other?

src/app/content/components/Toolbar.tsx
```diff
@@ -0,0 +1,164 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+// tslint:disable-next-line:variable-name
+const BarWrapper = styled.div`
+  width: 100%;
+  padding: 0 ${theme.padding.page.desktop}rem;
+  box-shadow: 0 0.2rem 0.2rem 0 rgba(0,0,0,0.14);
```
philschatz: It seems like the box shadow is included in several places. Could this be extracted out as a function? (maybe that would not help with maintainability)

philschatz (changes requested):
It looks really good! I just had a couple accessibility & maintenance concerns

src/app/components/NavBar.tsx
```diff
@@ -0,0 +1,89 @@
...
+    padding: 0 ${theme.padding.page.mobile}rem;
+  `)}
+`;
+
+// tslint:disable-next-line:variable-name
+const NavigationBar: SFC = ({}) =>
+  <BarWrapper>
+    <TopBar>
+      <FormattedMessage id='i18n:nav:logo:alt'>
+        {(msg: Element | string) => <LogoLink href='/'>
```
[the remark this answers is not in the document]
TomWoodward: right now the `OpenStax Homepage` alt is on the image, should we move it to the logo or have it on both?

src/app/content/components/BookBanner.tsx
TomWoodward: they're made up in the design, not sure what comment we should add?

src/app/content/components/BookBanner.tsx
TomWoodward: no this was intended to stop the text from going over the background image, but the background image got removed from the spec, so maybe we don't even need this anymore? i'll ask about it

src/app/content/components/Toolbar.tsx
TomWoodward: it is used twice i think in the navs, but i'm not sure it would really help to extract it since its one line and the different usages could easily want different shadows

src/app/content/components/BookBanner.tsx
TomWoodward: this is `1170 - 570 / 2` or `pageWidth - textWidth / 2`, i'll replace with the math

philschatz (approved):

PR #85, opened 2019-03-11
chore: developer homepage

TomWoodward:
try out that "view deployment" button


philschatz (approved):
This seems to:

- change the button styling
- add some styling to headings
- add a way to get the constructed route to add it to the dev page
- puts the update notification on the current page (rather than generate a link to content that causes the update notification to pop up)

PR #90, opened 2019-03-15
feat: toc behavior

TomWoodward:


src/app/content/components/Content.tsx
```diff
@@ -18,17 +32,118 @@ const Background = styled.div`
...
+/*
+ * this layout is a mess for these reasons:
+ * - the navs must have the default padding inside their containers so their
+ *   backgrounds go to the edge of the window.
+ *
+ * - the content wrapper must behave the same way as the navs in order to get
+ *   the sidebar in the right place to line up with the button in the toolbar.
+ *
+ * - the white background must then have negative margin to undo the default
+ *   padding so that it can get to the edge of the screen on small windows.
```
philschatz: It seems like this will work but is it unlikely that we will need to add styling to the body (or descendants) when a component changes state in general?

If not, then these undo styles make sense. I'm just worried that we will have many Undo's sprinkled in the code eventually.


**Or:** the ToC toggle state is in redux; could the components (or the root) pull that state out and style themselves appropriately? Or is that what you are referring to in:

> ... which means that a disturbing number of things need to know when the sidebar is open/closed

philschatz (commented):
A partial review (up to but not including ` src/app/content/components/Sidebar.tsx`)

script/prerender/prerender.tsx
```diff
@@ -193,6 +213,8 @@ function injectHTML(html: string, {body, styles, state, fonts, meta}: Options) {
     `<script>window.__PRELOADED_STATE__ = ${JSON.stringify(state).replace(/</g, '\\u003c')}</script>`
   );
 
+  html = html.replace(/(<script src="[^"]+\.chunk\.js"><\/script>)/, scripts.join('') + '$1');
```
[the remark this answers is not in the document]
TomWoodward: there are enough interconnected things that we ran into a dependency loop when unit testing the sidebar. it went something like

sidebar > stuff > contentlink > content route > content layout > sidebar

there are many things that need to reference routes for reverse routing but don't need the whole component tree from the route, so i decided to lazy-load the route components to break the dependency loop.

src/app/content/components/Content.tsx
TomWoodward: yes i was referring to the redux state, the components do pull the redux state and style themselves accordingly. i don't love having lots of styling all over the app based on the same variable, it feels disjointed to me, but in this case i couldn't centralize the sidebar open/close styles.

src/app/content/components/Sidebar.tsx
```diff
@@ -1,59 +1,106 @@
...
-const sidebarTransitionTime = 300;
 const sidebarPadding = 1;
 
-// TODO - magic numbers to be replaced in `top`, `height` when ToC gets real styling in openstax/unified#176
-const SidebarBody = styled.div<{isOpen: boolean}>`
-  top: 23rem;
-  height: calc(100vh - 23rem);
+const sidebarClosedStyle = css`
+  overflow-y: hidden;
+  transform: translateX(-${sidebarDesktopWidth}rem);
```
philschatz: Are these `translateX` here for the animations or to get around having parent elements know about the ToC state? (or both?)

src/app/content/components/Sidebar.tsx
```diff
@@ -1,59 +1,106 @@
...
   overflow-y: auto;
-  padding: ${sidebarPadding}rem;
-  transition: all ${sidebarTransitionTime}ms;
+  height: calc(100vh - ${navDesktopHeight + bookBannerDesktopHeight}rem);
+  transition:
+    transform ${sidebarTransitionTime}ms,
+    box-shadow ${sidebarTransitionTime}ms,
+    background-color ${sidebarTransitionTime}ms;
   background-color: ${theme.color.neutral.darker};
+  z-index: 3; /* stay above book content and overlay */
```
philschatz: This _might_ be an accessibility concern. I don't remember if it's just `tabindex` or if `z-index` is also problematic. I'll check.

src/app/content/components/Sidebar.tsx
```diff
@@ -1,59 +1,106 @@
...
+  margin-left: -50vw;
+  padding-left: 50vw;
+  width: calc(50vw + ${sidebarDesktopWidth}rem);
+  min-width: calc(50vw + ${sidebarDesktopWidth}rem);
+  box-shadow: 0.2rem 0 0.2rem 0 rgba(0, 0, 0, 0.1);
+  ${theme.breakpoints.mobile(css`
+    width: calc(50vw + ${sidebarMobileWidth}rem);
+    min-width: calc(50vw + ${sidebarMobileWidth}rem);
+    margin-top: -${toolbarMobileHeight}rem;
```
philschatz: All these negative signs still worry me about maintainability but at least we have unit test screenshots!

src/app/content/components/SidebarControl.tsx
```diff
@@ -1,43 +1,104 @@
...
+}
+
+// tslint:disable-next-line:variable-name
+const ListIcon = styled(ListOl)`
+  ${toolbarIconStyles};
+`;
+
+// tslint:disable-next-line:variable-name
+const ToCButtonText = styled.h3`
+  font-family: ${contentFont};
```
philschatz: do we have multiple fonts in REX (excluding the actual content since it's Roboto or whatever)

src/app/content/components/Toolbar.tsx
```diff
@@ -76,10 +53,24 @@ const SearchInputWrapper = styled.div`
   display: flex;
   align-items: center;
   margin-right: 4rem;
-  border-bottom: solid 0.1rem ${iconColor};
+  position: relative;
+
+  :after {
```
philschatz: What is this doing? It seems like it is just adding a line under the input box?

src/app/content/content.browserspec.ts
```diff
@@ -125,21 +129,17 @@ const clickTocLink = (href: string) => page.evaluate(async(href) => {
...
 
 const getSelectedTocSection = () => page.evaluate(() => {
-  const tocHeader = document && Array.from(document.querySelectorAll('h2'))
-    .find((node) => node.textContent === 'Table of Contents');
-
-  const li = tocHeader && tocHeader.parentElement
-    && tocHeader.parentElement.querySelector('li[aria-label="Current Page"]');
+  const toc = document && document.querySelector('[aria-label="Table of Contents"]');
 
+  const li = toc && toc.querySelector('li[aria-label="Current Page"]');
```
philschatz: are there more `[aria-label="Current Page"]` elements? Or is this just clarifying that the element should be an `li`?

src/app/content/routes.ts
```diff
@@ -12,7 +12,12 @@ interface State {
 }
 
 export const content: Route<Params, State> = {
-  component: components.Content,
+  component: Loadable({
+    loader: () => import(/* webpackChunkName: "Content" */ './components/Content'),
+    loading: () => null,
```
philschatz: Should an indicator (visual & aural) be added so folks (screenreader, selenium, slow-internet users) know that something is happening?

And can it error when the chunk cannot be loaded?

src/app/developer/routes.ts
```diff
@@ -1,11 +1,16 @@
...
 import { Route } from '../navigation/types';
-import Home from './components/Home';
 
 const ROUTES_PATH = '/';
 
 export const routes: Route<undefined, undefined> = {
-  component: Home,
+  component: Loadable({
+    loader: () => import(/* webpackChunkName: "DeveloperHome" */ './components/Home'),
+    loading: () => null,
```
philschatz: Similarly, can in indicator of loading/loaded be included?

And since this is the error page, maybe an `alert` (or some other fallback) if it fails to load?

src/app/errors/routes.ts
```diff
@@ -1,10 +1,15 @@
...
 import { Route } from '../navigation/types';
-import * as components from './components';
 
 const CATCH_ALL = '/(.*)';
 
 export const notFound: Route<undefined> = {
-  component: components.PageNotFound,
+  component: Loadable({
+    loader: () => import(/* webpackChunkName: "PageNotFound" */ './components/PageNotFound'),
+    loading: () => null,
```
philschatz: Similarly, can in indicator of loading/loaded be included?

src/app/content/components/Sidebar.tsx
TomWoodward: i think you're thinking of tabindex

src/app/content/components/Sidebar.tsx
TomWoodward: yea element positioning is a bit crazy in this layout, hopefully we won't have to do any more after this

src/app/content/components/SidebarControl.tsx
TomWoodward: not currently, but sometimes designers like to use different fonts for headings and stuff so i didn't want to limit our flexibility

src/app/content/components/Toolbar.tsx
TomWoodward: yea it makes the underline, initially it used border-bottom but then the content was pushed up by 1px and i didn't want to make the container 1px bigger than everything else, with the right alignment, and overflow visible just to make the line show up

src/app/content/content.browserspec.ts
TomWoodward: the li here is irrelevant, its just for readability, but it might be brittle, should we remove it?

src/app/content/routes.ts
TomWoodward: hmmmm. we should test something using aria-live to announce the page load

src/app/developer/routes.ts
TomWoodward: this is developer home so this route we can probably ignore

src/app/errors/routes.ts
TomWoodward: these error pages aren't actually used in production

src/app/content/routes.ts
philschatz: I _think_ that since the JS is loaded immediately aria-live is not necessary but I'll spin up NVDA and send it to a page & see if the whole page is read out

philschatz: NVDA reads it fine

philschatz (approved):
@TomWoodward and I discussed re-visiting whether a slightly different UI would help address the negative margin spaghetti code in the ToC. One possibility is to:

- move the ToC button _into_ the top navbar
- add padding to the top of the ToC so it works regardless of how many nav bars are open at the top of the page (or add 2 cases for the top of the ToC based on how many nav bars are visible)
- remove JS that computes the height of the ToC and use `position: fixed; bottom: 0`

PR #94, opened 2019-03-26
chore: load rex-specific CSS files

philschatz:
refs https://github.com/openstax/unified/issues/169

As part of https://github.com/openstax/cnx-recipes/pull/953 the CSS files were renamed. This is the matching change to read them in (and not silently error when the file is not found).

Discussion: There might be value in renaming the file in cnx-recipes to be `.json` or it might be useful to allow importing the package and exposing a `Map` to this repo.

See https://github.com/openstax/cnx-recipes/pull/964 for turning cnx-recipes into an importable package


TomWoodward (approved):

TomWoodward:
😬 previously i had mocked the book css so we didn't get the whole stylesheet in the snapshot probably want to change that so it affects the new way you're doing it

philschatz:
Thanks 😊 ! They should be fixed now.

PR #96, opened 2019-03-29
feat: toc refinements

TomWoodward:
first pr to address feedback on the ToC behavior.

- momentum scrolling on mobile in the toc 
- fix body scroll lock on mobile safari

there is no css solution for mobile safari scroll lock, so this keeps the css solution for everything else and adds js for mobile


snm8 (approved):
works fine on ios safari

philschatz (dismissed):

philschatz (approved):

PR #97, opened 2019-04-01
add intro stats and set color-theme meta

TomWoodward:


snm8 (approved):
Devika also saw and approved

src/app/content/hooks/receiveContent.ts
```diff
@@ -23,6 +24,7 @@ const hookBody: ActionHookBody<typeof receivePage | typeof receiveBook> = ({getS
   dispatch(setHead({
     meta: [
       {property: 'og:description', content: ''},
+      {name: 'theme-color', content: theme.color.primary[book.theme].base},
```
philschatz: It seems like the theme color is being added twice. Is that desired?

image

philschatz (changes requested):

src/app/content/hooks/receiveContent.ts
TomWoodward: i have a note to myself that setHead is being sent twice on navigation and that is causing some tags to be added twice (og:description as well)

i will deal with that as a separate change

philschatz (approved):

PR #98, opened 2019-04-02
feat: Back/Next Buttons

KarinaMendez2:


src/app/content/components/Content.tsx
```diff
@@ -139,6 +140,7 @@ const Content: React.SFC = () => <Layout>
               <HideOverflowAndRedoPadding>
                 <Page />
               </HideOverflowAndRedoPadding>
+              <PrevNextBar />
```
TomWoodward: i think you probably want to move this up one line next to `Page` so it lines up perfectly with the text

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  prevNext?: {
+    prev?: LinkedArchiveTreeSection;
+    next?: LinkedArchiveTreeSection;
+  };
+}
+
+// tslint:disable-next-line:variable-name
+const PrevPage = styled.span`
+  ${textRegularStyle}
+  ${linkStyle}
```
TomWoodward: you want `decoratedLinkStyle`, it looks a little different for links that have an icon and you won't have to override the border

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  border: none;
+`;
+
+// tslint:disable-next-line:variable-name
+const BarWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  background: ${theme.color.neutral.base};
```
TomWoodward: you probably don't need to specify the background color, i think the parent element does

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+const NextPage = styled.span`
+  ${textRegularStyle}
+  ${linkStyle}
+  border: none;
+`;
+
+// tslint:disable-next-line:variable-name
+const BarWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
```
TomWoodward: this won't work when there is only one link shown though right?

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+// tslint:disable-next-line:variable-name
+const PrevPage = styled.span`
+  ${textRegularStyle}
+  ${linkStyle}
+  border: none;
+`;
+
+// tslint:disable-next-line:variable-name
+const NextPage = styled.span`
```
TomWoodward: why not just style the ContentLink here? `styled(ContentLink)`? that way you don't need the extra elements in the markup

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+    const treeSection = findArchiveTreeSection(book, page.id);
+
+    if (!treeSection || !prevNext) {
+      return null;
+    }
+
+    if ( prevNext.prev) {
+      prev = <ContentLink book={book} page={prevNext.prev}><PrevPage><LeftArrow/>Previous</PrevPage></ContentLink>;
+    } else {
+      prev = <PrevPage><LeftArrow/>Previous</PrevPage>;
```
TomWoodward: this shouldn't display anything when there is no previous link

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -39,3 +39,36 @@ export const findArchiveTreeSection = (
...
+): {} | undefined => {
+  const flattenTree = flattenArchiveTree(book.tree);
+  const currentSection = findArchiveTreeSection(book, pageId);
+
+  const sections: { [s: string]: object; } = {};
+
+  if ( flattenTree && currentSection ) {
+    const index = flattenTree.findIndex((currentSection) =>
+    stripIdVersion(currentSection.shortId) === stripIdVersion(pageId)
+    || stripIdVersion(currentSection.id) === stripIdVersion(pageId));
```
TomWoodward: this is the same comparison logic that is already in the `findArchiveTreeSection` above, we could separate it out by defining a new function like 

```js
const sectionMatcher = (targetPageId) => (currentSection) =>
    stripIdVersion(currentSection.shortId) === stripIdVersion(targetPageId)
    || stripIdVersion(currentSection.id) === stripIdVersion(targetPageId));
```

and using that in both places

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -39,3 +39,36 @@ export const findArchiveTreeSection = (
     stripIdVersion(section.shortId) === stripIdVersion(pageId)
     || stripIdVersion(section.id) === stripIdVersion(pageId)
   );
+
+export const prevNextBookPage = (
+  book: {tree: ArchiveTree},
+  pageId: string
+): {} | undefined => {
+  const flattenTree = flattenArchiveTree(book.tree);
+  const currentSection = findArchiveTreeSection(book, pageId);
```
TomWoodward: i think this is unused

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -39,3 +39,36 @@ export const findArchiveTreeSection = (
...
+
+  const sections: { [s: string]: object; } = {};
```
TomWoodward: a stronger type for this would be:
```
{
    prev?: LinkedArchiveTreeSection;
    next?: LinkedArchiveTreeSection;
}
```

same as in the component

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -39,3 +39,36 @@ export const findArchiveTreeSection = (
...
+
+    } else if ( index > 0 && index === (flattenTree.length - 1)) {
+      sections.prev = flattenTree[index - 1];
+
+    } else if (index === 0) {
+      sections.next = flattenTree[index + 1];
+
+    } else {
+      sections.prev = {};
+      sections.next = {};
```
TomWoodward: you shouldn't need this else case, because the logic in the component handles `prev` and `next` being undefined, adding these empty objects will actually break it right?

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -39,3 +39,36 @@ export const findArchiveTreeSection = (
...
+
+    if ( index > 0 && index < (flattenTree.length - 1)) {
+      sections.prev = flattenTree[index - 1];
+      sections.next = flattenTree[index + 1];
+
+    } else if ( index > 0 && index === (flattenTree.length - 1)) {
+      sections.prev = flattenTree[index - 1];
+
+    } else if (index === 0) {
+      sections.next = flattenTree[index + 1];
```
TomWoodward: you can replace all these if/elseif statements with:

```typescript
type Sections = {prev?: LinkedArchiveTreeSection; next?: LinkedArchiveTreeSection;}
const sections: Sections = {
  prev: flattenTree[index - 1],
  next: flattenTree[index + 1]
};
```

the index offsets might be undefined, but thats fine.

TomWoodward (changes requested):

TomWoodward:
@KarinaMendez2 this needs master merged in and the react snapshots updated

TomWoodward:
@KarinaMendez2 i think the prev/next bar is supposed to be limited to the same size as the text, like this 

image

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+// tslint:disable-next-line:variable-name
+const NextLinkElement = styled(ContentLink)`
+  ${textRegularStyle}
+  ${decoratedLinkStyle}
+  float: right;
+`;
+
+// tslint:disable-next-line:variable-name
+const BarWrapper = styled.div`
```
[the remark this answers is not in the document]
TomWoodward: @KarinaMendez2 you can remove a bunch of the styling in here by pulling in `contentTextStyle` from `Page.tsx`, that is what the attribution uses to match the margin too

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  height: 4rem;
+  display: flex;
+  align-items: center;
+  max-width: 57rem;
+  margin: 5rem auto 6rem auto;
```
TomWoodward: this bottom margin appears too large because there is also margin on top of the attribution. you either need to reduce this margin to compensate or remove the margin on the attribution.

be careful not to mess up the box-shadow on the attribution, it might need that margin for the shadow to work

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  border-top: solid 0.1rem #e5e5e5;
+  border-bottom: solid 0.1rem #e5e5e5;
```
TomWoodward: this color should come from a variable

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+// tslint:disable-next-line:variable-name
+const LeftArrow = styled(ChevronLeft)`
+  height: 3rem;
+  width: 3rem;
+`;
+
+// tslint:disable-next-line:variable-name
+const RightArrow = styled(ChevronRight)`
+  height: 3rem;
+  width: 3rem;
```
TomWoodward: should probably have these icon styles in a variable

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+`;
+
+interface PropTypes {
```
TomWoodward: i like to put these kind of interface next to where they're used, can we move it down above `PrevNextBar`?

src/app/content/components/PrevNextBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  border-top: solid 0.1rem #e5e5e5;
+  border-bottom: solid 0.1rem #e5e5e5;
+  ${theme.breakpoints.mobile(css`
+    margin: 3.5rem auto 4.5rem auto;
+    border: 0;
+  `)}
+`;
+
+// tslint:disable-next-line:variable-name
+const LinksWrapper = styled.div`
```
TomWoodward: is this component actually necessary?

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -31,11 +31,34 @@ export const findDefaultBookPage = (book: {tree: ArchiveTree}) => {
...
+}
+
+export const prevNextBookPage = (
+  book: {tree: ArchiveTree},
+  pageId: string
+): Sections | undefined => {
+  const flattenTree = flattenArchiveTree(book.tree);
+  const index = flattenTree.findIndex(sectionMatcher(pageId));
+
+  if ( index !== -1) {
```
TomWoodward: you should remove this condition, then the function will always return `Sections` even if both `prev` and `next` are undefined. that will simplify the signatures for this function and the selector

TomWoodward (changes requested):

src/app/content/components/PrevNextBar.tsx
KarinaMendez2: For the way I styled it, it's necessary for my floats to work inside a div with `display flex`.

TomWoodward: if you put `justify-content: space-between;` on the flex wrapper, you can get rid of `LinksWrapper` and the floats

TomWoodward (changes requested):

src/app/content/components/PrevNextBar.screenshotspec.tsx
```diff
@@ -0,0 +1,121 @@
...
+  });
+
+});
+
+const scrollDown = () => page.evaluate(() => {
+  return window && document && document.documentElement && window.scrollBy(0, document.documentElement.scrollHeight);
+});
+
+// tslint:disable-next-line:no-shadowed-variable
+const clickNextLink = (href: string, elementToClick: string) => page.evaluate(async(href, elementToClick) => {
```
TomWoodward: [page.click](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageclickselector-options) is a thing that could replace a bunch of the stuff in this function i think

TomWoodward (approved):

PR #100, opened 2019-04-03
feat: close toc on navigate on mobile

TomWoodward:


snm8 (approved):

src/app/content/components/ContentLink.tsx
```diff
@@ -17,19 +17,23 @@ interface Props extends React.HTMLProps<HTMLAnchorElement> {
...
 // tslint:disable-next-line:variable-name
-export const ContentLink: SFC<Props> = ({book, page, currentPath, navigate, ...props}) => {
+export const ContentLink: SFC<Props> = ({book, page, currentPath, navigate, onClick, ...props}) => {
   const {url, params} = getBookPageUrlAndParams(book, page);
   const relativeUrl = toRelativeUrl(currentPath, url);
 
   return <a
     onClick={(e) => {
       e.preventDefault();
+      if (onClick) {
```
philschatz: Other code uses `onClick && onClick()` syntax; is there a reason for doing it differently here? Or is that just selective-memory?

src/app/content/components/Sidebar.tsx
```diff
@@ -204,5 +211,8 @@ export default connect(
   (state: AppState) => ({
     ...selectors.bookAndPage(state),
     isOpen: selectors.tocOpen(state),
+  }),
+  (dispatch: Dispatch) => ({
+    onNavigate: () => dispatch(resetToc()),
```
philschatz: It seems that the ToC collapses _before_ the page changes; is that desired? It seems a bit confusing if the user is on a slow internet connection.

philschatz (changes requested):
It seems that the ToC collapses _before_ the page changes; is that desired? It seems a bit confusing if the user is on a slow internet connection.

src/app/content/components/Sidebar.tsx
TomWoodward: i think the ideal solution would be to add a loading state to the content side of the page. given that i think its better to close the toc immediately after clicking

src/app/content/components/ContentLink.tsx
TomWoodward: the linter actually complains about doing that in a block. it fails the `no-unused-expression` rule. i think we might do that outside of blocks in `() => thing && thing()` expressions

src/app/content/components/Sidebar.tsx
philschatz: @snm8 Are there plans to design a `Loading...` UI when the user navigates between pages? I can create a story if not. It's not critical but it is one of the weird intermediate states that rex can be in.

philschatz (approved):

snm8:
Sure, we should have a standard one anyway that can be used across REX. Can you create another story for it, like you said?

PR #103, opened 2019-04-04
feat: min content height

TomWoodward:
on pages with very short content the white bg section will now fill the left side of the screen (instead of stopping at the end of the content and allowing the grey bg to show below)


snm8 (approved):

philschatz (commented):
Would it make sense to have a screenshot test for this case?

TomWoodward:
> Would it make sense to have a screenshot test for this case?

i thought we had one already, but i will add

philschatz (approved):

PR #104, opened 2019-04-04
chore: add stacktraces to actions

TomWoodward:
adds stacktrace to actions in dev to expose them in the redux developer tools. 

note, using a library instead of `(new Error()).stack` directly because the library resolves sourcemaps

this makes it easier to debug why an action is happening by exposing the exact line that is calling `dispatch`


src/app/index.tsx
```diff
@@ -86,6 +87,11 @@ export default (options: Options) => {
     ...hooks.map((hook) => hook(services)),
   ];
 
+  /* istanbul ignore next */
+  if (process.env.REACT_APP_ENV === 'development') {
```
philschatz: Why not include it in production? Our code is public & it could help debug errors that occur in production that are sent to sentry.

TomWoodward: i didn't include it in production because it adds an async middleware step to every action dispatch where it resolves the sourcemaps for the trace before allowing the action to continue, it seemed probably bad to do that.

we can definitely use the same library to decorate any messages or errors that we send to sentry when we make that service though.

philschatz (approved):

PR #111, opened 2019-04-09
feat: shrinking nav 3a

TomWoodward:
updates scrollbar to be smaller after scrolling down, the changes to attribution actually fixes openstax/unified#226, i had to fix that because the new nav behavior was breaking some screenshot tests.

initially i really wanted to have the one nav transform to be smaller as you scrolled down, but this led to a flash of weirdly styled content when loading a page that was already scrolled down. so now there are two navs, one for the top of the page and one that is sticky. the solution is css only except for some js that makes it look nicer if you're in the 60px where the two bars overlap.


src/app/content/components/BookBanner.tsx
```diff
@@ -66,14 +79,22 @@ const BookTitle = styled.a`
...
+    ${theme.breakpoints.mobile(css`
+      display: none;
+    `)}
+  `)}
 `;
 
 // tslint:disable-next-line:variable-name
-const BookChapter = styled.h1`
-  ${h3Style}
+const BookChapter = styled((props) => props.variant === 'mini' ? <h2 {...props} /> : <h1 {...props} />)`
```
philschatz: Why does the heading level change based on whether the `mini` is visible or not? It seems odd that the root of the heading tree (`H1`) would change based on the scroll position.

src/app/content/components/BookBanner.tsx
```diff
@@ -87,27 +108,58 @@ const BookChapter = styled.h1`
 `;
 
 // tslint:disable-next-line:variable-name
-const BarWrapper = styled.div<{theme: Book['theme']}>`
-  position: sticky;
+export const BarWrapper = styled.div<{theme: Book['theme'], up: boolean, variant: 'mini' | undefined}>`
```
philschatz: **Musing:** Why not use `'mini' | 'expanded'` to help clarify that there are 2 different options? 

I saw that `BookChapter` is just a vanilla styled element so it's probably not important.

philschatz (commented):
Just a question about why the heading level changes based on the scroll position. The other comments are just musings.

src/app/content/components/BookBanner.tsx
```diff
@@ -87,27 +108,58 @@ const BookChapter = styled.h1`
...
   top: 0;
   padding: 0 ${theme.padding.page.desktop}rem;
   box-shadow: 0 0.2rem 0.2rem 0 rgba(0, 0, 0, 0.1);
   display: flex;
   align-items: center;
-  height: ${bookBannerDesktopHeight}rem;
+  height: ${ifMiniNav(bookBannerDesktopMiniHeight, bookBannerDesktopBigHeight)}rem;
```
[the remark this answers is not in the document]
TomWoodward: i had it this way at first, but i was annoyed that the rendered css had all kinds of stuff overriding itself instead of just rendering what it meant

i can put it back this way if you think it is easier to read though

src/app/content/components/BookBanner.tsx
TomWoodward: it doesn't change, the mini and big both exist in the dom at the same time, so the mini is not an h1 because there can only be 1 h1

philschatz: Hmm, in that case it seems like only one of these is visible at any time. If that is the case, then should the visible one be `<h1>` while the invisible is switched to be `<h2>`?

philschatz (approved):

src/app/content/components/Toolbar.tsx
```diff
@@ -99,8 +99,9 @@ const PrintOptWrapper = styled.div`
 `;
 
 // tslint:disable-next-line:variable-name
-const PrintOptions = styled.h3`
+const PrintOptions = styled.span`
```
philschatz: This might need to be a `<button>` but since it's a no-op it doesn't matter for now

philschatz (approved):

PR #113, opened 2019-04-12
fix: notifications

TomWoodward:
react becomes confused if you call `hydrate` but do not render the exact same thing that the server rendered. because i was merging state from the query before rendering, it was becoming different. this led to the notification content replacing the book content.

note: when this is merged update testrail plans to use this query for notifications `?initialActions=[{"type":"Notification/updateAvailable"}]` instead of the current `initialState`

additionally, there was a problem where the styled-component classes were being applied in the wrong order only when serving pre-rendered pages. the overrides on the notification wrapper from the content layout were being applied, but they were being applied under the base notification styles, so they were doing nothing.

i updated all styled-components to use the babel macro, i'm not sure if this actually fixed anything but it might have helped. also now the class names are more legible so i'm just leaving these changes in.

the real fix was updating the notification wrapper to export a styled component, instead of a SFC that wraps styled components. apparently, internally, when you do `styled(thing)` it does something different if `thing` is a standard or styled component. so whatever it does to "merge" the two styles or whatever wasn't happening properly because there was that intermediary SFC stopping it from realizing that the child was a styled component.

in general we should make sure that if styled components are going to be extended with overriding styles, they need to be exported as the styled component, not a wrapper.


philschatz (approved):
I had wondered why the `styled` sometimes began with a `;`. Glad they're not necessary.

PR #114, opened 2019-04-15
feat: browser print

philschatz:
Refs openstax/unified#183 . Hide everything except the content when printing. 

**Note:** The chapter and section are in the `<title>` so they show up in the header of every printed page

image

# TODO

- [x] Add puppeteer sceenshot tests


src/app/content/components/Content.tsx
```diff
@@ -27,31 +27,35 @@ import { wrapperPadding } from './Wrapper';
 
 // tslint:disable-next-line:variable-name
 const Background = styled.div`
+@media screen {
```
philschatz: I can indent these properly once the PR is approved. Not indenting them makes it easier to pull in updates from `master`

src/app/content/components/Page.tsx
```diff
@@ -168,9 +168,10 @@ export class PageComponent extends Component<PropTypes> {
 
 export const contentTextStyle = css`
   ${bodyCopyRegularStyle}
+@media screen { /* full page width in print */
   max-width: ${contentTextWidth}rem;
   margin: 0 auto;
```
TomWoodward: i think this margin is somehow overriding the margin in the PrevNextBar here https://github.com/openstax/rex-web/blob/master/src/app/content/components/PrevNextBar.tsx#L47 you can see that component is pulling in `contentTextStyle` but applying its own margin after.

if you compare this branch to master, you see that these two styles, the max-width and margin, are now in their own class instead of being applied directly to the same class when you drop in the variable. maybe something to do with the css pre-processor handling the `@media` call? we may want to re-think how we use variables to dry styles like this in the future.

for now, since this `contentTextStyle` is used by other things like attribution, prev/next links, and all you really care about is the page component, maybe you could put a `@media print` override down in the component on line 177?

TomWoodward (changes requested):

TomWoodward:
@philschatz2 the mobile view is broken for me on this branch, i wonder why CI is passing? is it broken for you?

TomWoodward (changes requested):
the mobile media query seems broken now

TomWoodward (approved):
found a bug in the mobile nav, but looks like it is also on master.

linter is still failing, but otherwise 👍

philschatz:
Yep, I was waiting to fix the lint errors (indent all the CSS) until this PR was approved so I wouldn't have to constantly merge conflicts 😄

PR #118, opened 2019-04-17
fix: os logo

TomWoodward:
there was 3px of rando space under the image, making it block fixed it


philschatz (approved):
Why didn't any of the screenshots change?

TomWoodward:
@philschatz they're within the margin for error that we allow in CI...

PR #121, opened 2019-04-22
fix: SVG and alignment

KarinaMendez2:
Based on the solution I found, we would have to leave the arrows (Desktop & mobile) the same way, but we can add extra padding on desktop if it looks nicer.


TomWoodward:
@KarinaMendez2  i liked the `svgViewBox` solution, but it seemed like a lot of overriding stuff so i tried out a different way. both solutions are very tightly coupled to the icon itself, but i think the one i tried out is a little less complicated.

let me know if this works for you.

TomWoodward (approved):
@KarinaMendez2 i liked the svgViewBox solution, but it seemed like a lot of overriding stuff so i tried out a different way. both solutions are very tightly coupled to the icon itself, but i think the one i tried out is a little less complicated.

let me know if this works for you.

KarinaMendez2:
Cool! Works great, thanks.

PR #122, opened 2019-04-22
fix: security 

TomWoodward:
https://nvd.nist.gov/vuln/detail/CVE-2019-11324


philschatz (approved):
Since this file isn't in the root directory does that mean we can't use pyupbot to get automatic PR's? /cc @pumazi

PR #124, opened 2019-04-22
chore: move TypeScript definition files

philschatz:
so they do not clutter the src directory


TomWoodward (approved):
👍 on the `typings` directory, but i'm confused about deleting the `react-app-env.d.ts` and adding it to the `.gitignore`?

philschatz:
It looks like CRA regenerates the file so since it is autogenerated I removed it from git.

TomWoodward:
in that case the new annoying thing is that cra generates it in src and not in src/typings... but i guess we have to just deal with that

PR #127, opened 2019-04-25
feat: login

TomWoodward:


src/test/browserutils.ts
```diff
@@ -117,5 +117,5 @@ export const checkLighthouse = async(target: puppeteer.Browser, urlPath: string)
   expect(lhr.categories.customAccessibility.score).toBeGreaterThanOrEqual(1);
   expect(lhr.categories.accessibility.score).toBeGreaterThanOrEqual(1);
   expect(lhr.categories.seo.score).toBeGreaterThanOrEqual(0.8);
-  expect(lhr.categories['best-practices'].score).toBeGreaterThanOrEqual(0.93);
+  expect(lhr.categories['best-practices'].score).toBeGreaterThanOrEqual(0.87);
```
philschatz: What caused the drop?

src/test/setup.ts
```diff
@@ -21,6 +21,8 @@ expect.extend({
   toMatchImageSnapshot,
 });
 
+jest.mock('ally.js/style/focus-within');
```
philschatz: When is ally.js loaded in the real code? I probably missed it while reviewing.

philschatz (approved):
Why the switch from a down-caret to a hamburger menu when the screen is narrow?

Kapture 2019-05-06 at 16 20 17


The code looks fine aside from the other couple of comments in this review.

TomWoodward:
@philschatz it changes to a hamburger on small screens because the menu works as an overlay on small screens and the hamburger is more intuitive for that, openstax.org does the same thing

src/test/browserutils.ts
TomWoodward: the 403 from accounts indicating the user is not logged in, i couldn't find a way to make chrome ignore the error status in the request

```
Failed to load resource: the server responded with a status of 403 (Forbidden)
```

src/test/setup.ts
philschatz: Thanks! And it seems to be to support [:focus-within in IE](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within)

philschatz (approved):

PR #128, opened 2019-04-26
feat: style content with generic styles

philschatz:
cnx-recipes code remains because we will still slowly transition to using them

# TODO

- [x] fix the screenshot test (I was unable to get the iframe to render in my docker container)
- [x] add a flag on the book to support switching between the generic style and one defined in `cnx-recipes`
- [x] change screen width so generic style matches webview


generic-styles/_content-common.less
```diff
@@ -0,0 +1,512 @@
+// Constants from bootstrap/less/variables.less
```
TomWoodward: do we need this file at all? we only have baked content in rex right?

src/app/components/SkipToContentWrapper.spec.tsx
```diff
@@ -26,15 +26,15 @@ describe('SkipToContentWrapper', () => {
 
     it('fails when main is not wrapped in a SkipToContentWrapper', () => {
         const breaks = () => renderToDom(<MessageProvider>
-            <MainContent/>
+            <MainContent isGenericStyle={true}/>
```
TomWoodward: why even have a prop for this if it is always `true`?

src/app/content/components/Page.tsx
```diff
@@ -92,17 +94,33 @@ export class PageComponent extends Component<PropTypes> {
...
-          />
-        </div>
-      </MainContent>}
-    </BookStyles>;
+    const book = this.props.book && BOOKS[this.props.book.id] || null;
+    if (book && book.bookStyleName) {
+      const bookStyles = assertString(getBookStyles().get(book.bookStyleName),
+        `missing book style: "${book.bookStyleName}"`);
+      // tslint:disable-next-line:variable-name
+      const BookStyles = styled.div`
```
TomWoodward: is this whole top block only here to load the intro-business styles that we've been historically using? is this a real life use-case? defining a styled component like this is a performance no-no since it re-processes it on every render. also the wrapping divs for chapter, page are not required for any particularly good reason, just that they were already in the css.

if we're gonna need separate stylesheets for different books, i'd rather build that on the new system you put in place, and not have all this junk.

src/index.tsx
```diff
@@ -5,6 +5,7 @@ import Loadable from 'react-loadable';
 import createApp from './app';
 import { assertWindowDefined } from './app/utils';
 import config from './config';
+import './content.css';
```
TomWoodward: so we're unconditionally loading the generic styles, but then conditionally loading specific styles on top of that? and presumably the generic styles would then target a selector that didn't exist?

TomWoodward (changes requested):

src/app/components/SkipToContentWrapper.spec.tsx
philschatz: It is sometimes false. When the `config.json` contains an entry for a book style to use then the value is false.

src/index.tsx
philschatz: yes, the selector in `content.css` is `#main-content[data-is-generic="true"]` so the CSS is not applied when the book is not configured to use the generic style.

src/app/content/components/Page.tsx
philschatz: Hmm, ok. I can try to switch cnx-recipes so that it always provides a style (either generic or book-specific). It seems like this code will still need to be somewhat dynamic since it would have to ask `cnx-recipes` for the book style (by providing the book slug name)

generic-styles/_content-common.less
philschatz: From the discussions Unified folks (aka "me") would be fixing any problems with the generic styles. Since there is no way to preview the rendered HTML in `cnx-recipes` I put the code in here.

src/app/content/components/Page.tsx
TomWoodward: i thought the point of these styles was that they wouldn't involve cnx-recipes? can we just put book specific css files in rex? can we rip out all the cnx-recipes integration until we do the baked styles down the road?

src/app/components/MainContent.tsx
```diff
@@ -2,15 +2,16 @@ import React, { SFC } from 'react';
...
-  {({registerMainContent}) => <div
-    className={className}
-    id={MAIN_CONTENT_ID}
-    ref={registerMainContent}
-    tabIndex={-1}
-  >
-    {children}
-  </div>}
-</Consumer>;
+const MainContent: SFC<{className?: string}> =
```
TomWoodward: can we revert this file? looks like only formatting changes

src/app/content/components/Page.tsx
```diff
@@ -92,17 +91,11 @@ export class PageComponent extends Component<PropTypes> {
...
-          <div
-            data-type='page'
-            ref={(ref: any) => this.container = ref}
-            dangerouslySetInnerHTML={{ __html: this.getCleanContent()}}
-          />
-        </div>
-      </MainContent>}
-    </BookStyles>;
+    return <MainContent className={this.props.className}>
+      <div
```
TomWoodward: is this extra div super necessary?

src/config.development.js
```diff
@@ -3,10 +3,21 @@ module.exports = {
...
+    '14fb4ad7-39a1-4eee-ab6e-3ef2482e3e22':{defaultVersion:'15.1'}, // Anatomy & Physiology
+    '2e737be8-ea65-48c3-aa0a-9f35b4c6a966':{defaultVersion:'20.1'}, // Astronomy
+    '4abf04bf-93a0-45c3-9cbc-2cefd46e68cc':{defaultVersion:'10.16'}, // Psychology
+    // '9ab4ba6d-1e48-486d-a2de-38ae1617ca84':{defaultVersion:'4.25'}, // Principles of Accounting, Volume 1: Financial Accounting. Broken because contains link to external content.
+    '4e09771f-a8aa-40ce-9063-aa58cc24e77f':{defaultVersion:'8.5'}, // Introduction to Business
+    'bc498e1f-efe9-43a0-8dea-d3569ad09a82':{defaultVersion:'7.1'}, // Principles of Economics 2e
+    'caa57dab-41c7-455e-bd6f-f443cda5519c':{defaultVersion:'18.1'}, // Prealgebra
+    // 'fd53eae1-fa23-47c7-bb1b-972349835c3c':{defaultVersion:'10.3'}, // Precalculus. Broken because contains link to external content.
+    '914ac66e-e1ec-486d-8a9c-97b0f7a99774':{defaultVersion:'4.1'}, // Business Ethics
+    '9d8df601-4f12-4ac1-8224-b450bf739e5f':{defaultVersion:'5.1'}, // American Government 2e
```
TomWoodward: before merge we should add these to production config

generic-styles/content-baked.less
```diff
@@ -0,0 +1,598 @@
...
+//
+//     #content { @import 'this_file.less'; }
+
+@import "./_make-block.less";
+
+
+#main-content {
+
+  // Override the defaults so this matches webview
+  max-width: 96rem;
```
TomWoodward: should remove this when my pr is merged

generic-styles/content-baked.less
```diff
@@ -0,0 +1,598 @@
...
+  margin: auto;
```
TomWoodward: and this

generic-styles/content-baked.less
```diff
@@ -0,0 +1,598 @@
...
+    font-weight: normal; //to ensure that all terms with this tag will display unbolded
+  }
+
+  // It is already displayed at the top of the page in blue
+  // the `>` is because collated titles (like Key Concepts) contain descendent
+  // document-title elements.
+  [data-type="page"],
+  [data-type="composite-page"] {
+    > [data-type="document-title"] {
+      display: none;
```
TomWoodward: are these not removed by the regex?

TomWoodward (changes requested):
content margin in src/app/content/__image_snapshots__/content-screenshotspec-ts-content-attribution-looks-right-1-snap.png looks broken, we should figure that out

cnx has this style 
```
[data-is-baked="true"] .media-body>#content span[data-type="media"] img {
    display: block;
}
```

which is what left-aligns the images, i don't see anything targeting `[data-is-baked="true"]` in this pr, are we missing some styles?

generic-styles/content-baked.less
philschatz: They probably are; I just copied the styles over; we don't really report CSS coverage but I'll remove.

src/app/content/components/Page.tsx
philschatz: I only added it because `MainContent` did not allow `dangerouslySetInnerHTML` but I can try adding that prop to `MainContent`

philschatz: I tried moving it but `MainContent` would have 2 `ref=` which I wasn't sure how to handle

TomWoodward (commented):
looks fine, i'd like to merge the content size PR before this one though if thats ok

src/config.development.js
philschatz: I am not sure what is meant by this... `config.production.js` uses the `REACT_APP_BOOKS` environment variable and `unified-deployment` does not seem to contain a value for `REACT_APP_BOOKS`. 

https://rex-web.herokuapp.com/ does specify a custom set of books in `REACT_APP_BOOKS`.

TomWoodward: the production books are currently set here

https://github.com/openstax/rex-web/blob/master/Dockerfile#L68

i figure we want to expand that right?

TomWoodward (changes requested):
this comment https://github.com/openstax/rex-web/pull/128#discussion_r285697774 and the one about the missing cnx styles are still outstanding

philschatz:
@tomjw64 RE https://github.com/openstax/rex-web/pull/128#pullrequestreview-238933459 : I was using an older version of the webview styles so I just pushed a commit with a newer version (last updated like 4 days ago)

TomWoodward (approved):

tomjw64:
> @tomjw64 RE [#128 (review)](https://github.com/openstax/rex-web/pull/128#pullrequestreview-238933459) : I was using an older version of the webview styles so I just pushed a commit with a newer version (last updated like 4 days ago)

Wrong Tom?

PR #130, opened 2019-05-01
WIP: Automated test for checking the attribution link

m1yag1:


pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,39 @@
+from selenium import webdriver
+import unittest
```
m1yag1: imports are not ordered correctly. they should be like the following:

```
# standard library first
import unittest  # import x first
from time import sleep  # from x import y second

# pip installed things second separated by line break
import pytest   # import x first 
from selenium import webdriver  # from x import y second
```
This takes a bit of experience and practice. The order is standard library imports like `unittest` and `time` are first. The selenium module is installed via pip so it is second separated by a line break. Within those there is another order. The ones with `import x` are first with `from x import y` being last. Within their little groups they are sorted alphabetically.

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,39 @@
...
+    def setUp(self):
+        self.browser = webdriver.Chrome()
+        self.browser.implicitly_wait(3)
+
+    def teardown(self):
+        self.browser.quit()
+
+    # c476303 attribution section is initially collapsed, expands when clicked
+    def test_c476303_attribution_initial_status(self):
+        self.browser.get("https://rex-web.herokuapp.com/books/college-physics/pages/preface")
```
m1yag1: It's useful to place any items being passed into methods in their own variable that describes them better. For example:

```
page_url = "https://rex-web.herokuapp.com/books/college-physics/pages/preface"
attribute_xpath_locator = "//*[@id="root"]/div[2]/div[4]/div/div[2]/div/div/details"

self.browser.get(page_url)
attribute_element = self.browser.find_element_by_xpath(attribute_xpath_locator)
attribute_element.click()
```

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,39 @@
...
+        # self.assertEqual(self.browser.get_attribute(details),open)
+
+        # url = 'http://localhost:8000/analyse/'
+
+        org = self.browser.find_element_by_xpath(
+            '//*[@id="root"]/div[2]/div[4]/div/div[2]/div/div/details'
+        )
+        # Find the value of org?
+        # val = org.get_attribute("class")
+        y = org.is_selected()
```
m1yag1: Same thing goes here. It's hard to know what `y` is. I know you are playing around with selenium trying to figure things out. That's ok but it's still helpful to name variables so they convey their intent.

`is_org_selected = org.is_selected()` might be better.

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,39 @@
...
+
+        # class="Attribution__Details-sc-11isnv6-4 fPeJdR"
+        # is_open = "open" in org.get_attribute("class")
+        print(y)
```
m1yag1: This print can be replaced with an assert. This will allow you to test that the element is selected as you expect.

assertEqual(is_org_selected, True)

m1yag1: You may not realize it but your are doing a bit of TDD development here while writing a test. You are asserting something that is probably going to fail b/c you don't know if you have the right selector. This is great!

In true TDD fashion you can now write the code to get the test to pass. 💪

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,63 @@
...
+
+class TestCitationtexturl():
+    #def setUp(self):
+        #self.browser = webdriver.Chrome()
+        #self.browser.implicitly_wait(3)
+    #def teardown(self):
+       # self.browser.quit()
+
+
+
```
m1yag1: We need to add two more decorators here. We are currently using the pytest-testrail plugin which allows us to integrate with testrail. In order to let the plugin know we need to add a special marker to the test. 

`@markers.test_case("<test_case_id>")`
example: https://github.com/openstax/rex-web/blob/master/pytest-selenium/tests/test_toc.py#L5

The next marker we need to add tells the test framework that this test is "nondestructive". What this means is that this test is ok to run in production and that it doesn't do anything to manipulates data. If this marker is not here anytime the `base-url` matched on production it would skip this test.

`@markers.nondestructive`

pytest-selenium/tests/test_attribution_pom.py
```diff
@@ -0,0 +1,127 @@
...
+class TestCitationtexturl():
+
+#c476303 attribution section is initially collapsed, expands when clicked
+    @markers.test_case("C476303")
+    @markers.parametrize("book_slug,page_slug", [("college-physics", "preface")])
+    @markers.nondestructive
+
+    def test_attribution_initial_status(self, selenium, base_url, book_slug, page_slug):
+        
+#Given - the page {base_url}/books/college-physics/pages/preface
```
m1yag1: We always indent 4 spaces even if it is a comment. I'll show you how to not worry about formatting soon. We need to install the pre-commit hooks and before every commit you'll be able to run checks against your code to make sure you have no linting or formatting issues.

pytest-selenium/tests/test_attribution_pom.py
```diff
@@ -0,0 +1,127 @@
...
+#click on attribution again and verify it closes
+        attribution.attribution_click()
+        sleep(1)
+        assert not attribution.is_open
+        print (attribution.is_open) #none
+
+
+
+
+# c476304 citation/attribution section collapses by default when you navigate to a new page
```
m1yag1: In python we don't place comments above function names. We're encouraged to use something known as docstrings. The format for docstrings are the quotation marks after the function is defined like so:

```python
    def offscreen_click(self, element):
        """Clicks an offscreen element.

        Clicks the given element, even if it is offscreen, by sending the ENTER key.
        Returns the element.
        """
        # We actually navigate using the ENTER key because scrolling the page can be flaky
        # https://stackoverflow.com/a/39918249
        element.send_keys(Keys.ENTER)
        return element
    """

The general thing is the first line is a very short description. A longer description starts after a line break. We end with the closing quotations.

File: pytest-selenium/tests/test_attribution_pom.py (hunk 6 above)
Comment (Malar-Natarajan) [reply pytest-selenium/tests/test_attribution_pom.py]: Haha...I intentionally deleted the spaces for the given-when-then lines so that the scenario stands out and easily readable...I will change that though

File: pytest-selenium/pages/content.py [hunk 8]
```diff
@@ -50,19 +106,33 @@ def click_toc_toggle_button(self):
             self.offscreen_click(self.toc_toggle_button)
             return self.page.sidebar.wait_for_region_to_display()
 
+
+
     class SideBar(Region):
         _root_locator = (By.CSS_SELECTOR, "[aria-label='Table of Contents']")
-
+        #_TOC_section2_locator = (By.XPATH, "//nav/ol/li[3]/nav/ol/li[1]/a")
+        _TOC_section2_locator = (By.CSS_SELECTOR, '[href$=to]')
```
m1yag1: What is `_TOC_section2_locator`? is there not a more descriptive name. It's possible that there isn't.

Malar-Natarajan: I am trying to click on a link in TOC. Here I am referring to section 2 link in TOC

pytest-selenium/pages/content.py
```diff
@@ -27,6 +31,59 @@ def toolbar(self):
...
+        return self.driver.current_url
+    
+
+     
+
+
+    class Attribution(Region):
+        _attribute_status_locator = (By.XPATH, "//details[contains(@class,'Attribution__Details')]")
+        #_root_locator = (By.CSS_SELECTOR, '[data-testid="attribution-details"]')
+        _root_locator = (By.XPATH,"//summary[contains(@class,'Attribution__Summary')]")
```
m1yag1: As a matter of style we place the `_root_locator` as the first item.

pytest-selenium/tests/test_attribution_pages.py
```diff
@@ -0,0 +1,102 @@
...
+        #Expand attribution
+        attribute_element.click()
+        attribute = attribute_element.get_attribute("open")
+        print(attribute)  #true
+
+
+        #Navigate to Next page and verify attribution is closed by default
+        Next_element_locator = self.browser.find_element_by_css_selector("a[aria-label='Next Page']")
+        self.offscreen_click(Next_element_locator)
+        sleep(1)
```
m1yag1: We need to figure out a way to not use explicit sleeps. We want to use implicit sleeps. The same way a page is detected as loaded we can add for a page region. We can also look and see if the region is displayed.

pytest-selenium/tests/test_attribution_pom.py
```diff
@@ -0,0 +1,138 @@
...
+        
+    #AND the attribution section is expanded
+        attribution = content.attribution
+        sleep(1)
+        attribution.attribution_click()
+        url_after_attribution_click = content.current_url
+        #print(content.current_url)
+     
+    #Then - the urls in the the attribution should reference current page in rex
+        assert url_before_attribution_click == url_after_attribution_click
```
m1yag1: Awesome!

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,63 @@
...
+#c476303 attribution section is initially collapsed, expands when clicked
+    def test_c476303_attribution_initial_status(self, selenium, base_url):
+        
+        page_url = base_url + "/books/college-physics/pages/preface"
```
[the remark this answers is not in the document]
Malar-Natarajan: This is updated now with latest commit

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: This is done with latest commit

pytest-selenium/tests/test_attribution_pom.py
Malar-Natarajan: Updated the comments to indent 4 spaces

pytest-selenium/tests/test_attribution_pom.py
Malar-Natarajan: Corrected in few places. Need to discuss more

pytest-selenium/pages/content.py
Malar-Natarajan: I have for now used toc_element. But this part of the script needs rework after the TOC layout is finalized. So I am not spending more time here

pytest-selenium/pages/content.py
Malar-Natarajan: Done

pytest-selenium/tests/test_attribution_pages.py
Malar-Natarajan: Updated with region lookup in 2 test cases. C476304 still works only with sleep. I will look into this when reworking the script for TOC layout changes

pytest-selenium/tests/test_attribution_pom.py
Malar-Natarajan: Agreed

pytest-selenium/pages/content.py
```diff
@@ -27,6 +31,55 @@ def toolbar(self):
...
+            
+
+    @property
+    def current_url(self):
+        return self.driver.current_url
+     
+
+
+    class Attribution(Region):
+        _root_locator = (By.XPATH,"//summary[contains(@class,'Attribution__Summary')]")
```
TomWoodward: can we use css selector instead of xpath? i think they're in general more readable

TomWoodward: also i'm pretty sure there is a testid on this now

pytest-selenium/tests/test_attribution_pom.py
TomWoodward: the test assertion `the urls in the the attribution should reference current page in rex` is about making sure the urls in the attribution text reference the current page, i don't see that happening here?

Malar-Natarajan: The scenario is making sure page url doesn't change when we expand the attribution on a page:
I am storing the page url in `url_before_attribution_click'
Then click on Attribution and storing the new url in `url_after_attribution_click`
comparing the 2 url's in assert statement.

Did you mean this or am I not getting your question.

pytest-selenium/pages/content.py
m1yag1: Yes, I added one.

PR #131, opened 2019-05-03
Fixed table of contents button interactions

m1yag1:
* Refactored the content page object to include new regions that
describe the page. The regions are Navbar, Sidebar, Sidebar header, and
Toolbar.
* Refactored the test to account for the differences in resolutions and
the order of toggling the Sidebar. When the resolution is desktop we
click the toc button on the sidebar to close and the toc button on the
toolbar to open. It's the other way around for mobile and tablet
resolutions.

### DIFF
--- a/pytest-selenium/pages/content.py
+++ b/pytest-selenium/pages/content.py
@@ -4,46 +4,73 @@ from selenium.webdriver.support import expected_conditions as expected
 from pages.base import Page
 from regions.base import Region
 
-# TODO: Ensure only 1 <nav> element exists on the Page
-NAV_SELECTOR = "#root > div > div > div > div > nav"
-
 
 class Content(Page):
     URL_TEMPLATE = "/books/{book_slug}/pages/{page_slug}"
 
     _body_locator = (By.TAG_NAME, "body")
     _main_content_locator = (By.CSS_SELECTOR, "h1")
-    _table_of_contents_button_locator = (
-        By.CSS_SELECTOR,
-        "[aria-label='Click to close the Table of Contents'], [aria-label='Click to open the Table of Contents']",
-    )
-    _table_of_contents_nav_locator = (By.CSS_SELECTOR, NAV_SELECTOR)
+    _sidebar_locator = (By.CSS_SELECTOR, "[aria-label='Table of Contents")
 
     @property
     def loaded(self):
         return self.find_element(*self._body_locator).get_attribute("data-rex-loaded")
 
     @property
-    def table_of_contents_nav(self):
-        return self.find_element(*self._table_of_contents_nav_locator)
+    def navbar(self):
+        return self.NavBar(self)
 
     @property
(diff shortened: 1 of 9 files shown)


pytest-selenium/pages/content.py
```diff
@@ -4,46 +4,72 @@
...
+        return self.ToolBar(self)
 
     @property
-    def table_of_contents(self):
-        return self.TableOfContents(self)
+    def sidebar(self):
+        return self.SideBar(self)
+
+    class NavBar(Region):
+        _root_locator = (By.CSS_SELECTOR, "div[class^='NavBar__TopBar']")
```
TomWoodward: we should just add data selectors

pytest-selenium/tests/test_toc.py
```diff
@@ -1,23 +1,62 @@
...
-    #   AND: The table of contents toggle button is clicked
     content = Content(selenium, base_url, book_slug=book_slug, page_slug=page_slug).open()
 
-    toc = content.table_of_contents
-    content.click_table_of_contents_button()
+    toolbar = content.toolbar
+    sidebar = content.sidebar
+
+    # AND: Window width is 1024 or greater (Desktop)
+    if selenium.get_window_size()["width"] > 1024:
```
TomWoodward: magic numbers should be defined somewhere, maybe a top level region with a `isDesktop()` or `isMobile()`?

TomWoodward (changes requested):

pytest-selenium/pages/content.py
m1yag1: Can you show me how to add them? I actually don't think this is that terrible. It works really well and I don't see the prefix portion of these elements changing but I could be wrong.

m1yag1: The good thing is that the regions are now more specific. I really like that part. This object representation is a lot closer to how the page is really laid out. This is also much better than XPATH.

m1yag1: @TomWoodward ping

TomWoodward: the class names there are auto generated based on the component names, so if we ever rename a component all the ui tests will break. i would rather use the more explicit coupling of the data selectors.

m1yag1: Done.

TomWoodward (changes requested):
this looks fine, but needs the snaps updated

TomWoodward (approved):

TomWoodward:
@m1yag1 linter errors
```
ERROR: /home/travis/build/openstax/rex-web/src/app/components/NavBar.tsx[86, 25]: " should be '
ERROR: /home/travis/build/openstax/rex-web/src/app/content/components/Sidebar.tsx[254, 58]: " should be '
ERROR: /home/travis/build/openstax/rex-web/src/app/content/components/Toolbar.tsx[146, 23]: " should be '
```

PR #134, opened 2019-05-03
fix: notification display

TomWoodward:


src/app/notifications/components/Notifications.tsx
```diff
@@ -34,19 +34,26 @@ const connector = connect(
...
 );
 
+const notificationWidth = 30;
+
 // TODO - magic number to be replaced in `top` when scroll behavior is developed in openstax/unified#179
 export default styled(connector(Notifications))`
-  z-index: 2; /* on top of the navbar */
+  width: ${notificationWidth}rem;
+  height: 0;
+  margin-left: calc(100% - ${notificationWidth}rem);
```
philschatz: Why did `right: 0` not work? Is it because it would have covered the `Login` button?

philschatz (approved):

src/app/notifications/components/Notifications.tsx
TomWoodward: because its position sticky left/right don't really do anything (they do but not in this case)

PR #135, opened 2019-05-06
chore: update mobile breakpoint

TomWoodward:


src/app/content/components/PrevNextBar.tsx
```diff
@@ -35,10 +35,10 @@ interface HidingContentLinkProps {
   side: 'left' | 'right';
 }
 // tslint:disable-next-line:variable-name
-const HidingContentLinkComponent: React.SFC<HidingContentLinkProps> = ({page, book, ...props}) =>
+const HidingContentLinkComponent: React.SFC<HidingContentLinkProps> = ({page, book, side, ...props}) =>
```
philschatz: This same change occurs in a different PR; what does it do? It seems like it excludes the `side` field from being passed to `ContentLink`. Is it fixing a generic problem in `master`?

philschatz (approved):

src/app/content/components/PrevNextBar.tsx
TomWoodward: yea the change has been moving through open pr's by osmosis. it is fixing an issue on master that causes the snaps to fail

PR #142, opened 2019-05-14
fix selenium screen sizes

TomWoodward:
i python good

### DIFF
--- a/pytest-selenium/tests/test_toc.py
+++ b/pytest-selenium/tests/test_toc.py
@@ -1,3 +1,4 @@
+import pytest
 from pages.content import Content
 from . import markers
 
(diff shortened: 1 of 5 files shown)


pytest-selenium/tests/test_toc.py
```diff
@@ -59,3 +59,7 @@ def test_toc_toggle_button_opens_and_closes(selenium, base_url, book_slug, page_
         sidebar.header.click_toc_toggle_button()
 
         assert not sidebar.header.is_displayed
+
+    else:
+
+        assert false
```
m1yag1: This is a great update to the test! Yea, if it's neither mobile or desktop it will pass which is no bueno. this is is why it's great having more eyes on things. 

An improvement on this is to use `pytest.fail("Some Message")` This will fail the test and return the proper message on why.

m1yag1 (changes requested):
One small change but overall a great example of boy scout coding.

"Leave code better than you found it."

m1yag1 (approved):

PR #144, opened 2019-05-16
fix: double processing math

TomWoodward:


public/index.html
```diff
@@ -10,7 +10,7 @@
     -->
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
 
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&amp;delayStartupUntil=configured" async></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.2/MathJax.js?config=TeX-MML-AM_HTMLorMML-full&delayStartupUntil=configured" async></script>
```
philschatz: Oh weird, what problem was this creating?

philschatz (approved):
It's so speedy now!

public/index.html
TomWoodward: as far as i can tell this wasn't actually breaking anything, the real fix was adding `skipStartupTypeset: true,`, but this seemed weird so i fixed it

PR #145, opened 2019-05-16
chore: added browser test to verify attribution section shows url of current page

Malar-Natarajan:
C476302
Making sure that the page url does not change when we expand/collapse the citation/attribution section


m1yag1 (approved):

pytest-selenium/pages/content.py
```diff
@@ -73,3 +77,23 @@ def click_toc_toggle_button(self):
                 return self.wait.until(
                     expected.invisibility_of_element_located(self.toc_toggle_button)
                 )
+
+    class Attribution(Region):
+        _root_locator = (By.CSS_SELECTOR, '[data-testid="attribution-details"]')
+        _attribute_status_locator = (By.CSS_SELECTOR, 'summary[aria-label="Citation/Attribution"]')
```
TomWoodward: it would be better to use `'[data-testid="attribution-details"] summary'` here, in case of any future copy changes.

the name of this variable is also confusing to me

pytest-selenium/tests/test_attribution.py
```diff
@@ -0,0 +1,23 @@
+from pages.content import Content
+from tests import markers
+
+
+@markers.test_case("C476302")
```
TomWoodward: this doesn't match this test

TomWoodward (changes requested):
need to run `yarn test:unit -u` to update react snapshots

pytest-selenium/pages/content.py
m1yag1: the variable could be better. 👍 

how about `_attribution_open_locator`?

pytest-selenium/tests/test_attribution.py
m1yag1: do you mean this doesn't match the test case in test rail?

pytest-selenium/pages/content.py
TomWoodward: or `_attribution_toggle_locator` maybe?

pytest-selenium/tests/test_attribution.py
TomWoodward: yes, the test in testrail is testing something different

Malar-Natarajan: I think we need to talk to understand this conflict

Malar-Natarajan: I finally figured out what Tom is saying and yes my test is not checking the intended functionality. Thanks for pointing it out Tom. 
I actually need to check the urls that are within the citation section (section url) refer to the current page url

Malar-Natarajan: @TomWoodward I am rewriting this script.
Section URL within the attribution always defaults to current page.
Do we also know that the Book url & Access for free url always points to Chapter 1 or will there be a change to it?

TomWoodward: they point to the "Default page" of the book, which is different for different books but shouldn't change. its defined as the first page in the first chapter (recursively) so if there are units its the first page of the first chapter of the first unit. 

this is to avoid defaulting to the preface, which we generally try to ignore

m1yag1: i'm assuming that there will be some redirects to rex from openstax.org in the future. I'm seeing openstax.org in these links but they return 404 when clicked. 
image cc: @TomWoodward

Malar-Natarajan: Based on Tom's comments this test would just cover the section URL and I am writing separate tests for the Book URL and access for free URL

TomWoodward: @m1yag1 the openstax.org links are correct, that is what the rex url will be after launch

m1yag1: 👍 . it would be nice to continue to use the `base_url` we pass in to build the url to do the verification ... we can make it work.

Malar-Natarajan: @m1yag1 
Since the section url in attribution is pointing to production url, I did not find an easy way other than hardcoding the base url. But if its achievable the other way around, we can do that.

pytest-selenium/pages/content.py
Malar-Natarajan: updated to _attribution_toggle_locator

TomWoodward (approved):

PR #150, opened 2019-05-24
fix: mathjax rendering quirk

philschatz:
`lim f` was being rendered as `limf`. See https://github.com/openstax/webview/issues/2195#issuecomment-495681451

The latest version of MathJax is 2.7.5 (June 2018) and there is a v3 beta is coming along. I opted for 2.7.4 because that's the version being used on cnx.org currently. But we could just skip over to 2.7.5 or v3 since QA will be checking all the books anyway.


TomWoodward (approved):
/shrug

PR #153, opened 2019-05-28
fix: saucelabs

TomWoodward:


pytest-selenium/Makefile-py
```diff
@@ -42,6 +42,7 @@ test-rail-sauce: venv
 		--driver SauceLabs \
 		--capability avoidProxy True \
 		--capability browserName $(BROWSER) \
+		--capability screenResolution 2560x1600 \
```
TomWoodward: @m1yag1 is there any reason not to just set this to the max all the time?

m1yag1: Yea that should be fine.

m1yag1 (approved):

PR #154, opened 2019-05-29
fix: generic styles to match cnx

KarinaMendez2:
Refs openstax/unified#284


generic-styles/index.less
```diff
@@ -1,3 +1,31 @@
 #main-content {
   @import "./content-baked.less";
 }
+#main-content > .os-glossary-container > dl > dt {
```
philschatz: These can be nested inside the `#main-content {` block above. So something like:

```less
#main-content {
  @import "./content-baked.less";

  > .os-glossary-container > dl > dt {
    font-weight: 700;
  }
  // ...
}
```

generic-styles/index.less
```diff
@@ -1,3 +1,31 @@
...
+  font-weight: 700;
+}
+#main-content .insider-perspective h3 {
```
philschatz: Which problem was this addressing? It looks like `.insider-perspective` is a book-specific style and the term "insider-perspective" does not occur anywhere in [webview](https://github.com/openstax/webview).

Maybe this was originally a bootstrap style that applies to all `<h3>` elements?

src/app/content/genericstyles.screenshotspec.ts
```diff
@@ -0,0 +1,40 @@
+/** @jest-environment puppeteer */
+import {
+  fullPageScreenshot,
+  navigate,
+  setDesktopViewport,
+  setMobileViewport
+} from '../../test/browserutils';
+
+const CHEMISTRY_BOOK = '/books/chemistry-2e';
```
philschatz: Rather than relying on the chemistry book being built for the unit tests, tests would run faster and any changes would be clearer if a "fake" book was used for screenshot tests.

[src/app/components/NavBar/index.screenshotspec.ts#L9](https://github.com/openstax/rex-web/blob/master/src/app/components/NavBar/index.screenshotspec.ts#L9) can serve as an example.

src/app/content/genericstyles.screenshotspec.ts
```diff
@@ -0,0 +1,40 @@
...
+/*const SPLASH_IMAGE_FULL_SIZE = `${CHEMISTRY_BOOK}/pages/6-5-periodic-variations-in-element-properties`;
+const TABLE_SEPARATIONS = `${CHEMISTRY_BOOK}/pages/7-5-strengths-of-ionic-and-covalent-bonds`;
+const INSIDER_PERSPECTIVE_PAGE = '/books/american-government-2e/pages/2-4-the-ratification-of-the-constitution#fs-id1166227577868';
+const KEY_TERMS_BOLDED = '/books/business-ethics/pages/1-key-terms';*/
+
+describe('content', () => {
+  it('looks right', async() => {
+    setDesktopViewport(page);
+    await navigate(page, SPLASH_IMAGE_FULL_WIDTH);
+    const screen = await fullPageScreenshot(page);
```
philschatz: Additionally, adding `await page.emulateMedia('print');` before this line would help reduce the number of false positives in screenshots when the ToC or nav bar changes because they would not be included in the screenshot. See [src/app/content/print.screenshotspec.ts#L36](https://github.com/openstax/rex-web/blob/master/src/app/content/print.screenshotspec.ts#L36)

philschatz (commented):
The changes look good. I just had a couple of questions.

src/app/content/components/__snapshots__/PrevNextBar.spec.tsx.snap
```diff
@@ -511,8 +523,24 @@ exports[`PrevNextBar renders only \`prev\` element correctly 1`] = `
     </svg>
     Previous
   </a>
-  <span
-    aria-hidden={true}
-  />
+  <a
```
TomWoodward: this test is supposed to be testing when only the prev element is rendered, now we've added another page so it is no longer testing the right thing.

maybe we should move our new page to not be at the end of the book?

TomWoodward (changes requested):
i can't make the screenshots fail locally, dunno what is going on with travis. maybe if we make the fonts work on the print view it'll work better?

TomWoodward (approved):

TomWoodward:
stylelint is failing but otherwise looks good. definitely wait for @philschatz's review as well

philschatz (approved):
And nice screenshots in the diff!

kjdav:
This passes acceptance tests! But as Tom pointed out in another thread, there are conflicts to resolve.

PR #164, opened 2019-06-05
fix: Toc fixes

TomWoodward:


src/app/content/components/Sidebar/styled/index.tsx
```diff
@@ -11,26 +11,35 @@ import ContentLinkComponent from '../../ContentLink';
...
+  (element) => 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890.'.split('').reduce(
+    (result, char) => (result[char] = ((element.innerText = char) && element.getBoundingClientRect().width)) && result,
+    {}
+  )
+)(element = document.querySelector('[data-testid=toc] [aria-label="Current Page"] a')
+  .appendChild(document.createElement('span'))
+);
+ */
+const numberCharacterWidth = .7796875;
+const letterCharacterWidth = 1.0375;
```
philschatz: Each number _does_ have the same width in the current font. Interesting.

src/app/content/components/Sidebar/styled/index.tsx
```diff
@@ -11,26 +11,35 @@ import ContentLinkComponent from '../../ContentLink';
 
 export * from './wrapper';
 
-const numberCharacterWidth = 1;
-const numberPeriodWidth = 0.2;
+/* to regenerate these numbers, run this in a rex browser window
+(
+  (element) => 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890.'.split('').reduce(
```
philschatz: Since the code assumes all numbers have the same width, and it seems to use the widest letter it seems clearer to write:

```suggestion
  (element) => 'wW1234567890.'.split('').reduce(
```

philschatz (approved):

src/app/content/components/Sidebar/styled/index.tsx
TomWoodward: yea letters not so much though. since letters are less common in the sidebar i just went with one of the wider ones

PR #165, opened 2019-06-05
fix: firefox

TomWoodward:


src/app/content/components/Attribution.tsx
```diff
@@ -64,26 +54,18 @@ const Content = styled.div`
 `;
 
 // tslint:disable-next-line:variable-name
-const Details = styled.details`
+const AttributionDetails = styled(Details)`
```
philschatz: Why wouldn't this use the `DetailsComponent` ?

src/app/content/components/Sidebar/styled/index.tsx
```diff
@@ -131,7 +105,7 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
   }}
 `;
 
-type DetailsComponentProps = {defaultOpen: boolean} & React.HTMLProps<HTMLDetailsElement>;
+interface DetailsComponentProps {defaultOpen: boolean; open: boolean; }
```
philschatz: It seems like the definition of `DetailsComponent` is generic enough to apply to the nav bar _and_ the attribution. Is that not the case?

philschatz (commented):
Just a couple of questions about code organization, otherwise looks great

src/app/content/components/Sidebar/styled/index.tsx
```diff
@@ -11,6 +9,8 @@ import ContentLinkComponent from '../../ContentLink';
 
 export * from './wrapper';
 
+export {Summary, ExpandIcon, CollapseIcon} from '../../../../components/Details';
```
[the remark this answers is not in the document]
TomWoodward: Just In Case ™️ we decide to add some style overrides to those then we don't have to change the imports in the sidebar

if it bothers you we can remove, not a big deal

src/app/content/components/Attribution.tsx
TomWoodward: the DetailsComponent in sidebar only exists to provide the `defaultOpen` functionality so that when you nav around the changing props don't make the details re-render themselves and collapse when they shouldn't

i suppose it wouldn't _hurt_ to use it everywhere, but its also extra logic that nothing else needs

kjdav:
@TomWoodward Suah says this fixes it for her in Firefox and Edge

src/app/content/components/Sidebar/styled/index.tsx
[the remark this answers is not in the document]
philschatz: I just didn't expect to see it, that's all.

philschatz (approved):

kjdav:
@TomWoodward I think something broke? i don't see the skip to content button and I can't tab to login or print.

PR #166, opened 2019-06-05
feat: make mobile os nav bigger

TomWoodward:


TomWoodward:
need to bump this to 55px for both desktop and mobile

PR #169, opened 2019-06-06
fix: open in new tab

philschatz:
Command-clicking on links to make them open in a new tab did not work because the ContentLink handler consumed the event.

Oh, this was already reported 3 days ago: openstax/unified#301


TomWoodward (changes requested):
i think the logic in the `Page` component which handles clicking content links inside the body of the content will still be broken.

i think the logic in Page and in ContentLink is similar enough that we should try to share it between these two places

philschatz:
I fixed the links in `Page` content. I like the idea of combining the logic since it does the same thing in 2 different places but I had trouble resolving the fact that the event has 2 different types: React.MouseEvent<HTMLAnchorElement>` and `dom.MouseEvent`. Any ideas on resolving those 2 types? Maybe an Interface would work?

TomWoodward:
looking more closely, the two blocks of code are not as similar as i remembered. they're similar, but each has bits of code that the other does not and the shared logic is only like 3 statements (and the data for navigate is compiled differently). i think it is fine to leave them separate for now.

TomWoodward (approved):

PR #172, opened 2019-06-10
Test attribution status in new page

Malar-Natarajan:
GIVEN: A page URL in the format of {base_url}/books/{book_slug}/pages/{page_slug}
AND: The citation/attribution tab is open
WHEN: navigating via (toc link, next link, previous link)
THEN: the citation/attribution section is not open on the new page


pytest-selenium/pages/content.py
```diff
@@ -1,20 +1,36 @@
...
+    _next_locator = (By.CSS_SELECTOR, "[aria-label='Next Page']")
+    _previous_locator = (By.CSS_SELECTOR, "[aria-label='Previous Page']")
 
     @property
     def loaded(self):
         return self.find_element(*self._body_locator).get_attribute("data-rex-loaded")
 
+    @property
+    def title(self):
+        return self.find_element(*self._title_locator)
```
TomWoodward: titiles apply to all pages, not just content pages, this could go in the super class

pytest-selenium/pages/content.py
```diff
@@ -31,6 +47,24 @@ def sidebar(self):
...
+    @property
+    def section_url_within_attribution(self):
+        return self.find_element(*self._section_url_locator)
+
+    def click_next_link(self):
+        title_before_click = self.title.get_attribute("innerHTML")
+        self.offscreen_click(self.next_link)
+        self.wait.until(
+            lambda _: title_before_click != (self.title.get_attribute("innerHTML") or "")
+        )
```
TomWoodward: this `wait.util` on the title is neat, but maybe we should move it to a helper function so we don't have to put it in multiple places? and then if we ever update the logic for being able to tell when navigation is done it'll only be in one place?

pytest-selenium/tests/test_attribution.py
```diff
@@ -49,3 +50,59 @@ def test_attribution_collapsed_by_default_expands_when_clicked(
...
+    # WHEN: Navigating via next link
+    content.click_next_link()
+
+    # THEN: The citation/attribution section is not open on the new page
+    assert not attribution.is_open
+
+    attribution.click_attribution_link()
+
+    # WHEN: Navigating via Back link
+    content.click_previous_link()
```
TomWoodward: i think it might be better to have 3 separate tests for next/prev/toc

pytest-selenium/tests/test_attribution.py
```diff
@@ -49,3 +50,59 @@ def test_attribution_collapsed_by_default_expands_when_clicked(
...
+
+    if content.is_desktop:
+        chapter = toc.chapters[12]
+        chapter.click()
+        page = chapter.pages[4]
+        page.click()
+
+        # THEN: the citation/attribution section is not open on the new page
+        assert not attribution.is_open
+        attribution.click_attribution_link()
```
TomWoodward: why click it at the end?

pytest-selenium/tests/test_attribution.py
```diff
@@ -49,3 +50,59 @@ def test_attribution_collapsed_by_default_expands_when_clicked(
...
+
+    # THEN: The citation/attribution section is not open on the new page
+    assert not attribution.is_open
+
+    attribution.click_attribution_link()
+
+    # WHEN: Navigating via TOC link
+    toolbar = content.toolbar
+    toc = content.sidebar.toc
+    from time import sleep
```
TomWoodward: unused?

pytest-selenium/pages/content.py
```diff
@@ -78,6 +116,68 @@ def click_toc_toggle_button(self):
                     expected.invisibility_of_element_located(self.toc_toggle_button)
                 )
 
+        class TableOfContents(Region):
+            _root_locator = (By.CSS_SELECTOR, '[data-testid="toc"]')
+            _chapter_toggle = (By.CSS_SELECTOR, "ol li details")
+            _toc_chapter_name_locator = (By.XPATH, "(//ol/li/details/summary/div/span)")
+
+            @property
+            def is_chapter_expanded(self):
```
TomWoodward: this seems to be unused

pytest-selenium/pages/content.py
```diff
@@ -78,6 +116,68 @@ def click_toc_toggle_button(self):
...
+                return self.find_element(*self._chapter_toggle).get_attribute("open")
+
+            @property
+            def number_of_chapters(self):
```
TomWoodward: also this

pytest-selenium/pages/content.py
```diff
@@ -78,6 +116,68 @@ def click_toc_toggle_button(self):
...
+
+            @property
+            def chapters(self):
+                return [
+                    self.ContentChapter(self.page, self.root, index)
+                    for index in range(len(self.find_elements(*self._chapter_toggle)))
+                ]
+
+            @property
+            def chapter_name(self):
```
TomWoodward: and this

pytest-selenium/pages/content.py
```diff
@@ -78,6 +116,68 @@ def click_toc_toggle_button(self):
...
+
+                @property
+                def pages(self):
+                    return [
+                        self.ContentPage(self.page, self.root, index)
+                        for index in range(len(self.find_elements(*self._page_link_locator)))
+                    ]
+
+                @property
+                def number_of_pages(self):
```
TomWoodward: also unused?

pytest-selenium/pages/content.py
```diff
@@ -78,6 +116,68 @@ def click_toc_toggle_button(self):
...
+            @property
+            def is_chapter_expanded(self):
+                return self.find_element(*self._chapter_toggle).get_attribute("open")
+
+            @property
+            def number_of_chapters(self):
+                return len(self.find_elements(*self._chapter_toggle))
+
+            @property
+            def chapters(self):
```
TomWoodward: you have this organized such that the toc has chapters and the chapter have pages but that is not true, the toc can have pages directly and the chapters can have more chapters inside

pytest-selenium/regions/content_item.py
```diff
@@ -0,0 +1,40 @@
...
+    def __init__(self, page, parent_root, index):
+        root = parent_root.find_element(
+            By.XPATH, self._root_locator_template.format(index=index + 1)
+        )
+        self.parent_root = parent_root
+        self.index = index
+        super().__init__(page, root)
+
+    @property
+    def has_chapter_section(self):
```
TomWoodward: unused?

pytest-selenium/regions/content_item.py
```diff
@@ -0,0 +1,40 @@
...
+        return self.is_element_displayed(*self._chapter_section_span_locator)
+
+    @property
+    def chapter_section_span(self):
+        return self.find_element(*self._chapter_section_span_locator)
+
+    @property
+    def chapter_section(self):
```
TomWoodward: unused?

pytest-selenium/regions/content_item.py
```diff
@@ -0,0 +1,40 @@
...
+        return self.chapter_section_span.text
+
+    @property
+    def title_span(self):
+        return self.find_element(*self._title_span_locator)
+
+    @property
+    def title(self):
```
TomWoodward: uused?

TomWoodward (changes requested):

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: @TomWoodward 
Wouldn't that be too overwhelming to have just two lines of code in one test...I preferred this way because testrail test was designed like this to include all possible navigations in one test

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: @TomWoodward 
No great reason. But my hypothesis is just making sure attribution is clickable after navigating to a new page via TOC. 
I can remove that if it hurts you :-)

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: import sleep was my bad. I forgot to remove it.
The toolbar assignment is used with the the mobile resolution. I moved under the mobile If clause now.
The toc assignment is used in both desktop & mobile. So stays in the same place

pytest-selenium/pages/content.py
Malar-Natarajan: Removed

pytest-selenium/pages/content.py
Malar-Natarajan: removed

pytest-selenium/pages/content.py
Malar-Natarajan: removed

pytest-selenium/pages/content.py
Malar-Natarajan: removed

pytest-selenium/regions/content_item.py
Malar-Natarajan: removed

pytest-selenium/regions/content_item.py
Malar-Natarajan: removed. 
I was playing around the TOC elements and added all these. I will make sure not to add unised attributes/elements in future

pytest-selenium/regions/content_item.py
Malar-Natarajan: removed

pytest-selenium/pages/content.py
Malar-Natarajan: created a helper function

pytest-selenium/pages/content.py
Malar-Natarajan: created a helper function for title. I shall add this as super class when we have more pages coming into Rex

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: I removed the TOC part from this test since the TOC structure needs more time for me to put in a neat framework

pytest-selenium/pages/content.py
Malar-Natarajan: TOC is removed from this PR. But I will keep this point in mind when doing TOC

pytest-selenium/pages/base.py
```diff
@@ -44,3 +44,15 @@ def offscreen_click(self, element):
         # https://stackoverflow.com/a/39918249
         element.send_keys(Keys.ENTER)
         return element
+
+    def offscreen_click_and_wait_for_new_title_to_load(self, element):
```
TomWoodward: name binds to implementation, how about

```suggestion
    def click_and_wait_for_load(self, element):
```

pytest-selenium/pages/base.py
```diff
@@ -44,3 +44,15 @@ def offscreen_click(self, element):
...
+        """Clicks an offscreen element and waits for title to load.
+
+        Clicks the given element, even if it is offscreen, by sending the ENTER key.
+        Returns after loading the last element (title) of the page).
+        """
+        title_before_click = self.title_before_click
```
TomWoodward: this function is using other functions defined in the subclass, they need to be moved here

pytest-selenium/regions/base.py
```diff
@@ -16,4 +18,8 @@ def offscreen_click(self, element=None):
         Clicks the given element, even if it is offscreen, by sending the ENTER key.
         Returns the element.
         """
-        return self.page.offscreen_click(element or self.root)
+        # We actually navigate using the ENTER key because scrolling the page can be flaky
+        # https://stackoverflow.com/a/39918249
+        # return self.page.offscreen_click(element or self.root)
+        element.send_keys(Keys.ENTER)
```
TomWoodward: i wonder if this'll fix the tests in edge 🤔

pytest-selenium/regions/content_item.py
```diff
@@ -0,0 +1,28 @@
...
+
+class ContentItem(Region):
+    _chapter_section_span_locator = (By.CSS_SELECTOR, "span.os-number")
+    _title_span_locator = (By.CSS_SELECTOR, "span.os-text")
+
+    def __init__(self, page, parent_root, index):
+        root = parent_root.find_element(
+            By.XPATH, self._root_locator_template.format(index=index + 1)
+        )
+        self.parent_root = parent_root
```
TomWoodward: is this used for something?

TomWoodward (changes requested):

pytest-selenium/pages/base.py
Malar-Natarajan: Changed as mentioned

pytest-selenium/regions/content_item.py
Malar-Natarajan: Removed with lates commit. This file is not needed now since we removed the TOC part from this PR

m1yag1 (approved):

pytest-selenium/pages/base.py
```diff
@@ -44,3 +44,23 @@ def offscreen_click(self, element):
         # https://stackoverflow.com/a/39918249
         element.send_keys(Keys.ENTER)
         return element
+
+    def click_and_check_reference_object_has_changed(self, click_element, ref_object, ref_property):
```
TomWoodward: checking to make sure the page title or chapter title have been changed are nifty but not completely accurate ways to make sure the page has loaded, i liked the previous implementation of this method because it abstracted the exact method of telling when the page was loaded, and if we implement a better way in the future we'd only have to do it in one place.

pytest-selenium/pages/content.py
```diff
@@ -50,6 +78,10 @@ class BookBanner(Region):
         _chapter_title_locator = (By.CSS_SELECTOR, "div > h1 > span.os-text")
         _chapter_section_locator = (By.CSS_SELECTOR, "div > h1 > span.os-number")
 
+        @property
+        def book_title_element(self):
```
TomWoodward: didn't mike add a bookbanner region in #188 ? can't we use that here?

TomWoodward (changes requested):

pytest-selenium/pages/content.py
m1yag1: oh i think that's left over from a different thing i was trying. ooops.

pytest-selenium/pages/base.py
m1yag1: i don't see how it abstracted anything away as it was hardcoded to always use the title. It was also checking the title on the parent class which didn't have a title which was also confusing to me. in this case here, any element that is clicked that also causes a side effect can be checked. 

I'm in full agreement that there should be something better for detecting when a page is fully loaded. I wanted to suggest initially that maybe on these navigation transitions maybe the `rex-loaded` attribute should be removed and added again when the page is loaded. @TomWoodward

TomWoodward: yea i agree that a better solution is possible and we should do it, but it doesn't have to happen right now and we can just change it in this method when we get around to it

m1yag1: i agree. i've reverted my changes.

pytest-selenium/pages/base.py
TomWoodward: @Malar-Natarajan this still needs to be fixed

pytest-selenium/tests/test_attribution.py
```diff
@@ -49,3 +50,31 @@ def test_attribution_collapsed_by_default_expands_when_clicked(
...
+
+    attribution.click_attribution_link()
+
+    # WHEN: Navigating via Previous link
+    content.click_previous_link()
+
+    # THEN: The citation/attribution section is not open on the new page
+    assert not attribution.is_open
+
+    attribution.click_attribution_link()
```
TomWoodward: there is no assertion after this, so its not testing anything right? why is it here?

TomWoodward (changes requested):

pytest-selenium/pages/base.py
Malar-Natarajan: @TomWoodward did you mean that I commented the `offscreen_click()` subclass but forgot to remove it or did you mean some other changes?

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: my bad..removed it

pytest-selenium/pages/base.py
Malar-Natarajan: Updated in the latest commit

TomWoodward (approved):

PR #175, opened 2019-06-11
show search results

KarinaMendez2:
desktop and mobile


src/app/content/components/Toolbar.tsx
```diff
@@ -18,22 +20,57 @@ import SidebarControl from './SidebarControl';
...
+    padding: 0 ${theme.padding.page.mobile}rem;
+  `)}
 `;
 
 // tslint:disable-next-line:variable-name
 const SearchIcon = styled(Search)`
   ${toolbarIconStyles}
+
+  &[active] {
+    display: none;
```
TomWoodward: can we hide/show this in react instead of css?

src/app/content/components/Toolbar.tsx
```diff
@@ -18,22 +20,57 @@ import SidebarControl from './SidebarControl';
...
   ${toolbarIconStyles}
 `;
 
+// tslint:disable-next-line:variable-name
+const CloseIcon = styled(TimesCircle)`
+  ${toolbarIconStyles}
+  color: ${theme.color.primary.gray.lighter};
+  display: none;
+
+  &[show] {
```
TomWoodward: hide/show with react instead of adding html attributes?

src/app/content/components/Toolbar.tsx
```diff
@@ -55,43 +94,52 @@ const SearchInputWrapper = styled.div`
...
+  border-radius: 0.2rem;
 
-  ::after {
-    content: "";
-    position: absolute;
-    left: 0;
-    right: 0;
-    bottom: 0;
-    border-bottom: solid 0.1rem ${toolbarIconColor.base};
+  :focus-within {
```
TomWoodward: need to also use all the polyfill classes for focus-within

src/app/content/components/Toolbar.tsx
```diff
@@ -121,44 +169,145 @@ const SearchPrintWrapper = styled.div`
   display: flex;
   justify-content: center;
   align-items: center;
+  overflow: visible;
+
+  ${theme.breakpoints.mobile(css`
+    height: 100%;
+    ${SearchInputWrapper} {
+      border: none;
```
TomWoodward: why put these styles here instead of actually on the `SearchInputWrapper`?

src/app/content/components/Toolbar.tsx
```diff
@@ -121,44 +169,145 @@ const SearchPrintWrapper = styled.div`
...
 `;
 
+// tslint:disable-next-line:variable-name
+const MobileSearchWrapper = styled.div`
+  ${SearchInputWrapper} {
+    margin: 1rem 0;
+  }
+
+  &[show] {
+    display: none;
```
TomWoodward: can this hide/show with react logic instead of adding html attributes?

src/app/content/components/Toolbar.tsx
```diff
@@ -121,44 +169,145 @@ const SearchPrintWrapper = styled.div`
...
+    }
+  `)}
+`;
+
 // tslint:disable-next-line:variable-name
 const Toolbar: SFC = () => <BarWrapper>
   <TopBar data-testid='toolbar'>
     <SidebarControl />
-    <SearchPrintWrapper>
+    <SearchPrintWrapper id='SearchPrintWrapper'>
```
TomWoodward: what is this id for?

src/app/content/components/Toolbar.tsx
```diff
@@ -121,44 +169,145 @@ const SearchPrintWrapper = styled.div`
...
     </SearchPrintWrapper>
   </TopBar>
+  <MobileSearchWrapper id='mobileSearchWrapper'>
+    <FormattedMessage id='i18n:toolbar:search:placeholder'>
+        {(msg: Element | string) => <SearchInputWrapper>
+          <MobileSearchInput
+            aria-label={assertString(msg, 'placeholder must be a string')}
+            placeholder={assertString(msg, 'placeholder must be a string')}
+            onKeyDown={handleKeyDown}
+            data='searchValue' />
```
TomWoodward: what is this data for?

src/app/content/components/Toolbar.tsx
```diff
@@ -121,44 +169,145 @@ const SearchPrintWrapper = styled.div`
...
+          <CloseIcon className='closeIcon' onClick={clearSearch}/>
+        </SearchInputWrapper>}
+      </FormattedMessage>
+  </MobileSearchWrapper>
 </BarWrapper>;
 
 export default Toolbar;
+
+const toggleSearchIcon = (opt: string, target: string) => {
```
TomWoodward: pretty sure all this dom manipulation can go

src/app/content/components/Toolbar.tsx
TomWoodward: because there are two search input wrappers, and these are the styles for the mobile one

src/app/content/components/SidebarControl.tsx
```diff
@@ -24,9 +24,18 @@ interface MiddleProps {
   closeToc: () => void;
 }
 
+// tslint:disable-next-line:variable-name
+const ListIconStyles = css`
```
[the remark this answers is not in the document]
KarinaMendez2: this icon has different styles from the ones that were sharing in the toolbar (like the search and the close icon)

PR #180, opened 2019-06-13
chore: more books

TomWoodward:
and related fixes

### DIFF
--- a/script/prerender/prerender.tsx
+++ b/script/prerender/prerender.tsx
@@ -11,6 +11,7 @@ import createApp from '../../src/app';
 import { AppOptions } from '../../src/app';
 import { content } from '../../src/app/content/routes';
 import { flattenArchiveTree, getUrlParamForPageId, stripIdVersion } from '../../src/app/content/utils';
+import { developerHome } from '../../src/app/developer/routes';
 import { notFound } from '../../src/app/errors/routes';
 import * as errorSelectors from '../../src/app/errors/selectors';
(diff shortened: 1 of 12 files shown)


script/prerender/prerender.tsx
```diff
@@ -131,6 +132,7 @@ const preparePages: PreparePages = async(archiveLoader, osWebLoader) => {
 
   const pages: Pages = [
     {code: 404, page: {route: notFound}},
+    {code: 200, page: {route: developerHome}},
```
TomWoodward: This needs to not be included in production

TomWoodward: Didn't work anyway, just delete

src/app/content/utils.ts
```diff
@@ -8,16 +8,16 @@ export { stripIdVersion } from './utils/idUtils';
 export { scrollTocSectionIntoView } from './utils/domUtils';
 
 export const getContentPageReferences = (content: string) =>
-  (content.match(/\/contents\/([a-z0-9\-]+(@[\d\.]+)?)(:([a-z0-9\-]+(@[\d\.]+)?))?/g) || [])
+  (content.match(/"\/contents\/([a-z0-9\-]+(@[\d\.]+)?)(:([a-z0-9\-]+(@[\d\.]+)?))?/g) || [])
```
philschatz: What prompted the need to include the `"` in the match? And if it's because other text contained `/contents/` should this be `href="`

philschatz (approved):

PR #185, opened 2019-06-18
feat: add search hook and integrate into history state

TomWoodward:
this adds redux logic for searching, navigating after searching, and navigating to first search result automatically.

this doesn't add any UI, you can test by dispatching a search action manually in the browser console

```js
__APP_STORE.dispatch({type: 'Content/Search/request', payload: 'my cool search'})
```


src/app/content/search/hooks.ts
```diff
@@ -0,0 +1,48 @@
...
+    loadSearch(services, query);
+  }
+};
+
+export default [
+  actionHook(requestSearch, searchHookBody),
+];
+
+function loadSearch(services: AppServices & MiddlewareAPI, query: string | null) {
+  const savedState = services.history.location.state;
```
TomWoodward: don't use services.history

src/app/content/search/utils.ts
```diff
@@ -0,0 +1,27 @@
+import { SearchResult } from '@openstax/open-search-client';
+import { Location } from 'history';
+import { getIdVersion, stripIdVersion } from '../utils/idUtils';
+
+// TODO - sort by page order
+// TODO - sort by `pagePosition`
```
TomWoodward: will deal with these requirements when working on the results view

KarinaMendez2 (approved):
Works for me :+1:

PR #192, opened 2019-06-25
fix: h3 font size

berkalpyakici:


TomWoodward (approved):
code is fine, lets wait for deployment to finish and malar will check the content

PR #193, opened 2019-06-26
feat: add sentry and redux-sentry-middleware

nathanstitt:
And a small wrapper to configure and set the env for it.

I'm curious of the best way to get set the DEPLOY_ENV, I suspect that reading `global` like I am doing is the wrong way, but not sure how to do better.


src/config.js
```diff
@@ -15,6 +15,7 @@
 
 let config = {
   APP_ENV: process.env.REACT_APP_ENV,
+  DEPLOYED_ENV: global && global.location ? global.location.hostname.split('.')[0] : 'server',
   ACCOUNTS_URL: process.env.ACCOUNTS_URL || 'https://accounts.openstax.org',
```
nathanstitt: this seems bad to me.  Do you know of a better alternative?

src/helpers/Sentry.ts
```diff
@@ -0,0 +1,32 @@
+import * as Sentry from '@sentry/browser';
+import createSentryMiddleware from 'redux-sentry-middleware';
+import config from '../config';
+
+export default {
+
+  initializeWithMiddleware() {
+    Sentry.init({
+      dist: config.CODE_VERSION,
```
TomWoodward: release_id is more descriptive, that captures code version and books that are built

TomWoodward (changes requested):

src/config.js
nathanstitt: well, webpack will rewrite `global` to be `window` for the web.  Using `global` allows it to also be set for non-dom environments. 

We use the first part of the URL in tutor just so it's shorter but maybe that won't work well with with REX targeted domains.  I'll use the entire hostname for now

src/app/index.tsx
```diff
@@ -91,6 +92,8 @@ export default (options: AppOptions) => {
 
   const middleware: Middleware[] = [
     navigation.createMiddleware(routes, history),
+    Sentry.initializeWithMiddleware(),
```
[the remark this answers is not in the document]
nathanstitt: I don't think so, but maybe you see something I don't?  It runs fine in the specs at least.

src/config.js
TomWoodward: `typeof(window) === undefined` returns false in node, thats the pattern we follow throughout the code, doesn't rely on using `global`

src/app/index.tsx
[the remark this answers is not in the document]
TomWoodward: can you try running `REACT_APP_ENV=test yarn serve:static` to see if pre-rendering works?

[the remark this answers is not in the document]
nathanstitt: guessing you meant `start-static`, since that's the only command that has `static` in it.  I ran:
`REACT_APP_ENV=test yarn run start:static`
without any issues and then loaded the site from http://localhost:62873/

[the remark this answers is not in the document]
TomWoodward: yea sorry thats right, ok cool.

src/config.js
nathanstitt: Hm, so using `window` causes specs to fail even though it's behind the typeof check.
```
$ REACT_APP_ENV=test jest --coverage --testPathPattern="(\.|/)spec\.tsx?"
ReferenceError: window is not defined
    at Object.<anonymous> (/Users/nas/code/ox/rex-web/src/config.js:18:59)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
```

We could abandon the attempt at the DEPLOYED_ENV, Sentry will log whatever the url was so we'll still have the info.

TomWoodward: this is because i led you astray, `typeof` returns a string, so its `typeof(window) === 'undefined'`

nathanstitt: oh derp, I knew that.  That's what I get for blindly copy & pasting code 😁

TomWoodward (approved):
this is looking fine, just be aware that its gonna send stuff to sentry from local dev environments, we may want to throw a check somewhere for `REACT_APP_ENV=production` which will be true in every environment thats not local dev or heroku

TomWoodward (changes requested):

TomWoodward:
looks good now, but code coverage took a hit. after running `yarn test:unit` locally you can open the coverage report with `yarn coverage-report`

nathanstitt:
> looks good now, but code coverage took a hit.

Something really weird is going on with coverage.  I ran it as you'd suggested and noticed that the report was pointing out stuff that was no longer in the codebase.  I then thought that maybe coverage wasn't re-generating so I removed all the report files (`rm -r coverage/*`), and they're not re-created after test:unit

Thinking this might be something I'd introduced, I switched to `master` but see the same thing happen there.

It seems like jest dies right as it's generating the coverage. I see multiple:
```
Error [ERR_IPC_CHANNEL_CLOSED]: Channel closed
    at process.target.send (internal/child_process.js:627:16)
    at reportSuccess (/Users/nas/code/ox/rex-web/node_modules/jest-worker/build/child.js:59:11)
Emitted 'error' event at:
    at process.nextTick (internal/child_process.js:631:35)
    at process._tickCallback (internal/process/next_tick.js:61:11)
events.js:167
      throw er; // Unhandled 'error' event
```

TomWoodward:
sometimes jest workers get into a weird state and i have to do `killall node` to get new ones

TomWoodward:
@nathanstitt i'm still showing some things that need coverage:

```
Statements   : 99.78% ( 1360/1363 )
Branches     : 99.63% ( 534/536 )
Functions    : 100% ( 456/456 )
Lines        : 99.76% ( 1252/1255 )
```

TomWoodward:
@nathanstitt getting closer...

```
=============================== Coverage summary ===============================
Statements   : 99.93% ( 1362/1363 )
Branches     : 99.81% ( 535/536 )
Functions    : 100% ( 456/456 )
Lines        : 99.92% ( 1254/1255 )
================================================================================
```

src/config.development.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   RELEASE_ID: 'development',
   CODE_VERSION: 'development',
+  DEPLOYED_ENV: 'development',
```
TomWoodward: the heroku environments run in development but it'd be nice to be able to differentiate them in the logs. will we be supplying `hostname` in any other metadata? can we supply `APP_ENV` as metadata separately and not override `DEPLOYED_ENV` here?

TomWoodward (approved):

PR #195, opened 2019-06-26
feat: Search sidebar

TomWoodward:


src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+// tslint:disable-next-line:variable-name
+const CloseIcon = styled(Times)`
+  ${toolbarIconStyles}
+  color: ${theme.color.primary.gray.lighter};
+  margin-right: 1.4rem;
+`;
+
+// tslint:disable-next-line:variable-name
+const SearchResultsBar = styled.div`
```
TomWoodward: most of this is the same as the ToC wrapper, we should try to combine as much as possible

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+  flex-direction: column;
+  position: sticky;
+  background: ${searchResultsBarVariables.backgroundColor};
+
+  &:not([open]) {
+    display: none;
+  }
+
+  &[open] {
+    display: block;
```
TomWoodward: the sidebar has a transition that looks nicer than this, also use props instead of html attributes for this kind of thing

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  ${theme.breakpoints.mobile(css`
+    width: calc(50vw + ${searchResultsBarMobileWidth}vw);
+    min-width: calc(50vw + ${searchResultsBarMobileWidth}vw);
+    top: ${bookBannerMobileMiniHeight + toolbarMobileHeight
```
TomWoodward: we're gonna have to give this access to whether the search input is open on mobile and compensate so the top position is right whether its open or closed

probably do the same in the toc sidebar

thats sort of a pain...

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+          + toolbarSearchInputMobileHeight + (mobileSearchContainerMargin * 2)}rem;
+    height: calc(100vh - ${navMobileHeight + bookBannerMobileMiniHeight}rem);
```
TomWoodward: this height looks wrong just because its less complicated than the top calculation

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+    margin-left: calc(-50vw - ${sidebarMobileWidth}rem - ${searchResultsBarVariables.mainPaddingMobile}rem);
```
TomWoodward: i don't think you need this `searchResultsBarVariables.mainPaddingMobile` part here

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+// tslint:disable-next-line:variable-name
+const SearchResultsBar = styled.div`
+  top: calc(${bookBannerDesktopMiniHeight}rem + ${toolbarDesktopHeight}rem);
+  overflow-y: auto;
+  height: calc(100vh - ${navDesktopHeight + bookBannerDesktopMiniHeight}rem);
+  transition: transform 300ms ease-in-out,box-shadow 300ms ease-in-out,background-color 300ms ease-in-out;
+  background-color: #fafafa;
+  z-index: 4;
+  margin-left: calc(-50vw - ${sidebarDesktopWidth}rem - ${searchResultsBarVariables.mainPaddingDesktop}rem);
```
TomWoodward: i don't think you need this `searchResultsBarVariables.mainPaddingMobile` part here

src/app/content/components/constants.ts
```diff
@@ -9,10 +9,17 @@ export const sidebarDesktopWidth = 33.5;
...
+export const searchResultsBarDesktopWidth = 43.5;
+export const searchResultsBarMobileWidth = 100;
+
 export const toolbarIconColor = theme.color.primary.gray;
 export const toolbarDesktopHeight = 5;
 export const toolbarMobileHeight = 4;
 
+export const mobileSearchContainerMargin = 1;
+export const toolbarSearchInputDesktopHeight = 3.2;
+export const toolbarSearchInputMobileHeight = 3.2;
```
TomWoodward: if we also export variables for `toolbarSearchDesktopHeight` `toolbarSearchMobileOpenHeight` `toolbarSearchMobileClosedHeight` that might make logic elsewhere simpler

src/app/content/components/SearchResultsSidebar.tsx
```diff
@@ -0,0 +1,278 @@
...
+    display: none;
+  }
+
+  &[open] {
+    display: block;
+  }
+
+  ${theme.breakpoints.mobile(css`
+    width: calc(50vw + ${searchResultsBarMobileWidth}vw);
+    min-width: calc(50vw + ${searchResultsBarMobileWidth}vw);
```
TomWoodward: your search bar width should not be in `vw`

TomWoodward (commented):
looking good so far, the only big thing missing is that there is supposed to be a loading state while the query is processing using the bouncy openstax logo from osweb

src/app/components/Loader.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  to {
+      transform: matrix(0.99863, 0.052336, -0.052336, 0.99863, 0, -4);
+  }
+`;
+
+const LoadingIcon = styled.svg`
+  enable-background:new 0 0 57.6 39.1;
+                animation: ${loadSvg} 0.3s forwards 0.3s;
```
TomWoodward: whitespace

src/app/content/components/SearchResultsSidebar/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+import * as selectSearch from '../../search/selectors';
+import { SearchResultChapter, SearchResultContainer, SearchResultPage } from '../../search/types';
+import * as select from '../../selectors';
+import { Book } from '../../types';
+import { CollapseIcon, ExpandIcon, SummaryTitle, SummaryWrapper } from '../Sidebar/styled';
+import { SearchResultsBarWrapper } from './SearchResultsBarWrapper';
+import * as Styled from './styled';
+
+// tslint:disable-next-line:variable-name
+export const SearchResultContainers = (props: {containers: SearchResultContainer[],
```
TomWoodward: these aren't used in this file, we should probably move them to separate files

src/app/content/components/SearchResultsSidebar/index.tsx
```diff
@@ -0,0 +1,94 @@
+import { SearchResultHit } from '@openstax/open-search-client/dist/models/SearchResultHit';
```
TomWoodward: lets move these search sidebar files into `/search/components/`

src/app/content/components/SearchResultsSidebar/SearchResultsBarWrapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+  book?: Book;
+}
+
+export class SearchResultsBarWrapper extends Component<SearchResultsSidebarProps> {
+  public searchSidebar = React.createRef<HTMLElement>();
+
+  public render() {
+    const {query, totalHits, results, onClose, book, open, toggleSearchSidebar} = this.props;
+
+    return <Styled.SearchResultsBar ref={this.searchSidebar} open={open}>
```
TomWoodward: why does this have to have the `open` prop? won't it always be true?

src/app/content/components/SearchResultsSidebar/SearchResultsBarWrapper.tsx
```diff
@@ -0,0 +1,83 @@
...
+    const scrollHandler = () => {
+      const top = searchSidebar.getBoundingClientRect().top;
+      searchSidebar.style.setProperty('height', `calc(100vh - ${top}px)`);
+    };
+
+    const animation = () => requestAnimationFrame(scrollHandler);
+
+    window.addEventListener('scroll', animation, {passive: true});
+    window.addEventListener('resize', animation, {passive: true});
+    scrollHandler();
```
TomWoodward: try to move this logic into a util function to be shared with the `scrollHandler` in the ToC

src/app/content/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,210 @@
...
+  } from '../constants';
+import ContentLinkComponent from '../ContentLink';
+import { Summary } from '../Sidebar/styled';
+import { toolbarIconStyles } from '../Toolbar';
+
+const searchResultsBarVariables = {
+    backgroundColor: '#f1f1f1',
+    iconRightPadding: 0.7,
+    mainPaddingDesktop: 3,
+    mainPaddingMobile: 2,
```
TomWoodward: is this supposed to be the same as the page padding in the app theme?

src/app/content/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,210 @@
...
+    searchResultsBarDesktopWidth,
+    toolbarDesktopHeight,
+    toolbarMobileElementsHeight,
+  } from '../constants';
+import ContentLinkComponent from '../ContentLink';
+import { Summary } from '../Sidebar/styled';
+import { toolbarIconStyles } from '../Toolbar';
+
+const searchResultsBarVariables = {
+    backgroundColor: '#f1f1f1',
```
TomWoodward: can get this color from the app theme colors

src/app/content/components/Toolbar.tsx
```diff
@@ -3,22 +3,29 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { connect } from 'react-redux';
 import styled, { css } from 'styled-components/macro';
+import { ChevronLeft } from 'styled-icons/boxicons-regular/ChevronLeft';
 import { Print } from 'styled-icons/fa-solid/Print';
 import { Search } from 'styled-icons/fa-solid/Search';
 import { TimesCircle } from 'styled-icons/fa-solid/TimesCircle';
 import { maxNavWidth } from '../../components/NavBar';
-import { contentFont, textRegularLineHeight, textRegularSize, textRegularStyle } from '../../components/Typography';
+import { contentFont, textRegularLineHeight,
```
TomWoodward: weird formatting here

src/app/content/components/Toolbar.tsx
```diff
@@ -3,22 +3,29 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { connect } from 'react-redux';
 import styled, { css } from 'styled-components/macro';
+import { ChevronLeft } from 'styled-icons/boxicons-regular/ChevronLeft';
```
TomWoodward: use the font-awesome icons, not `boxicons`

src/app/content/components/Wrapper.tsx
```diff
@@ -1,9 +1,24 @@
...
 export { wrapperPadding } from '../../components/Layout';
 
-export default styled(LayoutBody)`
+// tslint:disable-next-line:variable-name
+export const Wrapper = styled(LayoutBody)`
   position: relative; /* for sidebar overlay */
   overflow: visible; /* so sidebar position: sticky works */
+  flex: 1;
+  ${(props) => {
+    if (props.open) {
```
TomWoodward: this could use a more descriptive name like `searchOpen` or something, and can be reformatted like

```
styled(asdf)`
  ${(props) => props.open && css`
    display: none;
  `}
`;

File: src/app/content/components/constants.ts [hunk 19]
```diff
@@ -9,10 +9,24 @@ export const sidebarDesktopWidth = 33.5;
 export const sidebarMobileWidth = 28.8;
 export const sidebarTransitionTime = 300;
 
+export const searchResultsBarDesktopWidth = 43.5;
+export const searchResultsBarMobileWidth = 100;
```
TomWoodward: unused

src/app/content/components/constants.ts
```diff
@@ -9,10 +9,24 @@ export const sidebarDesktopWidth = 33.5;
...
 export const toolbarDesktopHeight = 5;
 export const toolbarMobileHeight = 4;
 
+export const mobileSearchContainerMargin = 1;
+export const toolbarSearchInputDesktopHeight = 3.2;
+export const toolbarSearchInputMobileHeight = 3.2;
+export const toolbarSearchInputMobileWidth = 18.5;
+export const toolbarToggleSearchBarTextMobileWidth = 11.2;
+export const toolbarMobileSearchWrapperHeight = toolbarSearchInputMobileHeight
++ (mobileSearchContainerMargin * 2) + 0.1;
```
TomWoodward: where does this `0.1` come from?

src/app/content/components/constants.ts
```diff
@@ -9,10 +9,24 @@ export const sidebarDesktopWidth = 33.5;
...
+export const toolbarMobileElementsHeight = bookBannerMobileMiniHeight
```
TomWoodward:
```suggestion
export const searchSidebarTopOffset = bookBannerMobileMiniHeight
```

src/app/content/search/reducer.ts
```diff
@@ -7,17 +7,18 @@ import { State } from './types';
 
 export const initialState = {
   loading: false,
+  open: false,
```
TomWoodward: might consider renaming this `mobileOpen` since its only respected on mobile

src/app/content/search/reducer.ts
```diff
@@ -7,17 +7,18 @@ import { State } from './types';
...
 
 const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
   switch (action.type) {
     case getType(actions.requestSearch): {
-      return {...state, loading: true, query: action.payload};
+      return {...initialState, loading: true, query: action.payload, open: true};
     }
     case getType(actions.receiveSearchResults): {
-      return {...state, loading: false, results: action.payload};
+      return {...state, loading: false, results: action.payload, open: true};
```
TomWoodward: you shouldn't need to add `open: true` here right, because its already in `requestSearch`?

src/app/content/search/reducer.ts
```diff
@@ -27,6 +28,9 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
         ? initialState
         : state;
     }
+    case getType(actions.toggleSearchSidebar): {
+      return {...state, open: action.payload};
```
TomWoodward: if this is a toggle then it should be `return {...state, open: !state.open};`, if you need to be able to pass in the open/closed boolean to make something work then probably should create two actions `openResults` and `closeResults`

src/app/content/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,210 @@
...
+  ${theme.breakpoints.mobile(css`
+    display: none;
+  `)};
+`;
+
+// tslint:disable-next-line:variable-name
+export const HeaderQuery = styled.div`
+  max-width: calc(${searchResultsBarDesktopWidth - (textRegularLineHeight * 2)
+              - searchResultsBarVariables.mainPaddingDesktop
+              - searchResultsBarVariables.mainRightPaddingDesktop}rem)
```
TomWoodward: this is complicated, why is this complicated?

src/app/content/components/SearchResultsSidebar/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  onClose: () => void;
+  open: boolean;
+  toggleSearchSidebar: (open: boolean) => void;
+}
+
+export class SearchResultsSidebar extends Component<SearchResultsSidebarProps> {
+  public render() {
+    const {query, open} = this.props;
+
+    return query && open && <SearchResultsBarWrapper {...this.props}/>;
```
TomWoodward: we don't have spec for open/closing the sidebar on desktop, so the `open` should only be respected on mobile. meaning if you're on mobile and click a search result (closing the sidebar) then resize your screen up you should get the sidebar again.

i think this means we have to do the hiding of the sidebar when it is closed (but not cleared) in css rather than this react condition.

src/app/content/components/Toolbar.tsx
```diff
@@ -323,9 +389,15 @@ class Toolbar extends React.Component<{
...
+    open: selectSearch.open(state),
     query: selectSearch.query(state),
+    results: selectSearch.results(state),
   }),
   (dispatch: Dispatch) => ({
+    onClose: () => {
+      dispatch(clearSearch());
+    },
     search: (query: string) => dispatch(requestSearch(query)),
+    toggleSearchSidebar: (open: boolean) => dispatch(toggleSearchSidebar(open)),
```
TomWoodward: not sure if you only put this here for testing, but we don't want this here. when you click the magnifying glass we want to `clearSearch`, the only way to close the sidebar without clearing it is to click on one of the result links

src/app/content/components/constants.ts
KarinaMendez2: From the <hr/> border height that separates the toolbar from the mobile search input

src/app/content/search/reducer.ts
KarinaMendez2: right.

src/app/content/components/Toolbar.tsx
KarinaMendez2: We need the "openSearchResults" action for the "< Search Results" button that will open the results bar.

TomWoodward: right ok, but its not being used for that now

src/app/content/components/SearchResultsSidebar/styled.tsx
KarinaMendez2: probably, but I followed the invision design, says 30px, already removed the mainPaddingMobile that was unused.

src/app/content/components/SearchResultsSidebar/styled.tsx
KarinaMendez2: because i'm calculating the space for that div to start overlapping the text if it's a long search

src/app/content/search/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,208 @@
...
+export const LoadingWrapper = styled.div`
+  overflow: hidden;
+  right: 0;
+  top: 0;
+  position: absolute;
+  transition: opacity 0.2s;
+  width: 100%;
+  height: 100%;
+  background-color: ${searchResultsBarVariables.backgroundColor};
+  z-index: 5;
```
TomWoodward: is this necessary?

src/app/content/search/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,208 @@
...
+    border-bottom: solid 0.2rem ${searchResultsBarVariables.backgroundColor};
+  }
+`;
+
+// tslint:disable-next-line:variable-name
+export const SearchQueryAlignment = styled.div`
+  max-width: 26.5rem;
+  text-align: justify;
+  /*textRegularLineHeight is the close icon height*/
+  margin-top: calc(7rem - ${textRegularLineHeight}rem);
```
TomWoodward: why do you have to pull this up by the height of the icon? maybe the icon should be absolutely positioned? or this should be a row? i forget what it looks like '

src/app/content/search/components/SearchResultsSidebar/styled.tsx
```diff
@@ -0,0 +1,208 @@
...
+  ${theme.breakpoints.mobile(css`
+    display: none;
+  `)};
+`;
+
+// tslint:disable-next-line:variable-name
+export const HeaderQuery = styled.div`
+  max-width: calc(${searchResultsBarDesktopWidth - (textRegularLineHeight * 2)
+              - searchResultsBarVariables.mainPaddingDesktop
+              - searchResultsBarVariables.mainRightPaddingDesktop}rem)
```
TomWoodward: not sure what is up with this, seems like it should be simpler

src/app/content/search/hooks.spec.ts
```diff
@@ -127,7 +127,7 @@ describe('hooks', () => {
         match: {} as any,
       });
 
-      expect(dispatch).toHaveBeenCalledWith(requestSearch('qwer'));
+      expect(dispatch).not.toHaveBeenCalledWith(requestSearch('qwer'));
```
TomWoodward: i don't think you were supposed to change this. `'searches for saved query on POP if it is differet from current query'` sounds right to me

src/app/content/search/selectors.ts
```diff
@@ -10,3 +11,19 @@ export const query = createSelector(
   localState,
   (state) => state.query
 );
+
+export const totalHits = createSelector(
```
TomWoodward: seems easier to get this off of results, since if there are results you will definitely have `results.hits.total`, by making another selector that might return null you just have one more null check you have to do

src/app/theme.ts
```diff
@@ -22,6 +22,10 @@ const padding = {
   },
 };
 
+const searchSidebar = {
+  background: '#f1f1f1',
```
TomWoodward: sigh... since this color is literally only used in the search results sidebar maybe we should just define it at the top of that file

src/config.test.js
```diff
@@ -5,6 +5,8 @@ module.exports = {
 
   BOOKS: {'testbook1-uuid':{'defaultVersion':'1.0'}},
 
+  SEARCH_URL: process.env.SEARCH_URL || 'https://search-jul11.sandbox.openstax.org',
```
TomWoodward: you don't want this

src/setupProxy.js
```diff
@@ -68,6 +68,7 @@ const findFileIn = (baseDir, reqInfo) => {
   const queryFilePath = path.join(filePath,
     reqInfo.search ? encodeURIComponent(reqInfo.search) : ''
   );
+  console.log(queryFilePath);
```
TomWoodward: debug code

src/test/fixtures/open-search/api/v0/search/%3Fq%3Dlaksdkd%26books%3Dtestbook1-uuid%25401.0%26index_strategy%3Di1%26search_strategy%3Ds1
```diff
@@ -0,0 +1 @@
+{"overall_took":114,"took":4,"timed_out":false,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0},"hits":{"total":0,"hits":[]}}
```
TomWoodward: can you format this json just to make it a little nicer to read?

src/app/content/components/Toolbar.tsx
```diff
@@ -233,46 +244,99 @@ const TopBar = styled.div`
...
+  padding: 0;
+  color: #027eb5;
+  display: flex;
+  align-items: center;
+  overflow: visible;
+`;
+
+// tslint:disable-next-line:variable-name
+const InnerText = styled.div`
+  width: 11.2rem;
```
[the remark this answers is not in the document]
KarinaMendez2: this is the width of the text "search results", never changes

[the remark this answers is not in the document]
TomWoodward: but it doesn't have to be explicitly set right? it'll figure it out based on the font? that way if we change the font styles we don't have to figure out what the new width is?

src/app/content/search/components/SearchResultsSidebar/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  book?: Book;
+  query: string | null;
+  totalHits: number | null;
+  results: SearchResultContainer[] | null;
+  onClose: () => void;
+  mobileOpen: boolean;
+  closeSearchResults: () => void;
+}
+
+export class SearchResultsSidebar extends Component<SearchResultsSidebarProps> {
```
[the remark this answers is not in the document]
KarinaMendez2: We did this so that the ref in the SearchResultsBarWrapper would asume the element would never be null

src/app/content/search/hooks.spec.ts
KarinaMendez2: this test was failing, saw the response and changed it based on that, since we modified the thing that did a search on reload

src/app/content/components/Toolbar.tsx
```diff
@@ -116,20 +149,25 @@ const SearchInputWrapper = styled.form`
   color: ${toolbarIconColor.base};
   border: solid 0.1rem;
   border-radius: 0.2rem;
+  width: 38rem;
```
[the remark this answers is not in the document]
KarinaMendez2: from Devika: 1. SEARCH BOX: a. MVP The search box should be 380px wide.

src/app/content/components/Toolbar.tsx
```diff
@@ -233,87 +270,161 @@ const TopBar = styled.div`
...
-class Toolbar extends React.Component<{
-  search: typeof requestSearch, query: string | null}, {query: string, mobileOpen: boolean, formSubmitted: boolean
+// tslint:disable-next-line:variable-name
+const LeftArrow = styled(AngleLeft)`
+  width: 2.5rem;
+  height: 2.5rem;
+`;
+
+// tslint:disable-next-line:variable-name
+const ToggleSeachResultsText = styled.button`
```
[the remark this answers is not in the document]
KarinaMendez2: you suggested it to be a button before, since it's clickable.

KarinaMendez2 (approved):

PR #200, opened 2019-06-27
fix: problem spacing

berkalpyakici:


generic-styles/index.less
```diff
@@ -28,6 +28,15 @@
       margin-inline-start: 0;
     }
   }
+  code {
+    color: #c7254e;
+    background-color: #f9f2f4;
+    border-radius: 4px;
+  }
+  [data-type="composite-page"] [data-type="exercise"] [data-type="problem"] p:first-of-type a {
```
TomWoodward: is this the exact same selector that cnx uses? a little worried we're gonna miss some random type

generic-styles/index.less
```diff
@@ -28,6 +28,15 @@
       margin-inline-start: 0;
     }
   }
+  code {
```
TomWoodward: this fix is on a different pr, should remove this commit from this branch

TomWoodward (changes requested):

generic-styles/index.less
berkalpyakici: CNX has [data-is-baked="true"] .media-body>#content [data-type="composite-page"] [data-type="exercise"] [data-type="problem"] p:first-of-type a {...}

TomWoodward: ok yea this is 👍 then

PR #201, opened 2019-06-28
fix: font size during landscape orientation

andymcheng:


generic-styles/index.less
```diff
@@ -28,6 +28,11 @@
       margin-inline-start: 0;
     }
   }
+  code {
```
TomWoodward: need to remove this from this PR, its on the other one

generic-styles/index.less
```diff
@@ -41,3 +46,6 @@
     border-spacing: 0;
   }
 }
+html {
+  -webkit-text-size-adjust: 100%; /* Prevent font scaling in landscape while allowing user zoom */
```
TomWoodward: this is fine, but can we investigate maybe using the viewport meta tags instead?

TomWoodward (changes requested):

generic-styles/index.less
TomWoodward: you can use `text-size-adjust` without the webkit prefix, it'll figure it out

TomWoodward (changes requested):

TomWoodward (changes requested):
i totally missed this on my first review, but since this fix is for the whole app and on the `html` and not generic styles specific, please add it to `src/index.css` instead of here

TomWoodward (approved):

PR #202, opened 2019-06-28
Fix scrolling

Dantemss:
- Tested (not extensively) for scrolling to:
  - Figures
  - Tables
  - Equations
  - Inline items
  - Solution links

- Made some tests for the id regex
- Made browser tests for scrolling to the right place


src/app/content/components/Page.tsx
```diff
@@ -58,6 +58,8 @@ export class PageComponent extends Component<PropTypes> {
       .replace(/<a(.*?href="(https?:\/\/|\/\/).*?)>/g, '<a target="_blank" rel="noopener nofollow"$1>')
       // same as previous, but allow indexing links to relative content
       .replace(/<a(.*?href="\.\.\/.*?)>/g, '<a target="_blank"$1>')
+      // move figure and table ids up to the parent div
+      .replace(/(<div[^>]*)(>(?!<\/?div>)*?<(?:figure|table)[^>]*?)( id=[^\s>]*)/g, '$1$3$2')
```
TomWoodward: this works in chrome, but gives an error on https://regex101.com/ for the `*?`, substitution seems to work without it though

TomWoodward (changes requested):

src/app/content/components/Page.tsx
Dantemss: I'll see if it accepts putting the `*?` inside the negative lookahead.

Dantemss: Turns out... this expression would not match `<div><div></div><figure id="blah">` anyway... maybe I'll just simplify it to match the figure/table only if it is the first child of the div. We can worry about other cases when we stop doing this in regex.

TomWoodward: i think they are always the first child anyway, so that should be fine

src/index.css
```diff
@@ -15,6 +11,13 @@ body, #root, html {
   overflow: visible;
 }
 
+html {
+  font-size: 62.5%; /* Now 10px = 1rem! */
+  scroll-snap-type: y proximity;
+  scroll-padding-top: 120px;
+  overflow-y: scroll;
```
TomWoodward: as we talked about in slack, we should try to move the `120px` (size of the nav) to a place that logically should have knowledge of that, and use the variables to get it.

also `overflow-y: scroll;` would ideally be `overflow: auto;`, hopefully that doesn't break everything 😬

TomWoodward (changes requested):

src/index.css
Dantemss: I think `overflow: auto` should work fine... I'll give it a try. And move the offset to the main content component as discussed.

Dantemss: Seems I can put `overflow: auto` in the `MainContentWrapper` and it works fine. This is also right below the one element that really needs `overflow: visible` (the `div#root`)

Dantemss: But... hopefully this doesn't break mobile *looks at the monster layout in Content.tsx*

Dantemss: Ignore that last one, `overflow: auto` needs to be in the html tag and that seems to just work.

src/app/content/components/Content.tsx
```diff
@@ -148,29 +159,32 @@ const HideOverflowAndRedoPadding = isOpenConnector(styled.div`
...
-              <Footer/>
-            </MainContentWrapper>
-          </UndoPadding>
-        </ContentPane>
-      </CenteredContentRow>
-    </Wrapper>
-  </Background>
-</Layout>;
+const Content: React.SFC = () => <React.Fragment>
+  <GlobalStyle />
```
TomWoodward: can't you put the globalstyle inside the layout and not have the `React.Fragment` thing?

src/app/components/SkipToContentWrapper.tsx
```diff
@@ -19,7 +20,8 @@ export default class SkipToContentWrapper extends Component {
   private scrollToTarget = (event: React.MouseEvent<HTMLAnchorElement>) => {
     if (this.mainContent) {
       event.preventDefault();
-      scrollTo(this.mainContent);
+      scrollToContent(this.mainContent);
+      (this.mainContent as HTMLElement).focus();
```
TomWoodward: you shouldn't need this `as HTMLElement`, its already a `HTMLDivElement`, if you're getting the wrong `HTMLDivElement` you might have to import it from `lib.dom`

src/app/content/components/Content.screenshotspec.tsx
```diff
@@ -0,0 +1,43 @@
...
+    describe(testCase, () => {
+      beforeEach(() => {
+        const setViewport = testCases[testCase];
+        setViewport(page);
+      });
+
+      it('scrolls correctly to all elements', async() => {
+        await navigate(page, TEST_PAGE_URL);
+        await finishRender(page);
+        expect(await page.screenshot()).toMatchImageSnapshot(SNAPSHOT_OPTIONS);
```
TomWoodward: can you do these as browserspecs by measuring the page scrollTop after every scroll against a known or computed-good value? the screenshots are brittle and add a lot of bytes to the repo. i prefer to use the screenshots when something just can't be reasonably tested with math (like: `it('looks good')`)

src/app/content/components/Content.tsx
```diff
@@ -26,6 +27,16 @@ import { isOpenConnector, styleWhenSidebarClosed } from './utils/sidebar';
 import Wrapper from './Wrapper';
 import { wrapperPadding } from './Wrapper';
 
+// tslint:disable-next-line:variable-name
+const GlobalStyle = createGlobalStyle`
+  html {
+    scroll-padding-top: ${(bookBannerDesktopMiniHeight + toolbarDesktopHeight) * .625}rem;
```
TomWoodward: ` * .625` gives me anxiety, i see why it is necessary, can we extract to a constant or helper function that says what this is doing?

src/app/content/components/Page.spec.tsx
```diff
@@ -382,7 +441,7 @@ describe('Page', () => {
     expect(spy).toHaveBeenCalledWith(0, 0);
   });
 
-  it('scrolls to selected content on load', () => {
+  it('does not scroll to selected content on load', () => {
```
TomWoodward:
```suggestion
  it('does not scroll to selected content on initial load', () => {
```

src/app/content/components/utils/scrollToContent.ts
```diff
@@ -0,0 +1,21 @@
...
+const pxPerRem = 10;
+// height of navbar in px
+const desktopOffset = -pxPerRem * (bookBannerDesktopMiniHeight + toolbarDesktopHeight);
+const mobileOffset = -pxPerRem * (bookBannerMobileMiniHeight + toolbarMobileHeight);
+
+export default function scrollToContent(elem: Element | string) {
+    // innerWidth works best on Chrome
+    // check other browsers at https://ryanve.com/lab/dimensions/
+    // default to desktopOffset if window is undefined (at least the target is always visible)
+    const offset = window && window.innerWidth <= pxPerRem * mobileBreak ? mobileOffset : desktopOffset;
```
TomWoodward: this is what `window.matchMedia(theme.breakpoints.mobileQuery)` is for

but as i wrote above, i think it would be better to `window.getComputedStyle(document.documentElement).scrollPaddingTop`

TomWoodward (changes requested):

src/app/components/SkipToContentWrapper.tsx
```diff
@@ -1,6 +1,7 @@
+import { HTMLElement } from '@openstax/types/lib.dom';
 import React, { Component } from 'react';
 import { FormattedMessage } from 'react-intl';
-import scrollTo from '../../helpers/scrollTo';
+import scrollToContent from '../content/components/utils/scrollToContent';
```
TomWoodward: lost this comment on the first review:

the directory `content` is for book-content related stuff, this helper is an accessibility thing "focus the main content of the page". this should be decoupled from anything happening in the content directory because we'll use it on pages that are not in that module.

the `scrollToContent` helper should be in `app/utils`, and the logic should be decoupled from the knowledge of the navbars, maybe reading the computed `scroll-padding-top` off the html element would be more flexible?

src/app/content/components/Content.tsx
Dantemss: OK I'll do that.

src/app/components/SkipToContentWrapper.tsx
Dantemss: I think it was complaining that HTMLDivElement did not respond to focus... I'll try the import and see what happens.

src/app/content/components/Content.screenshotspec.tsx
Dantemss: Alright, I'll change them to browserspecs.

src/app/content/components/Content.tsx
Dantemss: I'll put it in constants.

src/app/content/components/utils/scrollToContent.ts
Dantemss: I will change it to use `getComputedStyle`. This should also fix the missing coverage for this statement.

src/app/components/SkipToContentWrapper.tsx
Dantemss: Yes, I agree that with the change to the helper to use `getComputedStyle` it can live outside the content dir.

src/app/content/components/Content.tsx
Dantemss: I guess I should ask... why is the `.625` necessary anyway?
Shouldn't rem in the `<html>` element just use its own font size?
This behavior seems... odd to me...
I'll test to see if `em` behaves better in the html element.

Dantemss: Seems like using `em` instead of `rem` here removes the need for the multiplier.

src/app/components/SkipToContentWrapper.tsx
Dantemss: If I add this import here, I also have to add it to SkipToContent.ts but then the mergeRefs function gets a type error because this HTMLDivElement is not assignable to the non-OpenStax HTMLElement and I'm not sure how to fix that one...

Dantemss:
@TomWoodward: I made all requested changes except the `HTMLDivElement` one because as mentioned I am not sure what to do about `mergeRefs` failing with that change.

TomWoodward (approved):

TomWoodward (approved):

PR #203, opened 2019-07-01
extend domVisitor to support multiple audit logics

TomWoodward:
allows easily adding different code for checking content across all rex books


philschatz (approved):
Yay for more audits!

PR #207, opened 2019-07-02
feat: add logic for generating search results data structure

TomWoodward:


KarinaMendez2 (commented):
Just one quick question, is there a reason why there's nested duplicated content in the results?

TomWoodward:
@KarinaMendez2 can you elaborate on `nested duplicated content in the results`?

KarinaMendez2 (approved):
I understand the structure more now :+1:

PR #208, opened 2019-07-02
fix: footnote numbers breaking line

berkalpyakici:


TomWoodward:
i didn't realize that this issue was related to the `::before` element added by the `:target` styles to offset for the nav. this will be fixed by dante's outstanding PR

PR #210, opened 2019-07-03
feat: use cnx page slug url

philschatz:
Refs openstax/unified#247


src/test/mocks/archiveLoader.ts
```diff
@@ -76,6 +77,7 @@ export default () => {
       localBooks[`${parentBook.id}@${parentBook.version}`].tree.contents.push({
         id: `${newPage.id}@${newPage.version}`,
         shortId: `${newPage.shortId}@${newPage.version}`,
+        slug: getUrlParamForPageId(parentBook, newPage.title),
```
TomWoodward: since this is generating the tree for this mock book, i don't think its good to rely on calculating the new slug. maybe we should require passing the slug as a third param?

TomWoodward (changes requested):
all of the functions in urlUtils including `getUrlParamForPageId` should be either deleted or rewritten to use the slugs in the book tree

src/app/content/utils/urlUtils.ts
```diff
@@ -74,15 +29,15 @@ export const getUrlParamForPageId = (book: Pick<Book, 'id' | 'tree' | 'title'>,
...
+  const result = assertDefined(treeSection.slug, 'Slug needs to be defined');
   getUrlParamForPageIdCache.set(cacheKey, result);
 
   return result;
 };
 
 export const getPageIdFromUrlParam = (book: Book, pageParam: string): string | undefined => {
   for (const section of flattenArchiveTree(book.tree)) {
-    const sectionParam = getUrlParamForPageTitle(section);
+    const sectionParam = assertDefined(section.slug, 'Slug needs to be defined');
```
TomWoodward: can we put the book/page information in these error messages?

TomWoodward (changes requested):

src/app/content/utils/urlUtils.ts
philschatz: yes, updated.

TomWoodward (approved):

PR #213, opened 2019-07-09
fix: footnote alignment

berkalpyakici:


TomWoodward:
https://github.com/openstax/rex-web/pull/208#issuecomment-508199861

PR #218, opened 2019-07-15
fix: changed css styling for tables

andymcheng:
modified rex css styling for tables in principles of economics book to match those in cnx


generic-styles/index.less
```diff
@@ -55,4 +74,18 @@
     border-collapse: collapse;
     border-spacing: 0;
   }
+  table thead:first-child tr:first-child th {
```
TomWoodward: this stuff is no longer needed right?

TomWoodward (changes requested):

TomWoodward (approved):

PR #221, opened 2019-07-17
fix: update toast

TomWoodward:


src/helpers/pollUpdates.ts
```diff
@@ -1,38 +1,78 @@
...
+  previousObservedReleaseId = releaseId;
+};
+
+export const poll = (store: Store) => async() => {
   const environment = await fetch('/rex/environment.json')
-    .then((response) => response.json())
-    .catch(() => ({release_id: RELEASE_ID}))
+    .then((response) => response.json() as Promise<Environment>)
+    .catch(() => null)
   ;
```
nathanstitt: Seems strange to me to call `.then` & `.catch` on a function that you're also using `await` on.  I'm guessing that the then will be called and then the await ?

nathanstitt (approved):
I'm guessing you know what you're doing on the await/then question, it's probably needed for some reason

src/helpers/pollUpdates.ts
TomWoodward: yea it'll await the final promise, easier to chain the call to `json` and more terse than a real try/catch

PR #223, opened 2019-07-18
Test navigation

Malar-Natarajan:
C477321: previous link is hidden if there is no previous page

C477322: next link is hidden if there is no next page


pytest-selenium/utils/utility.py
```diff
@@ -11,3 +14,16 @@ def books(self) -> [str]:
 
     def random_book_slug(self):
         return random.choice(self.books)
+
+
+class WaitForTitleChange(object):
+    def click_and_wait_for_load(self, element: WebElement = None) -> None:
```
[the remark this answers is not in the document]
TomWoodward: starting to get why you did this, because you're calling `click_and_wait_for_load` from both regions and pages, so you define `page_title` in page, and then you have a helper in the base region to also do `page_title`.

seems like it'd be overall simpler to have a helper in the base region to find the root page (called like, `page` or whatever), and then from any region code simply do `self.page.click_and_wait_for_load`

pytest-selenium/regions/toc.py
```diff
@@ -0,0 +1,34 @@
...
+from pages import content
+from pages.base import Page
+from regions.base import Region
+
+from utils.utility import WaitForTitleChange
+
+
+class TableOfContents(Region):
+    _root_locator = (By.CSS_SELECTOR, "ol")
+    _section_link_locator = (By.CSS_SELECTOR, "ol li a")
```
TomWoodward: is this the branch you wanted those extra selectors for? i can add them

pytest-selenium/tests/test_navigation.py
```diff
@@ -0,0 +1,59 @@
...
+@markers.nondestructive
+def test_previous_link_hidden_on_first_page(selenium, base_url, book_slug, page_slug):
+
+    # GIVEN: The page is loaded
+    content = Content(selenium, base_url, book_slug=book_slug, page_slug=page_slug).open()
+    toolbar = content.toolbar
+    toc = content.sidebar.toc
+
+    # confirm first page is selected
+    if toc.active_section.get_attribute("textContent") != toc.sections[0].section_title:
```
TomWoodward: can't you use the `section_title` helper on the first part?

pytest-selenium/tests/test_navigation.py
```diff
@@ -0,0 +1,59 @@
...
+        while content.is_mobile:
+            toolbar.click_toc_toggle_button()
```
TomWoodward: why is this in a `while` loop?

pytest-selenium/tests/test_navigation.py
```diff
@@ -0,0 +1,59 @@
...
+            break
+        toc.sections[0].click()
```
TomWoodward: would it be better to reorganize these helpers to look more like

```
if not toc.first_page.is_active:
  if content.is_mobile:
    toolbar.click_toc_toggle_button()
  toc.first_page.click()
```

pytest-selenium/tests/test_navigation.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # THEN: The "previous" link should be hidden
+    with pytest.raises(Exception) as exc_info:
+        assert not content.previous_link.is_displayed
```
TomWoodward: this `with pytest.raises(Exception)` feels like it should be in a helper, you have other helpers for similar things

TomWoodward (changes requested):

pytest-selenium/regions/toc.py
Malar-Natarajan: @TomWoodward Not really. But you can add here also.
I need data-testid `chapter` in the corresponding <li> tags.
and `unit` in the corresponding <li> tags

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: @TomWoodward 
Mike suggested that Assertions has to be within the test and not in the helper functions (as per pytest guidelines)

I still have the other test where assertions are in the helper because, it was checking the same assertion for 3 different elements and did not make sense to me to have 4 lines of code repeated in a test 3 times.
 
This assertion is checking a different exception just one time, so I thought I can comply to the rules of pytest :-). I dont mind doing it the other way though

TomWoodward: the whole process of catching the exception doesn't have to be directly linked to the assertion. if you did the catch within the `is_displayed` function and returned `false`, then the assertion just becomes `assert not content.previous_link.is_displayed` and all the logic is encapsulated in that function

TomWoodward (changes requested):

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: @TomWoodward May be I am not getting your point, but when I tried 
`assert not content.previous_link.is_displayed` fails with `nosuchelement` exception error even before calling the `is_displayed` function because the previous_link is not visible.

I also tried to call  `content.assert_element_not_displayed(content.previous_link)` where
    """def assert_element_not_displayed(self, element):
        with pytest.raises(Exception) as exc_info:
            assert not element.is_displayed
        exception_raised = exc_info.type
        assert "NoSuchElementException" in str(exception_raised)
        return False"""
which would again fail even before calling the helper, because the the argument (previous_link) that we are passing is not visible in the first place.

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: Updated to If loop

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: reorganized the helpers

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: @TomWoodward reorganized the helpers like you mentioned and the test really looks short :-)

pytest-selenium/utils/utility.py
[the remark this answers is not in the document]
Malar-Natarajan: @TomWoodward I tried to do what you suggest. But I am probably missing something and couldn't get a proper result.

pytest-selenium/tests/test_navigation.py
Malar-Natarajan: @TomWoodward Ignore my previous comment. I got what you are saying and have updated the test

TomWoodward (approved):
@Malar-Natarajan i made some changes, please make sure they look good to you

pytest-selenium/regions/toc.py
```diff
@@ -0,0 +1,48 @@
...
+
+        @property
+        def section_title(self):
+            return self.root.get_attribute("textContent")
+
+        @property
+        def is_active(self):
+            html = self.find_element(*self._is_active_locator).get_attribute("outerHTML")
+            try:
+                assert "Current Page" in html
```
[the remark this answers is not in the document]
Malar-Natarajan: @TomWoodward 
`html` would store this `<li class="styled__NavItem-sc-1bshw4i-2 gSPIoy" aria-label="Current Page"><a href="preface" class="styled__ContentLink-sc-1bshw4i-1 inAuJF"><span class="os-text">Preface</span></a></li>` which has the `current page` attribute

pytest-selenium/tests/test_toc.py
```diff
@@ -81,9 +81,12 @@ def test_toc_disables_interacting_with_content_on_mobile(selenium, base_url, boo
     toolbar.click_toc_toggle_button()
 
     # THEN: The links in the content should be disabled
-    content.assert_element_not_interactable(content.next_link)
-    content.assert_element_not_interactable(content.previous_link)
-    content.assert_element_not_interactable(attribution.attribution_link)
+    if content.next_link_is_displayed:
+        assert content.element_is_not_interactable(content.next_link)
+    if content.previous_link_is_displayed:
```
Malar-Natarajan: @TomWoodward I noticed this failure in testrail and was planning to do this assert once you approve this PR.
Now thanks for updating the failing test + approving the PR :-)

PR #226, opened 2019-07-19
fix notification toast during deployments

TomWoodward:


src/helpers/pollUpdates.ts
```diff
@@ -32,8 +32,8 @@ const processEnvironment = (store: Store, environment: Environment) => {
   const releaseId = environment.release_id;
 
   if (
-    environment.release_id !== RELEASE_ID
-    && (trustRelease() || previousObservedReleaseId !== releaseId)
+    (trustRelease() && environment.release_id !== RELEASE_ID)
+    || (previousObservedReleaseId && previousObservedReleaseId !== releaseId)
   ) {
```
nathanstitt: You've assigning `environment.release_id` to `releaseId` on line 32, but only using once in the condition and using the original `environment.release_id` in the other spot.

Also seems like you wouldn't need to test if `previousObservedReleaseId` is defined.  Even if it's null/undefined it wouldn't match `releaseId`, right? 

Seems like you could:
```js
 (trustRelease() && releaseId !== RELEASE_ID) || (releaseId  !== previousObservedReleaseId)
```

src/helpers/pollUpdates.spec.ts
```diff
@@ -15,12 +15,19 @@ describe('poll updates', () => {
...
     store = createTestStore();
     dispatch = jest.spyOn(store, 'dispatch');
     (global as any).fetch = mockFetch(200, {release_id: 'releaseid'});
+
+    function MockDate() {
+      return null;
+    }
+    MockDate.prototype.getTime = jest.fn(() => (new dateBackup()).getTime());
+    (global as any).Date = MockDate;
   });
```
nathanstitt: Tutor's has a handy time mocking helper for jest that you might take a look at: 
https://github.com/openstax/tutor-js/blob/master/tutor/specs/helpers/time-mock.js

It's used like this: https://github.com/openstax/tutor-js/blob/master/tutor/specs/screens/assignment-builder/homework.spec.js#L11

nathanstitt (changes requested):

src/helpers/pollUpdates.ts
TomWoodward: if i didn't null check the `previousObservedReleaseId` then on the initial pass `null` would be different from `environment.release_id` and it'd show the toast, which is what i'm trying to avoid. with the null check it noops on the first pass and by the second pass they'll equal one another (or it'll have changed, in which case showing the toast is correct)

src/helpers/pollUpdates.spec.ts
TomWoodward: i may look into a helper like this if i have to do it again.. but for now i'll leave as is

src/helpers/pollUpdates.ts
nathanstitt: oh yep that makes sense. thanks!

nathanstitt (approved):

PR #227, opened 2019-07-22
feat: Capture and display errors

nathanstitt:
* Adds try/catch blocks around store middleware
  * Displays errors as modal
* adds error boundary around portions of UI
 * Displays error card in place of components when error occurs
* Logs all errors to sentry


src/app/utils.ts
```diff
@@ -23,13 +25,17 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
...
+        try {
+          const promise = boundHook(action);
+          if (promise) {
+            services.promiseCollector.add(promise);
+          }
+        } catch (e) {
+          // how to get store ?
+
+          Sentry.captureException(e);
         }
```
nathanstitt: @TomWoodward this is where I'm stuck. The store isn't available anywhere that I can see. Where can I access it from, or should I change the arguments to add a "store" and then change all the spots where this is called from provide that?

TomWoodward: `stateHelpers` (defined a few lines up) has `getState` and `dispatch`

nathanstitt:
@TomWoodward take a look at this when you get a chance, it's a little different than what we'd discussed.  

This:
* Captures action errors and records them in the store
* Adds error boundaries around different parts of the app, and a top-level one as well. If the error boundary captures an error it sets it on the store and does not render it's children
     * We could render a placeholder instead.  if we go this route, I think the error boundary component should accept a `renderOnError` prop so each spot could customize it's own error appearance.
* There's a top-level ErrorModal component that displays any errors.  The modal is really plain and needs UI work.
  * If this looks like a good approach let's discuss if there's a existing implementation you'd like to use or if we should create our own.

src/app/content/components/Toolbar.tsx
```diff
@@ -277,6 +278,7 @@ class Toolbar extends React.Component<{
 
     return <BarWrapper>
       <TopBar data-testid='toolbar'>
+        <ErrorModal />
```
TomWoodward: think i'd move this into `app/components/Layout`

src/app/content/hooks/locationChange.spec.ts
```diff
@@ -13,7 +13,8 @@ import { receiveBook, receivePage } from '../actions';
 import * as routes from '../routes';
 import { formatBookData } from '../utils';
 
-const mockConfig = {BOOKS: {
+// tslint:disable-next-line
+var mockConfig = {BOOKS: { // var is needed so that the mock is hoisted
```
TomWoodward: this wasn't broken before, why does it need to be changed?

src/app/errors/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children: ReactNode;
+  recordError: (error: Error) => void;
+}
+
+class ErrorBoundary extends React.Component<Props> {
+
+  public componentDidCatch(error: Error) {
+    this.props.recordError(error);
+    // also log the error to an error reporting service
+    Sentry.captureException(error);
```
TomWoodward: probably put this line above the call to `recordError` to make sure it goes out before anything else might break

src/app/errors/reducer.ts
```diff
@@ -2,12 +2,18 @@ import { Reducer } from 'redux';
...
+import { clearCurrentError, recordError } from './actions';
 import { notFound } from './routes';
 import { State } from './types';
+
 export const initialState = {};
 
 const reducer: Reducer<State, AnyAction> = (state = {}, action) => {
   switch (action.type) {
+    case getType(clearCurrentError):
+      return { ...state, error: undefined };
```
TomWoodward: you could also `return initialState` here, not sure if that would be more correct or not

src/app/utils.ts
```diff
@@ -23,13 +26,19 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
       const result = next(action);
 
       if (matches(action)) {
-        const promise = boundHook(action);
-
-        if (promise) {
-          services.promiseCollector.add(promise);
+        const catchError = (e: Error) => {
+          stateHelpers.dispatch(recordError(e));
+          Sentry.captureException(e);
```
TomWoodward: swap these two lines

src/app/content/components/Content.tsx
```diff
@@ -159,30 +160,32 @@ const HideOverflowAndRedoPadding = isOpenConnector(styled.div`
...
-              <Attribution />
-              <Footer/>
-            </MainContentWrapper>
-          </UndoPadding>
-        </ContentPane>
-      </CenteredContentRow>
-    </Wrapper>
-  </Background>
-</Layout>;
+const Content: React.SFC = () => <ErrorBoundary>
```
TomWoodward: putting this boundary here isn't functionally different from the one around the whole route, lets push it down around wrapper on 169 so they still see the navs if something in the content/toc/sidebar breaks

src/app/components/Card.tsx
```diff
@@ -0,0 +1,52 @@
+import styled from 'styled-components';
+import { textStyle } from './Typography/base';
+
+const margin = 3.0;
+
+// tslint:disable-next-line:variable-name
+const Card = styled.div`
```
TomWoodward: if this `Card` component is supposed to be generic can you try to make it work with the notification popup? if not then you might as well just put these components in the ErrorModal component

TomWoodward (changes requested):

src/app/errors/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface Props {
+  children: ReactNode;
+  recordError: (error: Error) => void;
+}
+
+class ErrorBoundary extends React.Component<Props> {
+
+  public componentDidCatch(error: Error) {
+    this.props.recordError(error);
```
TomWoodward: recording the error in the state is nice because the pre-renderer needs a way to tell something went wrong, but the spec is to have the ErrorBoundary replace the content of the page with the error message, not pop the modal, so i think something around here has to change

TomWoodward (changes requested):

src/app/errors/components/ErrorBoundary.tsx
nathanstitt: Yeah, that's mainly what I wanted your opinion on.  We could just render the error instead of the content, and in fact that was what I did originally.

The problem with that is that it'll only work if the boundary wraps a large enough area to display the error.  If the wrapped content is small or maybe does something funky like render a modal itself, then replacing children with the error will be unpredictable.

Perhaps the solution is to have `ErrorBoundary` take a `renderErrorAsModal` for cases like that, and otherwise render errors in place of children.

Thoughts?

TomWoodward: i think sort of the point is that we never want the `ErrorBoundary` to trigger a modal, because it represents the failure to render the content, we would just replace the content. we only want the modal for errors happening outside of the render process.

agree that if we used error boundaries on smaller elements, the design would get weird. what if we add a render prop for `errorDisplay` or something, and make a new component for `FullPageError` that we use in all the places you've currently got. that way if we wanted to later add a boundary around something like a sidebar or a note card, that would be easier to organize

nathanstitt:
@tomjw64 this is failing a11y checks and I see that `master` is also.  https://travis-ci.com/openstax/rex-web/builds/122295712#L411

src/app/components/Card.tsx
nathanstitt: I took a look at doing so, but the notification popup is quite a bit different. it changes the layout to be a top bar depending on mobile size.

src/app/content/hooks/locationChange.spec.ts
nathanstitt: I have no idea 😭 without that we get error:
```
 FAIL  src/app/content/hooks/locationChange.spec.ts
  ● Test suite failed to run

    ReferenceError: mockConfig is not defined

      21 | };
      22 |
    > 23 | jest.mock('../../../config', () => mockConfig);
         |                                    ^
      24 |
      25 | describe('locationChange', () => {
      26 |   let store: Store;

      at Object.mockConfig [as user:/Users/nas/code/ox/rex-web/src/config.js:] (src/app/content/hooks/locationChange.spec.ts:23:36)
      at Object.<anonymous> (src/helpers/Sentry.ts:3:1)
      at Object.<anonymous> (src/app/errors/components/ErrorBoundary.tsx:3:1)
      at Object.<anonymous> (src/app/navigation/components/NavigationProvider.tsx:4:1)
      at Object.<anonymous> (src/app/navigation/components/index.ts:1:1)
      at Object.<anonymous> (src/app/navigation/index.ts:2:1)
      at Object.<anonymous> (src/app/errors/reducer.ts:3:1)
      at Object.<anonymous> (src/test/createTestStore.ts:6:1)
      at Object.<anonymous> (src/app/content/hooks/locationChange.spec.ts:5:1)
```

src/app/components/Card.tsx
TomWoodward: so can we put these components in the ErrorModal component then?

src/app/errors/components/ErrorBoundary.tsx
nathanstitt: Hey, I think this got missed.  I've went through and left the error card only displaying inline for right now on the theory that we should only do as much as is required.

If and when we need to display errors in a small region we can do as you suggest then.

TomWoodward (changes requested):
minor comment, gonna ask for feedback on the error page design later today

TomWoodward (changes requested):

TomWoodward (approved):

PR #228, opened 2019-07-22
fix: Cross browser scroll padding

TomWoodward:


src/app/utils.ts
```diff
@@ -94,8 +94,8 @@ export const assertDocumentElement = (message: string = 'BUG: Document Element i
 };
 
 export const scrollTo = (elem: Element | string) => {
-  const html = assertDocumentElement();
-  const padding = assertWindow().getComputedStyle(html).getPropertyValue('scroll-padding-top');
```
TomWoodward: @Dantemss i really really liked this approach, but had to re-think it because edge doesn't support scroll-padding at all so this returns empty all the time

src/app/components/ScrollOffset.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * the context of a scroll snap position
+ *
+ * https://bugs.webkit.org/show_bug.cgi?id=179379
+ */
+
+export default class ScrollOffset extends React.Component<ScrollOffsetProps> {
+
+  public getOffset = (window: Window) => -(window.matchMedia(theme.breakpoints.mobileQuery).matches
+    ? this.props.mobileOffset * 10
+    : this.props.desktopOffset * 10
```
TomWoodward: this won't work if the browser zoom is not default, not sure if there is a way to compensate for that

TomWoodward: or maybe it'll work for zoom but not if default font size has been changed 🤔

src/app/components/ScrollOffset.tsx
```diff
@@ -0,0 +1,114 @@
...
+  private checkScroll = (maxChecks: number = 1) => {
+    let scrolls = 0;
+    const handler = () => {
+      scrolls++;
+      if (scrolls >= maxChecks) {
+        assertWindow().removeEventListener('scroll', handler);
+      }
+      this.scrollForOffset();
+    };
+    assertWindow().addEventListener('scroll', handler);
```
Dantemss: Hmm wouldn't those smooth scrolling JS polyfills trigger this very fast?
Well if it passed acceptance testing I guess it's fine.

src/app/utils.ts
```diff
@@ -94,8 +94,16 @@ export const assertDocumentElement = (message: string = 'BUG: Document Element i
...
+  const body = assertDocument().body;
+  const padding = body.getAttribute('data-scroll-padding') || '0';
+  const offset = parseFloat(padding) || 0;
   return scrollToElement(elem, {offset});
 };
+
+export const remsToPx = (rems: number) => {
+  const bodyFontSize = typeof(window) === 'undefined'
+    ? 10
+    : parseFloat(window.getComputedStyle(window.document.body).fontSize || '') || 10;
```
Dantemss: I think rem is technically based on the font size of `<html>`, not `<body>`? Not that it matters as long as they are the same.

Dantemss (approved):
Obviously not a huge fan of the JS scroll hack but as a temporary solution until browsers get their act together it's very acceptable.

src/app/components/ScrollOffset.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  public render() {
+    return <GlobalStyle {...this.props} />;
+  }
+
+  private resizeHandler = () => {
+    const window = assertWindow();
+    const body = window.document.body;
+
+    body.setAttribute('data-scroll-padding', String(this.getOffset(window)));
```
Dantemss: Technically this is the opposite (negative) of scroll-padding so might be worth at least leaving a comment here for the future in case we go back to scroll-padding. But we'll probably figure out we need the minus sign pretty fast anyway.

src/app/components/ScrollOffset.tsx
TomWoodward: when we use the smooth scrolling library we set the offset manually, this is only for browser scroll on initial load and clicking anchors

src/app/utils.ts
TomWoodward: yea i should probably switch this to `html`, i think thats more correct

PR #229, opened 2019-07-22
Test to randomly select book slug

Malar-Natarajan:
Sandbox is going to host books that will be in the immediate release. 
So creating a utility to list the books thats available in sandbox. The test will choose a random book from the list


pytest-selenium/tests/conftest.py
```diff
@@ -2,6 +2,8 @@
 
 import pytest
 
+import sys
```
m1yag1: These imports are in the wrong order. `import sys` should be the very first import because it is in the standard library. According to PEP 8: 

```
Imports should be grouped in the following order:
Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.
```
For more information please see [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) the Python Style Guide

pytest-selenium/tests/test_toc.py
```diff
@@ -19,7 +21,10 @@ def test_toc_toggle_button_opens_and_closes(selenium, base_url, book_slug, page_
     """
     # GIVEN: The selenium driver, base_url, book_slug, and page_slug
 
-    # WHEN: The book and page URL is loaded
+    # WHEN: The page is loaded
+    book_list = utility.Library()
```
m1yag1: This would be better as a fixture. A suggestion for this test would be.

`def test_toc_toggle_button_opens_and_closes(selenium, base_url, random_book_slug, page_slug):`

You could implement `random_book_slug` as a fixture which would do these same actions and return what is needed. In other words, what is happening here is really a `GIVEN` and involves setting up the test. In most cases this can be taken care of by a fixture. You would place this in `conftest.py` and might look something like this:

```
@pytest.fixture
def random_book_slug():
    book_list = utility.Library()
    return book_list.random_book_slug()
```
A good reference for fixtures is the pytest documentation: http://doc.pytest.org/en/latest/fixture.html

Let me know if you want more help with this.

pytest-selenium/utils/utility.py
```diff
@@ -0,0 +1,14 @@
+import random
+from random import choice
```
m1yag1: You should only need the top import. `from random import choice` is redundant.

m1yag1 (commented):
Just stopping by ;). I left some comments that I thought would be helpful.

pytest-selenium/utils/utility.py
Malar-Natarajan: Updated

pytest-selenium/tests/test_toc.py
Malar-Natarajan: Thanks for the suggestion @m1yag1 . It works and I pushed the changes

pytest-selenium/tests/conftest.py
Malar-Natarajan: Updated

Malar-Natarajan:
@TomWoodward this is ready for review

TomWoodward (approved):
this looks good. it seems like if we ever did have a test that really cared about which content it was testing then we could just not use the random slug fixture. i'm not sure how the fixtures work, does this preclude us from having a parameter called `book_slug`? if so we should probably change the fixture to be called `random_book_slug`

Malar-Natarajan:
>  i'm not sure how the fixtures work, does this preclude us from having a parameter called `book_slug`? if so we should probably change the fixture to be called `random_book_slug`

@TomWoodward 
the book_slug parameter overrides the fixture. So at any point if we want to write a test for a specific book, we can just pass book_slug and the book name in the `markers.parametrize` and it would work.

PR #234, opened 2019-07-23
feat: add sitemap

TomWoodward:


script/prerender/contentPages.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const end = (new Date()).getTime();
+  const elapsedTime = end - start;
+  const elapsedMinutes = elapsedTime / 1000 / 60;
+
+  return {numPages, elapsedMinutes};
+};
+
+type MakeRenderPage = (services: AppOptions['services']) =>
+  (action: AnyMatch, expectedCode: number) => Promise<void>;
+const makeRenderPage: MakeRenderPage = (services) => async(action, expectedCode) => {
```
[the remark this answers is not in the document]
nathanstitt: if you're inclined to take CodeClimate's advice I think line 91-106 could be extracted into a `prepareHtml` function or something like that.

script/prerender/contentPages.tsx
```diff
@@ -0,0 +1,223 @@
...
+  html = html.replace(
+    '<div id="root"></div>',
+    `<div id="root">${body}</div>` +
+    `<script>window.__PRELOADED_STATE__ = ${JSON.stringify(state).replace(/</g, '\\u003c')}</script>`
+  );
+
+  html = html.replace(/(<script src="[^"]+\.chunk\.js"><\/script>)/, scripts.join('') + '$1');
+
+  return html;
+}
```
nathanstitt: We've chatted about this before, but I'm super curious how well the regex based html manipulation will work.  I've always heard "never use regexes on HTML".  If it does prove buggy there's several DOM parsers that could be used but they're probably slower than regex.

script/prerender/fileUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+const prefixPath = (filePath: string) => path.join(ASSET_DIR, filePath);
+const directoryExists = (filePath: string) => fs.existsSync(filePath) && fs.lstatSync(filePath).isDirectory();
+
+export const assetDirectoryExists = flow(prefixPath, directoryExists);
+
+function makeDirectories(filepath: string) {
+  const dirname = path.dirname(filepath);
+  if (fs.existsSync(dirname)) {
+    return true;
+  }
```
nathanstitt: Might check if the directory can be written to using `fs.access`

nathanstitt (approved):
LGTM, a few minor comments

package.json
```diff
@@ -130,6 +130,7 @@
     "less": "^3.9.0",
     "lighthouse": "^5.1.0",
     "md5-file": "^4.0.0",
+    "moment": "^2.24.0",
     "node-fetch": "^2.3.0",
```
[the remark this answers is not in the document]
TomWoodward: i'm literally only using it for the `format` function, so if the lib you posted does similar stuff i can switch

script/prerender/contentPages.tsx
TomWoodward: yea using regex on html is... limited, only useful for simple replacements, nothing that involves any structural knowledge of the document. for now all the regex manipulations are simpler, or at least easier, than booting a dom parser, processing the whole document, and doing some traversal/manipulations on it.

if i ever start to feel like i'm putting a square peg in a round hole i'll definitely think about switching to using dom traversal for the manipulations.

script/prerender/fileUtils.ts
TomWoodward: assuming fs.mkdirSync throws when it isn't writable, i'm fine with that, it should never happen.

PR #235, opened 2019-07-23
bump lodash version because security

TomWoodward:
https://github.com/lodash/lodash/pull/4336


nathanstitt (approved):
FYI, in tutor we had a jest-axe package that hadn't updated in awhile and had lodash locked to the bad version.  We used `resolutions` to force it to update.
https://github.com/openstax/tutor-js/blob/master/package.json#L179-L181

PR #243, opened 2019-07-31
Fix prerender

TomWoodward:


script/prerender/contentPages.tsx
```diff
@@ -203,9 +196,31 @@ function injectHTML(html: string, {body, styles, state, fonts, meta, modules, ti
...
+   * the numbered chunks.
+   *
+   * it would probably be better to wait for completion of any chunks
+   * that are requested dynamically, but i'm not seeing a way in
+   * webpack's api to do that.
+   */
+  const extractAssets = () => Object.keys(assetManifest.files)
+    .filter((asset) =>
+      // chunks requested by react-loadable
+      modules.indexOf(asset.replace('.js', '')) > -1
```
nathanstitt: any reason not to use `includes` rather than the old `indexOf != -1` ?

nathanstitt (approved):
LGTM, minor comment

script/prerender/contentPages.tsx
TomWoodward: only that i forgot `includes` is a thing

PR #244, opened 2019-07-31
look up URL in the canonical book

philschatz:
does not seem to always work for pages in "Chemistry Atoms First"

Refs openstax/unified#86

# TODO

- [x] fix tests
- [x] add tests


TomWoodward:
@philschatz i glanced over this and don't really understand some of the bits, but its draft so i'm not trying very hard. let me know if you'd like me to give feedback on anything in particular.

src/app/developer/components/utils.ts
```diff
@@ -1,6 +1,41 @@
...
-import { formatBookData } from '../../content/utils';
+import { formatBookData, getUrlParamForPageId } from '../../content/utils';
 import { AppServices } from '../../types';
+import { assertDefined } from '../../utils';
+
+// from https://docs.google.com/spreadsheets/d/1Hj5vm2AbEiLgxgcbNRi550InzFTLvBOv16-Fw6P-c5E/edit#gid=2066350488
+export const CANONICAL_MAP: ObjectLiteral<string[] | undefined> = {
+  /* Algebra & Trigonometry */ '13ac107a-f15f-49d2-97e8-60ab2e3b519c' : [
+    /* College Algebra */ '9b08c294-057f-4201-9f48-5d6ad992740d',
+    /* Precalculus */ 'fd53eae1-fa23-47c7-bb1b-972349835c3c',
```
TomWoodward: maybe we should break down and have *one* file that just maps book names to uids so we dont have to have comments all over the place

src/app/developer/components/utils.ts
```diff
@@ -10,11 +45,39 @@ export async function getBooks(
...
+) {
+  const canonicals: string[] = CANONICAL_MAP[bookId] || [];
+  for (const id of canonicals) {
+    const version = assertDefined(BOOKS[id], `Could not find ${id} in BOOKS config`).defaultVersion;
+
+    const canonicalBook = await getBook(archiveLoader, osWebLoader, id, version);
+    const pageInBook = getUrlParamForPageId(canonicalBook, pageShortId, true);
+
+    if (pageInBook) {
+      return {bookSlug: canonicalBook.slug, pageSlug: pageInBook};
```
TomWoodward: why are you passing around the page params instead of the ArchivePage object? there are helpers to construct urls that you can use when you do the urling

TomWoodward (changes requested):

src/app/content/utils/urlUtils.ts
```diff
@@ -63,7 +63,11 @@ export const getBookPageUrlAndParams = (book: Book, page: Pick<Page, 'id' | 'sho
 };
 
 const getUrlParamForPageIdCache = new Map();
-export const getUrlParamForPageId = (book: Pick<Book, 'id' | 'tree' | 'title'>, pageId: string): string => {
+export const getUrlParamForPageId = (
+  book: Pick<Book, 'id' | 'tree' | 'title'>,
+  pageId: string,
+  emptyIfNotFound: boolean = false): string => {
```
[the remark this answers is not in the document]
philschatz: I wanted the entry to be cached since it seemed like the cache was there for a reason but I can use `findArchiveTreeNode` directly instead.

src/app/content/utils/canonicalUrl.ts
```diff
@@ -0,0 +1,40 @@
+import { CANONICAL_MAP } from '../../../canonicalBookMap';
+import { BOOKS } from '../../../config';
+import { AppServices } from '../../types';
+import { assertDefined } from '../../utils';
+import { formatBookData } from '../utils';
+import { findArchiveTreeNode } from './archiveTreeUtils';
+import { getUrlParamForPageTitle } from './urlUtils';
+
+export async function getBook(
```
TomWoodward: on my other branch i added a function that does the same thing as this, https://github.com/openstax/rex-web/pull/258/files#diff-72e23ff60301d48e0ffae98f76dd57b4R30 content/utils is probably a better home for a generic book loader rather than `canonicalUrl` helper

src/app/head/reducer.spec.ts
```diff
@@ -3,6 +3,9 @@ import reducer, { initialState } from './reducer';
 
 describe('head reducer', () => {
   const action = setHead({
+    link: [
```
TomWoodward: should this be `links` plural?

src/app/head/selectors.ts
```diff
@@ -11,6 +11,11 @@ export const meta = createSelector(
   (state) => state.meta
 );
 
+export const link = createSelector(
```
TomWoodward: again make more sense plural?

TomWoodward (changes requested):

src/app/head/reducer.spec.ts
philschatz: I set it to be singular because `meta` was singular and I thought `meta` and `link` referred to the element name.

I was also confused when I saw `meta` as singular, so making them both plural makes sense to me.

TomWoodward (approved):
@philschatz i made some changes, please make sure they look good to you

philschatz:
Nice, a `makeUnifiedBookLoader`. Is there a reason `archiveLoader` and `osWebLoader` are still referred to separately in the rest of the code? It seems a little confusing that there are now 3 "loaders" floating around.

👍 on the changes though, merge away!

TomWoodward:
there are some places where one or the other loader is specifically needed, but i agree we could add the unified loader to the services object and use that in most places

kjdav:
can you please redeploy this?

kjdav:
@philschatz can we hardcode openstax.org in the urls? malar pointed out that osweb hardcodes it; it seems good to do that

kjdav:
Verified by Malar and me. Can be merged when checks pass @philschatz

PR #253, opened 2019-08-06
Fix content diff

TomWoodward:
in the diff script
- ignore term ids
- format html to sort attributes
- deal with a content bug that breaks `prettyhtml`
- add `-c commit.gpgsign=false` because i have that set in my environment and it makes the script annoying

in page.tsx:
- fix the content bug

this content bug is one of those things that comes up because i'm using regex to manipulate html, the regex targeting anchors would never mistakenly target `<annotation-xml` if it was using a proper dom parser, but it works now and **_i don't wanna hear any i told you so-s_**


nathanstitt (approved):
LGTM, but there's a lot of magic sed commands that I'm only vaguely familiar with 😁

PR #267, opened 2019-08-13
FE selenium test: login and logout

Malar-Natarajan:
C477326
> GIVEN currently logged out of {base_url}/accounts
> AND the page {base_url}/books/college-physics/pages/preface   
> WHEN clicking the login link
> THEN the browser redirects to accounts/login  
> AND after successful login, redirects back to the preface page
> AND the user menu displayed in the nav includes Account Profile and Logout
> AND reloading does not reset the state back to logged out

C477327
> GIVEN currently logged in at {base_url}/accounts
> AND the page {base_url}/books/college-physics/pages/preface   
> WHEN clicking the logout link
> THEN the page reloads
> AND the login link is visible in the nav
> AND reloading does not reset the state back to logged in

Updated the book library with current Rex books


Malar-Natarajan:
@TomWoodward @m1yag1 This is ready for review

Malar-Natarajan:
@TomWoodward 
I have stored the user information in store.yaml and encrypted using encode.py. While running locally I send the password to decrypt as an argument.
Please review and once its approved, we can add these 2 variables in the  I think in the concourse env file or wherever you store the testrail credentials 
SECURE_STORE_FILE=
# SECURE_STORE_PASSWORD=

pytest-selenium/pages/base.py
```diff
@@ -75,3 +75,15 @@ def height(self, element):
...
+    def offscreen_click(self, element=None):
+        """Clicks an offscreen element (or the region's root).
+
+        Clicks the given element, even if it is offscreen, by sending the ENTER key.
+        Returns the element.
+        """
+        # We actually navigate using the ENTER key because scrolling the page can be flaky
+        # https://stackoverflow.com/a/39918249
+        # return self.page.offscreen_click(element or self.root)
+        element.send_keys(Keys.ENTER)
```
gregfitch: This may raise an exception if the `element=None` or the `find_element` in use by `element` fails; is this handled elsewhere or is a raised exception acceptable when running the test (ie: not in an `Exception` as opposed to an assertion statement)?

pytest-selenium/pages/content.py
```diff
@@ -127,11 +127,50 @@ def click_previous_link(self):
     class NavBar(Region):
         _root_locator = (By.CSS_SELECTOR, '[data-testid="navbar"]')
         _openstax_logo_link_locator = (By.CSS_SELECTOR, "div > a")
+        _user_nav_locator = (By.CSS_SELECTOR, '[data-testid="user-nav"]')
+        _login_locator = (By.CSS_SELECTOR, "[data-testid='nav-login']")
+        _user_nav_toggle_locator = (By.CSS_SELECTOR, '[data-testid="user-nav-toggle"]')
+        _account_profile_locator = (By.XPATH, "//a[contains(text(), 'Account Profile')]")
+        _logout_locator = (By.XPATH, "//a[contains(text(), 'Log out')]")
```
gregfitch: Is there a style preference within the tests for single- or double-quoted strings?

pytest-selenium/pages/content.py
```diff
@@ -127,11 +127,50 @@ def click_previous_link(self):
...
+        def user_nav(self):
+            return self.find_element(*self._user_nav_locator)
+
+        @property
+        def login(self):
+            return self.find_element(*self._login_locator)
+
+        @property
+        def user_is_not_logged_in(self):
+            return bool(self.login)
```
gregfitch: If the `self.login` property isn't found, this will raise an exception instead of returning a `bool`. Use a `try` block to catch the `NoSuchElementException` possibility or `return bool(self.find_elements(*self._login_locator))` as `bool([])` (if the login isn't found) would evaluate as `False`.

pytest-selenium/pages/content.py
```diff
@@ -127,11 +127,50 @@ def click_previous_link(self):
...
+
+        @property
+        def user_is_logged_in(self):
+            return bool(self.find_element(*self._user_nav_toggle_locator))
```
gregfitch: Same issue as above.

pytest-selenium/tests/test_login.py
```diff
@@ -0,0 +1,55 @@
...
+    # AND: After successful login, redirects back to the preface page
+    assert page_url_before_login == selenium.current_url
+
+    # AND: The user menu displayed in the nav includes Account Profile and Logout
+    if content.is_mobile:
+        user_nav.click_user_name()
+    user_nav.hover_over_user_name()
+
+    assert user_nav.account_profile.is_displayed()
+    assert user_nav.logout.is_displayed()
```
gregfitch: Word of warning: Firefox may fail here as it may consider `user_nav` a `dead element` because the page has navigated away from the book page to log in.

pytest-selenium/pages/content.py
Malar-Natarajan: Updated the current test to have double quotes for the locators. will maintain uniformity for future tests as well

pytest-selenium/pages/content.py
Malar-Natarajan: Done. But I have it like this with timeout exception:
```            try:
                self.wait.until(expected.visibility_of_element_located(self._login_locator))
                return bool(self.find_element(*self._login_locator))
            except TimeoutException:
                 return bool([])
```
because firefox was failing while looking for elements before the page loaded the usernav elements. Let me know if there is a better way

pytest-selenium/pages/content.py
Malar-Natarajan: same as above

pytest-selenium/pages/base.py
Malar-Natarajan: I think its ok as long the test is failing when it should. We haven't been catching exceptions in this case.

gregfitch (approved):

PR #274, opened 2019-08-22
google analytics tracking

jpslav:
Google analytics tracking

The code does send page view events to multiple google analytics properties.

## TODO
* [x] Change the `analytics_debug.js` back to `analytics.js`
* [ ] ~set the user ID~
* [x] Fix the pollUpdates spec
* [x] Use queue time for pending commands https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#queueTime


src/helpers/pollUpdates.ts
```diff
@@ -53,7 +76,7 @@ export const poll = (store: Store) => async() => {
 };
 
 export default (store: Store): Cancel => {
-  if (APP_ENV !== 'production') {
+  if (APP_ENV == 'production') {
```
jpslav: Wonder if this guard can be reworked?  Locally I made a `public/rex/environment.json` file to have something to pull GA IDs from.  I had to disable this guard here so that it would be fetched.

src/app/content/hooks/locationChange/index.ts
```diff
@@ -2,10 +2,12 @@ import { RouteHookBody } from '../../../navigation/types';
 import { content } from '../../routes';
 import { syncSearch } from '../../search/hooks';
 import resolveContent from './resolveContent';
+import trackPageView from './trackPageView';
 
 const hookBody: RouteHookBody<typeof content> = (services) => async(action) => {
   await resolveContent(services, action.match);
   await syncSearch(services)(action);
+  await trackPageView(action.match);
```
jpslav: Note sure if this is the right place for this.  And pretty sure that `await` isn't needed/desired?

src/app/content/hooks/locationChange/trackPageView.ts
```diff
@@ -0,0 +1,11 @@
+import googleAnalyticsClient from '../../../../gateways/googleAnalyticsClient';
+import { Match } from '../../../navigation/types';
+import { content } from '../../routes';
+
+import { assertWindow } from '../../../../app/utils';
+
+export default async(match: Match<typeof content>) => {
+  // match.route.getUrl(match.params) is the path, how to get the domain?
+  console.log(match); // to hide unused variable error
+  googleAnalyticsClient.trackPageView(assertWindow().location.href);
```
jpslav: tried to use `match` but couldn't figure out the domain.  is `window.location` good enough here?

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    if (this.isReadyForCommands()) {
+      this.flushPendingCommands();
+      this.commandEachTracker(command);
+    }
+    else {
+      this.saveCommandForLater(command);
+    }
+  };
+
```
nathanstitt: I'd suggest calling this something other than `ga`, perhaps `record` or `track`?

Rather than call `this.isReadyForCommands()` on every call, how about getting clever and have the something like:

```javascript
class GoogleAnalyticsClient {
  // by default 'record' simply records the command for later recording
  public record(cmd) { this.pending.push(cmd); } 
  // this method will record to the real GA
  private recordReal(cmd) { ... }

 setTrackingIds(...) {
    flush();
   // replace the record method with the real tracking 
   this.record = this.recordReal;
 } 
}
```

 I have no idea if typescript will agree with this, it will probably require some casting

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+      this.GA('create', id, 'auto', trackerName);
+    }
+  }
+
+}
+
+let singleton = new GoogleAnalyticsClient();
+
+export { GoogleAnalyticsClient };
+export default singleton;
```
nathanstitt: singleton should be `const`,  as a general rule everything in JS is `const` now unless there's a real need to re-assign it

src/app/content/hooks/locationChange/trackPageView.ts
```diff
@@ -0,0 +1,11 @@
+import googleAnalyticsClient from '../../../../gateways/googleAnalyticsClient';
```
TomWoodward: we don't want this `trackPageView` file to be content specific, i'd move it to `src/app/hooks/trackPageView.ts`

src/app/content/hooks/locationChange/index.ts
TomWoodward: i'd make a new hook in `src/app/hooks/trackPageView.ts` and put the call there to match any locationChange action

src/app/content/hooks/locationChange/trackPageView.ts
TomWoodward: does it want the full domain in the url or is just the path enough? if you need everything `window.location` is probably fine, or we could add it to the `Navigation/locationChange` payload

src/gateways/googleAnalyticsClient.ts
jpslav: Good idea on flushing in `setTrackingIds`.  

I kind of like exposing `ga` so that clients can know they can call whatever they would have called on the normal `ga` function.  Know what I mean?

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  private flushPendingCommands() {
+    for (let pendingCommand of this.pendingCommands) {
+      this.commandEachTracker(pendingCommand);
+    }
+    this.pendingCommands = [];
+  }
+
+  private commandEachTracker(command: Command) {
+    for (let trackerName of this.trackerNames) {
```
TomWoodward: const instead of let here

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  public setUserId(id: string) {
+    this.ga('set', 'userid', id);
+  }
+
+  public trackPageView(url: string) {
+    this.ga('send', 'pageview', url);
+  }
+
+  public setTrackingIds(ids: Array<string>) {
```
TomWoodward: should flush after setting the tracking ids, not wait until the next request comes in

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+    // The client may not be fully initialized by the time the first
+    // command comes in; in that case, save the command until that
+    // initialization.
+
+    if (this.isReadyForCommands()) {
+      this.flushPendingCommands();
+      this.commandEachTracker(command);
+    }
+    else {
+      this.saveCommandForLater(command);
```
TomWoodward: when saving these for later, wont the timestamps potentially get messed up?

src/gateways/googleAnalyticsClient.ts
TomWoodward: not sure about swapping out the functions, but you could also have the `ga` function always queue the command and then always call flush, then have flush noop if the client isn't ready

src/gateways/googleAnalyticsClient.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  };
+
+  public setUserId(id: string) {
+    this.ga('set', 'userid', id);
```
TomWoodward: yea i agree making this function called `ga` is confusing with the `window.ga`

src/helpers/pollUpdates.ts
TomWoodward: i've been developing all this `pollUpdates` stuff pretty much purely relying on unit tests, but amanda also requested that this somehow work in development so maybe we should figure that out. i'm not sure how to set it up in a super useful way though

TomWoodward (changes requested):

src/app/content/hooks/locationChange/trackPageView.ts
jpslav: you are right, path is enough, so I'm using `locationChange.payload.location.pathname`

src/gateways/googleAnalyticsClient.ts
jpslav: There's a thing about setting the queue time but I haven't figured it out yet (namely, how setting this field interacts with sending the pageview): https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#queueTime

src/gateways/googleAnalyticsClient.ts
jpslav: called ours `gaProxy` and made a new `ga` that really just wraps the `window.ga`.

src/helpers/pollUpdates.ts
jpslav: I changed it to guard specifically against the `test` env and when in development it checks for the `public/rex/environment.json` file and only proceeds if it is there.

src/gateways/googleAnalyticsClient.ts
jpslav: This is done

TomWoodward (changes requested):

jpslav:
Isn’t that what his `/i` did? (case insensitive?)

TomWoodward (approved):

PR #275, opened 2019-08-22
FE Test: Verify book url & access for free url in attribution

Malar-Natarajan:
1. Added new test:
C480905: Book url in citation section points to default page
Given - the page {base_url}/books/college-physics/pages/preface
When -  the attribution section is expanded
Then - the book url in the the citation section should reference Default page of the book (first page in the first chapter (recursively) so if there are units its the first page of the first chapter of the first unit.)

2. Added new test:
C480906: Access for free url in attribution section points to Default page
Given - the page {base_url}/books/college-physics/pages/preface
When -  the attribution section is expanded
Then - the Access for free url in the the attribution section should reference Default page of the book (first page in the first chapter (recursively) so if there are units its the first page of the first chapter of the first unit.)

3. Updated the test C476304 with the acceptance that was left out earlier:
WHEN: Navigating via TOC link
THEN: The citation/attribution section is not open on the new page


pytest-selenium/regions/toc.py
```diff
@@ -1,14 +1,22 @@
...
-from pages import content
-from pages.base import Page
 from regions.base import Region
 
 
 class TableOfContents(Region):
     _root_locator = (By.CSS_SELECTOR, "ol")
     _section_link_locator = (By.CSS_SELECTOR, "ol li a")
+    _active_section_locator = (By.CSS_SELECTOR, "[aria-label='Current Page']")
+    _default_page_locator = (By.CSS_SELECTOR, "li:nth-child(2) li:nth-of-type(1) a")
```
TomWoodward: this logic will not always be right for selecting the default page, and i would prefer to parameterize the default page. i have a thing where i try to avoid using the thing i'm testing to generate my assertions for the thing i'm testing.

@m1yag1 what do you think of this?

TomWoodward (changes requested):

pytest-selenium/regions/toc.py
m1yag1: yep, i agree.

On a separate note, the page object is meant to help expose information for the test. If this logic isn't always going to work then I wonder why that is the case? We do want the page to operate as close as possible to the actual page. It might not be the best to generate test data from but it is our interface w/ the page and useful for other things.

TomWoodward: in this case the default page just might not always be the first anchor in the second `li`, it depends on the structure of the book

Malar-Natarajan: @TomWoodward Agree....This css works perfect for books without units. For books with units it gives multiple results but still the first hit would be the default page. So I used this instead of using xpaths with AND/OR condition

TomWoodward: @Malar-Natarajan can we parameterize it so that we remove the dependency on the book structure?

Malar-Natarajan:
@TomWoodward I have used dict to store the default page for each book. Can you please review and let me know your thoughts

pytest-selenium/utils/utility.py
```diff
@@ -3,14 +3,27 @@
 
 class Library(object):
```
TomWoodward: i don't understand why this is a class, having to instantiate it before accessing the book utilities is annoying. can we make everything static? or lose the class and just export the functions?

pytest-selenium/tests/test_attribution.py
```diff
@@ -76,3 +78,58 @@ def test_attribution_collapses_on_navigating_to_new_page(selenium, base_url, boo
...
+):
+    # GIVEN: A page is loaded
+    content = Content(selenium, base_url, book_slug=book_slug, page_slug=page_slug).open()
+    attribution = content.attribution
+
+    # WHEN: The attribution section is expanded
+    attribution.click_attribution_link()
+
+    # THEN: The access for free at url in the the citation section should reference the default page of the book
+    default_page_slug = content.default_page(book_slug)
```
TomWoodward: the `default_page` utility seems like an unnecessary wrapper, you could make it so you just do `default_page_slug = utils.get_default_page_slug(book_slug)` right here

pytest-selenium/utils/utility.py
```diff
@@ -3,14 +3,27 @@
...
     def __init__(self):
-        self._book_slug_list = ["chemistry-2e", "chemistry-atoms-first-2e"]
+        self._book_dict = {
```
TomWoodward: it would be more clear to format this like

```
self._books = {
  "chemistry-2e": {
    "default_page":  "1-introduction"
  },
```

this is more extensible for when we want more stuff about the books, and includes a description about what the slug value is. the way it is now it would have to be named something like `_book_default_pages` but then also using it as the authority on what books are available is weird.

pytest-selenium/utils/utility.py
```diff
@@ -3,14 +3,27 @@
...
+        }
 
     @property
     def books(self) -> [str]:
-        return self._book_slug_list
+        return self._book_dict
 
     def random_book_slug(self):
-        return random.choice(self.books)
+        test-attribution-book_url
```
TomWoodward: what does this line do? is this a syntax error?

TomWoodward (changes requested):

pytest-selenium/utils/utility.py
Malar-Natarajan: it is a human error while resolving conflicts. I removed that

pytest-selenium/utils/utility.py
Malar-Natarajan: Tom, I removed the get_default_page out of the class so it no longer needs the instantiation.

I still have the book dict as a class because, I was having trouble with the book slug fixture while getting rid of the class and also while getting item values from the dict

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: removed the default_page from base/pages .

pytest-selenium/utils/utility.py
Malar-Natarajan: Updated the book library

Malar-Natarajan:
@TomWoodward this is ready for review

TomWoodward (approved):

PR #280, opened 2019-08-28
shrink rice logo file size

jpslav:
Shrinks the Rice logo in the footer from 1.1 MB to 8.5 KB without much if any loss in quality.


jpslav (commented):
Looks good @TomWoodward

PR #292, opened 2019-09-03
ft. search toolbar fixes

KarinaMendez2:


src/app/content/components/Toolbar/index.tsx
```diff
@@ -30,6 +30,16 @@ interface State {
 class Toolbar extends React.Component<Props, State> {
   public state = { query: '', formSubmitted: false };
 
+  constructor(props: any) {
+    super(props);
+    this.state = { query: '', formSubmitted: false };
+    this.onChange = this.onChange.bind(this);
```
TomWoodward: because the function is a `=>` function you don't need to bind it like this

TomWoodward (changes requested):

TomWoodward (approved):

PR #302, opened 2019-09-09
fix: tabbing through hidden toc

KarinaMendez2:
when toc and search sidebars are closed


src/app/content/components/Page.tsx
```diff
@@ -358,6 +360,8 @@ export default connect(
       : undefined
     ,
     searchResults: selectSearch.currentPageResults(state),
+    searchResultsOpen: selectSearch.searchResultsOpen(state),
+    tocIsOpen: select.tocOpen(state),
```
TomWoodward: are these still here for a reason?

TomWoodward (changes requested):
works good, just not sure why those extra props are still there

TomWoodward (approved):

PR #304, opened 2019-09-09
fix: mobile-scroll-lock and add to search

TomWoodward:


src/app/components/MobileScrollLock.tsx
```diff
@@ -11,7 +11,7 @@ import OnScroll, { OnScrollCallback } from './OnScroll';
 const MobileScrollLockBodyClass = createGlobalStyle`
   body {
     ${theme.breakpoints.mobile(css`
-      overflow: hidden;
+      overflow: hidden !important;
```
TomWoodward: @nathanstitt i'm overriding styles from the index.css here, because they're both global i'm assuming styled is injecting this one below the other one and they have the same specificity (i guess, even though this one has a media query and the other one doesn't) 

do you know of any tricks that would help me prioritize this override without using the `important`?

nathanstitt: hm, styled-components should be injecting it's styles at the end of `head`.  Where is `index.css` loaded from? if it's in the `body`, it could be moved into `head` and then styled-components would appear after it and have a higher priority.

https://www.styled-components.com/docs/advanced#issues-with-specificity

nathanstitt: Something else you could consider is adding a class to `body` using JS via `Window.matchMedia`.  Then instead of using media queries you could match on `.body.is-mobile` or something like that.

TomWoodward: couldn't get webpack to change the order it added the style tags, but changed to use a class hack so you can at least control how important the thing is, this resembles the recommended precedence hack from https://www.styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity

nathanstitt (approved):
Looks good, to me the `!important` isn't a deal breaker and is sometimes the best option.  Would be nice to add a comment explaining it though

PR #309, opened 2019-09-09
fix: Try to fix math

TomWoodward:
i could reproduce this issue by navigating back and forth between chem 2e pages 3.1 and 3.2 and eventually realized the problem was just when the pages were cached. i'm not totally sure that is the same issue that was reported though.

i also worked out a way to check to see if math was processed and if not try to process it again, but i ripped that back out again because it wasn't necessary for my fix, can put it back if the original issue is unresolved.

i wasn't able to reproduce the line spacing part


kjdav:
approved for merging when it can be merged

philschatz (commented):
Why would MathJax be called when there is processed math on the Page? If that occurs, it seems like there is a timing problem and filtering the processed math would continue to hide the underlying timing problem.

jpslav (approved):

TomWoodward:
@philschatz the issue was that sometimes you'd call mathjax and it wouldn't process the math. in this case it was not mathjax's fault i was just managing the update badly, but there have been reports of mathjax failing to process everything if there is a lot of math on the page. if we encounter that i'm saying we can run the typesetting, then check to see if there is remaining unprocessed math, and if so run it again. but i took that out for now because it wasn't needed for the fix i was working on

PR #310, opened 2019-09-09
Content focus on update

KarinaMendez2:


TomWoodward (changes requested):

src/app/content/components/Page.tsx
```diff
@@ -42,6 +42,8 @@ interface PropTypes {
   searchResults: SearchResultHit[];
   search: RouteState<typeof content>['search'];
   services: AppServices;
+  tocIsOpen: boolean;
+  searchResultsOpen: boolean;
```
TomWoodward: i think these extra props came over from your other branch

TomWoodward (changes requested):

TomWoodward (approved):

PR #311, opened 2019-09-12
fix: Edge polyfills + refactor everything

TomWoodward:


src/app/content/components/Page.tsx
```diff
@@ -30,6 +30,10 @@ import { toRelativeUrl } from '../utils/urlUtils';
 import { contentTextWidth } from './constants';
 import allImagesLoaded from './utils/allImagesLoaded';
 
+if (typeof(document) !== 'undefined') {
+  import(/* webpackChunkName: "NodeList.forEach" */ 'mdn-polyfills/NodeList.prototype.forEach');
+}
```
jpslav: Useful to add a comment re what this is doing? (might be hard for future readers or current me to figure out :-)

src/app/content/utils/domUtils.ts
```diff
@@ -1,5 +1,9 @@
 import { HTMLElement } from '@openstax/types/lib.dom';
 
+if (typeof(document) !== 'undefined') {
+  import(/* webpackChunkName: "Node.children" */ 'mdn-polyfills/Node.prototype.children');
+}
```
jpslav: Can this be put in some shared place (it is duplicated in another file above) with a comment explaning it?

src/app/developer/components/Notifications.tsx
```diff
@@ -19,7 +19,7 @@ const Notifications = ({updateAvailable, error, sendMessages}: Props) => {
   const [showError, setError] = useState(false);
 
   if (showError) {
-    throw new Error();
+    throw new Error('test error');
```
jpslav: is this `'test error'` supposed to stick around or was it a debugging thing?

src/app/head/hooks/locationChange.ts
```diff
@@ -2,6 +2,10 @@ import { locationChange } from '../../navigation/actions';
 import { ActionHookBody } from '../../types';
 import { actionHook, assertDefined } from '../../utils';
 
+if (typeof(document) !== 'undefined') {
+  import(/* webpackChunkName: "NodeList.forEach" */ 'mdn-polyfills/NodeList.prototype.forEach');
+}
```
jpslav: ditto this re comment above

jpslav (changes requested):

src/app/content/utils/domUtils.ts
TomWoodward: not duplicated, different polyfill

src/app/developer/components/Notifications.tsx
TomWoodward: no this is irl a test error, having an actual message will help filter sentry

src/app/head/hooks/locationChange.ts
TomWoodward: this one is duplicated from Page.tsx, but i like to do a thing where i include the polyfill in each place that its used, so that:
- if the code is split its always included when its needed
- its close to the code that its needed for so if you update the code to need or not need polyfills you can change them right there

that said i started a pattern on #315 that might be more clear and save 4 lines of code

jpslav (approved):

PR #312, opened 2019-09-12
fix:safari intl

TomWoodward:


philschatz (approved):
How did these errors surface? From a sentry entry? I'm just curious how informative the Entry was

KarinaMendez2 (approved):

PR #313, opened 2019-09-13
fix: Unhandled rejection

TomWoodward:
at first i thought maybe the memoization was messing up the promise chain, even though that makes no sense. that wasn't the case but i also noticed that memoizing the fetch functions was memoizing failed requests, so if your network is flaky and a page request fails it just never tries again, and that sucked so i left the change in this PR.

in the sentry logs for the unhandled rejection it just says `CustomEvent` and the only occurrence of custom event that i could find was in the ally.js polyfill, so i've swapped it out and hopefully the errors will go away, i don't have solid evidence anything will be fixed so this is more exploratory than anything.


philschatz (approved):
🤞 that CustomEvents are gone

PR #340, opened 2019-10-03
fix: remove mathjax plugins

TomWoodward:


kjdav:
Discussed with Tom that in order to see the fix, you have to right click math and then disable some settings under the Accessibility menu bc Mathjax is setting a cookie to remember those settings. So Tom is going to see if he can figure disable that cookie after we make this change, so that end users see the fix without having to adjust their settings manually.

Discussed also that we are changing the REX configuration to match what CNX effectively has. Although CNX has math-sre enabled, Tom says it's not doing anything so we are not taking anything away from users that they have today.

KarinaMendez2 (approved):

PR #343, opened 2019-10-07
chore: more books

TomWoodward:


kjdav:
Can you please update this to get the latest version of US History @TomWoodward?

PR #362, opened 2019-11-04
My highlights button and pop up

KarinaMendez2:
Initial styles for user not logged in for my highlights pop up, created selectors for my highlights opened and closed actions


src/app/components/Layout.tsx
```diff
@@ -10,6 +11,7 @@ import SkipToContentWrapper from './SkipToContentWrapper';
 const Layout: SFC = ({ children }) => <SkipToContentWrapper>
   <NavBar />
   <ErrorModal />
+  <HighlightsPopUp />
```
TomWoodward: this should be in `/src/app/content/components/Content.tsx`

src/app/content/components/Toolbar/HighlightsPopUp.tsx
```diff
@@ -0,0 +1,441 @@
...
+        <Modal>
+          <Mask>
+            <Wrapper>
+              <Header>
+                <FormattedMessage id='i18n:toolbar:highlights:popup:heading'>
+                  {(msg: Element | string) => msg}
+                </FormattedMessage>
+                <CloseIcon data-testid='close-highlights-popup' onClick={() => this.props.closeMyHighlights()}/>
+              </Header>
+              {this.props.user ? this.myHighlights() : this.loginForHighlights()}
```
TomWoodward: we should split the `myHighlights` and `loginForHighlights` content into separate files, this file is getting really big already.

src/app/content/components/Toolbar/index.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+import createTestStore from '../../../../test/createTestStore';
+import { receiveUser } from '../../../auth/actions';
+import { User } from '../../../auth/types';
+import MessageProvider from '../../../MessageProvider';
+import { Store } from '../../../types';
+import { closeMyHighlights, openMyHighlights } from '../../highlights/actions';
+import HighlightButton from './HighlightButton';
+import HighlightsPopUp from './HighlightsPopUp';
+
+describe('MyHighlights', () => {
```
TomWoodward: this file has tests for both the highlights button and the modal popup, neither of those tests really have anything to do with the toolbar but the test file is called `Toolbar/index.spec.tsx`.

i think these tests look good, but they should be separated out for the components that they're testing.

src/app/content/reducer.ts
```diff
@@ -17,6 +17,7 @@ import { getPageSlug } from './utils/archiveTreeUtils';
 export const initialState = {
   highlights: initialHighlightState,
   loading: {},
+  myHighlightsOpen: initialHighlightState.myHighlightsOpen,
```
TomWoodward: this is already dealt with in `src/app/content/highlights/reducer.ts`, you don't need it here

src/app/content/types.ts
```diff
@@ -11,6 +11,7 @@ export interface Params {
 
 export interface State {
   tocOpen: boolean | null;
+  myHighlightsOpen: boolean;
```
TomWoodward: or here

TomWoodward (changes requested):

src/app/content/components/Content.tsx
```diff
@@ -167,6 +168,7 @@ const Content = ({mobileExpanded}: {mobileExpanded: boolean}) => <Layout>
   />
   <Background>
     <BookBanner />
+    <HighlightsPopUp />
```
TomWoodward: might as well put this inside the error boundary one line down

src/app/content/components/Toolbar/MyHighlights/HighlightStyles.tsx
```diff
@@ -0,0 +1,297 @@
+import React from 'react';
```
TomWoodward: please move all the modal stuff out of `Toolbar` and into `content/highlights/components`

src/app/content/components/Toolbar/MyHighlights/HighlightStyles.tsx
```diff
@@ -0,0 +1,297 @@
...
+  defaultOffset: 3.2,
+  green: 'rgba(99, 165, 36)',
+  height: 8,
+  left: 32.8,
+  opacity: '0.85',
+  tooltip: {
+    height: 8.8,
+    width: 18,
+  },
+  width: 24.8 + 5, /* to allow text to fit in one line with tooltip */
```
TomWoodward: why not just set this to `29.8`? why do the math here like this?

src/app/content/components/Toolbar/MyHighlights/HighlightsPopUp.tsx
```diff
@@ -0,0 +1,146 @@
...
+  closeMyHighlights: () => void;
+  user?: User;
+  loggedOut: boolean;
+  currentPath: string;
+}
+
+class HighlightsPopUp extends Component<Props> {
+  public loginForHighlights = () => {
+    return <Styled.PopupBody>
+      <Styled.LoginText values={{loginLink: '/accounts/login?r=' + this.props.currentPath}} />
```
TomWoodward: i added `loginLink` to authSelectors so you don't have to have `'/accounts/login?r='` hardcoded here anymore

src/app/content/highlights/reducer.ts
```diff
@@ -24,6 +25,10 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.createHighlight): {
       return {...state, highlights: [...state.highlights, action.payload]};
     }
+    case getType(actions.openMyHighlights):
+      return {...state, myHighlightsOpen: true};
+    case getType(actions.closeMyHighlights):
+      return {...state, myHighlightsOpen: false};
```
TomWoodward: should probably also set `myHighlightsOpen: false` on `locationChange` up on line 23

TomWoodward (changes requested):

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -0,0 +1,297 @@
...
+  z-index: 1;
+  width: 100%;
+  background: ${theme.color.neutral.base};
+  margin: 3rem;
+  border-radius: 0.5rem;
+`;
+
+// tslint:disable-next-line:variable-name
+export const Header = styled(H3)`
+  background: #002569;
```
TomWoodward: this is supposed to match the book color, don't have to change it now but you should make a follow up PR before we forget

TomWoodward (approved):

PR #372, opened 2019-11-14
FE Selenium test: Rex & osweb connecting login scenarios

Malar-Natarajan:
C477329: Logout with multiple openstax components open in different tabs 
GIVEN: Openstax.org and REX are open in different tabs in logged-in state
WHEN: click the logout link in openstax.org
THEN: REX tab will stay the same logged-in state
AND: REX tab goes to logged-out state on a reload 

C477328: Login state while redirected from openstax.org
GIVEN: User is logged into {base_url}
WHEN: user clicks 'view online' option for a book that is available in REX (Eg. college physics)
THEN: the book is opened in REX with the header showing <Hi {firstname}> with the same user as openstax.org
AND: the user stays logged-in while navigating to other pages in REX

Added more books to the Utility file


pytest-selenium/pages/osweb.py
```diff
@@ -1,14 +1,112 @@
...
+
+    @property
+    def mobile_user_nav(self):
+        return self.find_element(*self._mobile_user_nav_locator)
+
+    @property
+    def user_is_logged_in(self):
+        if self.is_desktop:
+            if self.is_element_present(*self._user_nav_locator):
+                return True
```
gregfitch: instead of "if bool: return bool" just "return bool"
`return self.is_element_present(*self._user_nav_locator)`

pytest-selenium/pages/osweb.py
```diff
@@ -1,14 +1,112 @@
...
+        elif self.is_mobile:
+            self.mobile_user_nav.click()
+            if self.is_element_present(*self._mobile_user_nav_locator):
+                self.mobile_user_nav.click()
+                self.wait.until(
+                    expected.invisibility_of_element_located(self._mobile_user_nav_loaded_locator)
+                )
+                return True
```
gregfitch: This should probably be in a `try:` block. If the `wait` fails, instead of returning `False` (not logged in) you'll get a `TimeoutException`.

pytest-selenium/pages/osweb.py
```diff
@@ -1,14 +1,112 @@
...
+            return self.find_element(*self._view_online_desktop_locator)
+        elif self.is_mobile:
+            return self.find_element(*self._view_online_mobile_locator)
+
+    def click_login(self):
+        if self.is_desktop:
+            self.login.click()
+        elif self.is_mobile:
+            self.click_mobile_user_nav()
+            self.login.click()
```
gregfitch: Is there a third state possible? (ie not desktop nor mobile) If yes, ignore this :) otherwise the `elif self.is_mobile:` should probably be `else:` so each execution path is covered.

pytest-selenium/tests/test_login.py
```diff
@@ -48,3 +51,81 @@ def test_login_and_logout(selenium, base_url, book_slug, page_slug, email, passw
...
+    rex_nav.click_login()
+
+    accounts = Login(selenium)
+    accounts.login(email, password)
+
+    # AND: Open osweb url in a new tab
+    rex.open_new_tab()
+    rex.switch_to_window(1)
+
+    sleep(1)
```
gregfitch: The `sleep` should be in the object, not the test (unless you're executing specific things like accessibility keyboard navigation).

PR #379, opened 2019-11-19
ft. my highlights popup - body scroll lock

KarinaMendez2:


src/app/content/highlights/components/HighlightsPopUp.tsx
```diff
@@ -121,26 +122,31 @@ class HighlightsPopUp extends Component<Props> {
...
-                data-testid='close-highlights-popup'
-                onClick={() => this.props.closeMyHighlights()}
-              />
-            </Styled.Header>
-            {this.props.user ? this.myHighlights() : this.loginForHighlights()}
-          </Styled.Wrapper>
-        </Styled.Mask>
-      </Styled.Modal>
+      <React.Fragment>
+        <ScrollLock overlay={false} mobileOnly={false} />
```
TomWoodward: looks like `ScrollLock` will render a mask, so you can probably use it inside the modal instead of the custom mask, just put it above the wrapper and i think it'll work. bonus, it has a click handler you can assign to close the modal

you also won't need the fragment

TomWoodward (changes requested):

TomWoodward (approved):

PR #390, opened 2019-12-09
ft. Highlights popup - Scroll lock mask for highlights

KarinaMendez2:


src/app/components/ScrollLock.tsx
```diff
@@ -31,26 +31,30 @@ const fadeIn = keyframes`
 `;
 
 // tslint:disable-next-line:variable-name
-const Overlay = styled.div`
+export const Overlay = styled.div`
   animation: ${sidebarTransitionTime}ms ${fadeIn} ease-out;
   background-color: ${Color(theme.color.primary.gray.base).alpha(0.75).string()};
-  z-index: ${theme.zIndex.overlay}; /* stay above book content */
+  z-index: ${(props: {isModal?: boolean}) => props.isModal ? theme.zIndex.highlights : theme.zIndex.overlay};
```
TomWoodward: why do you need this?

TomWoodward: oh maybe because the default ToC overlay doesn't cover the book banner? is that what you're trying to fix here? would be better to just pass in the correct zIndex, rather than switching internally here. i would default to `theme.zIndex.overlay` and move that above all the navs and junk, then change the ToC to pass in a zIndex which is correct for it (below banner or whatever)

TomWoodward (changes requested):
looking good, but i think that zIndex logic needs to be organized differently

TomWoodward (approved):

PR #393, opened 2019-12-11
chore: add some utilities that will be used when paginating highlight summary page

TomWoodward:


src/app/content/highlights/selectors.ts
```diff
@@ -35,3 +49,30 @@ export const summaryIsLoading = createSelector(
...
+    mapValues(size)
+  )
+);
+
+const filteredTotalCountsPerPageInSummary = createSelector(
+  localState,
+  (state) => state.summary.filteredTotalCounts
+);
+
+export const totalCountsPerPageInSummary = createSelector(
```
TomWoodward: this selector isn't currently used, but should be used to disable chapter filters that have no highlights in any of their pages

src/app/content/highlights/utils/paginationUtils.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return Object.keys(reduceUntilPageSize(pages, addPageCount, {} as CountsPerSource));
+};
+
+/*
+ * CountsPerSource returned by api does not filter by source id, only color,
+ * so in order for our filteredTotalCounts to reflect reality we need to filter
+ * its page ids by the filter chapters
+ */
+export const filterCountsPerSourceByChapters = (
```
TomWoodward: this should be used when populating `filteredTotalCounts`

src/app/content/highlights/utils/paginationUtils.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/*
+ * in order to avoid passing all page ids for all selected chapters in
+ * api requests to get highlights, we calculate how many ids are necessary
+ * to fill a response based on the given CountsPerSource
+ */
+export const getNextPageSources = (
+  remainingCounts: CountsPerSource,
+  tree: ArchiveTree,
+  nextPageSize: number
```
TomWoodward: remainingCounts comes from selectors.remainingSourceCounts, tree comes from book.tree, book is required to be known when getting highlights because it is used as the scopeId. nextPageSize can be anything but should be the same as the perPage passed to getHighlights

PR #396, opened 2019-12-12
Skip changing ids

janKozie1:
Note: 
Most of the changes are in Highlighter's repo, I've only added a new config option here

### DIFF
--- a/src/app/content/components/Page/highlightManager.ts
+++ b/src/app/content/components/Page/highlightManager.ts
@@ -59,14 +59,15 @@ const onSelectHighlight = (
 });
 
 const createHighlighter = (services: Omit<Services, 'highlighter'>) => {
+
   const highlighter: Highlighter = new Highlighter(services.container, {
-    onClick: (...args) => onClickHighlight({...services, highlighter}, ...args),
-    onSelect: (...args) => onSelectHighlight({...services, highlighter}, ...args),
+    onClick: (...args) => onClickHighlight({ ...services, highlighter }, ...args),
+    onSelect: (...args) => onSelectHighlight({ ...services, highlighter }, ...args),
+    skipIDsBy: [new RegExp(/^\d+$/), new RegExp(/^term/)],
     snapMathJax: true,
     snapTableRows: true,
     snapWords: true,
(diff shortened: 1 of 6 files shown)


src/app/content/components/Page/highlightManager.ts
```diff
@@ -59,22 +59,23 @@ const onSelectHighlight = (
 });
 
 const createHighlighter = (services: Omit<Services, 'highlighter'>) => {
+
   const highlighter: Highlighter = new Highlighter(services.container, {
-    onClick: (...args) => onClickHighlight({...services, highlighter}, ...args),
-    onSelect: (...args) => onSelectHighlight({...services, highlighter}, ...args),
+    onClick: (...args) => onClickHighlight({ ...services, highlighter }, ...args),
+    onSelect: (...args) => onSelectHighlight({ ...services, highlighter }, ...args),
+    skipIDsBy: [new RegExp(/^\d+$/), new RegExp(/^term/)],
```
TomWoodward:
```suggestion
    skipIDsBy: /^(\d+$|term)/,
```

src/app/content/components/Page/highlightManager.ts
```diff
@@ -137,11 +138,11 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
...
-                ...listHighlights.filter(
-                  (highlight) => !listPendingHighlight || highlight.id !== listPendingHighlight.id
-                ),
-                listPendingHighlight,
-              ]
+              ...listHighlights.filter(
+                (highlight) => !listPendingHighlight || highlight.id !== listPendingHighlight.id
+              ),
+              listPendingHighlight,
+            ]
```
TomWoodward: please avoid changing unrelated whitespace, it makes the diff harder to read, if it makes your life easier you can try using the `git add -p` option that allows you to stage chunks of diffs instead of whole files

TomWoodward (changes requested):
pretty sure the checks are failing because of type mismatch on the new options, they'll start passing when we merge openstax/highlighter#31, publish a new version, and add it here

PR #398, opened 2019-12-13
Filters

PiotrKozlowski:
#https://github.com/openstax/unified/issues/587

- Filters are saved in redux store
- Filters are displayed
- New actions are created for updating filters and receiving summary highlights
- Display summary highlights
- Loading state when changing filters
- When there is no colors or chapters selected then highlights will not be shown
- On ~~locationChange~~ summary popup open, current page / chapter will be added to filters

TODO:
- [x] Add more test and fixed those which are broken
- [x] Check if everything is displaying correctly on mobile and if not then fix it


src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
...
+// tslint:disable-next-line: variable-name
+const Highlights = () => {
+  const book = useSelector(bookSelector);
+  const highlights = useSelector(summaryHighlights);
+  const isLoading = useSelector(summaryIsLoading);
+
+  if (book && Object.keys(highlights).length > 0) {
+    return <Styled.Highlights isLoading={isLoading}>
+      {isLoading ? <Styled.LoaderWrapper><Loader/></Styled.LoaderWrapper> : null}
+      {highlights[book.tree.id]
```
TomWoodward: there will never be highlights associated with the book id, its always a page id

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
...
+        && <SectionHighlights
+          section={book.tree}
+          highlights={highlights}
+        />}
+      {book.tree.contents.map((chapter) => (
```
TomWoodward: this won't accurately find chapters, you need to flatten the tree and filter with `archiveTreeUtils.archiveTreeSectionIsChapter`, different books have different table of contents formatting and they're not always the first children of the book.

its also possible to have highlights in pages that are not in a chapter

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
...
+interface SectionHighlightsProps {
+  section: ArchiveTreeSection;
+  highlights: SummaryHighlights;
+}
+
+// tslint:disable-next-line: variable-name
+const SectionHighlights = ({ section, highlights }: SectionHighlightsProps) => (
+  <>
+    <Styled.HighlightsChapter>
+      {stripHtmlAndTrim(section.title)}
```
TomWoodward: pretty sure you don't want to do this

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
...
+    </Styled.HighlightsChapter>
+    <Styled.HighlightWrapper>
+    {(section as ArchiveTree).contents.map((page) => {
+      const pageId = stripIdVersion(page.id);
+      if (!highlights[section.id][pageId]) { return null; }
+      return <div key={page.id}>
+        <Styled.HighlightSection>
+          {stripHtmlAndTrim(page.title)}
```
TomWoodward: or this

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
...
+export default Highlights;
+
+interface SectionHighlightsProps {
+  section: ArchiveTreeSection;
+  highlights: SummaryHighlights;
+}
+
+// tslint:disable-next-line: variable-name
+const SectionHighlights = ({ section, highlights }: SectionHighlightsProps) => (
+  <>
```
TomWoodward: <React.Fragment>

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
```
TomWoodward: breaking things into smaller files is cool, but you're gonna have hella conflicts with karina's changes

src/app/content/highlights/components/SummaryPopup/ChapterFilter.tsx
```diff
@@ -67,22 +66,28 @@ const chunk = <T extends any>(sections: T[]) => {
...
-    )
-  );
-  const [selectedChapters, setSelectedChapters] = React.useState<string[]>(currentChapter ? [currentChapter.id] : []);
+
+  const selectedChapters = useSelector(chaptersFilter);
+  const dispatch = useDispatch();
+
+  const setSelectedChapters = (chapterIds: string[]) => {
+    dispatch(setChaptersFilter(chapterIds));
+    dispatch(filtersChange());
```
TomWoodward: shouldn't need this

src/app/content/highlights/components/SummaryPopup/ChapterFilter.tsx
```diff
@@ -114,6 +116,7 @@ export default styled(ChapterFilter)`
   outline: none;
   max-height: 72rem;
   overflow: auto;
+  z-index: 2;
```
TomWoodward: this is being fixe on another card, make sure merge leaves us with only 1 z-index directive

src/app/content/highlights/components/SummaryPopup/ColorFilter.tsx
```diff
@@ -1,23 +1,41 @@
...
 // tslint:disable-next-line:variable-name
 const ColorFilter = ({className}: Props) => {
-  const allColors = highlightStyles.map((style) => style.label);
-  const [selectedColors, setSelectedColors] = React.useState<HighlightColorEnum[]>(allColors);
+  const selectedColors = useSelector(colorsFilter);
+  const dispatch = useDispatch();
+
+  const setSelectedColors = (colors: HighlightColorEnum[]) => {
+    dispatch(setColorsFilter(colors));
+    dispatch(filtersChange());
```
TomWoodward: shouldn't need this

src/app/content/highlights/components/SummaryPopup/ColorFilter.tsx
```diff
@@ -48,6 +63,7 @@ export default styled(ColorFilter)`
   font-size: 1.4rem;
   padding: 0.8rem 1.6rem;
   outline: none;
+  z-index: 2;
```
TomWoodward: same as in chapters

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,161 @@
...
+    body = <FormattedMessage id={`i18n:highlighting:colors:${props.color}`}>
+      {(msg: Element | string) => msg}
+    </FormattedMessage>;
+  } else {
+    body = <span dangerouslySetInnerHTML={{ __html: props.title }}/>;
+  }
+
+  return <li>
+    <RemoveIcon onClick={handleClick}><Times /></RemoveIcon>
+    <ItemLabel>{body}</ItemLabel>
```
TomWoodward: there is no need to have two nested spans for the chapter/page titles here

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,161 @@
...
+    width: 1rem;
+  }
+`;
+
+// tslint:disable-next-line: variable-name
+const ItemLabel = styled.span`
+  max-width: 80px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
TomWoodward: this truncation is causing some weird spacing, the element with the truncated text appears to have more right margin than the others 

image

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const onRemoveItem = (type: 'colors' | 'chapters', label: HighlightColorEnum | string) => {
+    if (type === 'colors') {
+      setSelectedColors(selectedColors.filter(not(match(label))));
+    } else if (type === 'chapters') {
+      setSelectedChapters(selectedChapters.filter(not(match(label))));
+    }
+  };
+
+  React.useEffect(() => {
```
TomWoodward: i think you're doing this so that it doesn't reevaluate the whole tree anytime the component is re-rendered, it would be cleaner to use a selector to do this.

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }, [book]);
+
+  return <ul className={className}>
+    {selectedChapters.map((chapterId) => sections.has(chapterId) && <FiltersListItem
+      key={chapterId}
+      type='chapters'
+      title={sections.get(chapterId)!.title}
+      chapterId={chapterId}
+      onRemoveItem={onRemoveItem}
```
TomWoodward: you're already assigning this separately for chapters and colors, i'm not sure what value the shared onRemoveItem function is providing, you're merging the data flow just to have to split it out again.

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,161 @@
...
+  font-size: 1.4rem;
+  display: flex;
+  flex-wrap: wrap;
+  width: 100%;
+  margin: 0;
+  padding: 2rem 0;
+  list-style: none;
+
+  ${theme.breakpoints.mobile(css`
+    display: none;
```
TomWoodward: 👍 good catch on the mobile designs!

src/app/content/highlights/components/SummaryPopup/__snapshots__/ChapterFilter.spec.tsx.snap
```diff
@@ -248,7 +249,7 @@ input:checked + .c10 .c11 {
         className="c7 c8"
       >
         <input
-          checked={true}
+          checked={false}
```
TomWoodward: this seems worrying

src/app/content/highlights/hooks/addCurrentPageToSummaryFilters.ts
```diff
@@ -0,0 +1,35 @@
...
+  const {book, page} = bookAndPage(state);
+  const selectedChapters = select.chaptersFilter(state);
+  if (!book || !page || typeof(window) === 'undefined') {
+    return;
+  }
+
+  // Add current page or their parent to the chaptersFilter
+  const treeNode = findArchiveTreeNode(book.tree, page.id);
+  if (!treeNode) { return; }
+  const nodeParent = treeNode.parent!;
```
TomWoodward: chapters aren't always one level deep, sometimes they have groups of pages inside them. i'd stick to iterating over the chapters and checking to see if the current page id is any descendant of them.

src/app/content/highlights/hooks/filtersChange.ts
```diff
@@ -0,0 +1,77 @@
...
+    colors: selectedColors as unknown as GetHighlightsColorsEnum[],
+    perPage: 30,
+    scopeId: book.id,
+    sourceIds,
+    sourceType: GetHighlightsSourceTypeEnum.OpenstaxPage,
+  });
+
+  if (highlights.data && book) {
+    const pages = archiveTreeUtils.findTreePages(book.tree);
+    const summaryHighlights: SummaryHighlights = {};
```
TomWoodward: grouping highlights to  be keyed by chapter and page seems like a good thing to split out a utility function for

src/app/content/highlights/hooks/filtersChange.ts
```diff
@@ -0,0 +1,77 @@
...
+      } else {
+        summaryHighlights[chapterId] = {
+          [pageId]: [h],
+        };
+      }
+    }
+
+    dispatch(receiveSummaryHighlights(summaryHighlights));
+  } else {
+    dispatch(setIsLoadingSummary(false));
```
TomWoodward: i'm trying to decide if its a good idea to still `dispatch(receiveSummaryHighlights({}));` in this case... i think its a good idea... but i'm not sure

src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -25,6 +25,8 @@ const hookBody = ({dispatch, getState, highlightClient}: MiddlewareAPI & AppServ
   if (highlights.data) {
     dispatch(receiveHighlights(highlights.data));
   }
+
+  dispatch(addCurrentPageToSummaryFilters());
```
TomWoodward: you could just compose this logic here by calling a function, no need to dispatch an extra method

src/app/content/highlights/reducer.ts
```diff
@@ -9,13 +9,20 @@ import * as actions from './actions';
 import { highlightingFeatureFlag } from './constants';
 import { State } from './types';
 
+const defaultColors = [
+  HighlightColorEnum.Blue,
+  HighlightColorEnum.Green,
+  HighlightColorEnum.Pink,
+  HighlightColorEnum.Purple,
+  HighlightColorEnum.Yellow,
+];
```
TomWoodward: you could derive this from the constant like you did in ColorFilter instead of hardcoding it

src/app/content/highlights/reducer.ts
```diff
@@ -80,6 +87,34 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      const newState = {...state};
+      newState.summary.loading = action.payload;
+      return newState;
+    }
+    case getType(actions.receiveSummaryHighlights): {
+      const newState = {...state};
+      const { chapters, colors } = newState.summary.filters;
+      if (chapters.length === 0 || colors.length === 0) {
+        // When we make api call without filters it is returning all highlights
+        // so we manually set it to empty object.
```
TomWoodward: the reducer shouldn't care about this, the hook should detect this case and short circuit before actually loading all highlights

TomWoodward (changes requested):
when i tested on the review environment some highlights were loaded into the redux state but the dialog remained empty, i think that logic in the Highlights component was not working correctly but i'm not exactly sure how

src/app/content/highlights/actions.ts
```diff
@@ -11,3 +11,10 @@ export const receiveHighlights = createStandardAction('Content/Highlight/receive
 
 export const openMyHighlights = createStandardAction('Content/openMyHighlights')<void>();
 export const closeMyHighlights = createStandardAction('Content/closeMyHighlights')<void>();
+
+export const setColorsFilter = createStandardAction('Content/setColorsFilter')<HighlightColorEnum[]>();
+export const setChaptersFilter = createStandardAction('Content/setChaptersFilter')<string[]>();
+export const filtersChange = createStandardAction('Content/filtersChange')<void>();
+export const setIsLoadingSummary = createStandardAction('Content/setIsLoadingSummary')<boolean>();
```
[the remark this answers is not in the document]
PiotrKozlowski: I've merge `setColorsFilter` and `setChaptersFilter` to `setSummaryFilters` and removed `setIsLoadingSummary` + hooked `filtersChange` to `setSummaryFilters`

src/app/content/highlights/actions.ts
```diff
@@ -11,3 +11,10 @@ export const receiveHighlights = createStandardAction('Content/Highlight/receive
...
+export const receiveSummaryHighlights = createStandardAction('Content/receiveSummaryHighlights')<SummaryHighlights>();
+export const addCurrentPageToSummaryFilters = createStandardAction('Content/addCurrentPageToSummaryFilters')<void>();
```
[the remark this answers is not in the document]
PiotrKozlowski: I've extracted this so I can use it in tests for selecting given page, but maybe I can somehow call `locationChange` in tests?

src/app/content/highlights/components/Highlights.tsx
PiotrKozlowski: Placeholder interface for `summary.highlights` was `{ [chapterId]: { [pageId]: Highlight[] } }` so for example for Preface I'm setting book's id as a chapterId, but I can change summary.highlights to be just a map with pageIds which should be easier to manage

src/app/content/highlights/components/Highlights.tsx
PiotrKozlowski: I assume I'll have to do this anyway so let me know which way you see better for project management

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
PiotrKozlowski: This is actually happening only for this `Function...` chapter. Probably because of different width of letters. I've could make like in cnx.org homepage for book excerptions. Let me know if you have something else in mind.
image

src/app/content/highlights/components/Highlights.tsx
TomWoodward: i was worried that re-sorting the pages into their chapters all the time would be annoying and heavy, but we could have chapter OR page ids on the top level of the object, and then have subobject for chapters.

or if you're thinking of something that you think you work well just do it your way and i will look at it.

src/app/content/highlights/components/Highlights.tsx
TomWoodward: you can continue as you are, just be very aware of karina's changes to the old component files

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
TomWoodward: ok, if its just because of the letter width that makes sense, i'll look more closely with different filters

src/app/content/highlights/hooks/addCurrentPageToSummaryFilters.ts
PiotrKozlowski: Ok, now I'm iterating over sections

src/app/content/highlights/hooks/filtersChange.ts
PiotrKozlowski: It feels like this is okay if we decide to remove `setIsLoadingSummary` (which I already did :D) 
I've restructured this a little so now it should be easier to ready

src/app/content/highlights/hooks/locationChange.ts
PiotrKozlowski: I've made it separate to make tests easier to write, like:
```
it('initially has the selected page checked', () => {
    store.dispatch(receiveBook(book));
    hook(addCurrentPageToSummaryFilters());
```

Can I achive the same by calling `receiveUser` (`locationChange` is hooked to this)? I'll check

src/app/content/highlights/components/SummaryPopup/__snapshots__/ChapterFilter.spec.tsx.snap
PiotrKozlowski: I've added only `hook(addCurrentPageToSummaryFilters());` to those tests so I'm not sure why output has changed

src/app/content/highlights/reducer.ts
```diff
@@ -32,11 +34,23 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      const highlight = {
         ...action.payload,
         color: action.payload.color as string as HighlightColorEnum,
         sourceType: action.payload.sourceType as string as HighlightSourceTypeEnum,
-      }]};
+      };
+      newState.highlights = [...state.highlights || [], highlight];
+
+      const { chapterId, pageId } = action.meta;
+      updateSummaryHighlights(newState.summary.highlights, {
```
TomWoodward: this should return the new object, not modify it by reference

src/app/content/highlights/reducer.ts
```diff
@@ -47,27 +61,104 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      const { color, annotation } = action.payload.highlight;
+
+      const newState = {...state};
+      const dataToUpdate = {
+        ...action.payload.highlight,
+        color: action.payload.highlight.color as string as HighlightColorEnum,
       };
+      let newHighlight: Highlight | undefined;
+
+      newState.highlights = newState.highlights!.map((h) => {
```
TomWoodward: please don't use `!`, just handle the null check

src/app/content/highlights/reducer.ts
```diff
@@ -47,27 +61,104 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      });
+
+      const { chapterId, pageId } = action.meta;
+      const { filters: { chapters, colors }, highlights } = newState.summary;
+
+      // If highlight's chapter is not in summary filters stop here...
+      if (!chapters.includes(chapterId)) { return newState; }
+
+      // If only annotation was changed just update summary highlights...
+      if (!color && annotation) {
```
TomWoodward: all this branch and bound logic is nuts and should be extracted to a helper method if it can't be simplified

src/app/content/hooks/receiveContent.ts
```diff
@@ -10,7 +10,7 @@ import * as select from '../selectors';
 import { getCanonicalUrlParams } from '../utils/canonicalUrl';
 import getCleanContent from '../utils/getCleanContent';
 
-const stripHtmlAndTrim = (str: string) => str
+export const stripHtmlAndTrim = (str: string) => str
```
TomWoodward: don't need to export this

src/app/content/types.ts
```diff
@@ -107,3 +107,5 @@ export interface ArchivePage {
 }
 
 export type ArchiveContent = ArchivePage | ArchiveBook;
+
+export type BookSections = Map<string, LinkedArchiveTree | LinkedArchiveTreeSection>;
```
TomWoodward: the way you're using this its a highlighting specific thing based on the filter options available for highlighting, i'd keep it in the highlighting types

src/app/content/utils.ts
```diff
@@ -42,3 +48,24 @@ export const preloadedPageIdIs = (window: Window, id: string) => window.__PRELOA
...
   && window.__PRELOADED_STATE__.content.page
   && window.__PRELOADED_STATE__.content.page.id === id;
+
+export const mapBookSections = (book: Book) => {
+  return new Map(flattenArchiveTree(book.tree).filter((section) =>
+    (section.parent && archiveTreeSectionIsBook(section.parent))
+    || archiveTreeSectionIsChapter(section)).map((s) => [s.id, s]));
+};
+
+export const getCurrentChapter = (sections: BookSections, page: Page) => {
```
TomWoodward: since this is acting on a given page, not necessarily the _current_ page, i'd call it `getPageChapter`, but this isn't actually getting the page chapter, its getting the highlighting filter option for a page, so it should probably be named something else and moved into the highlighting module.

it would be nice to have a generic `getPageChapter` function though, i think we could use that elsewhere

TomWoodward (changes requested):

TomWoodward:
i think the spacing between chapters has gotten lost somehwere 
image

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,102 @@
...
+// tslint:disable-next-line: variable-name
+const Highlights = () => {
+  const sections = useSelector(bookSections);
+  const highlights = useSelector(summaryHighlights);
+  const isLoading = useSelector(summaryIsLoading);
+
+  if (sections.size > 0 && Object.keys(highlights).length > 0) {
+    return <Styled.Highlights isLoading={isLoading}>
+      {isLoading ? <Styled.LoaderWrapper><Loader/></Styled.LoaderWrapper> : null}
+      {Array.from(sections).map(([id, section]) => {
```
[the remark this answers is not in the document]
PiotrKozlowski: sections are type of `Map` so it need to be first converted to array for rendering

src/app/content/highlights/reducer.ts
PiotrKozlowski: Those are for cases which I can see when updating highlight which may be or may not be in summary highlights. I've moved logic outside reducer

src/app/content/highlights/actions.ts
```diff
@@ -1,13 +1,26 @@
...
-import { HighlightData } from './types';
+import { HighlightData, SummaryFilters, SummaryHighlights } from './types';
 
 export const focusHighlight = createStandardAction('Content/Highlight/focus')<string>();
 export const clearFocusedHighlight = createStandardAction('Content/Highlight/clear')();
-export const createHighlight = createStandardAction('Content/Highlight/create')<NewHighlight & {id: string}>();
-export const deleteHighlight = createStandardAction('Content/Highlight/delete')<string>();
-export const updateHighlight = createStandardAction('Content/Highlight/update')<UpdateHighlightRequest>();
+export const createHighlight = createStandardAction('Content/Highlight/create')<NewHighlight & {id: string}, {
+  chapterId: string,
```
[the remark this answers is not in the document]
PiotrKozlowski: If we want to make it without additional hook then it should be always provided. I'll add selector for current chapter / filterOptionId and then it should be easier to use

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
PiotrKozlowski: If we are going to use `dangerouslySetInnerHTML` then it's needed

src/app/content/utils.ts
PiotrKozlowski: I've renamed it to `getHighlightLocationForPage` and moved it to highlight utils

src/app/content/highlights/hooks/filtersChange.ts
```diff
@@ -0,0 +1,90 @@
...
+    return;
+  }
+
+  const pages = archiveTreeUtils.findTreePages(book.tree);
+
+  for (const h of highlights.data) {
+    const pageId = stripIdVersion(h.sourceId);
+
+    // SectionId can be the same as pageId for ex. for Preface
+    if (sections.has(pageId)) {
```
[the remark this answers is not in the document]
PiotrKozlowski: I've used `highlightLocations`, `locations`,  `locationIds`, `locationId`

src/app/content/selectors.ts
```diff
@@ -17,6 +24,16 @@ export const book = createSelector(
   (state) => state.book
 );
 
+export const bookSections = createSelector(
```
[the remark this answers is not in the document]
PiotrKozlowski: I've used `highlightLocations`

src/app/content/highlights/components/Card.tsx
```diff
@@ -78,7 +81,17 @@ const Card = (props: Props) => {
     return null;
   }
 
-  const onRemove = () => props.data && props.remove(props.data.id);
+  const chapter = getCurrentChapter(sections, page);
```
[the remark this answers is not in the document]
PiotrKozlowski: I've resigned from this and just used `getHighlightLocationForPage`

src/app/content/highlights/components/Highlights.tsx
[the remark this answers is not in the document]
TomWoodward: what does `Array.from(Map)` return? would `Map.keys`, `Map.values`, or `Map.entries` work? at least then you know what you're getting

src/app/content/highlights/components/Card.tsx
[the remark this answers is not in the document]
TomWoodward: on nomenclature, highlightLocation still doesn't describe what we're talking about. what it is is a highlight filter option id, where filter options can be either colors or book parts (chapter or page) `getHighlightFilterForPage` might make sense, but is potentially confusing because you don't know if its a color or location filer, which is why i proposed `highlightLocationFilter` so in this case `getHighlightLocationFilterForPage`, since you're going to be importing it from some `highlights/utils.ts` file, maybe just `getLocationFilterForPage` would be adequate

src/app/content/selectors.ts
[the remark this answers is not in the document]
TomWoodward: i'm not really happy with `highlightLocations`, see my other comment

TomWoodward (changes requested):
this is getting close, most of the comments are about nomenclature at this point. some minor feedback on data handling. i think the only thing that might lead to a significant change is the question about whether updating the filters when the user navigates to a new chapter will trigger a reload of the modal data, and if so maybe we should consider not updating the filters until they open the modal, otherwise we might be loading a bunch of data they don't care about

src/app/content/highlights/components/Highlights.tsx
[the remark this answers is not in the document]
TomWoodward: never mind i just looked it up and all those functions return a map iterator so they wouldn't help. i'm now wondering why we're using a map at all

src/app/content/highlights/utils/addCurrentPageToSummaryFilters.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getHighlightLocationForPage } from '../utils';
+
+const addCurrentPageToSummaryFilters = ({
+  dispatch, getState,
+}: MiddlewareAPI & AppServices) => {
+  const state = getState();
+  const page = pageSelector(state);
+  const locations = select.highlightLocations(state);
+  const filters = select.summaryFilters(state);
+  if (!locations.size || !page || typeof(window) === 'undefined') {
```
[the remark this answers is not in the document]
TomWoodward: after thinking more about the other comments about moving this logic to occur when the modal is opened, not on navigation, we probably want to keep this check as you have it written now. 

the idea is that when someone opens the dialog it should show them their current chapter, not necessarily every chapter they've navigated through since loading the page, the same applies to the chapter of the page that was loaded.

so forget the previous comment.

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -44,6 +65,10 @@ export const HighlightSection = styled.div`
   display: flex;
   align-items: center;
   font-weight: bold;
+
+  > span {
```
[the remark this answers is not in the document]
PiotrKozlowski: since we are using `dangerouslySetInnerHTML` we have additional spans with `.os-number`, `.os-divider` and `.os-text` classes. I'll list it like this:
```
 > .os-number,
 > .os-divider,
 > .os-text {
    overflow: hidden;
 }
```

src/app/content/highlights/components/Card.tsx
```diff
@@ -78,7 +81,17 @@ const Card = (props: Props) => {
     return null;
   }
 
-  const onRemove = () => props.data && props.remove(props.data.id);
+  const location = getHighlightLocationForPage(locations, page) || page;
```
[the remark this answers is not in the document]
PiotrKozlowski: I'll change it to `getHighlightLocationFilterForPage`

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -0,0 +1,131 @@
...
+      colors: filters.colors.filter(not(match(color))),
+    }));
+  };
+
+  return <ul className={className}>
+    {filters.locationIds.map((locationId) => locations.has(locationId) && <FiltersListChapter
+      key={locationId}
+      title={locations.get(locationId)!.title}
+      chapterId={locationId}
+      onRemove={onRemoveChapter}
```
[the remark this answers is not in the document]
PiotrKozlowski: In our project we've decided to use [jsx-no-bind](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md), so this is why I'm using it, but I can ofc change it

src/app/content/highlights/utils/getHighlightLocationsFromBook.ts
```diff
@@ -0,0 +1,14 @@
+import { ArchiveBook, Book } from '../../types';
+import {
+  archiveTreeSectionIsBook,
+  archiveTreeSectionIsChapter,
+  flattenArchiveTree,
+} from '../../utils/archiveTreeUtils';
+
+const getHighlightLocationsFromBook = (book: Book | ArchiveBook) => {
+  return new Map(flattenArchiveTree(book.tree).filter((section) =>
```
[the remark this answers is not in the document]
PiotrKozlowski: Previously it was just an array which would not be efficient for things like `locationFitlers.get(page.id)` so I've decided to make it an object, but then we wouldn't have access to `locationFilters.size` without `Object.keys().length`

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -0,0 +1,141 @@
...
+}
+
+export const addSummaryHighlight = (summaryHighlights: SummaryHighlights, data: DataAdd) => {
+  const newHighlights = {...summaryHighlights};
+  const { locationId, pageId, highlight } = data;
+  const chId = locationId || pageId;
+
+  if (newHighlights[chId]) {
+    if (newHighlights[chId][pageId]) {
+      newHighlights[chId][pageId].push(highlight);
```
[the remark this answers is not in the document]
PiotrKozlowski: I know that, but I had to not think when I was writing this... Changed.

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,128 @@
...
+  ) {
+    return <Styled.Highlights>
+      <HStyled.GeneralCenterText>
+        <FormattedMessage id='i18n:toolbar:highlights:popup:heading:no-highlights'>
+          {(msg: Element | string) => msg}
+        </FormattedMessage>
+        <span>
+          <FormattedMessage
+            id='i18n:toolbar:highlights:popup:heading:no-highlights-tip'
+            defaultMessage='Try selecting different chapter or color filters to see different results.'
```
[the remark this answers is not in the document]
PiotrKozlowski: Not sure how this library works, but when I was using [Fluent](https://projectfluent.org/) I had to specify message because when provided `id` was missing it would break app :D

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,128 @@
...
+            values={{strong: (str: string) => <strong>{str}</strong>}}
```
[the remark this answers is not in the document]
PiotrKozlowski: It is replacing `<strong>...</strong>` from text i `messages.json` to normal HTML Elements

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const pageIdIsSameAsSectionId = highlights[location.id][location.id];
+  return (
+    <React.Fragment>
+      {!pageIdIsSameAsSectionId && <Styled.HighlightsChapter
+        dangerouslySetInnerHTML={{ __html: location.title }}
+      />}
+      {Object.entries(highlights[location.id]).map(([pageId, pageHighlights]) => {
+        pageId = stripIdVersion(pageId);
+        const page = archiveTreeSectionIsChapter(location)
+          ? findArchiveTreeNode(location, pageId)!
```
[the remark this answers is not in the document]
PiotrKozlowski: Should I in this case write separate test case to handle this branch?

src/app/content/highlights/utils/addCurrentPageToSummaryFilters.ts
```diff
@@ -0,0 +1,30 @@
...
+    return;
+  }
+
+  const chapterToAdd = getHighlightLocationForPage(locations, page);
+  const idToAdd = chapterToAdd ? stripIdVersion(chapterToAdd.id) : null;
+
+  if (idToAdd && !filters.locationIds.includes(idToAdd)) {
+    dispatch(setSummaryFilters({
```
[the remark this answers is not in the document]
PiotrKozlowski: It is. Ok.

src/app/content/highlights/utils/getHighlightLocationForPage.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (!location) {
+    for (const section of locations.values()) {
+      if (archiveTreeSectionIsChapter(section) && findArchiveTreeNode(section, pageId)) {
+        location = section;
+        break;
+      }
+    }
+  }
+
+  return location;
```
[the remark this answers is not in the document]
PiotrKozlowski: I've modified `getHighlightLocationFilters` which is now returning only chapters and pages so we can met design criteria about display:
image

src/app/content/highlights/components/SummaryPopup/__snapshots__/ChapterFilter.spec.tsx.snap
PiotrKozlowski: My mistake, I've forgot to add `addCurrentPageToSummaryFilters` to this test case. Fixed.

src/app/content/highlights/components/Highlights.tsx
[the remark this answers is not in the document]
TomWoodward: there will be an error if the message isn't found in the configuration, but that just means we need to add the message to the configuration

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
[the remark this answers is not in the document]
TomWoodward: thank you, i had assumed you were only targeting the os-text, so if its important that you hit all of them its much nicer to have it explicitly written out

src/app/content/highlights/utils/getHighlightLocationsFromBook.ts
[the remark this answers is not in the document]
TomWoodward: what do we need the size for>

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -296,6 +296,9 @@ export const GeneralCenterText = styled(GeneralTextWrapper)`
   text-align: center;
 `;
 
+// tslint:disable-next-line:variable-name
+export const GeneralCenterTextWrapper = styled.span``;
```
TomWoodward: do you really need this styled component with no styles?

src/app/content/highlights/components/Highlights.tsx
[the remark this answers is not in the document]
TomWoodward: if something really _should_ never happen, its ok to throw an exception (for example by using the assertDefined utility), that makes testing easier because it doesn't create a branch that needs coverage. 

if it is reasonably possible for the null case to actually occur, then yea, you'd need to add a short circuit and add a test case for that branch.

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -36,69 +36,59 @@ interface FiltersListColorProps {
...
 
   return <ul className={className}>
-    {filters.locationIds.map((locationId) => locations.has(locationId) && <FiltersListChapter
+    {filters.locationIds.map((locationId) => locationFilters.has(locationId) && <FiltersListChapter
       key={locationId}
-      title={locations.get(locationId)!.title}
-      chapterId={locationId}
+      title={locationFilters.get(locationId)!.title}
+      locationId={locationId}
       onRemove={onRemoveChapter}
```
TomWoodward: if you make this `onRemove={onRemoveChapter(locationId)}` and change the handler to 

```
const onRemoveChapter = (locationId: string) => () => {
    dispatch(setSummaryFilters({
      ...filters,
      locationIds: filters.locationIds.filter(not(match(locationId))),
    }));
  };
```

then the `FiltersListChapter` component doesn't need any knowledge of the `locationId`

src/app/content/highlights/reducer.ts
```diff
@@ -38,63 +38,82 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
       const highlight = {
         ...action.payload,
         color: action.payload.color as string as HighlightColorEnum,
         sourceType: action.payload.sourceType as string as HighlightSourceTypeEnum,
       };
-      newState.highlights = [...state.highlights || [], highlight];
 
-      if (newState.summary.filters.colors.includes(highlight.color)) {
-        newState.summary.highlights = addSummaryHighlight(newState.summary.highlights, {
+      let newSummaryHighlights = {};
```
TomWoodward: you could leave this initialized to undefined and the below logic would be simpler

TomWoodward (changes requested):
changes are looking good, i left some minor comments and this branch needs conflicts resolved with master, after that i'll read it again and hopefully we'll be good to merge

src/app/content/highlights/components/HighlightStyles.tsx
PiotrKozlowski: I've used this because `htmlMessage` is accepting `ComponentType`, but I can also use `() => <span/>`

src/app/content/highlights/utils/getHighlightLocationsFromBook.ts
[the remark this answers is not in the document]
PiotrKozlowski: I'm using this in `Highlights.tsx` to determine if list of highlights should be displayed, in `addCurrentPageToSummaryFilters.ts` to return earlier if map is empty and `getHighlightLocationFilters.spec.ts` to check if empty map was returned for one of test cases

TomWoodward (approved):
this looks good, karina, devika, and kim should also review on the test environment before we merge. needs master merged in again (new conflicts)

also https://travis-ci.com/openstax/rex-web/builds/143394916 is showing some linter errors

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
TomWoodward: i'm not sure if i haven't been explaining myself well or if you're just ignoring me because you really really hate currying, but regardless its fine, i think everything is looking good now.

PiotrKozlowski: I'm confused, I think I've applied your suggestion with one change:
I did `onRemove={() => onRemoveChapter(locationId)}` instead of `onRemove={onRemoveChapter(locationId)}` because with the second option `onRemoveChapter` function will be called immediately after render

PiotrKozlowski:
After merging master thare are some large margins around titles.
It is related to @KarinaMendez2 pr - https://github.com/openstax/rex-web/pull/407/
If thats valid then ok, if not then let me know or just push change to this pr directly
image

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
TomWoodward: not if you also change the definition as i put in the comment so that the initial call returns a new function that takes no arguments, the second function is then invoked by the handler. this way of stashing data in the parameters of previously called functions is called currying, although this isn't a perfect example.

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { stripIdVersion } from '../../utils/idUtils';
+import { highlightLocationFilters, summaryFilters, summaryHighlights, summaryIsLoading } from '../selectors';
+import { SummaryHighlights } from '../types';
+import * as HStyled from './HighlightStyles';
+import * as Styled from './ShowMyHighlightsStyles';
+
+// tslint:disable-next-line: variable-name
+const NoHighlightsTip = htmlMessage(
+  'i18n:toolbar:highlights:popup:heading:no-highlights-tip',
+  () => <span/>
```
TomWoodward: i think this is never showing up because you're not forwarding props here

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
PiotrKozlowski: Ahh, I wasn't looking at `=> () =>`

src/app/content/highlights/components/Highlights.tsx
PiotrKozlowski: Not sure what you want me to do here

PiotrKozlowski: Ok, I've missed that it is no longer displaying this tip.
I've removed this and just added `FormattedHTMLMessage`

PR #400, opened 2019-12-18
Unclickable logo on mobile 

janKozie1:


janKozie1:
I'm not sure where is the 5px coming from. Before i had an error with Google Analytics but it seems that it changed despite no changes in code?

TomWoodward (approved):

TomWoodward:
the google analytics test sometimes has false negatives, i haven't tracked it down yet

PR #401, opened 2019-12-19
ft. My highlights popup - tab behavior 

KarinaMendez2:


src/app/content/components/Toolbar/HighlightButton.tsx
```diff
@@ -22,6 +22,12 @@ const MyHighlightsWrapper = styled(PlainButton)`
   align-items: center;
   justify-content: center;
   margin-right: 2rem;
+  height: auto;
+
+  :focus {
+    outline: auto -webkit-focus-ring-color;
```
TomWoodward: what is this for?

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -229,7 +230,12 @@ export const InfoIconWrapper = styled.span`
 `;
 
 // tslint:disable-next-line:variable-name
-export const CloseIcon = styled((props) => <Times {...props} aria-hidden='true' focusable='false' />)`
+export const CloseIconWrapper = styled(PlainButton)`
+  cursor: pointer;
```
TomWoodward: does plainbutton not have this already?

TomWoodward (changes requested):

src/app/content/components/Toolbar/HighlightButton.tsx
KarinaMendez2: if I dont put this, the outline is gray on focus, I assumed blue outline is expected.

src/app/content/highlights/components/HighlightStyles.tsx
KarinaMendez2: lint was complaining about it being empty so I just added that to put something

src/app/content/components/Toolbar/HighlightButton.tsx
TomWoodward: i would leave it as default, i see normal behavior is grey outline when focused and blue when active, but i'm not sure, we should just leave whatever the browser does by default since we have no custom states

src/app/content/highlights/components/HighlightStyles.tsx
TomWoodward: so why do you have to make a new styled component at all? you can just `export const CloseIconWrapper = PlainButton` and then you can still switch back to styled component if you decide you need something

TomWoodward (changes requested):

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -93,6 +93,11 @@ export const PrintOptWrapper = styled(PlainButton)`
   display: flex;
   justify-content: center;
   align-items: center;
+  height: auto;
+
+  :focus {
```
TomWoodward: didn't we talk about removing this?

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -159,6 +159,7 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
     </ButtonGroup>}
     {confirmingDelete && data && <Confirmation
       data-testid='confirm-delete'
+      data-analytics-label='delete-note'
```
TomWoodward: i don't think this is going to do anything here

src/app/content/highlights/components/ColorPicker.tsx
```diff
@@ -64,7 +64,7 @@ const ColorButton = styled(({className, size, style, ...props}: ColorButtonProps
 // tslint:disable-next-line:variable-name
 const ColorPicker = ({className, ...props}: Props) => {
 
-  return <div className={className}>
+  return <div className={className} data-analytics-region='highlight-color-picker'>
```
TomWoodward: this is nice, but i don't think anything in here is going to trigger any events will it?

TomWoodward (changes requested):

TomWoodward (approved):

PR #402, opened 2019-12-20
Fixed ctrl click not opening in a new tab

janKozie1:
I'm not sure if ctrl+click should be universal across all systems since the issue mentioned Windows. I've also noticed that it doesn't work on Linux so that's why i included it in the check.

### DIFF
--- a/src/app/content/components/ContentLink.tsx
+++ b/src/app/content/components/ContentLink.tsx
@@ -50,9 +50,11 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
   return <a
     ref={myForwardedRef}
     onClick={(e) => {
-      if (e.metaKey) {
+      const shouldUseCtrl = navigator && /windows|linux/i.test(navigator.userAgent);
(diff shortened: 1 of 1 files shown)


src/app/content/components/ContentLink.tsx
```diff
@@ -50,9 +50,11 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
   return <a
     ref={myForwardedRef}
     onClick={(e) => {
-      if (e.metaKey) {
+      const shouldUseCtrl = navigator && /windows|linux/i.test(navigator.userAgent);
```
TomWoodward: other routers just bail if any modifiers are pressed, i think we should do the same https://github.com/ReactTraining/react-router/blob/ea44618e68f6a112e48404b2ea0da3e207daf4f0/packages/react-router-dom/modules/Link.js#L18

indeed shift and alt also change the behavior and we would not want to mess with that

TomWoodward (changes requested):

TomWoodward (approved):

PR #404, opened 2020-01-03
Fire clearSearchHook conditionally also on openToc

PiotrKozlowski:
https://github.com/openstax/unified/issues/640

I've just hooked `clearSearchHook` to both `clearSearch` and `openToc`

### DIFF
--- a/src/app/content/components/__snapshots__/Content.spec.tsx.snap
+++ b/src/app/content/components/__snapshots__/Content.spec.tsx.snap
@@ -3221,7 +3221,7 @@ li[aria-label="Current Page"] .c44 {
                           dangerouslySetInnerHTML={
                             Object {
                               "__html": "
-  <div data-html=\\"copyright\\">© 1999-2019, Rice University. Except where otherwise noted, textbooks on this site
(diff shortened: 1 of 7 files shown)


src/app/content/components/__snapshots__/Content.spec.tsx.snap
```diff
@@ -3221,7 +3221,7 @@ li[aria-label="Current Page"] .c44 {
                           dangerouslySetInnerHTML={
                             Object {
                               "__html": "
-  <div data-html=\\"copyright\\">© 1999-2019, Rice University. Except where otherwise noted, textbooks on this site
+  <div data-html=\\"copyright\\">© 1999-2020, Rice University. Except where otherwise noted, textbooks on this site
```
TomWoodward: 😆

TomWoodward (approved):
we've got a pattern emerging here of denormalized logic between these hooks and the reducers (knowledge of which actions close the search results) that i don't love, but what you've done here is a clean extension of what was already there so i say lets keep it for now, but keep in mind that if it gets any more complicated we're gonna want to re-organize it a little.

PR #408, opened 2020-01-07
Initial Highlight editing in SummaryPopup

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/588

Initial editing menu for highlights in summary popup

TODO:
- [x] Check on mobile
- [x] Close context menu on click outside
~~- [ ] Add tests (After merging filters pr and updating this branch with master)~~ I'll do this in separate pr for easier review


src/app/content/highlights/components/Highlights.tsx
```diff
@@ -90,6 +98,62 @@ interface SectionHighlightsProps {
 
 // tslint:disable-next-line: variable-name
 export const SectionHighlights = ({ location, highlights }: SectionHighlightsProps) => {
+  const [highlightIdToEdit, setHighlightIdToEdit] = React.useState<string | null>(null);
```
TomWoodward: lets move all this logic down into a per-highlight component. the only reason i can think of to have the logic in the list is to do some special handling for editing or blocking the editing of multiple highlights at once, but suhani has said we don't need to do that

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -134,3 +206,145 @@ export const SectionHighlights = ({ location, highlights }: SectionHighlightsPro
...
+  { annotation, isEditable, onSave, onCancel }: HighlightAnnotationProps
+) => {
+  const [anno, setAnno] = React.useState(annotation);
+
+  const onKeyDown = (e: React.KeyboardEvent) => {
+    if (e.key === 'Escape') {
+      onCancel();
+      return;
+    } else if (e.key === 'Enter' && !e.shiftKey) {
+      onSave(anno);
```
TomWoodward: if this isn't in the spec (i don't think it is?) please don't do any special enter key handling

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -134,3 +206,145 @@ export const SectionHighlights = ({ location, highlights }: SectionHighlightsPro
...
+}
+
+// tslint:disable-next-line:variable-name
+const HighlightAnnotation = (
+  { annotation, isEditable, onSave, onCancel }: HighlightAnnotationProps
+) => {
+  const [anno, setAnno] = React.useState(annotation);
+
+  const onKeyDown = (e: React.KeyboardEvent) => {
+    if (e.key === 'Escape') {
```
TomWoodward: handling the esc key is nice but is actually a much larger problem we need to address with using it to close dropdowns, dialogs, i would not add this for now and wait until we address the problem as a whole

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -134,3 +206,145 @@ export const SectionHighlights = ({ location, highlights }: SectionHighlightsPro
...
+  </Styled.HighlightNote>;
+};
+
+interface HighlightDeleteWrapperProps {
+  onCancel: () => void;
+  onDelete: () => void;
+}
+
+// tslint:disable-next-line:variable-name
+const HighlightDeleteWrapper = ({
```
TomWoodward: i think you split these components into new files in a different PR, but you can go ahead and do it on this branch, it would be easier to read with them in different files

TomWoodward (changes requested):

src/app/content/highlights/components/Highlights.tsx
PiotrKozlowski: ok, sorry, I'm just used to working without design specs when review is ending at: please implement also this, this and this :D

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -126,6 +143,106 @@ export const HighlightContentWrapper = styled.div`
...
+  background-color: ${theme.color.neutral.formBackground};
+`;
+
+// tslint:disable-next-line:variable-name
+export const HighlightDeleteWrapper = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
```
TomWoodward: i usually specify all of top/bottom/left/right to 0 instead of with/height 100 to take up the whole container, do you know if there is a difference in behavior? i think this is fine i'm just curious.

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -126,6 +143,106 @@ export const HighlightContentWrapper = styled.div`
...
+
+// tslint:disable-next-line:variable-name
+export const HighlightToggleEditContent = styled.div`
+  z-index: 2;
+  border: 1px solid ${theme.color.neutral.formBorder};
+  background-color: ${theme.color.neutral.formBackground};
+`;
+
+// tslint:disable-next-line:variable-name
+export const HighlightDeleteWrapper = styled.div`
```
TomWoodward: since `HighlightDeleteWrapper` has its own file, i would put the styles for it in there

src/app/content/highlights/reducer.ts
```diff
@@ -74,26 +77,22 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.updateHighlight): {
       if (!state.highlights) { return state; }
 
-      const oldHiglightIndex = state.highlights.findIndex(
-        (highlight) => highlight.id === action.payload.id);
-      if (oldHiglightIndex < 0) { return state; }
+      const currentHighlight = getHighlightByIdFromHighlights(state.highlights, action.payload.id)
+        || getHighlightByIdFromSummaryHighlights(state.summary.highlights, action.payload.id);
```
TomWoodward: i'm not sure this logic is working correctly. i think you should process the current page highlights and the summary highlights separately, and each one should be in a function that correctly noops and returns the input state if no action is required. 

`updateSummaryHighlightsDependOnFilters` it has a noop if the chapter doesn't match filters, but doesn't have one for changing a color that did not match filters to another color that also doesn't match filters.


edit: the `updateHighlights` function does this already, it returns the input state if no action is necessary, so there is no reason to separately check to see if action is necessary on its behalf.

src/app/content/highlights/utils/getHighlightByIdFromHighlights.ts
```diff
@@ -0,0 +1,17 @@
...
+  let foundHighlight: Highlight | undefined;
+
+  const oldHiglightIndex = highlights.findIndex(
+    (highlight) => highlight.id === id);
+
+  if (oldHiglightIndex < 0) { return foundHighlight; }
+
+  foundHighlight = {...highlights[oldHiglightIndex]};
+
+  return foundHighlight;
```
TomWoodward: this function seems excessively complicated. before the index was being fetched so that it could then be used to write back in the new highlight object, in this case there is no reason to findIndex you could use the regular find. you're also copying the object here which is undocumented and unnecessary

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -183,3 +183,22 @@ export const addOneToTotalCounts = (
...
 };
+
+export const getHighlightByIdFromSummaryHighlights = (
+  summaryHighlights: SummaryHighlights, id: string
+): Highlight | undefined => {
+  let foundHighlight: Highlight | undefined;
+
+  for (const data of Object.values(summaryHighlights)) {
+    for (const highlights of Object.values(data)) {
+      const index = highlights.findIndex((highlight) => highlight.id === id);
```
TomWoodward: again, a regular find would work here

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -183,3 +183,22 @@ export const addOneToTotalCounts = (
...
+      if (index !== -1) {
+        foundHighlight = highlights[index];
+        break;
```
TomWoodward: just return the highlight, no need for all the breaks

TomWoodward (changes requested):

src/app/content/highlights/components/HighlightData.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }, {
+      locationFilterId,
+      pageId,
+    }));
+    cancelEdit();
+  };
+
+  const updateColor = (color: string) => {
+    dispatch(updateHighlight({
+      highlight: {color: color as HighlightUpdateColorEnum},
```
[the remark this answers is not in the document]
PiotrKozlowski: If I didn't pass color when I was updating annotation then color was removed but it worked properly in this case. However now both cases works properly (I can update annotation or color alone). Something has changed?

src/app/content/highlights/components/HighlightDeleteWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onDelete,
+  onCancel,
+}: HighlightDeleteWrapperProps) => <Styled.HighlightDeleteWrapper>
+  <FormattedMessage id='i18n:highlighting:confirmation:delete-both'>
+    {(msg: string) => <span>{msg}</span>}
+  </FormattedMessage>
+  <Styled.HighlightEditButtons>
+    <FormattedMessage id='i18n:highlighting:button:delete'>
+      {(msg: Element | string) => <Button
+        data-testid='delete'
```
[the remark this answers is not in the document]
PiotrKozlowski: I'm (pr with tests) but I can target those buttons differently. I've just saw in other components that this is how it is used

src/app/content/highlights/components/HighlightToggleEdit.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: HighlightToggleEditProps) => <Styled.HighlightToggleEdit>
+  <Dropdown
+    toggle={<MenuToggle/>}
+    className='highlight-toggle-edit'
+  >
+    <Styled.HighlightToggleEditContent>
+      <ColorPicker
+        color={color}
+        onChange={onColorChange}
+        onRemove={() => onDelete()}
```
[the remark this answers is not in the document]
PiotrKozlowski: Lets wait for the answer
https://openstax.slack.com/archives/C8V1XET1R/p1579095008268700

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -126,6 +143,106 @@ export const HighlightContentWrapper = styled.div`
...
+  ${DropdownFocusWrapper} {
+    svg {
+      color: ${theme.color.neutral.darkest};
+    }
+  }
+
+  ${ColorPicker} {
+    label {
+      margin: 0.6rem;
+      width: 1.6rem;
```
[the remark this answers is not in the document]
PiotrKozlowski: Those are smaller in specs: https://projects.invisionapp.com/d/main?origin=v7#/console/18130098/376272522/preview?scrollOffset=468

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
PiotrKozlowski: I don't know, I'm just used to using this way

src/app/content/highlights/components/HighlightData.tsx
```diff
@@ -0,0 +1,90 @@
...
+import * as Styled from './ShowMyHighlightsStyles';
+
+interface HighlightDataProps {
+  highlight: Highlight;
+  locationFilterId: string;
+  pageId: string;
+}
+
+// tslint:disable-next-line:variable-name
+const HighlightData = ({ highlight, locationFilterId, pageId }: HighlightDataProps) => {
```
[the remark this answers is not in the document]
PiotrKozlowski: Should I move also `Highlights.tsx` and `ShowMyHighlights.tsx` since those should be there also, but they will make huge diff for you to check

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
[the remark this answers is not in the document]
PiotrKozlowski: oh, I see, there is props.size to set

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -126,6 +143,106 @@ export const HighlightContentWrapper = styled.div`
...
+    padding: 0;
+
+    li {
+      display: flex;
+
+      a {
+        width: 100%;
+
+        svg {
+          width: 15px;
```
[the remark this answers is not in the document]
PiotrKozlowski: I'm just used to this, but I've spitted them

src/app/content/highlights/reducer.ts
PiotrKozlowski: I'll add test for this case in #424 pr when this pr become approved

src/app/content/highlights/components/HighlightData.tsx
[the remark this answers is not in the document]
TomWoodward: i have not moved them because i'm afraid of conflicts with people, they really should be moved, but i'm not in a hurry

src/app/content/highlights/components/HighlightDeleteWrapper.tsx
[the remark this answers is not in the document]
TomWoodward: i like using the test ids, i just want to make sure they're actually used

src/app/content/highlights/components/ShowMyHighlights.tsx
```diff
@@ -77,6 +77,7 @@ class ShowMyHighlights extends Component<ShowMyHighlightsProps, { showGoToTop: b
       <Styled.ShowMyHighlightsBody
         ref={this.myHighlightsBodyRef}
         data-testid='show-myhighlights-body'
+        data-analytics-region='highlighting-my-highlights-popup'
```
TomWoodward: lets try to keep this short
```suggestion
        data-analytics-region='MH popup'
```

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -52,6 +52,7 @@ export const ShowMyHighlightsBody = styled(PopupBody)`
 export const HighlightWrapper = styled.div`
   margin: 1.6rem ${popupPadding}rem;
   border: solid 0.1rem ${theme.color.neutral.darkest};
+  overflow: unset;
```
TomWoodward: can you set this to `hidden` or `visible` or whatever you want here

src/app/content/highlights/components/SummaryPopup/HighlightDeleteWrapper.tsx
```diff
@@ -0,0 +1,65 @@
...
+        data-analytics-label='delete'
+        size='medium'
+        variant='primary'
+        onClick={onDelete}
+      >{msg}</Button>}
+    </FormattedMessage>
+    <FormattedMessage id='i18n:highlighting:button:cancel'>
+      {(msg: Element | string) => <Button
+        size='medium'
+        data-analytics-label='cancel'
```
TomWoodward: i like these short button labels, but to keep them in context we probably want 
`data-analytics-region="MH delete"` on the `StyledHighlightDeleteWrapper`

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -0,0 +1,128 @@
...
+      highlight: {annotation},
+      id: highlight.id,
+    }, {
+      locationFilterId,
+      pageId,
+    }));
+    setIsEditing(false);
+  };
+
+  const updateColor = (color: string) => {
```
TomWoodward: this color should be `HighlightUpdateColorEnum` or whatever it actually is like `HighlightColorEnum`

src/app/content/highlights/utils/updateHighlights.ts
```diff
@@ -0,0 +1,19 @@
+import { UpdateHighlightRequest } from '@openstax/highlighter/dist/api';
+import { HighlightData } from '../types';
+
+const updateHighlights = (highlights: HighlightData[], data: UpdateHighlightRequest) => {
```
TomWoodward: is this util used?

TomWoodward (changes requested):
looking really great in general, left some minor comments, also needs conflicts resolved

src/app/content/highlights/components/SummaryPopup/HighlightDeleteWrapper.tsx
TomWoodward: i realized that this is the second region i'm asking you to add for the same section of code, really both regions are good to have, i'm sorry the repeated comments are probably annoying though

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
PiotrKozlowski: I've updated ContextMenu and ColorPicker components to use `HighlightColorEnum` so it is more clear conversion

TomWoodward:
i think this is good, but it needs master merged in and conflicts resolved

PiotrKozlowski:
I think I've resolved conflicts, but I can't test it since highlight api is not working

TomWoodward:
some of the snaps need updating @PiotrKozlowski

PiotrKozlowski:
@TomWoodward I've updated this branch and `add-tests-to-588`

TomWoodward:
@PiotrKozlowski i think this branch's code is fine but we can't merge. is your intent to merge #424 into this branch first?

TomWoodward (approved):

TomWoodward:
waiting for qa/product review before merging

TomWoodward (approved):

TomWoodward:
@PiotrKozlowski please look at the fixes i added, i'm submitting to qa now

PiotrKozlowski:
@TomWoodward I've updated snapshots and according to https://github.com/openstax/unified/issues/588#issuecomment-586110403 we shouldn't add anything more in this pr, right?

PR #412, opened 2020-01-08
Fixed menu offset

janKozie1:


TomWoodward (changes requested):
needs react snaps updated with `yarn test:unit -u`

TomWoodward (approved):

janKozie1:
I'm not sure where do those 5px changes come from. They seem to be quite common? Is there a standard way to fix those? Just update the snapshots?

TomWoodward:
it happens when the math processing is not totally finished, the snapshots wait for things to be loaded and javascript to finish running but it is very difficult to tell if chrome has actually rendered the stuff yet. if you watch on your screen you can see the math going through several phases of rendering and each one has a different size.

if i get failures in the screenshots like this, or some of the other tests like lighthouse or GA, especially if the two travis tests (branch and PR) don't agree on the results, i just re-run them.

PR #416, opened 2020-01-09
Disable chapters without highlights

PiotrKozlowski:
- Add `receiveHighlightsTotalCounts` and `setHighlightsTotalCountsPerLocation` actions
- Add `mergeHighlightsTotalCounts` util function
- Use those functiosn in `locationChange` hook to fetch data and propagate state
- Handle `disable` prop on `<Checkbox>` component
- When clicking `All` in `<ChapterFilter>` only chapters with highlights will be selected
- Update state on highlight create / delete

TODO:
- [x] Tests


src/app/components/Checkbox.tsx
```diff
@@ -40,11 +41,22 @@ const CustomCheckbox = styled(
 `;
 
 // tslint:disable-next-line:variable-name
-const Checkbox = ({children, className, ...props}: React.PropsWithChildren<Props>) => <label className={className}>
+const Label = styled.label`
+  ${(props) => props.disabled ? `
```
TomWoodward: its a little confusing to me that these styles aren't with the other checkbox styles below

src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -16,16 +17,28 @@ const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
...
-    sourceType: GetHighlightsSourceTypeEnum.OpenstaxPage,
-  });
+  const [highlights, totalCounts] = await Promise.all([
+    await highlightClient.getHighlights({
+      perPage: 100,
+      scopeId: book.id,
+      sourceIds: [page.id],
+      sourceType: GetHighlightsSourceTypeEnum.OpenstaxPage,
+    }),
+    await highlightClient.getHighlightsSummary({
```
TomWoodward: we don't have to do this on every navigation, you could move it to an initializer that only happens once on app init. or you could put a check here that only queries it if its not already there

src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -16,16 +17,28 @@ const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
...
+  ]);
 
   if (highlights.data) {
     dispatch(receiveHighlights(highlights.data));
   }
+
+  if (totalCounts.countsPerSource) {
+    dispatch(receiveHighlightsTotalCounts(totalCounts.countsPerSource));
+    const mergedTotalCounts = mergeHighlightsTotalCounts(book, totalCounts.countsPerSource);
+    dispatch(setHighlightsTotalCountsPerLocation(mergedTotalCounts));
```
TomWoodward: do you think it makes sense to store both of these? or should we just make a selector for the processed version? (selectors are memoized, so it'll still only be calculated once)

its good to maintain access to both the per page and the per location filter versions of this data though, because for pagination we'll need the first

src/app/content/highlights/reducer.ts
```diff
@@ -52,13 +54,31 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      const { locationFilterId, pageId } = action.meta;
+
+      const totalCountsPerPage = {...state.totalCountsPerPage};
+      if (totalCountsPerPage[pageId]) {
+        totalCountsPerPage[pageId] += 1;
+      } else {
+        totalCountsPerPage[pageId] = 1;
+      }
+
+      const totalCountsPerLocation = {...state.summary.totalCountsPerLocation};
```
TomWoodward: you wouldn't have to do this second part if you used a selector, its probably marginally more resource efficient to do it this way thought, simple addition rather than doing the whole object processing again. 

i could go either way, if you leave it like this please move some of this logic to a util function

src/app/content/highlights/hooks/locationChange.ts
PiotrKozlowski: I've added check for equality between those two objects which can be helpful in cases when user is working on multiple tabs

TomWoodward: there are way more problems with supporting highlighting in multiple tabs, i would rather not have the extra network request on every navigation

TomWoodward:
@PiotrKozlowski i would like to focus on making very small and easy to review PRs. i have a tentative goal of 200 lines changed not counting snapshots or the yarn.lock file. this one is 374 and it does feel like it could be more manageable to review as separate small changes.

we don't have to change anything on this pr, but please in the future consider breaking down your changes into smaller pull requests

for instance using this change as a case study, adding the:
- disabled styles to the checkbox
- call for the summary from the api with action receiveHighlightsTotalCounts and reducer
- util function for mergeHighlightsTotalCounts with unit tests

could all have been made as separate PRs at the same time without blocking each other

then a follow up PR for adding setHighlightsTotalCountsPerLocation with reducer, would have built on top of patterns from the previous pr, being less likely to have major review feedback. tying together the rest of the functionality would have also been minor.

src/app/content/highlights/hooks/locationChange.ts
PiotrKozlowski: Ok, now I'm doing:
```
if (Object.keys(totalCountsInState).length > 0) { return; }
```
however there is something strange going on, because `totalCountsInState` is always an empty object even if state is propagated with data. When I console.log state in `locationChange` then I can see that `totalCountsPerLocation` are propagated, but `..perPage` are not.
Any ideas?

PiotrKozlowski:
@TomWoodward My approach was: Make logic ~150 lines > request review > push tests 

I'll try splitting functionalities into smaller pr's next time

TomWoodward:
yes but filters is merged now, so you should change the base to master and resolve conflicts

src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -26,6 +28,19 @@ const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
   if (highlights.data) {
     dispatch(receiveHighlights(highlights.data));
   }
+
+  if (Object.keys(totalCountsInState).length > 0) { return; }
```
TomWoodward: what if the data was previously fetched and the user just has no highlights? can we initialize it to null and check for that instead?

TomWoodward (changes requested):

src/app/content/highlights/hooks/locationChange.ts
PiotrKozlowski: I've update this, but it is still not solving issue that this hook does not have last version of state which causes summary request at each location change even that totalCountsPerPage are in state

src/app/content/highlights/selectors.ts
```diff
@@ -32,6 +32,17 @@ export const highlightLocationFilters = createSelector(
   : new Map() as HighlightLocationFilters
 );
 
+export const totalCountsPerPage = createSelector(
+  localState,
+  (state) => state.summary.totalCountsPerPage
+);
+
+export const totalCountsPerLocation = createSelector(
+  localState,
```
TomWoodward: because you're depending on `localState` here the selector will be re-evaluated anytime the highlight state changes at all, if you put `totalCountsPerPage` here, it'll only re-evaluate when that value changes

TomWoodward (changes requested):

TomWoodward (approved):

PR #417, opened 2020-01-09
Printing styles for highlights

janKozie1:


src/app/content/components/Toolbar/PrintButton.tsx
```diff
@@ -9,29 +9,32 @@ import { Book } from '../../types';
...
-                data-testid='print'
-              >
-                <Styled.PrintIcon />
-                <Styled.PrintOptions>{msg}</Styled.PrintOptions>
-              </Styled.PrintOptWrapper>
-            )}
-          </FormattedMessage>
-        )}
-      </FormattedMessage>
+      <div className={this.props.className}>
```
TomWoodward: is there a reason to have this extra div rather than putting the class on `Styled.PrintOptWrapper`?

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -32,6 +33,33 @@ export const stickyNoteMeasures = {
   width: 29.8, /* to allow text to fit in one line with tooltip */
 };
 
+const printStyles = {
```
TomWoodward: please just put these inline so you don't have to keep going up/down to tell what anything does

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -32,6 +33,33 @@ export const stickyNoteMeasures = {
...
+    position: relative;
+    overflow: visible;
+    height: max-content;
+    max-width: auto;
+    width: 100%;
+    margin: 0;
+    border-radius: 0;
+  `,
+  page: css`
+    margin: 15mm 10mm;
```
TomWoodward: is there not default margin on the page?

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -32,6 +33,33 @@ export const stickyNoteMeasures = {
   width: 29.8, /* to allow text to fit in one line with tooltip */
 };
 
+const printStyles = {
+  modal: css`
+    position: relative;
```
TomWoodward: if you moved these styles to the popup body, could you ignore the modal?

src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -134,6 +178,16 @@ export const BlueStickyNote = styled(StickyNote)`
     left: -${stickyNoteMeasures.bulletSize / 2}rem;
     background: ${stickyNoteMeasures.blue};
   }
+
+  @media print {
```
TomWoodward: is it really necessary to have these styles on the sticky note? if you did have to do something special for showing the empty state in print, i would say just hide the whole thing instead, there is no reason to be printing that

src/app/content/highlights/components/ShowMyHighlightsStyles.tsx
```diff
@@ -3,8 +3,29 @@ import { AngleUp } from 'styled-icons/fa-solid/AngleUp';
...
 import { h4Style } from '../../../components/Typography/headings';
 import theme from '../../../theme';
+import PrintButton from '../../components/Toolbar/PrintButton';
+import { disablePrint } from '../../components/utils/disablePrint';
 import { highlightStyles } from '../constants';
 import { PopupBody, popupBodyPadding, popupPadding } from './HighlightStyles';
+import { mobilePadding } from './SummaryPopup/constants';
+
+const printStyles = {
+  highlightOuterWrapper: css`
```
TomWoodward: same thing, please put these inline

src/app/content/highlights/components/ShowMyHighlights.tsx
```diff
@@ -57,7 +57,10 @@ class ShowMyHighlights extends Component<{}, { showGoToTop: boolean }> {
         ref={this.myHighlightsBodyRef}
         data-testid='show-myhighlights-body'
       >
-        <Filters />
+        <Styled.HighlightsToolbar>
+          <Filters />
+          <Styled.HiglightsPrintButton />
```
TomWoodward: i think you might want to add the `HiglightsPrintButton` within the filters component, because filters has a two row design and the print button is supposed to be on the first row. currently the way you have this the button is taking away space where the second row of filters is supposed to go

TomWoodward (changes requested):

src/app/content/highlights/components/HighlightStyles.tsx
janKozie1: I don't think so. It needs the position changed so it the whole content of highlights is being printed, not only one page.

TomWoodward:
@janKozie1 here is some more feedback on the styles

[asdf.pdf](https://github.com/openstax/rex-web/files/4067397/asdf.pdf)

janKozie1:
@TomWoodward I've removed the backgrounds

TomWoodward:
@janKozie1  i think we should remove the whole "My Highlights and Notes" header, and we should add the filter breadcrumbs that tell you what filters have been applied, just the labels though without the "x" button.

janKozie1:
@TomWoodward I've added the breadcrumbs

src/app/content/components/Toolbar/PrintButton.tsx
```diff
@@ -9,8 +9,9 @@ import { Book } from '../../types';
 import * as Styled from './styled';
 
 interface Props {
-  currentPath: string;
-  book: Book | undefined;
+  currentPath?: string;
+  book?: Book | undefined;
+  className?: string;
```
TomWoodward: why make all this stuff optional?

src/app/content/highlights/components/SummaryPopup/Filters.tsx
```diff
@@ -63,7 +73,7 @@ export default styled(Filters)`
   display: flex;
   flex-flow: row wrap;
   align-items: center;
-  padding: 2rem ${popupPadding}rem 0 ${popupPadding}rem;
+  padding: 2.1rem ${popupPadding}rem 0 ${popupPadding}rem;
```
TomWoodward: whats going on with this padding change?

TomWoodward (changes requested):

TomWoodward:
there are some borders showing up on the print page, in pdf i can see one along the left side, and in print preview i see one along the top of the page @janKozie1

janKozie1:
@TomWoodward it should be gone now.

TomWoodward (dismissed):

TomWoodward (approved):

PR #423, opened 2020-01-12
Routing by UUID

janKozie1:
Still want to add tests.


src/app/content/routes.ts
```diff
@@ -5,7 +5,15 @@ import { SelectedResult } from './search/types';
 import { Params } from './types';
 
 const CONTENT_PATH = '/books/:book/pages/:page';
-const VERSIONED_CONTENT_PATH = '/books/:book([^@/]+)@:version/pages/:page';
+const UUID_CONTENT_PATH =
```
TomWoodward: this might read better with the uuid separated out like:

```
const MATCH_UUID = '[\\da-z]{8}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{12}';
const UUID_CONTENT_PATH =  `/books/:uuid(${MATCH_UUID})@:version/pages/:page`;
```

src/app/content/types.ts
```diff
@@ -4,7 +4,8 @@ import { content } from './routes';
 import { State as SearchState } from './search/types';
 
 export interface Params {
-  book: string;
+  book?: string;
+  uuid?: string;
   page: string;
   version?: string;
```
TomWoodward: you can make this a union so it is a little more descriptive. like:
```
interface SlugParams {
  book: string;
  page: string;
}
interface VersionedSlugParams extends SlugParams {
  version: string;
}
interface UuidParams {
  uuid: string;
  version: string;
}
export type Params = SlugParams | VersionedSlugParams | UuidParams
```

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -67,16 +67,26 @@ const resolveBookReference = async(
   match: Match<typeof content>
 ): Promise<[string, string, string | undefined]> => {
   const state = getState();
-  const bookSlug = match.params.book;
   const currentBook = select.book(state);
 
+  if (!match.params.uuid && !match.params.book) {
+    throw new Error('Nor a book nor a uuid found');
```
TomWoodward: if you do the union thing, you shouldn't have to handle this case because ts will figure out narrowing the union type. probably.

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -67,16 +67,26 @@ const resolveBookReference = async(
...
+  }
+
+  const bookSlug = match.params.book
+    ? match.params.book
+    : currentBook && currentBook.id === match.params.uuid
+      ? currentBook.id
```
TomWoodward: this should be `slug` right?

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -67,16 +67,26 @@ const resolveBookReference = async(
...
+      : match.params.uuid && await osWebLoader.getBookSlugFromId(match.params.uuid);
```
TomWoodward: pretty sure if you do the union this you won't need `match.params.uuid && ` because ts will know that either params.book or params.uuid will be defined

TomWoodward (changes requested):

src/app/content/routes.ts
```diff
@@ -29,11 +24,10 @@ export const content: Route<Params, State> = {
     modules: ['Content'],
   }),
   getUrl: (params: Params): string => {
-    const { uuid, version } = params;
-    if (uuid && version) {
+    if ('uuid' in params) {
```
TomWoodward: this is snazzy

TomWoodward (approved):

src/test/mocks/archiveLoader.ts
```diff
@@ -7,6 +7,10 @@ export const book = JSON.parse(
   fs.readFileSync(path.resolve(__dirname, '../fixtures/contents/testbook1-shortid'), 'utf8')
 ) as ArchiveBook;
 
+export const validUUIDBook = JSON.parse(
```
TomWoodward: for what did you add this new book?

TomWoodward (approved):
i want to do some more testing tomorrow, but this is in general looking good

src/test/mocks/archiveLoader.ts
janKozie1: I've looked at the other books and their ID's don't seem to be the same format as the one used for routing. I thought that by adding a book with a format that the route expects, i can test it.

src/app/content/hooks/locationChange/resolveContent.spec.ts
```diff
@@ -105,4 +117,25 @@ describe('locationChange', () => {
...
+
+    match.params = versionedSlugParams;
+    helpers.archiveLoader.mockBook({
+      ...book,
+      version: undefined as any as string,
+    });
+    helpers.archiveLoader.mockPage({
+      ...book,
+      version: undefined as any as string,
+    }, page, 'test-page-1');
```
TomWoodward: i don't think these calls to `mockBook` and `mockPage` are currently doing anything, because you're creating them based on the `book` but you're not querying that book. if you use that book variable as your base, but do 

```
helpers.archiveLoader.mockBook({
      ...book,
      id: 'longidin-vali-dfor-mat1-111111111111'
      version: undefined as any as string,
    });
```

then you will not need to add the whole json file

TomWoodward (changes requested):

TomWoodward (approved):

PR #431, opened 2020-01-14
refactor disabling location filters

TomWoodward:
we never actually cared about the counts per location filter, only which ones had any content, so this is a little simpler and we won't have to update it when the CountsPerSource value becomes a map of colors


src/app/content/highlights/utils/getHighlightLocationFiltersWithContent.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  return Object.entries(totalCounts).reduce((result, [pageId]) => {
+    const location = getHighlightLocationFilterForPage(locationFilters, pageId);
+
+    if (location && !result.has(location.id)) {
+      result.set(location.id, location);
+    }
+
+    return result;
+  }, new Map() as HighlightLocationFilters);
```
PiotrKozlowski: This could be `Set<LocationId>` since we are using it only for disabling Checkbox

PiotrKozlowski (approved):

src/app/content/highlights/utils/getHighlightLocationFiltersWithContent.ts
TomWoodward: i did it this way to get the constant time access of Map.has, but it looks like Set.has does the same, so i'll change it

PR #436, opened 2020-01-17
[WIP] Initial work for scrolling to target

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/589

This is how it works now:
When you click <kb>Goto Highlight</kbd> or click on figure link, url will change which will trigger the change in `scrollTarget` selector (`app/navigation/selectors.ts`)

In the `PageComponent` now I'm calling: (in `componentDidUpdate`)
```
const currentScrollTarget = this.getCurrentScrollTarget(prevProps);
this.scrollTargetManager(prevProps.scrollTarget, currentScrollTarget);
```

I've moved some previous logic to `getCurrentScrollTarget` and added more data to create `ScrollTargetProp` based on `searchResults`, `scrollTarget` selector and other information.

Searching for content will also trigger url change (`/content/search/hooks.ts`) but it will not be used to determine scroll target because to retrieve `searchResults`, more complex logic from `searchHighlightManager` is required.

### Old comment, read on your own risk

The logic for searching, scrolling, updating highlights, etc. is split into multiple files and locations, so it is confusing to work with, but I have an idea of how we could simplify it.

We can use `options` from `makeCallHistoryMethodAction` (`navigation/actions.ts`) to pass everything we need.

In `receiveSearchHook` (`search/hooks`) I've added:
```
const options = {
    hash: selectedResult.result.source.elementId,
    search: `search=${query}`,
  }
```
which will add `?search=asdf#elementId` to the url.

Then `PageComponent` (`content/components/`) will trigger `scrollToTarget` from `scrollToTargetManager` with our hash. This will work for every hash in url.

`syncSearch` hook will work as before with one change: if there is search query detected in the URL and there is no `savedSearch` it will request a new search which will highlight elements on the screen. Users screen, thanks to `hash` from URL will be already scrolled to the proper place.

About extending this with highlight navigation:
We could pass 
```
  hash: highlightAnchorId,
  search: `highlight=${highlightId}`,
```
which would create `/url?highlight=highlightId#highlightAnchorId`

This would automatically scroll the user to the highlight location and then we have to only select this highlight.

If this is okay then we could add a hook to `callHistoryMethod` which will select found highlights.
Probably we would move to this hook also some logic from search hooks and some of it can be removed.

Those this sound good?

Btw. Do we need to keep `savedSearch` in the location state?


TomWoodward:
@PiotrKozlowski i haven't looked at your changes yet, but based on the description you provided:

i like the way you are updating the url with the hash and query param, and how it works for search and highlights, however, we can't simply rely on the scroll to the element id to solve all our scrolling needs. sometimes the parent elements are quite large, and if the search result or highlight is not near the top of it, the scroll will not end up in the right place.

so i think we need to come up with a way to decide what to scroll to based on the url, and organize the logic for how to find different types of scroll targets.

I'm sorry that it has taken me so long to get you feedback on this issue.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -148,6 +148,7 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
       }
       return null;
     },
+    highlighter,
```
TomWoodward: i don't see you using this anywhere, why did you expose it?

src/app/content/components/Page/scrollTargetManager.ts
```diff
@@ -36,17 +36,28 @@ const scrollToTop = () => {
...
+const scrollTargetManager = (container: HTMLElement) => (
+  previous: ScrollTargetProp,
+  current: ScrollTargetProp
+) => {
   if (previous.page !== current.page) {
-    scrollToTargetOrTop(container, current.hash);
-  } else if (previous.hash !== current.hash) {
-    scrollToTarget(container, current.hash);
+    scrollToTargetOrTop(container, current.target);
+  } else if (current.target && (!previous.target || previous.target.value !== current.target.value)) {
```
TomWoodward: could you do something like `} else if (!isEqual(previous.target, current.target)) {`?

src/app/content/components/Page/highlightManager.ts
PiotrKozlowski: thanks, those were leftovers from the previous approach

PiotrKozlowski:
@TomWoodward This seems reasonable, I'll try in separate pr

PiotrKozlowski:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/rex-web/702

PR #443, opened 2020-01-21
Displaying books with no osweb data

janKozie1:


src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -170,20 +179,27 @@ const resolveExternalBookReference = async(
...
   reference: ReturnType<typeof getContentPageReferences>[0]
 ) => {
   const targetBook: Book = archiveTreeContainsNode(book.tree, reference.pageUid)
     ? book
     : await resolveExternalBookReference(services, book, page, reference.pageUid);
 
-  return {
-    match: reference.match,
-    params: {
+  const params = hasOSWebData(targetBook)
```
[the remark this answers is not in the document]
TomWoodward: looks like `getBookPageUrlAndParams` already exists but you can't use it because you don't have the right page type. we could split out `getUrlParamForBook` in the same way we have `getUrlParamForPageId`

src/gateways/createOSWebLoader.ts
```diff
@@ -29,30 +29,28 @@ export default (prefix: string) => {
...
-    }
-    return data.items[0];
-  };
+  const firstRecord = (data: OSWebResponse) => data.items[0];
 
   const cache = new Map();
 
-  const cacheRecord = (record: OSWebBook) => {
+  const cacheRecord = (record?: OSWebBook) => {
+    if (!record) {
```
TomWoodward: we still want to cache the empty response, so that we don't try to query it again

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -65,25 +71,29 @@ const resolveBook = async(
...
   const bookSlug = 'book' in match.params
     ? match.params.book
-    : currentBook && currentBook.id === match.params.uuid
+    : currentBook && hasOSWebData(currentBook) && currentBook.id === match.params.uuid
       ? currentBook.slug
       : await osWebLoader.getBookSlugFromId(match.params.uuid);
 
   if (match.state && match.state.bookUid && match.state.bookVersion) {
-    return [bookSlug, match.state.bookUid, match.state.bookVersion];
+      return [match.state.bookUid, bookSlug,  match.state.bookVersion];
```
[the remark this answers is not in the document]
janKozie1: I've changed the return type to `Promise<[string, string?, string?]>` and only bookUid should always exist, i think? Should i change it to `Promise<[string | undefined, string, string | undefined>`?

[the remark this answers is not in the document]
TomWoodward: yea, that would be better

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -65,25 +71,29 @@ const resolveBook = async(
...
   }
 
   const bookUid  = 'uuid' in match.params
     ? match.params.uuid
-    : currentBook && currentBook.slug === bookSlug
+    : currentBook && hasOSWebData(currentBook) && currentBook.slug === bookSlug
       ? currentBook.id
-      : await osWebLoader.getBookIdFromSlug(bookSlug);
+      : bookSlug && await osWebLoader.getBookIdFromSlug(bookSlug);
```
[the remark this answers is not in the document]
janKozie1: I'm not sure about the exception. If the route was matched by slug and the slug doesn't have an associated uuid in osweb, then doesn't it mean something went wrong?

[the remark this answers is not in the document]
TomWoodward: good point, its possible for uuid to come out as undefined if a slug route is used with a slug that doesn't exist, please update the error message to that effect.

TomWoodward (changes requested):

script/prerender/contentPages.tsx
```diff
@@ -27,22 +27,22 @@ import { assetDirectoryExists, readAssetFile, writeAssetFile } from './fileUtils
 
 export async function prepareContentPage(
   bookLoader: ReturnType<AppServices['archiveLoader']['book']>,
-  bookSlug: string,
+  book: BookWithOSWebData,
   pageId: string,
   pageSlug: string
 ) {
-  const book = await bookLoader.load();
+  const loadedBook = await bookLoader.load();
```
TomWoodward: there is no reason to call load again since we have the book passed in now

script/prerender/contentPages.tsx
```diff
@@ -27,22 +27,22 @@ import { assetDirectoryExists, readAssetFile, writeAssetFile } from './fileUtils
...
     params: {
-      book: bookSlug,
+      book: book.slug,
       page: pageSlug,
     },
     route: content,
     state: {
-      bookUid: book.id,
-      bookVersion: book.version,
+      bookUid: loadedBook.id,
```
TomWoodward: you can use `book` here

src/app/content/components/Attribution.tsx
```diff
@@ -125,27 +126,27 @@ class Attribution extends Component<Props> {
...
     const {book} = this.props;
 
-    return <AttributionDetails
+    return hasOSWebData(book) ? <AttributionDetails
       ref={this.container}
       data-testid='attribution-details'
       data-analytics-region='attribution'
     >
-      <FormattedMessage id='i18n:attribution:toggle'>
+      <FormattedMessage id='i18n:attribution:to)ggle'>
```
TomWoodward: typo?

src/app/content/components/BookBanner.tsx
```diff
@@ -47,7 +48,7 @@ const LeftArrow = styled(ChevronLeft)`
 
 interface PropTypes {
   pageNode?: ArchiveTreeSection;
-  book?: Book;
+  book?: BookWithOSWebData | Book;
```
TomWoodward:
```suggestion
  book?: Book;
```

src/app/content/hooks/locationChange/resolveContent.spec.ts
```diff
@@ -160,4 +162,12 @@ describe('locationChange', () => {
     expect(helpers.archiveLoader.mock.loadBook).toHaveBeenCalledWith('longidin-vali-dfor-mat1-111111111111', '1.0');
   });
 
+  it('throws if there is no uuid', async() => {
+    try {
+      helpers.osWebLoader.getBookIdFromSlug.mockImplementation(() => Promise.resolve(undefined) as any) ;
+      await hook(helpers, match);
+    } catch (err) {
+      expect(err.message).toEqual(`No uuid provided or ${testBookSlug} doesn't have one`);
```
TomWoodward: this needs updating right?

src/app/content/hooks/locationChange/resolveContent.spec.ts
```diff
@@ -160,4 +162,12 @@ describe('locationChange', () => {
     expect(helpers.archiveLoader.mock.loadBook).toHaveBeenCalledWith('longidin-vali-dfor-mat1-111111111111', '1.0');
   });
 
+  it('throws if there is no uuid', async() => {
+    try {
+      helpers.osWebLoader.getBookIdFromSlug.mockImplementation(() => Promise.resolve(undefined) as any) ;
```
TomWoodward: whitespace on the `;`

src/app/content/hooks/locationChange/resolveContent.spec.ts
```diff
@@ -147,6 +148,7 @@ describe('locationChange', () => {
   });
 
   it('uses uuid if present', async() => {
+    helpers.osWebLoader.getBookSlugFromId.mockImplementation(() => Promise.resolve(undefined) as any)  ;
```
TomWoodward: whitespace on the `;`

TomWoodward (changes requested):
i pushed some changes directly in addition to these comments, please look them over and let me know if anything seems weird or you have any questions

src/app/content/components/BookBanner.tsx
```diff
@@ -139,7 +143,7 @@ export const BarWrapper = styled.div<{colorSchema: Book['theme'] | undefined , u
   position: ${ifMiniNav('sticky', 'relative' /* stay above mini nav */)};
   z-index: ${ifMiniNav(theme.zIndex.navbar - 2, theme.zIndex.navbar - 1)};
   overflow: hidden;
-  ${(props: {colorSchema: Book['theme'] | undefined }) => props.colorSchema && css`
+  ${(props: {colorSchema: BookWithOSWebData['theme'] | undefined }) => props.colorSchema && css`
```
PiotrKozlowski: This is a good moment to refactor this to `colorSchema?: BookWithOSWebData['theme']`

src/app/content/components/BookBanner.tsx
```diff
@@ -192,18 +196,23 @@ export class BookBanner extends Component<PropTypes, {scrollTransition: boolean}
...
-    const {pageNode, book} = this.props;
+    const { pageNode, book } = this.props;
+    const defaultTheme = 'blue' as BookWithOSWebData['theme'];
 
     if (!book || !pageNode) {
       return <BarWrapper colorSchema={undefined} up={false} />;
     }
 
-    const bookUrl = bookDetailsUrl(book);
+    const bookUrl = hasOSWebData(book) ? bookDetailsUrl(book) : '/errors/404';
```
PiotrKozlowski: `'errors/404'` probably should be some kind of constant, but I'm not familiar with this code base so maybe this is not a case here

src/app/content/components/BookBanner.tsx
janKozie1: I'm not exactly sure if its possible for `colorSchema`  to be undefined now except if it was set explicitly to it which i don't think is meant to happen.

PiotrKozlowski: If it can't then you can remove `| undefined`

src/app/content/utils/canonicalUrl.ts
```diff
@@ -22,7 +23,7 @@ export async function getCanonicalUrlParams(
 
     const canonicalBook = await getBook(id, version);
     const treeSection = findArchiveTreeNode(canonicalBook.tree, pageShortId);
-    if (treeSection) {
+    if (treeSection && hasOSWebData(canonicalBook)) {
```
TomWoodward: it would be fine for the unified book loader to only return `BookWithOSWebData`, and for that function to internally throw otherwise.

TomWoodward (changes requested):

janKozie1:
@TomWoodward updated

TomWoodward (approved):

TomWoodward:
@janKozie1 i made some changes and fixed tests but there is still some coverage needed that i don't have time to get to today. please take a look at my changes to make sure they look good to you, and if you want to work on coverage maybe we can merge this tomorrow.

janKozie1:
@TomWoodward i've updated the tests

TomWoodward:
@janKozie1 im merging this because it does most of the necessary work, is not getting pretty big, and is in a nice state right now with tests and everything is working.

however, we have another problem, all books that i've tested that don't exist in osweb are also missing slugs from the archive data for the individual pages. for the sake of testing arbitrary content it seems that we'll need to find a way to identify the pages other than the slugs

i guess we'll need to add more routing to deal with identifying pages by id instead of slug

PR #447, opened 2020-01-23
Fix disabling colors

PiotrKozlowski:
API is returning all colors with the summary so we can't use `Object.keys()` to determine if those colors have highlights.

I've also fixed `select all` in `ColorFilters` to select only colors with highlights.


TomWoodward:
where are you seeing that it returns all colors? it looks like the empty ones are omitted to me

image

TomWoodward (approved):

PR #449, opened 2020-01-23
Solution toggle aria-label missing in initial load state

PiotrKozlowski:
Add `aria-label="show solution"` in initial load state
Transoform content before rendering


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -29,9 +31,14 @@ export default class PageComponent extends Component<PagePropTypes> {
...
 
-  public getCleanContent = () => {
+  public getTrasnformedContent = () => {
     const {book, page, services} = this.props;
-    return getCleanContent(book, page, services.archiveLoader, contentLinks.reduceReferences(this.props.contentLinks));
+    const cleanContent = parser.parseFromString(
+      getCleanContent(book, page, services.archiveLoader, contentLinks.reduceReferences(this.props.contentLinks)),
+      'text/html'
+    );
+    transformContent(cleanContent.body, this.props.intl);
```
TomWoodward: the logic inside `transformContent` is depending on the document global and is failing in prerendering, we need to pass in the document that `parseFromString` returns

TomWoodward (changes requested):

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -1,10 +1,11 @@
-import { HTMLButtonElement, HTMLElement } from '@openstax/types/lib.dom';
+import { Document, HTMLButtonElement, HTMLElement } from '@openstax/types/lib.dom';
 import { IntlShape } from 'react-intl';
 import { assertDefined, assertNotNull } from '../../../utils';
 
 // from https://github.com/openstax/webview/blob/f95b1d0696a70f0b61d83a85c173102e248354cd
 // .../src/scripts/modules/media/body/body.coffee#L123
-export const transformContent = (rootEl: HTMLElement, intl: IntlShape) => {
+export const transformContent = (document: Document, intl: IntlShape) => {
+  const rootEl = document.body;
```
TomWoodward: can we pass element separately? i feel like this is a big assumption that the document is only going to contain things we want to transform

TomWoodward (changes requested):

PiotrKozlowski:
@TomWoodward I think now this is correct, but there were problems with deploying so I'm not totally sure if the output from Travis is valid.

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -1,10 +1,11 @@
-import { HTMLButtonElement, HTMLElement } from '@openstax/types/lib.dom';
+import { Document, HTMLButtonElement, HTMLElement } from '@openstax/types/lib.dom';
 import { IntlShape } from 'react-intl';
 import { assertDefined, assertNotNull } from '../../../utils';
 
 // from https://github.com/openstax/webview/blob/f95b1d0696a70f0b61d83a85c173102e248354cd
 // .../src/scripts/modules/media/body/body.coffee#L123
-export const transformContent = (rootEl: HTMLElement, intl: IntlShape) => {
+// We are passing Document because it is required to prerender.
+export const transformContent = (_document: Document, rootEl: HTMLElement, intl: IntlShape) => {
```
TomWoodward: you're passing document but you're not using it for anything, you need to pass the reference down so it is used in places like this https://github.com/openstax/rex-web/pull/449/files#diff-3684e43af97114ebe9e858e13d39add5R56

TomWoodward (changes requested):

src/app/content/components/Page/contentDOMTransformations.ts
PiotrKozlowski: Omg, now this makes sense. I could sware that `document` was not used in this file (I had to mistype it when I was searching) and I thought that prerendering script is somehow getting access to this...

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -44,7 +45,7 @@ function addScopeToTables(rootEl: HTMLElement) {
 }
 
 // Wrap title and content elements in header and section elements, respectively
-function wrapElements(rootEl: HTMLElement) {
+function wrapElements(document: Document, rootEl: HTMLElement) {
```
TomWoodward: now that you're passing this in the `assertDefined` is no longer necessary

TomWoodward (changes requested):

TomWoodward (approved):

PR #451, opened 2020-01-23
Focus indicator fix on chrome

janKozie1:
I think either this or disable the dotted outline on buttons for firefox.


TomWoodward:
we want to use whatever the browser default active state is @janKozie1, the buy button was correct already, but MH and Print buttons were not getting an active state at all, presumably we are suppressing them, the issue is to make the MH and Print buttons use the default browser active state

TomWoodward (approved):

PR #461, opened 2020-01-28
Add scrollOffset when scrolling to an element

PiotrKozlowski:
I've added additional offset when scrolling to an element.

Now, this looks like this:
image

Instead of this:
image

### DIFF
--- a/src/app/content/components/Content.spec.tsx
+++ b/src/app/content/components/Content.spec.tsx
@@ -1,64 +1,67 @@
-import React from 'react';
-import { Provider } from 'react-redux';
-import renderer from 'react-test-renderer';
-import createTestServices from '../../../test/createTestServices';
-import createTestStore from '../../../test/createTestStore';
-import {
-  book,
-  shortPage
-} from '../../../test/mocks/archiveLoader';
-import { mockCmsBook } from '../../../test/mocks/osWebLoader';
-import ScrollLock from '../../components/ScrollLock';
-import ScrollOffset from '../../components/ScrollOffset';
-import * as Services from '../../context/Services';
-import MessageProvider from '../../MessageProvider';
-import { locationChange } from '../../navigation/actions';
-import { Store } from '../../types';
-import { assertWindow } from '../../utils';
-import { openToc, receiveBook, receivePage } from '../actions';
-import { content } from '../routes';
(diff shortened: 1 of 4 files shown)


src/app/content/components/constants.ts
```diff
@@ -5,6 +5,8 @@ export const bookBannerDesktopMiniHeight = 7;
 export const bookBannerMobileBigHeight = 10.4;
 export const bookBannerMobileMiniHeight = 6;
 
+export const scrollOffset = 4;
```
janKozie1: Might be better to set it to `3`. It will prevent it from showing a bit of the text above the figure

`.os-figure {
      margin: 3rem 0;
    }`

src/app/content/components/Content.spec.tsx
```diff
@@ -232,15 +235,15 @@ describe('content', () => {
...
+        .findByProps({ "aria-label": "Click to close the Table of Contents" })
         .props.onClick();
     });
 
     expect(component.root.findByType(TableOfContents).props.isOpen).toBe(false);
 
     renderer.act(() => {
       component.root
-        .findByProps({ 'aria-label': 'Click to open the Table of Contents' })
+        .findByProps({ "aria-label": "Click to open the Table of Contents" })
```
janKozie1: A lot of changes from `'` to `"` in this file, not sure if desired seeing that `'` is more commonly used

janKozie1 (changes requested):

TomWoodward (approved):

src/app/content/components/Content.spec.tsx
TomWoodward: yea i think the snapshot updater did that, i would prefer not to have all the random changes, but i don't care a lot

TomWoodward:
looks like `src/app/content/components/Content.browserspec.tsx` needs to be updated @PiotrKozlowski

PiotrKozlowski:
@TomWoodward updated

janKozie1 (approved):

TomWoodward:
@PiotrKozlowski the `src/app/content/components/Content.browserspec.tsx` tests are still failing

PiotrKozlowski:
Ok, now it works... cc @TomWoodward

PR #463, opened 2020-01-29
Add CTA Popup

PiotrKozlowski:
TODO:
- [x] Fix responsivity (Separate pr for this)
- [x] Track when we display the popup (Separate pr for this)
- [x] Add animation
- [x] Add tests


src/app/content/components/CallToActionPopup/assets/bottom-layer.svg
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TomWoodward: i like how you organized these files

src/app/content/components/CallToActionPopup/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const loginLink = useSelector(loginLinkSelector);
+  const signupLink = useSelector(signupLinkSelector);
+
+  if (!showPopup) { return null; }
+
+  const closePopup = () => {
+    dispatch(closeCallToActionPopup());
+  };
+
+  return <CTAWrapper data-analytics-region='CTA popup'>
```
TomWoodward:
```suggestion
  return <CTAWrapper data-analytics-region='signup CTA'>
```

src/app/content/components/CallToActionPopup/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+              {(msg: Element | string) => <a
+                data-testid='log-in'
+                data-analytics-label='log-in'
+                href={loginLink}
+              >{msg}</a>}
+            </FormattedMessage>
+          </CTATextLink>
+        </CTAButtons>
+      </CTAText>
+      <CTAGraphic src={ctaGraphic} alt='' />
```
TomWoodward: i don't think you need the alt if you put like, "role=presentation" or something like that, i'm not sure the exact value. but the empty alt seems not so good

src/app/content/components/CallToActionPopup/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <CTAMiddleLayer src={middleLayer} alt='' />
+      <CTABottomLayer src={bottomLayer} alt='' />
+    </CTABackground>
+  </CTAWrapper>;
+};
+
+export default CallToActionPopup;
+
+// tslint:disable-next-line: variable-name
+const CtaPara = htmlMessage('i18n:cta:text', (props) => <CTAParagraph {...props} />);
```
TomWoodward: can you not just do this ?
```suggestion
const CtaPara = htmlMessage('i18n:cta:text', CTAParagraph);
```

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+
+// tslint:disable-next-line: variable-name
+export const CTAWrapper = styled.div`
+  position: fixed;
+  left: 0;
+  bottom: 0;
+  z-index: 99;
+  width: 100%;
+
+  @media screen and (max-width: 1200px) {
```
TomWoodward: there is a helper for mobile styles that you should use instead of this. 1200px is the mobile breakpoint but it is not good to use px in media queries

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+import { textRegularStyle } from '../../../components/Typography';
+import { textStyle } from '../../../components/Typography/base';
+import theme from '../../../theme';
+
+// tslint:disable-next-line: variable-name
+export const CTAWrapper = styled.div`
+  position: fixed;
+  left: 0;
+  bottom: 0;
+  z-index: 99;
```
TomWoodward: please put this in the z-index section of the theme

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  @media screen and (max-width: 1200px) {
+    display: none;
+  }
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTAContent = styled.div`
+  position: absolute;
+  z-index: 3;
```
TomWoodward: this you can leave inline because it is part of the fixed subtree, not the global z-index stack

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+  bottom: 0;
+  width: 100%;
+  display flex;
+  justify-content: center;
+  align-items: center;
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTAText = styled.div``;
```
TomWoodward: pretty sure this'll throw some lint errors or something, if this doesn't need to be a styled component you might want to just use an inline div

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTAParagraph = styled.p`
+  ${textStyle}
+  font-size: 1.8rem;
+  line-height: 3rem;
+  margin: 1rem 0;
+  color: ${theme.color.text.white};
+  max-width: 570px;
```
TomWoodward: we normally use `rem`s for sizes and spacing, so that they work with different default text sizes and zooming
```suggestion
  max-width: 57rem;
```

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTAButtons = styled.div`
+  display: flex;
+  align-items: center;
+  ${Button} {
+    width: 170px;
```
TomWoodward:
```suggestion
    width: 17rem;
```

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+export const CTAText = styled.div``;
+
+// tslint:disable-next-line: variable-name
+export const CTAHeading = styled.h2`
+  ${textStyle}
+  font-size: 3.6rem;
+  line-height: 4rem;
+  color: ${theme.color.text.white};
+  margin: 0;
+  max-width: 570px;
```
TomWoodward:
```suggestion
  max-width: 57rem;
```

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+  margin-left: 1.5rem;
+  a {
+    color: ${theme.color.text.white};
+    text-decoration: underline;
+    margin-left: 0.5rem;
+  }
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTACloseButton = styled(Times)`
```
TomWoodward: should probably make this actually a button

src/app/content/components/CallToActionPopup/styles.ts
```diff
@@ -0,0 +1,124 @@
...
+  cursor: pointer;
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTAGraphic = styled.img`
+  margin-left: 3rem;
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTATopLayer = styled.img`
```
TomWoodward: for all of these you could do
```suggestion
export const CTATopLayer = styled((props) => <img {...props} src={thesrc} />)`
```
that way you're defining what it looks like all in one place, and the index will just arrange them

src/app/content/hooks/locationChange/index.ts
```diff
@@ -1,14 +1,29 @@
...
 import { syncSearch } from '../../search/hooks';
+import { showCTAPopup } from '../../selectors';
 import resolveContent from './resolveContent';
 
 const hookBody: RouteHookBody<typeof content> = (services) => async(action) => {
   await resolveContent(services, action.match);
   const search = syncSearch(services)(action);
   const highlights = loadHighlights(services)();
 
+  const state = services.getState();
```
TomWoodward: this logic is ok, but can we make a separate hook to manage it?

TomWoodward (changes requested):

src/app/content/components/CallToActionPopup/index.tsx
PiotrKozlowski: Linter asked me to add empty alt for presentation purposes, but I see that role presentation also works and is more descriptive

src/app/content/components/CallToActionPopup/styles.ts
PiotrKozlowski: I've changed this to `1` but it is needed if I want to keep components in current order because if I would like this to be on top then I would have to move `<CTABackground>` at the beginning

src/app/content/components/CallToActionPopup/index.tsx
PiotrKozlowski: Actually liner is shouting about this :D 
```
Prefer alt="" over a presentational role. First rule of aria is to not use aria if it can be achieved via native HTML  jsx-a11y/alt-text
```

src/app/content/components/CallToActionPopup/styles.ts
PiotrKozlowski: I had to change `styles.ts` to `styles.tsx`, don't be scared about diff :D

src/app/content/hooks/locationChange/index.ts
PiotrKozlowski: Will this require to wait for `await resolveContent(services, action.match);` in both of those hooks?

src/app/content/components/CallToActionPopup/index.tsx
TomWoodward: i mean if the linter says it it must be true

src/app/content/hooks/locationChange/index.ts
TomWoodward: no, the logic you're adding doesn't care if the book is loaded or not, so it shouldn't matter. you will have to add something to make sure the user query is finished, i think we have something for that somewhere

TomWoodward: looks like in order to accomplish this you might want to actually hook into `receiveLoggedOut`

PiotrKozlowski: The specs say:
> banner appears 5s after page load with some kind of animation

This is why I've dispatched this after receiveContent

Would you like me still to hook this to `receiveLoggedOut` - I assume this hook should be placed in `/content/hooks/` rather than `/auth/hooks/`?

TomWoodward: Yea when they say page load they mean after the initial page load not after everything is loaded. 

Yes you should hook into receiveloggedout and put that in content, because it is controlling content stuff. You'll need to so something else with your timer too, because it should be 5 seconds from page load not 5 seconds from the user query finishing

src/app/content/hooks/signupCTAHooks.ts
```diff
@@ -0,0 +1,31 @@
...
+import { showCTAPopup } from '../selectors';
+
+export const receiveLoggedOutHookBody: ActionHookBody<typeof receiveLoggedOut> = (services) => async() => {
+  const state = services.getState();
+  const showCTA = showCTAPopup(state);
+
+  if (showCTA === null && !assertWindow().matchMedia(theme.breakpoints.mobileQuery).matches) {
+    setTimeout(() => {
+      services.dispatch(openCallToActionPopup());
+    }, 5000);
```
TomWoodward: this is waiting 5s after the auth request finishes, not 5s after page load. you could make it better by saving a timestamp at the top level of this file and doing some subtraction here. beware of passing negative numbers into settimeout

src/app/content/hooks/signupCTAHooks.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+};
+
+export const locationChangeHookBody: RouteHookBody<typeof content> = (services) => async() => {
```
TomWoodward: you don't actually care what route this is, you could use an action hook on locationChange

src/app/theme.ts
```diff
@@ -112,6 +112,7 @@ export default {
     'highlightSummaryPopup',
     'errorPopup',
     'focusedHiddenLink',
+    'ctaPopup',
```
TomWoodward: lets move this between `highlightSummaryPopup` and `errorPopup`

TomWoodward (changes requested):

TomWoodward (approved):

src/app/content/components/CallToActionPopup/styles.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+// tslint:disable-next-line: variable-name
+export const CTAGraphic = styled((props) => <img src={ctaGraphic} alt='' {...props} />)`
+  margin-left: 3rem;
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTATopLayer = styled((props) => <img src={topLayer} alt='' {...props} />)`
+  position: absolute;
+  right: -5px;
```
TomWoodward: rem

src/app/content/components/CallToActionPopup/styles.tsx
```diff
@@ -0,0 +1,151 @@
...
+  left: 0;
+  bottom: 0;
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTABottomLayer = styled((props) => <img src={bottomLayer} alt='' {...props} />)`
+  width: 100%;
+  position: absolute;
+  left: 0;
+  bottom: 20px;
```
TomWoodward: rem

src/app/content/components/CallToActionPopup/styles.tsx
```diff
@@ -0,0 +1,151 @@
...
+`;
+
+// tslint:disable-next-line: variable-name
+export const CTABackground = styled.div`
+  position: relative;
+  height: 420px;
```
TomWoodward: rem

TomWoodward (changes requested):

TomWoodward (approved):

PR #464, opened 2020-01-29
add analytics hook

TomWoodward:
the `withServices` HoC was causing a ref warning. i had been thinking about making a hook to tie together the analytics things anyway, so this fixes the warning and hopefully simplifies adding new analytics things


KarinaMendez2 (approved):
Awesome :+1:

PR #465, opened 2020-01-29
Work on displaying the search error 

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/499


src/app/notifications/utils.ts
```diff
@@ -16,3 +18,32 @@ export const shouldLoadAppMessage = (message: Message) => {
...
+    return {
+      ...state,
+      queuelessNotifications: state.queuelessNotifications.filter(({type}) => type !== notification.type),
+    };
+  }
+  return {
+    ...state,
+    notificationQueue: state.notificationQueue.filter(({type}) => type !== notification.type),
+  };
+}
```
PiotrKozlowski: newline

src/app/notifications/guards.ts
```diff
@@ -1,5 +1,12 @@
...
-import { AnyNotification, AppMessageNotification } from './types';
+
+import { AnyNotification, AppMessageNotification, QueuelessNotifications } from './types';
 
 export const isAppMessage = (message: AnyNotification): message is AppMessageNotification =>
   message.type === appMessageType;
+
+export const isQueuelessNotification =
+  (notification: AnyNotification): notification is QueuelessNotifications =>
+    notification.type === getType(actions.searchFailure)
```
PiotrKozlowski: Not sure if the name `isSearchFailureNotification` would be better.
Probably in the future, there will be more notifications which should be displayed outside of the queue, but for everything with `queueless` was confusing for me

+ newline

janKozie1: I'm not sure. I was expecting this kind of notification to not be one-time thing only. I will change it if needed.

janKozie1:
There is currently a bug: seems like adding a `positing:fixed` element to a scrolled page triggers the scroll event.

PiotrKozlowski:
> There is currently a bug: seems like adding a `positing:fixed` element to a scrolled page triggers the scroll event.

You can try `position: sticky; top: 0;`

src/app/notifications/components/SearchFailure.tsx
```diff
@@ -0,0 +1,173 @@
...
+          </CloseButton>
+        </BannerBody>
+      </BannerBodyWrapper>
+    );
+  }
+}
+
+export default connect(
+  () => ({
+  }),
```
PiotrKozlowski: You should be able to pass `undefined` here

src/app/notifications/components/SearchFailure.tsx
```diff
@@ -0,0 +1,173 @@
+import React from 'react';
```
PiotrKozlowski: This probably is not must-have, but I'm more into using a structure like `components/SearchFailure/index.tsx` and then make a separate file with style for this component like `../styled.ts(x)`
It would probably nice to make this a functional component as this is recommended by the react team

src/app/notifications/components/SearchFailure.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+// tslint:disable-next-line:variable-name
+const CloseIcon = styled((props) => <Times {...props} aria-hidden='true' focusable='false' />)`
+  width: 1.8rem;
+  height: 1.8rem;
+  cursor: pointer;
+`;
+
+// tslint:disable-next-line:variable-name
+const CloseButton = styled.button`
```
PiotrKozlowski: This could be probably `styled(PlainButton)` since it has similar styles.

src/app/notifications/reducer.spec.ts
```diff
@@ -6,13 +6,21 @@ describe('notifications reducer', () => {
...
-    expect(newState).toBe(state);
+  it('doesn\'t duplicate standard notifications', () => {
+    const stateWithNotifications = {
+      notificationQueue: [actions.updateAvailable()],
+      queuelessNotifications: [actions.searchFailure()],
+    };
+
+    const newState = flow(
+      (state) => reducer(state, actions.updateAvailable()),
+      (state) => reducer(state,  actions.searchFailure())
```
PiotrKozlowski: double space

PiotrKozlowski (changes requested):

src/app/notifications/components/SearchFailure.tsx
janKozie1: It was a functional component before but i had trouble testing the clean up function in useEffect, I will try again

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,98 @@
...
+  width: 1.8rem;
+  height: 1.8rem;
+  cursor: pointer;
+`;
+
+// tslint:disable-next-line:variable-name
+export const CloseButton = styled(PlainButton)`
+  background: transparent;
+  border: 0;
+  padding: 0;
```
PiotrKozlowski: I think PlainButton is implementing some of those attrs

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,98 @@
...
+  color: ${closeIconClor};
+
+  &:hover {
+    color: ${hoveredCloseIconColor};
+  }
+`;
```
PiotrKozlowski: newline

src/app/notifications/components/SearchFailure.tsx
PiotrKozlowski: I'll accept without this change, so what works best for you

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+    window.removeEventListener('click', startFadeOut);
+    window.removeEventListener('scroll', startFadeOut);
+    if (autoCloseTimeout.current) {
+      clearTimeout(autoCloseTimeout.current);
+    }
+  };
+
+  React.useEffect(() => {
+    autoCloseTimeout.current = setTimeout(startFadeOut, clearErrorAfter);
```
TomWoodward: you could make a custom effect `useTimeout`

```
const useTimeout = (delay, cb, deps) => React.useEffect(() => {
  const timeout = setTimeout(cb, delay);
  return () => clearTimeout(timeout);
}, deps)
```

src/app/notifications/components/Notifications.tsx
```diff
@@ -38,11 +47,13 @@ const renderNotification = (
 
 export class Notifications extends Component<Props> {
   public render() {
-    const { notification, className } = this.props;
-    return notification ? renderNotification(notification, className) : null;
+    const { notifications: {firstInQueue, queuelessNotifications}, className } = this.props;
+    const notificationsToRender = [...queuelessNotifications, firstInQueue || null ];
```
TomWoodward: you've changed a lot of logic to make this `queuelessNotifications` thing happen this way. in general, i think the logic you wrote looks fine, but this is our first notification that works outside the normal queue, and we have no idea what the requirements will be for future notifications that may or may not work within the queue or may or may not have a _separate_ queue that works with the search error. i just don't think we have enough information yet to make the kinds of design changes that you've made here. 

i think a simpler solution that we can use until we have more use cases is to not handle SearchFailure here in the notifications component. i would _try_ to set something up so that `searchHighlightManager` controls when to display it and hook that up within PageComponent in order to avoid having a notification action for searchFailure at all. if that doesn't work for whatever reason we could have `searchHighlightManager` dispatch something (maybe notification action, but in that case make sure that it is filtered out of `notificationForDisplay`) and have `SearchFailure` internally decide whether to return null or not (keeping it out of any other notification queuing logic) and just unconditionally throw a `<SearchFailure />` wherever it has to be to make the design work

TomWoodward (changes requested):
a lot of the changes here look good, but it addresses the wrong criteria for when to display the error. the issue described in the task was to handle when the target search result could not be found to scroll to, not when the api encounters an error. if you follow the instructions malar has in the card you should be able to see what shes talking about. the relevant code is in `searchHighlightManager.ts` in the `selectResult` function

janKozie1:
I thought about something like this, although the lack of state in PageComponent scares me. Is it because it already rerenders often? I wasn't able to replicate the error that Malar described but based on the description you gave me i assumed that it happens when 
`const firstSelectedHighlight = selectedHighlights && selectedHighlights[0];`
is undefined.

I thought about rendering the error into a ref but i don't think it's achievable by something else than `ReactDOM.render`. 

If introducing state to the PageComponent is a bad idea, please let me know and i will try to implement it some other way. @TomWoodward

PiotrKozlowski:
We could actually think about optimizing `PageComponent`, because it is now rerendering 2 times for every click on a highlight. cc @janKozie1 @TomWoodward

TomWoodward:
@PiotrKozlowski what are the two triggers? focusing the highlight will do it once, what is the other trigger?

in general i'm not _super_ worried about triggering re-renders because each update handler should be internally nooping if none of its arguments have changed, but if there are any updates that are happening in sequence for no reason, and could be combined into one update, that would be better.

PiotrKozlowski:
@TomWoodward I didn't found why it is happening, but when I've added:
```
console.log('compare', JSON.stringify(prevProps) === JSON.stringify(this.props))
```
inside `componentDidUpdate` it is logging `false` for the first update and `true` for the second which makes me wonder why it is happening because if `props` didn't change and there is no `state` it should not re-render for the second time.

TomWoodward:
@PiotrKozlowski the props having the same value doesn't mean they aren't new references, which will trigger a re-render. props changing without changing value probably means there is stuff we can easily optimize though, when i navigate to a new page (which i've already visited, so there is no loading) i get 6 re-renders, which is a lot of extra. putting some effort into having only one re-render when content is changed might also help some issues we've had with math processing.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -65,11 +68,29 @@ export default class PageComponent extends Component<PagePropTypes> {
...
     const highlgihtsAddedOrRemoved = this.highlightManager.update();
-    this.searchHighlightManager.update(prevProps.searchHighlights, this.props.searchHighlights, {
-      forceRedraw: highlgihtsAddedOrRemoved,
-    });
+
+    this.searchHighlightManager.update(
+      prevProps.searchHighlights,
+      this.props.searchHighlights,
+      { forceRedraw: highlgihtsAddedOrRemoved },
+      this.onHighlightSelect
```
TomWoodward: you can add this to the options object with forceRedraw

src/app/notifications/components/Notifications.spec.tsx
```diff
@@ -4,7 +4,7 @@ import renderer from 'react-test-renderer';
 import createTestStore from '../../../test/createTestStore';
 import MessageProvider from '../../MessageProvider';
 import { Store } from '../../types';
-import { acceptCookies, receiveMessages, updateAvailable } from '../actions';
+import { acceptCookies, receiveMessages, updateAvailable  } from '../actions';
```
TomWoodward: extra space

src/app/content/search/reducer.ts
```diff
@@ -30,6 +30,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.openMobileToolbar): {
       return {...state, mobileToolbarOpen: true};
     }
+
```
TomWoodward: extra space

src/app/notifications/components/SearchFailure/SearchFailure.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { renderToDom } from '../../../../test/reactutils';
+import MessageProvider from '../../../MessageProvider';
+import { Store } from '../../../types';
+import { assertWindow } from '../../../utils';
+import { dismissNotification } from '../../actions';
+import { clearErrorAfter } from './styles';
+
+jest.mock('react', () => {
+    const react = (jest as any).requireActual('react');
+    return { ...react, useEffect: react.useLayoutEffect };
```
TomWoodward: too many indents

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+} from './styles';
+
+interface Props {
+  dismiss: () => void;
+}
+
+const window = assertWindow();
+
+// tslint:disable-next-line:variable-name
+const SearchFailure = memo(({ dismiss }: Props) => {
```
TomWoodward: why are you using memo here?

src/app/notifications/reducer.ts
```diff
@@ -27,8 +27,8 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.updateAvailable):
     case getType(actions.acceptCookies):
       return state.find(({type}) => type === action.type)
-        ? state
-        : [...state, action];
+      ? state
```
TomWoodward: this whitespace was right before

src/app/notifications/components/__snapshots__/Notifications.spec.tsx.snap
```diff
@@ -444,7 +444,168 @@ exports[`Notifications matches snapshot 2`] = `
 `;
 
 exports[`Notifications matches snapshot 3`] = `
-.c6 {
+.c3 {
```
TomWoodward: i don't see why these changes are here, do you need to update it?

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,94 @@
...
+  margin: 0;
+  height: 0;
+  z-index: ${theme.zIndex.contentNotifications};
+  overflow: visible;
+  position: sticky;
+  top: ${bookBannerDesktopMiniHeight + toolbarDesktopHeight}rem;
+
+  @media (max-width: ${inlineDisplayBreak}) {
+    top: ${bookBannerMobileMiniHeight + toolbarMobileHeight + toolbarMobileSearchWrapperHeight}rem;
+    z-index: calc(${theme.zIndex.searchSidebar} + 1);
```
TomWoodward: i don't think you want this here isn't the one from line 39 correct on mobile too?

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  clearErrorAfter,
+  CloseButton,
+  CloseIcon
+} from './styles';
+
+interface Props {
+  dismiss: () => void;
+}
+
+const window = assertWindow();
```
TomWoodward: this is gonna break everything if you put it up here, put it down in the function body and it will work i think

src/app/utils.ts
```diff
@@ -212,3 +211,27 @@ export const onEscHandler = (element: React.RefObject<HTMLElement>, isEnabled: b
...
+      element.addEventListener(event, cb);
+    } else {
+      element.removeEventListener(event, cb);
+    }
+
+    return () => element.removeEventListener(event, cb);
+  }, [element, isEnabled, cb, event]);
+};
+
+export const useTimeout = (delay: number, cb: () => void, deps: any[]) => React.useEffect(() => {
```
TomWoodward: what is the type of the deps argument to useEffect? is it `any[]`?

src/app/utils.ts
```diff
@@ -212,3 +211,27 @@ export const onEscHandler = (element: React.RefObject<HTMLElement>, isEnabled: b
 export const useOnEsc = (element: React.RefObject<HTMLElement>, isEnabled: boolean, cb: () => void) => {
   React.useEffect(onEscHandler(element, isEnabled, cb), [element, isEnabled]);
 };
+
+export const useOnDOMEvent = (
+  element: React.RefObject<HTMLElement>['current'] | Window ,
```
TomWoodward: i think if a ref is being used you'd want to pass the refObject, not the html element

src/app/utils.ts
```diff
@@ -212,3 +211,27 @@ export const onEscHandler = (element: React.RefObject<HTMLElement>, isEnabled: b
...
+  event: keyof HTMLElementEventMap,
+  cb: () => void
+) => {
+  React.useEffect(() => {
+    if (!element) { return; }
+
+    if (isEnabled) {
+      element.addEventListener(event, cb);
+    } else {
+      element.removeEventListener(event, cb);
```
TomWoodward: you don't have to remove it, it will have been removed by the previous teardown

TomWoodward (changes requested):

src/app/notifications/components/SearchFailure/index.tsx
janKozie1: I thought that since callback that triggers the render gets executed few times it might cause some issues with the timeout, my bad.

TomWoodward: no worries, i was wondering if i was missing something

src/app/utils.ts
janKozie1: Its `ReadonlyArray<any>` but i've updated it. Not sure how to fix the linter errors about dependency list not being an array literal though.

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -18,10 +18,16 @@ interface Props {
 const SearchFailure = ({ dismiss }: Props) => {
   const window = assertWindow();
   const [isFadingOut, setIsFadingOut] = React.useState(false);
+  const [shouldAllowDismissal, setShouldAllowDismissal] = React.useState(false);
```
janKozie1: I've added this because clicking on a search result on mobile triggers a scroll event which will make the message fade out immediately. Not sure if this is the best solution.

TomWoodward: i haven't looked at your code yet, but i would think that maybe adding another timeout to accomplish "the message must be shown for x amount of time before we will automatically dismiss it" would be a good solution

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -65,11 +68,31 @@ export default class PageComponent extends Component<PagePropTypes> {
...
     const highlgihtsAddedOrRemoved = this.highlightManager.update();
-    this.searchHighlightManager.update(prevProps.searchHighlights, this.props.searchHighlights, {
-      forceRedraw: highlgihtsAddedOrRemoved,
-    });
+
+    this.searchHighlightManager.update(
+      prevProps.searchHighlights,
+      this.props.searchHighlights,
+      {
+        cb: this.onHighlightSelect,
```
TomWoodward: can we call this option `onSelect` or something like that so it is more descriptive than `cb`

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -25,6 +27,7 @@ const parser = new DOMParser();
 
 export default class PageComponent extends Component<PagePropTypes> {
```
TomWoodward: you need to pass the state type as the second generic parameter to Component

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -65,11 +68,31 @@ export default class PageComponent extends Component<PagePropTypes> {
     }
 
     const highlgihtsAddedOrRemoved = this.highlightManager.update();
-    this.searchHighlightManager.update(prevProps.searchHighlights, this.props.searchHighlights, {
-      forceRedraw: highlgihtsAddedOrRemoved,
-    });
+
+    this.searchHighlightManager.update(
+      prevProps.searchHighlights,
+      this.props.searchHighlights,
```
TomWoodward: you can revert this whitespace

src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -39,7 +40,12 @@ const updateResults = (services: Services, previous: HighlightProp, current: Hig
   services.searchResultMap = highlightResults(services.highlighter, current.searchResults);
 };
 
-const selectResult = (services: Services, previous: HighlightProp, current: HighlightProp, options: Options) => {
+const selectResult = (
+  services: Services,
+  previous: HighlightProp,
+  current: HighlightProp,
+  options: Options
```
TomWoodward: you didn't actually change any of this did you?

src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -64,9 +70,17 @@ const selectResult = (services: Services, previous: HighlightProp, current: High
...
+  if (options.cb) {
+    options.cb(firstSelectedHighlight);
+  }
 };
 
-const handleUpdate = (services: Services) => (previous: HighlightProp, current: HighlightProp, options: Options) => {
+const handleUpdate = (services: Services) => (
+  previous: HighlightProp,
+  current: HighlightProp,
+  options: Options,
```
TomWoodward: or this?

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface Props {
+  dismiss: () => void;
+}
+
+// tslint:disable-next-line:variable-name
+const SearchFailure = ({ dismiss }: Props) => {
+  const window = assertWindow();
+  const [isFadingOut, setIsFadingOut] = React.useState(false);
+  const [shouldAllowDismissal, setShouldAllowDismissal] = React.useState(false);
```
TomWoodward: instead of `shouldAllowDismissal` this is more like `shouldAutoDismiss`

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const startFadeOut = () => {
+    if (shouldAllowDismissal) {
+      setIsFadingOut(true);
+    }
+  };
+
+  useTimeout(clearErrorAfter, startFadeOut, [shouldAllowDismissal]);
```
TomWoodward: this is going to dismiss after startFadeOut + 100 because of the re-bind after shouldAllowDismissal changes, i think you can probably just not pass any dependencies and it'll be more clear

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,93 @@
...
+import theme from '../../../theme';
+import { inlineDisplayBreak } from '../../theme';
+import { Header } from '../Card';
+
+export const clearErrorAfter = 3200;
+const fadeOutDuration = 1000;
+
+const bannerBackground = '#F8E8EB';
+const errorBorderColor = '#E297A0';
+const closeIconClor = '#EDBFC5';
```
TomWoodward:
```suggestion
const closeIconColor = '#EDBFC5';
```

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,93 @@
...
+} from '../../../content/components/constants';
+import { disablePrint } from '../../../content/components/utils/disablePrint';
+import theme from '../../../theme';
+import { inlineDisplayBreak } from '../../theme';
+import { Header } from '../Card';
+
+export const clearErrorAfter = 3200;
+const fadeOutDuration = 1000;
+
+const bannerBackground = '#F8E8EB';
```
TomWoodward: unless these colors are totally made up randomly they should be pulled out of app/theme.colors

src/app/notifications/types.ts
```diff
@@ -17,7 +17,7 @@ export interface AppMessageNotification {
 
 export type Messages = Message[];
 
-export type AnyNotification = ActionType<Pick<typeof actions, 'updateAvailable' | 'acceptCookies' >>
+export type AnyNotification =  ActionType<Pick<typeof actions, 'updateAvailable' | 'acceptCookies' >>
```
TomWoodward: bad space

src/app/notifications/reducer.ts
```diff
@@ -3,9 +3,9 @@ import { getType } from 'typesafe-actions';
 import { ActionType } from 'typesafe-actions';
 import { AnyAction } from '../types';
 import * as actions from './actions';
-import { AnyNotification, Message, State } from './types';
+import { AnyNotification, Message , State } from './types';
```
TomWoodward: bad space

src/app/utils.ts
```diff
@@ -212,3 +211,27 @@ export const onEscHandler = (element: React.RefObject<HTMLElement>, isEnabled: b
 export const useOnEsc = (element: React.RefObject<HTMLElement>, isEnabled: boolean, cb: () => void) => {
   React.useEffect(onEscHandler(element, isEnabled, cb), [element, isEnabled]);
 };
+
+export const useOnDOMEvent = (
```
TomWoodward: this is a react hook so please put it in reactUtils

TomWoodward (changes requested):

TomWoodward:
also this https://openstax.slack.com/archives/C8V1XET1R/p1583358680161400 about stacking with other notifications on mobile

janKozie1:
I've noticed that there are some tests in react utils that dont test anything:
```
 it('doesn\'t register event listener when ref.current doesn\'t exist', () => {
    utils.onFocusLostHandler({ current: null }, true, () => null)();
    expect(addEventListener).not.toHaveBeenCalled();
  });
  it('doesn\'t register event listener when os disabled', () => {
    utils.onFocusLostHandler({ current: null }, false, () => null)();
    expect(addEventListener).not.toHaveBeenCalled();
  });``` Should i change those?

**Comment by janKozie1** [comment]
I've changed the timeout hooks because i _think_ that with the previous implementation, there was no way of keeping the timeout at X  seconds and updating it. 

Problem surfaced when the callback depended on stateful value that was being set in a different timeout, and because it wasnt getting updated, the callback still had the stale value.
If i tried updating it, no matter what it would set the timeout again.

Hopefully this works (i have no idea how to test it tho)

**Comment by PiotrKozlowski** [comment]
@janKozie1 To test `setTimeout` you can use `jest.useFakeTimer()` and `jest.runTimesrToTime(delay)` and then expect that some function has been called or not

File: src/app/reactUtils.spec.tsx [hunk 36]
```diff
@@ -81,6 +82,46 @@ describe('useFocusLost', () => {
...
+
+    callback = jest.fn();
+
+    Component = () => {
+      const [delay, setDelay] = React.useState(1000);
+      utils.useTimeout(delay, callback, []);
+
+      return <button onClick={() => setDelay(2000)}></button>;
+    };
+  });
```
janKozie1: @PiotrKozlowski what do you think about testing hooks in an actual component and not extracting their body into handlers? The hook allows the `delay` to change but there is no a component that does that (yet?), which lead me to creating a small test component that was doing just that.

PiotrKozlowski: I think that's okay. You could also add tests for this hook in tests for `SearchFailure` component where this hook is used

janKozie1: Well all of its other functionality (preserving the timeout, calling callback and etc.) is actually tested there. Just didnt have a way of testing the delay change.

PiotrKozlowski: oh, then I think it's totally fine to create this test component like you did

src/app/notifications/components/SearchFailure/index.tsx
```diff
@@ -32,18 +32,20 @@ const SearchFailure = ({ dismiss, mobileToolbarOpen, selectedHighlight }: Props)
...
-    setShouldAutoDismiss(false);
-    resetAutoDismiss();
-    resetErrorClearing();
-  }, [selectedHighlight]);
+    setImmediate(() => {
+      setIsFadingOut(false);
+      setShouldAutoDismiss(false);
+      resetErrorClearing();
+      resetAutoDismiss();
+    });
```
janKozie1: I'm not happy with this solution, but im struggling to find an another way. Problem is that this runs before(?) `useOnDOMEvent`. The callback gets updated but `useOnDOMEvent` still has the stale value which causes it to dismiss it anyway.

TomWoodward: i'm not understanding the problem you're having. what callback gets updated? what updates it? what value is stale?

janKozie1: I'm not exactly sure if i understand correctly whats happening.
The `useEffect` should basically reset the timeouts and state so that the component is renewed.
If the error is shown, and the click causes the same error:
1. First event listener for `useOnDOMEvent` executes which i expect
2. Then there is no rerender which im not sure why
3. Then the useEffect runs and resets the component
4. Then rerender happens which the correct values
5. Then an another rerender happens with values like the 1) was executed now which causes the fade out to start immediately.

Im experimenting now with the `useOnDOMEvent` to listen to mousedown rather than click.

TomWoodward: if you write a failing test i will help to figure this out, but if you want to try a few other things on your own, i would try:
- rearranging the order of the stuff in the useEffect, react usually batches state updates, but if it doesn't i could see that affecting things
- useEffect runs async while useLayoutEffect runs directly after the dom update, you might have better luck with useLayoutEffect in this case

i suspect that you're getting two renders because of the async nature of useEffect, you're getting one render as a result of the click event callback, and another render from the useEffect callback, and their state changes are clobbering each other. useState has a thing for when your new state is a derivative of your old state, you can pass a function into the update function so that your initial values are guaranteed to be consistent at the time that react decides to run the update, but this isn't helpful in your case because your new state is a derivative of _other_ states in your component, if you were to combine your states into one `useState` call, you might be able to use that feature and have more consistent behavior. the recommendation is in general to use separate `useState` calls, but because your states interact with one another, i think it is a special case.

src/app/content/components/Page.spec.tsx
```diff
@@ -620,10 +620,28 @@ describe('Page', () => {
     store.dispatch(receiveSearchResults(makeSearchResults([hit])));
     store.dispatch(selectSearchResult({result: hit, highlight: 0}));
 
-    // after images are loaded
+    await Promise.resolve();
+    await Promise.resolve();
```
PiotrKozlowski, line 624: Only one of these is required, tested with `yarn test:unit`

src/app/content/components/Page.spec.tsx
```diff
@@ -620,10 +620,28 @@ describe('Page', () => {
...
+
+    // click again for selectedSearchResult to update
+    store.dispatch(selectSearchResult({result: hit, highlight: 0}));
+
+    await Promise.resolve();
     await Promise.resolve();
```
PiotrKozlowski, line 630: Both of these can be removed, tested with `yarn test:unit`

PiotrKozlowski (changes requested):
This pr looks good, I've left some minor comments

src/app/reactUtils.ts
```diff
@@ -36,6 +37,61 @@ export const useFocusLost = (ref: React.RefObject<HTMLElement>, isEnabled: boole
...
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [...deps, callback]);
+
+  useEffect(() => {
+      reset();
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [delay]);
```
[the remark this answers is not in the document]
janKozie1: I wanted to give control of when the timeout actually resets vs when its callback gets updated.
In my opinion there are some cases where you would like the callback to have fresh values but still run after X seconds from its initial start.

I think deps in this hook actually serve no role. The callback gets updated on its own so its closure is always up to date.

src/app/notifications/components/SearchFailure/styles.tsx
```diff
@@ -0,0 +1,100 @@
...
+    opacity: 0;
+  }
+`;
+
+// tslint:disable-next-line:variable-name
+export const BannerBodyWrapper = styled.div`
+  width: 100%;
+  margin: 0;
+  height: 0;
+  z-index: ${theme.zIndex.contentNotifications - 1};
```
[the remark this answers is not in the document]
janKozie1: Well i felt like it still should behave like a notification in relation to other elements on the page and it's z index only matters for the other notifications. I'm not really sure about adding it there since book banner does something similar.

PiotrKozlowski (approved):

src/helpers/rangy.ts
```diff
@@ -16,7 +16,7 @@ export const findTextInRange = (
   text: string,
   range: RangyRange = rangy.createRange()
 ): RangyRange[] => {
-  const foundMatch = range.findText(text, {
+  const foundMatch = range.findText(text.trim(), {
```
janKozie1, line 19: Reason for this is mainly [this line](https://github.com/timdown/rangy/blob/1e55169d2e4d1d9458c2a87119addf47a8265276/src/modules/rangy-textrange.js#L885)
Because of it every text that has a leading space and is right after math is never found. 
Some tags inside mathml fit [this requirement](https://github.com/timdown/rangy/blob/1e55169d2e4d1d9458c2a87119addf47a8265276/src/modules/rangy-textrange.js#L675) which injects a new line to the string that rangy is using to compare with the query.

src/app/content/components/Page.spec.tsx
```diff
@@ -804,6 +820,58 @@ describe('Page', () => {
...
+    jest.spyOn(highlightUtils, 'highlightData').mockReturnValue(() => ({} as any));
+
+    // normally, search result selection handler would noop if the
+    // search result is the same. This makes it think that a new highlight was
+    // added and will force reselection
+
+    renderer.act(() => {
+      store.dispatch(createHighlight({} as any, {} as any));
+      store.dispatch(selectSearchResult(searchResultToSelect));
+    });
```
janKozie1, line 865: I'm not sure why those two dispatches show a warning that they are not inside a `renderer.act`

PiotrKozlowski (changes requested):

src/app/notifications/components/SearchFailure/styles.tsx
janKozie1: Yes, the colors were previously in theme, now just constants in this file

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -67,12 +74,39 @@ export default class PageComponent extends Component<PagePropTypes> {
...
+    this.setState({
+      hasSearchError: true,
+      selectedSearchResult: current,
+    });
+  };
+
+  public dismissError = () => {
+    this.setState({
+      hasSearchError: false,
+    });
```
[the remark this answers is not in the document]
janKozie1: No, doing so is one of the reasons why search error appears when creating a new highlight ( `onHighlightSelect` gets ran again, so i think its necessary to keep it )

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -67,12 +74,39 @@ export default class PageComponent extends Component<PagePropTypes> {
...
+  public onHighlightSelect: OptionsCallback = ({current, selectedHighlight}) => {
+    if (selectedHighlight) {
+      this.setState({
+        hasSearchError: false,
+      });
+
+      return;
+    }
+
+    if (current === this.state.selectedSearchResult) { return; }
```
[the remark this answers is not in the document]
janKozie1: I'm pretty sure that the search results themselves don't have an Id. I think generating an id for a selected search result based on things that make it unique, could be easier than checking all of the highlights' ids.

src/app/content/search/utils.ts
```diff
@@ -147,11 +148,20 @@ export const highlightResults = (
...
+        .map((range) => {
+          const highlight = new Highlight(range.nativeRange, {content: range.toString()});
+
+          try {
+            highlighter.highlight(highlight);
+          } catch (e) {
+            Sentry.captureException(e);
+          }
+
+          return highlight;
```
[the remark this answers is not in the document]
janKozie1: I think it's okay to just return highlight after attaching it with `attachHighlight` since it will still get filtered and it avoids additional unnecessary loops.

src/app/content/components/utils/attachHighlight.ts
```diff
@@ -5,12 +5,16 @@ const attachHighlight = (
...
-    Sentry.captureException(new Error(`Highlight with id: ${highlight.id} has not been attached.`));
+  try {
+    highlighter.highlight(highlight);
+    const result = highlighter.getHighlight(highlight.id);
+    if (!result || !result.isAttached()) {
+      Sentry.captureException(new Error(`Highlight with id: ${highlight.id} has not been attached.`));
+    }
+    return result;
+  } catch (e) {
+    Sentry.captureException(e);
```
PiotrKozlowski, line 16: Now we are calling Sentry with two different messages. It would be nice to call it only once, maybe using `let results: Highlight | undefined` at the beginning?
I think that `attachHighlight` should now require optional arg `errorMsg: (highlight) => string`
Default message would be `Highlight with id: ${highlight.id} has not been attached.`
This is needed to provide miningfull sentry error because search highlights have randomly generated ids

PiotrKozlowski (changes requested):
Only one minor thing and I'll accept

src/app/content/components/utils/attachHighlight.ts
janKozie1: What do you think about passing an optional callback instead? It would still be possible to throw a custom error from there and add more logic if necessary.

src/app/content/search/utils.ts
```diff
@@ -147,11 +148,18 @@ export const highlightResults = (
...
-        const highlight = new Highlight(range.nativeRange, {content: range.toString()});
-        highlighter.highlight(highlight);
-        return highlight;
-      });
+      const highlights = getHighlightRanges(element, highlightText)
+        .map((range) => {
+          const highlight = new Highlight(range.nativeRange, {content: range.toString()});
+
+          attachHighlight(highlight, highlighter, (failedHighlight) =>
+            `Search result highlight with id: ${failedHighlight.id} has not been attached`
```
PiotrKozlowski, line 156: Maybe @TomWoodward can decide if this message is okay and if not then what data should you pass to make it more clear

PiotrKozlowski (approved):
I'm not sure about an error message text, but I'll let Tom decide about this.

src/app/content/search/utils.ts
TomWoodward: the id is not useful because for search it will be random. the useful info would be the search term and the page we're on, if you can easily get that in there.

if not, just put a fixed string like "Search result failed to highlight" and we might be able to get the other relevant data out of sentry if we try hard enough

src/app/content/components/utils/attachHighlight.ts
```diff
@@ -1,16 +1,22 @@
...
+  try {
+    highlighter.highlight(highlight);
+    const result = highlighter.getHighlight(highlight.id);
+    if (!result || !result.isAttached()) {
+      throw new Error(`Highlight with id: ${highlight.id} has not been attached.`);
+    }
+    return result;
+  } catch (e) {
+    const error = errorMsg ? new Error(errorMsg(highlight)) : e;
+    Sentry.captureException(error);
```
TomWoodward, line 18: this logic seems very odd, first, are there any exceptions you're trying to catch here other than the one thrown above? why not just capture it there instead of throwing it? also, why default the message above and then override it here, why not put the custom message in on line 13?

TomWoodward (changes requested):

src/app/content/components/utils/attachHighlight.ts
janKozie1: I'm pretty sure that `higlighter.highlight` throws error like `cannot read property matches of null` when search results fails to highlight too. If i were to default it there then it is possible that it wouldn't actually record that error but rather the one i just mentioned.
cc. @TomWoodward

TomWoodward: if `highlight` throws `cannot read property matches of null` that is an indication of another issue that we'd like to have in sentry, and your change that you have in place will replace it with this more generic error.

TomWoodward (approved):

PR #471, opened 2020-01-31
Track showCreate/Login for highlighting

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/808
More analytics for highlighting


src/app/content/highlights/components/EditCard.tsx
```diff
@@ -61,13 +61,24 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
...
       onBlur();
     }
   };
 
+  React.useEffect(() => {
+    if (authenticated) {
+      trackShowCreate();
+    } else {
+      trackShowLogin();
+    }
```
TomWoodward: isn't this going to trigger `trackShowCreate` on page load for every highlight that already exists?

TomWoodward (changes requested):

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -69,6 +71,16 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
     }
   };
 
+  React.useEffect(() => {
+    if (!isFocused) { return; }
```
TomWoodward: this will still log pass when it shouldn't once you do the thing about showing all cards. i think the actual check you want is to see if `data` is defined

TomWoodward (changes requested):

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -69,6 +71,16 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
...
 
+  React.useEffect(() => {
+    if (data) { return; }
+    if (authenticated) {
+      trackShowCreate();
+    } else {
+      trackShowLogin();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [data]);
```
TomWoodward: this'll make it re-run when data is changed. not that that probably matters but you might as well leave it out and then it'll only run on mount

TomWoodward (changes requested):

TomWoodward (approved):
wait for qa before merge

PR #477, opened 2020-02-03
Fix highlights by chapter order 

janKozie1:
Is this approach fine? I will add the tests if it's okay.

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/743


src/app/content/highlights/components/Highlights.tsx
```diff
@@ -88,47 +88,56 @@ interface SectionHighlightsProps {
...
-            );
-          })}
+      {Object.entries(highlights[location.id])
+        .sort(([pageA], [pageB]) => {
+          if (archiveTreeSectionIsChapter(location)) {
+            const orderedPages = findTreePages(location);
+            return orderedPages.findIndex(({id}) => id === pageA) - orderedPages.findIndex(({id}) => id === pageB);
+          }
+          return 0;
+        })
```
PiotrKozlowski: Can you move this sorting logic to `summaryHighlights` selector `src/app/content/highlights/selectors.ts`?
Also: `const orderedPages = findTreePages(location)` should be probably before `sort` so it is not recalculated for every item.

PiotrKozlowski (changes requested):

src/app/content/highlights/components/Highlights.tsx
PiotrKozlowski: or actually create new selector based on `summaryHighlights` so it will be recalculated only if `summaryHighlights` has changed

PiotrKozlowski: you can also replace `const locationFilters = useSelector(selectors.highlightLocationFilters);` with `const locationFilters = useSelector(selectors.highlightLocationFiltersWithContent);` so we will map only for locations with highlights

TomWoodward:
@janKozie1 can you explain what the actual error was with safari?

PiotrKozlowski:
Recording which @annaludwin posted seems not to be related to Safari
I think it might be related to this code: from `updateSummaryHighlightsDependOnFilters`
```
let newHighlights: SummaryHighlights = {
    ...summaryHighlights,
    [locationFilterId]: {
      ...summaryHighlights[locationFilterId],
      [pageId]: [...(summaryHighlights[locationFilterId] || {})[pageId] || []],
    },
  };
```
we are adding `[pageId]` at the and of object which then is rendered also at the end?

This code will be removed with https://github.com/openstax/rex-web/pull/478 so this actually will not need any more work?

janKozie1:
I was able to replicate  in Calculus Volume 1 - functions and graphs by:
- adding a highlight in 1.3
- adding a highlight in 1.5
- adding a highlight in 1.1

src/app/content/highlights/selectors.ts
```diff
@@ -95,6 +96,14 @@ export const highlightLocationFilters = createSelector(
     : new Map() as HighlightLocationFilters
 );
 
+export const orderedSummaryHighlights = createSelector(
+  localState,
+  highlightLocationFilters,
+  (state, locationFilters) => {
+    return getSortedSummeryHighlights(state.summary.highlights, locationFilters);
```
TomWoodward: if what you want is `highlights` you should depend on `summaryHighlights` not `localState` so the memoization works better

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -67,6 +69,34 @@ export const addSummaryHighlight = (summaryHighlights: SummaryHighlights, data:
   return newHighlights;
 };
 
+export const getSortedSummeryHighlights =
```
TomWoodward:
```suggestion
export const getSortedSummaryHighlights =
```

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -67,6 +69,34 @@ export const addSummaryHighlight = (summaryHighlights: SummaryHighlights, data:
...
+  (summaryHighlights: SummaryHighlights | null, locationFilters: HighlightLocationFilters ) => {
```
TomWoodward: does this function return `SummaryHighlights`? objects have no implicit order in javascript, so if this is supposed to sort things we should really be returning an array

TomWoodward (changes requested):

src/app/content/highlights/selectors.ts
```diff
@@ -78,14 +79,14 @@ export const summaryColorFilters = createSelector(
...
   localState,
-  (state) => state.summary.highlights
+  (state) => state.summary.pagination
 );
 
-export const summaryPagination = createSelector(
+export const summaryHighlights = createSelector(
   localState,
-  (state) => state.summary.pagination
+  (state) => state.summary.highlights
```
PiotrKozlowski: It seems like this change is unnecessary

src/app/content/highlights/utils/summaryHighlightsUtils.ts
```diff
@@ -67,6 +70,46 @@ export const addSummaryHighlight = (summaryHighlights: SummaryHighlights, data:
...
+          pages:  Object.entries(summaryHighlights[locationID]).sort(([pageA], [pageB]) => {
+            return orderedPages.findIndex(({ id }) => id === pageA) - orderedPages.findIndex(({ id }) => id === pageB);
+          }).reduce((previousPages, [pageId, highlights]) => {
+            return [...previousPages, {
+              highlights,
+              pageId,
+            }];
+          }, [] as OrderedSummaryHighlights[0]['pages']),
+        },
+      ];
```
PiotrKozlowski: Can we simplify this to something like: ?
```
const pages: OrderedSummaryHighlights[0]['pages'] = [];

      for (const page of orderedPages) {
        if (!summaryHighlights[locationID][page.id]) { continue; }
        pages.push({
          highlights: summaryHighlights[locationID][page.id],
          pageId: page.id,
        });
      }

      return [
        ...previousLocations,
        {
          location,
          pages,
        },
      ];
```

PiotrKozlowski (changes requested):
It looks good, I've left two comments, but I think you can start adding tests

janKozie1:
I don't think the test for `SectionHighlights` to throw is necessary anymore since the pages are based on `findTreePages(location)`

PiotrKozlowski (approved):

src/app/content/highlights/components/Highlights.tsx
```diff
@@ -81,56 +77,58 @@ const Highlights = () => {
...
-                    </Styled.HighlightNote>
-                  ) : null}
-                </Styled.HighlightContentWrapper>
-              </Styled.HighlightOuterWrapper>
-            );
-          })}
-        </Styled.HighlightWrapper>;
-      })}
+      {pages
+        .map(({pageId, highlights}) => {
```
TomWoodward: this line break doesn't really have to be here

TomWoodward (changes requested):

janKozie1:
@TomWoodward updated

TomWoodward (approved):
wait for qa review before merging

PR #479, opened 2020-02-03
Fix responsivity on Firefox

PiotrKozlowski:
This is fixing responsivity issue on Firefox for pages like this `books/chemistry-2e/pages/1-2-phases-and-classification-of-matter`

for: https://github.com/openstax/unified/issues/829


src/app/content/components/Content.tsx
```diff
@@ -118,7 +118,7 @@ const OuterWrapper = styled.div`
   @media screen {
     display: flex;
     flex-direction: row;
-    overflow: visible;
+    overflow: auto;
```
TomWoodward: you can't specify this here, it breaks the position sticky on sidebars

PiotrKozlowski:
@TomWoodward Change which I've just pushed seems to keep current behavior in Chrome and in Firefox it will add scrollbar only for large element - in this example on the table - not for the entire page.

While searching for a solution that will add scrollbar for an entire page I've run into this case which seems okay to me. Let me know I should search for something else.

PiotrKozlowski:
@TomWoodward please take a look at: https://github.com/openstax/rex-web/pull/479#issuecomment-581899527

janKozie1:
Still can see a horizontal scrollbar on some pages, especially when they are focused 
Screenshot from 2020-08-25 19-29-22

janKozie1:
Few things:

- firefox seems to have trouble calculating the layout properly when there is a vertical scrollbar on desktop devices. When i hid the scrollbar with `scrollbar-width: none` it seemed to be laid out correctly (this is not the case for mobile devices since their scrollbar is "above" the content, not next to it).

- there is a `min-width: 30.4rem` set on the `#root` which which will cause overflow on mobile devices smaller than that.

- also it seems like i can make the horizontal scrollbar appear on small resolutions on chrome, the changes I've pushed seem to fix it too.

PiotrKozlowski:
@janKozie1
My bad, this pr is old and I forget to comment on what Tom said to me when I've started it:

> it works properly in chrome
> the way it works in chrome is the way we want it to work

https://rex-web.herokuapp.com/books/chemistry-2e/pages/1-2-phases-and-classification-of-matter

so basically he wanted the main content to be scrollable instead of a wider element only, on both firefox and chrome

But maybe something has changed? Because this approach seems okay to me and has a simpler implementation
Maybe we should ask Kim or/and Suhani

RoyEJohnson (approved):
This looks good to me.

TomWoodward:
@PiotrKozlowski the solution here won't work because there are things other than tables (notably, math) that might be wider than the screen. 

for the question of if we want the whole page to scroll horizontally or the particular wide element, i think scrolling the particular element is easier to do on desktop while scrolling the whole page is easier on mobile, but potentially less good looking with the way the text width appears.

i'm open to doing the scrolling on either the whole content or the individual elements, but we'd need to make sure we handle all the different kinds of elements that can be too large, while the currently solution works for anything, so i'm closing this PR for now.

PR #481, opened 2020-02-04
Initial display of all cards

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/581

TODO:
- [x] fix top offset for highlights next to each other and next highlights
- [x] Decision: Cards without a note should be displayed only when focused. ~How to display cards for highlights without notes? Should text area be visible?~
- [x] add animations
- [x] focus highlight on click in card
- [x] fix z-index for dropdowns
- [x] Update card height when displaying confirm delete overlay
- [x] Disable this behavior on smaller screens
- [x] #488  When creating a new highlight, card should be in the proper position, not at the end. (separate pr)
- [x] ~truncate text of notes if they are not focused (separate pr)~ This is actually already done
- [x] Add tests - (separate pr when this is approved)
- [x] Fix issue from the image: (the last card should be at highlight level)
image


src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -11,20 +15,109 @@ interface Props {
...
+      }
+
+      if (cardsHeights.get(highlight.id)) {
+        lastVisibleCardPosition = stackedTopOffset;
+        lastVisibleCardHeight = cardsHeights.get(highlight.id)!;
+      }
+
+      newPositions.set(highlight.id, stackedTopOffset);
+
+      setCardsPositions(newPositions);
```
TomWoodward: don't you want to do this outside the loop?

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -11,20 +15,109 @@ interface Props {
...
+    }
+  }, [highlights, cardsHeights, container]);
+
+  React.useEffect(() => {
+    if (displayAllCards) {
+      updatePositions();
+    } else {
+      onBlur();
```
TomWoodward: if i'm on a small screen, so `displayAllCards` is false, and i edit a highlight note, as soon as i edit the text area the save buttons appear, changing the height of the card, so `onHeightChange` is called, which will create a new `updatePositions` handler, triggering this effect.

so is this removing focus from cards when you edit them on small screens? or am i missing something?

TomWoodward (changes requested):
you're also able to resize the textarea of the edit note, not sure how we should handle that

TomWoodward:
just tested this and the animations are super snazzy but sometimes a little laggy, we might want to experiment with the order we call our setstates to produce less jitter

TomWoodward:
@PiotrKozlowski some people are reporting getting an error `Cannot read property 'style' of null` but i can't reproduce, i think it has something to do with a race condition before the book is loaded but i'm not sure

src/app/content/highlights/components/Card.tsx
```diff
@@ -75,6 +84,25 @@ const Card = (props: Props) => {
     }
   }, [props.highlight, annotation]);
 
+  const prevHeight = element.current && element.current.offsetHeight;
```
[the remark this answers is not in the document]
PiotrKozlowski: I'll remove this line because `onHeightChange` will be called either from `DisplayNote` or `EditCard`

src/app/content/highlights/components/Card.tsx
```diff
@@ -75,6 +84,25 @@ const Card = (props: Props) => {
...
+    if (!annotation && !isFocused) {
+      props.onHeightChange(props.highlight.id, { current: null } as React.RefObject<HTMLElement>);
+      return;
+    }
+
+    const currentHeight = element.current && element.current.offsetHeight;
+    if (currentHeight && prevHeight !== currentHeight) {
+      props.onHeightChange(props.highlight.id, element);
+    }
+  }, [element, editing, annotation, isFocused, prevHeight]);
```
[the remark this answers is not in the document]
PiotrKozlowski: Actually only `annotation` and `isFocused`, because I'll call `onHeightChange` here only if we do not render this Card

src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -53,6 +57,11 @@ const DisplayNote = React.forwardRef<HTMLElement, Props>((
     }
   }, [isFocused]);
 
+  React.useEffect(() => {
+    const refElement = confirmationRef.current ? confirmationRef : element;
+    onHeightChange(highlight.id, refElement);
+  }, [element, highlight, confirmationRef, confirmingDelete]);
```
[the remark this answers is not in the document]
PiotrKozlowski: It's was here initially because `Confirmation` has position: absolute and I wanted to have direct access. 
I've moved logic with `onHeightChange` to `DisplayNote` and `EditCard`

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -86,6 +88,12 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
 
   React.useEffect(onClickOutside(element, isFocused, blurIfNotEditing), [isFocused, editingAnnotation]);
 
+  React.useEffect(() => {
+    if (element.current) {
+      onHeightChange(highlight.id, element);
```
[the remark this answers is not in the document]
PiotrKozlowski: I still need this call in `Card`:
```
if (!annotation && !isFocused) {
      props.onHeightChange({ current: null } as React.RefObject<HTMLElement>);
    }
```
but you're right, rest is unnecessary.

src/app/content/highlights/components/Card.tsx
```diff
@@ -271,8 +324,9 @@ const StyledCard = styled(Card)`
     `;
   }}
 
-  @media (max-width: ${remsToEms(contentTextWidth + sidebarDesktopWidth + additionalWidthForCard)}em) {
+  @media (max-width: ${mediaQueryBreakToStopDisplaingAllCards}) {
     /* the window is too small to show note cards next to content when the toc is open */
+    animation: none;
```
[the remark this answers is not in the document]
PiotrKozlowski: Those shall inherit

src/app/content/highlights/components/Card.tsx
```diff
@@ -227,14 +256,38 @@ const mobileDisplay = css`
...
+  ${fadeInAnimation}
   position: absolute;
   padding: ${cardPadding}rem;
   ${cardBorder}
   ${rightSideDisplay}
   ${disablePrint}
 
+  transition: all 0.3s;
+
+  ${DropdownList} {
```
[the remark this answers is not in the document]
PiotrKozlowski: I guess it would remove animation for EditCard and Confirmation so either those should have separate styles or I can just animate Card

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -11,20 +15,109 @@ interface Props {
...
+    const position = cardsPositions.get(id);
+    if (typeof position !== 'number' || !highlight) { return; }
+
+    const topOffset = assertDefined(
+      getHighlightTopOffset(container, highlight),
+      `Couldn't get top offset for highlights`
+    );
+
+    if (position > topOffset) {
+      element.current!.style.top = `-${position - topOffset}px`;
```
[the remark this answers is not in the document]
PiotrKozlowski: I thought this would be easier to implement for now and in the next step we could just modify how `updatePositions` behave when we pass additional parameter like `focusId: highlightId`

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -11,20 +15,109 @@ interface Props {
...
-  ? <div className={className}>
+const Wrapper = ({highlights, className, container, highlighter}: Props) => {
+  const displayAllCards = !assertWindow().matchMedia(displayAllCardsMediaQuery).matches;
+
+  const element = React.useRef<HTMLElement>(null);
+  const [cardsPositions, setCardsPositions] = React.useState<Map<string, number>>(new Map());
+  const [cardsHeights, setCardsHeights] = React.useState<Map<string, number>>(new Map());
+
+  const onHeightChange = (id: string, ref: React.RefObject<HTMLElement>) => {
+    if (!displayAllCards) { return; }
```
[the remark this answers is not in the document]
PiotrKozlowski: > do you really need to short circuit these functions?

nope

> isn't it fine for them to update anyway since the other cards are all hidden by css?

I would say no because it would still consume cpu power and not provide any benefit for the user

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -11,20 +15,109 @@ interface Props {
...
+
+      if (topOffset < (lastVisibleCardPosition + lastVisibleCardHeight + remsToPx(cardMarginBottom))) {
+        stackedTopOffset = stackedTopOffset
+          + lastVisibleCardHeight
+          + (index > 0 ? remsToPx(cardMarginBottom) : 0);
+      } else {
+        stackedTopOffset = topOffset;
+      }
+
+      if (cardsHeights.get(highlight.id)) {
```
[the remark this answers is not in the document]
PiotrKozlowski: It will be `undefined` for the first render, before cards will fire `onHeightChange` which should happen after they render

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: It does not (firefox ubuntu), but probably there is a way to avoid calling this function.
For now, I'll change its name to `resetTopOffset`

PiotrKozlowski:
@TomWoodward I think I've responded to all fo your comments.

About jitter - it seems like the biggest issue is when we click on a highlight without a note which has hidden Card at the beginning and needs to be inserted between other cards - at the same time other cards need to update their position and we need to update the position of CardWrapper.

This could be easier if I would have access to all Cards heights in CardWrapper, but it seems hard because it can change in multiple situations like Confirmation dialog, typing in a note (shows Save and cancel button), cancel editing, etc.

Anyway - if current changes look good to you, I would like to spend some time on simplifying logic in Card, EditNote and DisplayNote components because I feel there is too much going on. 

About resizing textarea - we could either add resize listener and debounce `onHeightChange` or we could just allow this behavior and just change stacking order so after resizing textarea is above cards which are below.

src/app/content/highlights/components/CardWrapper.tsx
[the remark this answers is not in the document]
TomWoodward: in that slack thread others are ok with this current behavior, so lets leave it the way you have it

src/app/content/highlights/components/CardWrapper.tsx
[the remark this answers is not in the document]
TomWoodward: but in order to actually tell if the cards are all shown, your logic will have to become much more complicated to compensate for the different breakpoints for each sidebar's state. i think it might be worth the extra (probably nominal) processing to avoid this complexity.

src/app/content/highlights/components/CardWrapper.tsx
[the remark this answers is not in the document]
TomWoodward: would it be more correct to noop earlier in the logic for that case then?

TomWoodward (changes requested):
left some comments on the existing conversations. the overall logic looks good, if you have stuff you want to refactor please go ahead and do that before i read the whole thing again

src/app/content/highlights/components/CardWrapper.tsx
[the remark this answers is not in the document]
PiotrKozlowski: Currently, it will calculate positions for cards depends on the default heigh value and then it will update them only if they are different so I think it is okay to do not noop this

PiotrKozlowski:
@TomWoodward Please re-review

src/app/content/components/Page/highlightManager.ts
```diff
@@ -113,9 +113,21 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
     pendingHighlight = highlight;
     if (setListPendingHighlight) {
       setListPendingHighlight(highlight);
+      getProp().focus(highlight.id);
```
TomWoodward: this is done already on line 57, why does it need to be done twice?

src/app/content/highlights/components/Card.tsx
```diff
@@ -109,189 +133,41 @@ const Card = (props: Props) => {
   };
 
   const commonProps = {
-    className: props.className,
-    isFocused: props.isFocused,
-    onBlur: props.blur,
+    className,
+    isFocused,
+    onBlur: blur,
```
TomWoodward: since an onBlur function that does something totally different exists, you might want to do some more refactoring to make this name consistent

src/app/content/highlights/components/Card.tsx
```diff
@@ -109,189 +133,41 @@ const Card = (props: Props) => {
...
+      !editing && style && annotation ? <DisplayNote
+        {...commonProps}
+        style={style}
+        note={annotation}
+        onEdit={() => setEditing(true)}
+      /> : <EditCard
+        {...commonProps}
+        highlight={highlight}
+        authenticated={!!user}
+        loginLink={loginLink}
```
TomWoodward: if some of this stuff, like loginLink and authenticated are not used elsewhere, you could select them within `EditCard`. this file already has a LOT of props, so anything we can do to help manage that would be nice

src/app/content/highlights/components/Card.tsx
```diff
@@ -4,101 +4,125 @@ import { HTMLElement } from '@openstax/types/lib.dom';
...
+  highlighter,
+  loginLink,
+  isFocused,
+  onFocus,
+  onBlur,
+  onHeightChange,
+  page,
+  remove,
+  save,
+  user,
```
TomWoodward: i would prefer to use `props.` repeatedly rather than use up 20 lines of space on this destructuring

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,116 @@
...
+    const position = cardsPositions.get(highlight.id);
+    if (typeof position !== 'number') { return; }
+
+    const topOffset = assertDefined(
+      getHighlightTopOffset(container, highlight),
+      `Couldn't get top offset for highlights`
+    );
+
+    if (position > topOffset) {
+      element.current!.style.top = `-${position - topOffset}px`;
```
TomWoodward: please don't use `!` in production code

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,116 @@
...
+    }
+
+    // This will be undefined for pendingHighlight
+    if (highlight.elements[0]) {
+      scrollIntoView(highlight.elements[0] as HTMLElement);
+    }
+  };
+
+  const resetTopOffset = () => {
+    element.current!.style.top = '0';
```
TomWoodward: and here, just do the null check

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,116 @@
...
+  };
+
+  const updatePositions = React.useCallback(() => {
+    const newPositions: Map<string, number> = new Map();
+
+    let lastVisibleCardPosition: number = 0;
+    let lastVisibleCardHeight: number = 0;
+
+    for (const [index, highlight] of highlights.entries()) {
+      const topOffset = assertDefined(
```
TomWoodward: this isn't going to change during the process right? you could hoist it out of the loop?

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,116 @@
...
+      }
+
+      newPositions.set(highlight.id, stackedTopOffset);
+    }
+
+    setCardsPositions(newPositions);
+  }, [highlights, cardsHeights, container]);
+
+  React.useEffect(() => {
+    updatePositions();
```
TomWoodward: i feel like there is no reason not to put the function body in the effect directly, is there?

src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -34,24 +34,35 @@ interface Props {
...
   useOnClickOutside(element, isFocused, onBlur);
 
   React.useEffect(() => {
     if (!isFocused) {
       setConfirmingDelete(false);
+    } else {
+      onFocus();
     }
-  }, [isFocused]);
+  }, [isFocused, onFocus]);
```
TomWoodward: i don't think you want `onFocus` in the dep array

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -86,6 +90,15 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
...
   React.useEffect(onClickOutside(element, isFocused, blurIfNotEditing), [isFocused, editingAnnotation]);
 
+  React.useEffect(() => {
+    if (element.current) {
+      onHeightChange(element);
+    }
+    if (isFocused) {
+      onFocus();
+    }
+  }, [onFocus, onHeightChange, element, editingAnnotation, isFocused]);
```
TomWoodward: you don't want the callback functions in the deps array

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -0,0 +1,205 @@
...
+          right: 0;
+          bottom: unset;
+          width: unset;
+          height: ${cardPadding / 2}rem;
+        }
+     `)}
+    `;
+  }}
+
+  /* for some reason isTocOpen prop will not refresh after first page load and is null */
```
TomWoodward: tocOpen === null indicates that the state hasn't been set by the user, which means its open on desktop and closed on mobile.

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -0,0 +1,205 @@
...
+  transition: all 0.3s;
+
+  ${(props: WrapperProps) => (props.isTocOpen === null || props.isTocOpen) && css`
+    @media ${minimalWidthForCardsWithToc} {
+      top: 0 !important;
+    }
+  `}
+
+  ${(props: WrapperProps) => props.hasQuery && css`
+    @media (max-width: ${minimalWidthForCards}em) {
```
TomWoodward: shouldn't this use `minimalWidthForCardsWithSearchResults`?

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -0,0 +1,205 @@
...
+export const mainWrapperStyles = css`
+  position: relative;
+  overflow: visible;
+  z-index: ${theme.zIndex.highlightInlineCard};
+  top: 0;
+  transition: all 0.3s;
+
+  ${(props: WrapperProps) => (props.isTocOpen === null || props.isTocOpen) && css`
+    @media ${minimalWidthForCardsWithToc} {
+      top: 0 !important;
```
TomWoodward: why do you have important on all of these

TomWoodward (changes requested):

src/app/content/highlights/components/cardStyles.ts
PiotrKozlowski: This `top` value is set by javascript so it has a bigger priority that css

TomWoodward: oh and it needs to be overridden for the inline display to work properly... interesting.... can you leave a comment in here like `override js inline styles for display of all cards while using overlap display` or something?

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -55,20 +56,30 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
-  const updatePositions = React.useCallback(() => {
+  const getTopOffsetForHighlight = (highlight: Highlight) => {
+    if (topOffsets.has(highlight.id)) {
+      return assertDefined(topOffsets.get(highlight.id), 'this has to be defined');
+    } else {
+      const offset = assertDefined(
+        getHighlightTopOffset(container, highlight),
+        `Couldn't get top offset for highlight with an id: ${highlight.id}`
+      );
+      setTopOffsets((state) => new Map(state).set(highlight.id, offset));
```
TomWoodward: we're modifying state in a loop now, thats probably not great, i don't think this was here before, why are we adding it?

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -147,7 +147,7 @@ export const mainCardStyles = css`
     `;
   }}
 
-  /* for some reason isTocOpen prop will not refresh after first page load and is null */
+  /* isTocOpen means toc is open on the desktop and for mobile, there is a different media query */
```
TomWoodward: isTocOpen is used by both desktop and mobile. for highlight cards, within these highlighting media queries, we know we're on a desktop screen, so we know that `isTocOpen === null` means the sidebar is shown. thats why the condition is `props.isTocOpen === null || props.isTocOpen`.

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -194,7 +194,7 @@ export const mainWrapperStyles = css`
   `}
 
   ${(props: WrapperProps) => props.hasQuery && css`
-    @media (max-width: ${minimalWidthForCards}em) {
+    @media (max-width: ${minimalWidthForCardsWithSearchResults}em) {
       top: 0 !important;
     }
   `}
```
TomWoodward: don't you need another query here to handle when all sidebars are closed but the screen is too small to show the cards on the sides? the same as above on line 171?9

TomWoodward (changes requested):

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: Based on https://github.com/openstax/rex-web/pull/481#discussion_r395847393 I've wanted to memorize these values.
I think `setStates` should be merged and dispatched after the loop ends but I can't test it because servers are not working currently

src/app/content/highlights/components/cardStyles.ts
PiotrKozlowski: I think the mobile breakpoint is large enough to cover this case. I'll check when servers are back

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: So I've checked it and these calls are dispatched after loop ends and are merged into single setState (ex. 9 calls will cause only one update of component after the state update).
If we want to avoid additional re-render (1 is from setting state by `updatePositions` and second is for those `setStates`) then I could make a `Map` outside of a component and clean it before component mounts in `useEffect`. What do you think?

src/app/content/highlights/components/cardStyles.ts
PiotrKozlowski: You were right. I've had to add one more breakpoint and fix an existing one

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,121 @@
...
+      return offset;
+    }
+  };
+
+  React.useEffect(() => {
+    const newPositions: Map<string, number> = new Map();
+
+    let lastVisibleCardPosition: number = 0;
+    let lastVisibleCardHeight: number = 0;
```
janKozie1, line 76: I don't think they need to be typed if you are initializing them with 0's

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -192,10 +187,10 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
             e.preventDefault();
             setEditing(false);
 
-            if (pendingAnnotation === '' && data.annotation) {
+            if (pendingAnnotation === '' && assertDefined(props.data, 'props.data should be defined').annotation) {
               setConfirmingDelete(true);
             } else {
-              saveAnnotation(data);
+              saveAnnotation(assertDefined(props.data, 'props.data should be defined'));
```
janKozie1, line 193: I think you could move this assertion above the if - else

janKozie1 (commented):
Just minor things

PiotrKozlowski:
@janKozie1 I've applied changes as you suggested

TomWoodward:
@PiotrKozlowski looks like there are a bunch of test failures related to the `onFocus` function

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,121 @@
...
+
+  const onFocus = (highlight: Highlight) => {
+    const position = cardsPositions.get(highlight.id);
+    if (typeof position !== 'number') { return; }
+
+    const topOffset = getTopOffsetForHighlight(highlight);
+
+    if (position > topOffset) {
+      assertNotNull(element.current, 'element.current can\'t be null').style.top = `-${position - topOffset}px`;
+    }
```
janKozie1, line 47: If it's possible, you should be animating with `transformY` instead of `top` since i'm pretty sure it doesn't make the browser repaint

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,121 @@
...
+      setCardsHeights((previous) => new Map(previous).set(id, height === null ? 0 : height));
+    }
+  };
+
+  const onFocus = (highlight: Highlight) => {
+    const position = cardsPositions.get(highlight.id);
+    if (typeof position !== 'number') { return; }
+
+    const topOffset = getTopOffsetForHighlight(highlight);
+
```
janKozie1, line 44: What about instead of calculating the top offset, make the container a grid and here you could pass index of the highlight and set `transformY` to `highlightIndex / allHighlghts.length * containerHeight`

janKozie1 (commented):
It looks good, i left some more comments. 
Is it expected that i need to click twice on a card for it to animate if there is a different card that is already focused?

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: Not sure if I understand correctly. `topOffset` is a value for topOffset of a <span> for highlight inside a text and `position` is topOffset for a card with a note (on the right side).
`containerHeight` is not constant because some notes may be longer and then their `position`s will be updated accordingly

PiotrKozlowski:
@janKozie1 
> Is it expected that i need to click twice on a card for it to animate if there is a different card that is already focused?

No. It's a bug caused by `useOnClickOutside` which is calling `onBlur` after new highlight is focused.

PiotrKozlowski:
@janKozie1 I think I've resolved all of your comments and fixed one more bug

janKozie1:
The changes look good. 
I'm not sure what can be done about the scroll jumping when focusing a highlight on page and not through cards. There is `isHtmlElementWithHighlight` that could potentially be used with `onClickOutside` to prevent blur (and let highlightigh manager focus the new one), but it would probably require more changes.

Also, if text that isnt a highlight gets selected its edit box doesnt get scrolled. In bad cases it could result in it appearing off screen. Maybe you could use previous highlight's position? Its far from a good solution but im not sure what else could be done.

PiotrKozlowski:
@janKozie1 
> I'm not sure what can be done about the scroll jumping when focusing a highlight on page and not through cards.

I think I'm not seeing any "jumping", can your provide some video?

> Also, if text that isnt a highlight gets selected its edit box doesnt get scrolled. In bad cases it could result in it appearing off screen. Maybe you could use previous highlight's position? Its far from a good solution but im not sure what else could be done.

After some investiagtion: This is happening because `EditCard` is calling `onHeightChange` and `onFocus`, then `cardPositions` is updated but state update is async so it calls `onFocus` first.
I'm trying to find a solution for this

TomWoodward:
i got an error trying to reproduce what @janKozie1 said. had a note selected, then selected some random text
image


the animations are awesome though 👍

PiotrKozlowski:
@TomWoodward On pr with tests I've removed one statement but it seems that due to how state updates are happening `position` can be `undefined` for pending highlights.
I've pushed a change and now it works correctly.

janKozie1:
@PiotrKozlowski this is what i've meant https://drive.google.com/file/d/1rO7TAkwIxVKwC26tjzvdUQxx6UYxdnav/view?usp=sharing

PiotrKozlowski:
@janKozie1 I've pushed a change which hopefully fixes your issue. Could you please check?

src/app/content/highlights/components/CardWrapper.spec.tsx
```diff
@@ -19,8 +40,140 @@ describe('CardWrapper', () => {
...
+    // Update positions - currently all cards are in the same position
+    // after onHeightChange their positions will be recalculated
+    renderer.act(() => {
+      component.root.findAllByType(Card)
+        .forEach((card) => card.props.onHeightChange({ current: { offsetHeight: 100 }}));
+    });
+
+    // Position of last card is now 340px - all highlights are positioned 100px from the top
+    // Which mean that position of first card is at 100px offsetTop. Since all cards have 100px height
+    // and there is 20px margin between cards, we end up we having 340px offsetTop for third card.
```
janKozie1, line 124: The second `we` is unnecessary

janKozie1 (changes requested):
Just minor things, mainly about tests

src/app/content/components/Page/highlightManager.spec.tsx
```diff
@@ -59,15 +64,25 @@ describe('highlightManager', () => {
...
+    expect(() => component.root.findByType(CardWrapper.type)).not.toThrow();
   });
 
   it('CardList is rendered after update', () => {
     const {CardList, update} = highlightManager(element, () => prop);
+    const mockHighlight = { id: '123', setStyle: jest.fn() } as unknown as HighlightData;
+    prop.highlights = [mockHighlight];
+    Highlighter.mock.instances[0].getHighlight
+      .mockReturnValueOnce(mockHighlight)
+      .mockReturnValueOnce(mockHighlight)
```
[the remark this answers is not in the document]
PiotrKozlowski: Some leftovers after testing different things. I've just removed this mocks

janKozie1 (approved):

janKozie1:
All bugs found on ubuntu, chrome

1. https://drive.google.com/open?id=1AuPQnjzGsyd2FqldfAGdXr77CURGQUz4
2. https://drive.google.com/open?id=1k6Q_On-P8Aa1yaP0xZhvjZ3O0dmVJZgi
3. https://drive.google.com/open?id=1ez0DNk6-7ljnUFwRvnrFH1wpWn2MS3pW
4. https://drive.google.com/open?id=1plYiZKQiw_w5eq4ha95tkB1fwTAIZBQe
5. https://drive.google.com/open?id=1LSJE95WdxezrwD7dQBIqI5l69BTK34kd

They are all related to the cards not positioning properly + discard modal not showing.
I'm not sure what should be done about 5. Before this PR, the edit card modal stays in its 
position if it has any input. 

cc @PiotrKozlowski

janKozie1 (changes requested):

src/app/content/highlights/components/Card.tsx
```diff
@@ -310,15 +159,13 @@ export default connect(
     data: selectHighlights.highlights(state).find((search) => search.id === ownProps.highlight.id),
     hasQuery: !!selectSearch.query(state),
     isFocused: selectHighlights.focused(state) === ownProps.highlight.id,
-    isOpen: contentSelect.tocOpen(state),
-    loginLink: selectAuth.loginLink(state),
-    user: selectAuth.user(state),
+    isTocOpen: contentSelect.tocOpen(state),
```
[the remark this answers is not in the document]
PiotrKozlowski: yes, in the styles

src/app/content/highlights/components/Card.tsx
```diff
@@ -83,16 +70,30 @@ const Card = (props: Props) => {
...
     return null;
   }
 
-  const location = getHighlightLocationFilterForPage(locationFilters, page);
-  if (!location) { return null; }
-
-  const locationFilterId = stripIdVersion(location.id);
+  const handleClickOnCard = async(e: React.MouseEvent) => {
+    if (!props.isFocused && (!hasUnsavedHighlight || await showConfirmation())) {
+      e.preventDefault();
```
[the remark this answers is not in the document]
PiotrKozlowski: It was added here: https://github.com/openstax/rex-web/commit/60aefe2a5c46b31630fe8cefbdbc83c33464b4e5but it might not be required anymore due to change how resetting top offset for card wrapper works now

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,139 @@
...
+        getHighlightTopOffset(container, highlight),
+        `Couldn't get top offset for highlight with an id: ${highlight.id}`
+      );
+      setTopOffsets((state) => new Map(state).set(highlight.id, offset));
+      return offset;
+    }
+  };
+
+  const getCardsPositions = React.useCallback(() => {
+    const newPositions: Map<string, number> = new Map();
```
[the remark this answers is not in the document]
PiotrKozlowski: It was used before in `const position = cardsPositions.get(focusedHighlight.id);`
instead of `cardsPositions` I was using `getCardsPositions` but it seems I've changed it at some point. I'll check what is better

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -0,0 +1,213 @@
...
+    display: none;
+  `}
+`;
+
+const rightSideDisplay = css`
+  left: calc(100% - ((100% - ${contentTextWidth}rem) / 2) + ${cardContentMargin}rem);
+  right: unset;
+  top: ${(props: CardProps) => `${props.topOffset || getHighlightBottomOffset(props.container, props.highlight)}px;`}
+  ${(props: CardProps) => !!props.isFocused && css`
+    left: calc(100% - ((100% - ${contentTextWidth}rem) / 2) + ${cardFocusedContentMargin}rem);
```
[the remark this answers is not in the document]
PiotrKozlowski: Animating these through `transform` is a nightmare for me, so unless is necessary I would rather keep it unchanged

src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -78,7 +79,7 @@ const DisplayNote = React.forwardRef<HTMLElement, DisplayNoteProps>((
     </Dropdown>
     <CloseIcon onClick={onBlur} />
     <label>Note:</label>
-    <TruncatedText text={note} isFocused={isFocused} />
+    <TruncatedText text={note} isFocused={isFocused} onChange={() => setTextToggle(!textToggle)} />
```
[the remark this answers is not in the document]
PiotrKozlowski: In this case, it doesn't matter because it is the only function which is modifying this state, but overall it's good practc to use previous state, I'll change this

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -30,8 +29,11 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
   const [cardsPositions, setCardsPositions] = React.useState<Map<string, number>>(new Map());
   const [cardsHeights, setCardsHeights] = React.useState<Map<string, number>>(new Map());
   const [topOffsets, setTopOffsets] = React.useState<Map<string, number>>(new Map());
-  const [focusedHighlight, setFocusedHighlight] = React.useState<Highlight | null>(null);
   const [width] = useDebouncedWindowSize();
+  const focusedId = useSelector(focused);
+  const focusedHighlight = React.useMemo(
+    () => highlights.find((highlight) => highlight.id === focusedId),
+    [focusedId, highlights]);
```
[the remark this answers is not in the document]
PiotrKozlowski: These `highlights` come from `highlightManager` so they are different type than highlights in the redux state

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -1,30 +1,140 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [focusedHighlight, cardsPositions]);
+
+  return highlights.length
+    ? <div className={className} ref={element}>
+      {highlights.map((highlight, index) => {
+        // Rerender cards on window resize to trigger their height calculations.
+        // It's better to do it here than using hooks like useDebouncedWindowSize for each Card.
+        const key = generateHighlightKey(highlight, width, hasQuery, isTocOpen);
```
janKozie1, line 117: Won't this make the extra logic run on mobile as well? Maybe you want to pick the generated key or just highlight.id depending on the width?

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  React.useLayoutEffect(() => {
+    const updateSize = () => {
+      clearTimeout(timeout.current);
+      timeout.current = setTimeout(() => {
+        setSize([window.innerWidth, window.innerHeight]);
+      }, 50);
+    };
+    window.addEventListener('resize', updateSize);
+    updateSize();
```
janKozie1, line 20: Maybe initialize the state with current width and height instead of calling the `updateSize` manually?

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    return () => window.removeEventListener('resize', updateSize);
```
janKozie1, line 21: I think you should also clear the timeout in the clean up function.

janKozie1 (approved):
It looks ok to me.
I'm not sure about using the key to trigger rerender like that thought.

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: I think this cleanup function is called only when a component unmounts which mean that this state should be cleared anyway, right?

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: There was an issue when a user has screen width set to some value on which cards are hidden by default, but when he closes the ToC then all cards are displayed on the right side.
This means that cards have `0` height at the beginning, and then when they are displayed they have to update their heights by calling `onHeightChange` - it is not being called automatically because cards were already rendered but they were hidden using CSS.

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: oh, wait, will it still try to call `setSize` after unmounting?
This will probably never cause an error because it would be hard to unmount this component while resizing window, but still, it may happen, nice catch Jan ;)

janKozie1: It will remove the listener, but the timeout might still be running -> attempt at updating the state of an unmounted component, (even though the state is theoretically in a hook) which gives the `warning: can't perform state update on an unmounted component'.

janKozie1 (commented):
Just minor things

TomWoodward (approved):
i have not read this recently, but the functionality looks good.

i see one issue where the cards flicker while resizing the window, but i think we will probably want to address that in follow up

PR #485, opened 2020-02-06
Show discard modal 

janKozie1:
Well i thought about something like this. The confirmation is just a function, but I'm not sure if it's okay to add a new dom node to render the modal in.

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/746


src/app/content/components/Page/highlightManager.ts
```diff
@@ -36,10 +38,14 @@ export type HighlightProp = ReturnType<typeof mapStateToHighlightProp>
   & ReturnType<typeof mapDispatchToHighlightProp>;
 
 // deferred so any cards that are going to blur themselves will have done so before this is processed
-const onClickHighlight = (services: Services, highlight: Highlight | undefined) => defer(() => {
-  if (!highlight || services.getProp().focused) {
+const onClickHighlight = (services: Services, highlight: Highlight | undefined) => defer(async() => {
+  if (!highlight || services.getProp().focused === highlight.id) {
     return;
   }
+  if (services.getProp().focused && services.getProp().hasUnsavedHighlight) {
```
TomWoodward:
```suggestion
  if (services.getProp().focused && services.getProp().hasUnsavedHighlight && await showConfirmation()) {
```

src/app/content/components/Page/highlightManager.ts
```diff
@@ -49,11 +55,16 @@ const onSelectHighlight = (
...
-) => defer(() => {
-  if (highlights.length > 0 || !highlight || services.getProp().focused) {
+) => defer(async() => {
+  if (highlights.length > 0 || !highlight) {
     return;
   }
 
+  if (services.getProp().focused && services.getProp().hasUnsavedHighlight) {
+    const didConfirm = await showConfirmation();
+    return didConfirm && services.getProp().clearFocus();
```
TomWoodward: don't you want to return if they _don't_ confirm, and just let it do the stuff below if they do?

src/app/content/highlights/actions.ts
```diff
@@ -23,6 +23,8 @@ export const updateHighlight = createStandardAction('Content/Highlights/update')
   locationFilterId: string,
   pageId: string,
 }>();
+export const editStateChange = createStandardAction('Content/Highlights/editStateChange')<boolean>();
```
TomWoodward:
```suggestion
export const setAnnotationChangesPending = createStandardAction('Content/Highlights/setAnnotationChangesPending')<boolean>();
```

src/app/content/highlights/components/DiscardModal.tsx
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import styled from 'styled-components/macro';
+import Button from '../../../components/Button';
+import theme from '../../../theme';
+import DiscardCard, { Footer } from './DiscardCard';
+
+// tslint:disable-next-line:variable-name
+const Mask = styled.div`
```
TomWoodward: did you get any design guidance from anyone for this stuff?

src/app/content/highlights/components/DiscardModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+  onAnswer: (answer: boolean) => void;
+}
+
+// tslint:disable-next-line:variable-name
+const DiscardModal = ({ onAnswer }: PropTypes) => {
+  const confirm = () => onAnswer(true);
+  const deny = () => onAnswer(false);
+
+  return (
+    <Modal className='discard-changes'>
```
TomWoodward: class name is unused?

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -43,18 +45,20 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
...
+    onEditStateChange,
     onRemove,
     onSave,
   }: Props,
   ref
 ) => {
   const defaultAnnotation = () => data && data.annotation ? data.annotation : '';
   const [pendingAnnotation, setPendingAnnotation] = React.useState<string>(defaultAnnotation());
-  const [editingAnnotation, setEditing] = React.useState<boolean>(!!data && !!data.annotation);
+  const [editingAnnotation, setEditing] = React.useState<boolean>(false);
```
TomWoodward: you don't want to change this

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -68,6 +72,13 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
       onBlur();
     }
   };
+ 
+  React.useEffect(() => {
+    if (!isFocused && hasUnsavedHighlight) {
```
TomWoodward: why do you need `hasUnsavedHighlight` here? can't you just reset everything whenever you lose focus?

also you can probably call `cancelEditing` instead of duplicating this

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -109,6 +120,20 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
     onCancel();
   };
 
+  const updateUnsavedHighlightStatus = (newValue: string) => {
+    if (!data) {
+      return;
+    }
+
+    if (data.annotation !== newValue && !hasUnsavedHighlight) {
```
TomWoodward: you're gonna hit that undefined/empty string situation here too. might be nice to extract `const currentValue = data.annotation || ''` at the top of the logic

src/app/content/highlights/hooks/preventHighlightFromDiscarding.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Middleware } from '../../../types';
+import { clearFocusedHighlight } from '../actions';
+import showConfirmation from '../components/utils/showConfirmation';
+import * as select from '../selectors';
+
+const actionsToIntercept = new Set<AnyAction['type']>([
+    callHistoryMethod,
+].map(getType));
+
+export default (): Middleware => ({dispatch, getState}) => (next: Dispatch) => async(action: AnyAction) => {
```
TomWoodward: you can't do this will action hooks, and while with redux middleware you can stop the app from responding to history events, you can't stop the browser for processing them, so that would result in some really weird state.

i was thinking that we would handle navigation events in the links such as the ContentLink component and the contentLinkHandler helper. this doesn't help with the back button, but stopping the back button from functioning is a thing that browsers **really really** don't want you to do, so lets not worry about it for now

TomWoodward (changes requested):

src/app/content/highlights/components/DiscardModal.tsx
janKozie1: I only assumed that since the text format matches the error modal, it should look the same way, my bad.
https://openstax.slack.com/archives/C8V1XET1R/p1581460059190200

src/app/content/components/Page/highlightManager.ts
```diff
@@ -49,8 +55,12 @@ const onSelectHighlight = (
...
   highlight: Highlight | undefined
-) => defer(() => {
-  if (highlights.length > 0 || !highlight || services.getProp().focused) {
+) => defer(async() => {
+  if (highlights.length > 0 || !highlight) {
+    return;
+  }
+
+  if (services.getProp().focused && services.getProp().hasUnsavedHighlight && !await showConfirmation()) {
     return;
```
janKozie1: I thought that it might be a good idea to add `assertWindow().getSelection().removeAllRangers()` here but I'm having trouble testing it. I'm assuming it's because of 
`  afterEach(() => {
    delete window.document.getSelection;
  });`?

src/app/content/highlights/reducer.ts
```diff
@@ -107,6 +112,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
 
       return {
         ...state,
+        hasUnsavedHighlight: onlyColorChanged ? true : false,
```
PiotrKozlowski: `hasUnsavedHighlight: Boolean(onlyColorChanged)`

src/app/content/components/Page/highlightManager.ts
```diff
@@ -5,11 +5,13 @@ import flow from 'lodash/fp/flow';
 import React from 'react';
 import { isDefined } from '../../../guards';
 import { AppState, Dispatch } from '../../../types';
+import { assertWindow } from '../../../utils';
```
PiotrKozlowski: it is not used

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -37,39 +41,47 @@ const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageRefere
...
-  const href = anchor.getAttribute('href');
+export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => ContentLinkProp) =>
+  async(e: MouseEvent) => {
+    const {references, navigate, book, page, locationState, currentPath, hasUnsavedHighlight} = getProps();
+    const href = anchor.getAttribute('href');
 
-  if (!href || !book || !page || e.metaKey) {
-    return;
-  }
+    if (!href || !book || !page || isModifiedClick(e)) {
```
PiotrKozlowski: those are called "modifier keys" so probably something like `isModifierKey(e)` will suit more

PiotrKozlowski:
Some lint errors:
```
./src/app/content/highlights/components/EditCard.tsx
  Line 82:    The 'cancelEditing' function makes the dependencies of useEffect Hook (at line 92) change on every render. To fix this, wrap the 'cancelEditing' definition into its own useCallback() Hook  react-hooks/exhaustive-deps
  Line 145:  Unexpected mix of '&&' and '||'                                                                                                                                                              no-mixed-operators
  Line 145:  Unexpected mix of '&&' and '||'                                                                                                                                                              no-mixed-operators
```

janKozie1:
Not sure if it's better to `useCallback` in those two places or just disable the rule.

TomWoodward (changes requested):
@janKozie1 i saw in the conversation that you copied these modal designs from the error popup, but it doesn't look like you re-used any of the react components. if this popup and the error popup have the same design, they should really be using some shared react components

janKozie1:
Is making a generic confirmation modal like https://github.com/openstax/rex-web/pull/485/commits/282694f3fa1edf0fb7f6f94e0d85e1f20409874a a good idea? I could go with that or just share the styled components between error and discard modal. @TomWoodward

TomWoodward:
@janKozie1 a generic confirmation modal is a good idea, but even in that case it should share styles with the error modal, since they look basically the same

janKozie1:
I've meant that I want to make the error modal use the generic confirmation modal inside it instead @TomWoodward

TomWoodward:
@janKozie1 yea that sounds, good, just don't call it a confirmation modal in that case. just call it `modal`

src/app/content/components/ContentLink.spec.tsx
```diff
@@ -130,4 +133,22 @@ describe('ContentLink', () => {
...
+
+    const event = {
+      preventDefault: jest.fn(),
+    };
+
+    component.root.findByType('a').props.onClick(event);
+
+    expect(dispatch).not.toHaveBeenCalledWith(push(expect.anything()));
+    expect(event.preventDefault).toHaveBeenCalled();
+    expect(clickSpy).not.toHaveBeenCalled();
```
TomWoodward: these aren't testing anything. `onClick` is an async function and you're not waiting for it to complete, so the functions wouldn't be called either way.

you should make a corresponding test where the confirmation is displayed but you do confirm it, and you'll see the issue.

src/app/content/components/Page.spec.tsx
```diff
@@ -566,6 +568,25 @@ describe('Page', () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
+  it('noops if user decides not to discard', async() => {
```
TomWoodward: you should add this in the `src/app/content/components/Page/contentLinkHandler.spec.ts`, i'm trying not to add things to the page spec that could go in a more specific place because this spec file is so big

src/app/content/components/ContentLink.tsx
```diff
@@ -49,14 +56,18 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
 
   return <a
     ref={myForwardedRef}
-    onClick={(e) => {
-      const { metaKey, altKey, ctrlKey, shiftKey } = e;
+    onClick={async(e) => {
 
-      if (metaKey || altKey || ctrlKey || shiftKey) {
+      if (isClickWithModifierKeys(e)) {
```
TomWoodward: nice extraction 👍

src/app/content/components/Page/highlightManager.ts
```diff
@@ -49,8 +54,12 @@ const onSelectHighlight = (
...
   highlights: Highlight[],
   highlight: Highlight | undefined
-) => defer(() => {
-  if (highlights.length > 0 || !highlight || services.getProp().focused) {
+) => defer(async() => {
+  if (highlights.length > 0 || !highlight) {
+    return;
+  }
+
+  if (services.getProp().focused && services.getProp().hasUnsavedHighlight && !await showConfirmation()) {
```
TomWoodward: do you really need the `services.getProp().focused && ` part of this? does it not get cleared when focus is lost?

src/app/content/components/Page/highlightManager.ts
TomWoodward: thats a good idea, go ahead and add it. i can help with testing if necessary. i think the selection object is mocked so it might be a case of needing to add that `removeAllRangers` to the mock in order to test that it is called (and not have the code fail when it is not there)

src/app/content/components/utils/isClickWithModifierKeys.ts
```diff
@@ -0,0 +1,3 @@
+import { MouseEvent } from '@openstax/types/lib.dom';
+
+export default  (e: React.MouseEvent | MouseEvent) => e.shiftKey || e.ctrlKey || e.metaKey || e.altKey;
```
TomWoodward: this could go in `app/domUtils.ts`, unless that file is really big already

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -71,6 +79,18 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
     }
   };
 
+  const cancelEditing = React.useCallback(() => {
```
TomWoodward: don't put useCallback here, we don't need it to be memoized, that seems like its asking for trouble

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -71,6 +79,18 @@ const EditCard = React.forwardRef<HTMLElement, Props>((
...
+    setPendingAnnotation(defaultAnnotation());
+    setEditing(false);
+    onCancel();
+  }, [defaultAnnotation, onCancel]);
+
+  React.useEffect(() => {
+    if (!isFocused) {
+      cancelEditing();
+    }
+  }, [isFocused, cancelEditing]);
```
TomWoodward: don't provide `cancelEditing` as a dependency. if the linter complains disable it.

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -0,0 +1,30 @@
...
+import ReactDOM from 'react-dom';
+import MessageProvider from '../../../../MessageProvider';
+import { assertDocument, assertNotNull } from '../../../../utils';
+import ConfirmationModal from '../ConfirmationModal';
+
+export default async() => {
+  const document = assertDocument();
+  const domNode = document.createElement('div');
+
+  domNode.id = `dialog-${Math.random().toString(36).substring(7)}`;
```
TomWoodward: the uuid library is already a dependency through the highlighter, you can add it to use here

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const root = assertNotNull(document.getElementById('root'), 'root element not found');
+  root.insertAdjacentElement('afterend', domNode);
+
+  const userChoice: boolean = await new Promise((resolve) => {
+    const confirm = () => resolve(true);
+    const deny = () => resolve(false);
+    ReactDOM.render(
+      <MessageProvider>
+        <ConfirmationModal {...{deny, confirm}} />
```
TomWoodward: you can just pass these, using object destructuring just to avoid typing 9 characters seems like unnecessary obfuscation

src/app/content/highlights/reducer.ts
```diff
@@ -83,6 +85,9 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
       if (oldHiglightIndex < 0) { return state; }
 
       const oldHighlight = state.highlights[oldHiglightIndex];
+      const onlyColorChanged =
```
TomWoodward: this variable name would make sense if you weren't tacking on the current value of `hasUnsavedHighlight`, but as written maybe you should just call it `hasUnsavedHighlight`

src/app/errors/components/ErrorModal.tsx
```diff
@@ -49,26 +31,25 @@ const ErrorModal = ({ error, clearError }: PropTypes) => {
...
-          clearError={clearError}
-        />
-      </CardWrapper>
-      <Mask />
-    </Modal>
+    <Modal
+      className='error-modal'
+      headerTextId='i18n:error:boundary:heading'
+      bodyTextId='i18n:error:boundary:sub-heading'
+      body={
```
TomWoodward: i'd put the body as the children

src/app/components/Modal.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import * as Styled from './ModalStyles';
+
+interface PropTypes {
+  onModalClose: () => void;
+  footer: JSX.Element | null;
+  body: JSX.Element | null;
+  headerTextId: string;
+  bodyTextId: string;
```
TomWoodward: call this what it is. `subheading` or `bodyheading`

src/app/messages/en/messages.json
```diff
@@ -31,11 +35,13 @@
   "i18n:highlighting:login:link": "Log in",
   "i18n:highlighting:button:cancel": "Cancel",
   "i18n:highlighting:button:delete": "Delete",
+  "i18n:highlighting:button:discard-changes": "Discard changes",
```
TomWoodward: unsued?

src/app/messages/en/messages.json
```diff
@@ -31,11 +35,13 @@
...
   "i18n:highlighting:button:save": "Save",
   "i18n:highlighting:card:placeholder": "Add a note...",
   "i18n:highlighting:card:show-more": "show more",
   "i18n:highlighting:confirmation:delete-both": "Are you sure you want to delete this note and highlight?",
   "i18n:highlighting:confirmation:delete-note": "Are you sure you want to delete this note?",
+  "i18n:highlighting:confirmation:discard-changes": "Discard unsaved changes? You have an unsaved note on this page.",
```
TomWoodward: unused?

TomWoodward (changes requested):
there are a lot of tests for `'noops if user decides not to discard changes'` but none for `'continues if prompted to discard changes and discards'`, if you add a condition like this you should test both sides of it

src/app/errors/components/ErrorCard.spec.tsx
```diff
@@ -1,14 +0,0 @@
-import React from 'react';
-import renderer from 'react-test-renderer';
-import MessageProvider from '../../MessageProvider';
-import ErrorCard from './ErrorCard';
-
-describe('ErrorCard', () => {
-
-  it('matches snapshot', () => {
-    const tree = renderer
-      .create(<MessageProvider><ErrorCard /></MessageProvider>)
```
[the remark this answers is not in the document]
janKozie1: I'm not sure what do you mean. ErrorCard doesn't exist anymore, how can there be a spec for it? Should i split `Modal` into `Modal` and `Card`?

src/app/content/components/Page/contentLinkHandler.spec.ts
```diff
@@ -1,116 +1,183 @@
...
+      hasUnsavedHighlight: false,
       locationState: {} as any,
       navigate: jest.fn(),
       page,
       references: [],
     };
+  });
 
-    handler = contentLinkHandler(anchor, () => prop);
+  describe('without unsaved highlight', () => {
```
janKozie1: Those two tests here fail but say that there is no visual difference between expected and what was the function called with. I'm not exactly sure why does that make it fail.

I was able write the test for `.removeAllRanges()` but I'm struggling a lot with mocking the returned value from `showConfirmation` in tests for highlight manager so that there is a test for both confirm/deny scenarios.

src/app/errors/components/ErrorCard.spec.tsx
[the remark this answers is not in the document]
TomWoodward: no i was looking at the wrong thing. i should have been looking at `src/app/errors/components/__snapshots__/ErrorModal.spec.tsx.snap`. because you're refactoring the modal components that snapshot is useful because you've totally replaced the implementation of the component but it should still be the same in the snapshot. looking there now, i think everything looks good

src/app/components/Modal/styles.tsx
```diff
@@ -69,9 +58,39 @@ const Body = styled.div`
...
+
+// tslint:disable-next-line:variable-name
+export const CardWrapper = styled.div`
+  z-index: 1;
+`;
+
 // tslint:disable-next-line:variable-name
 export const Footer = styled.div`
   display: flex;
+  justify-content: space-between;
```
TomWoodward: i worry about the assumptions you're making using this to arrange your buttons, especially since in reality someone implementing this component can pass arbitrary content with any number of children to the footer. for instance with 3 buttons this would look super weird. i think we should either:

- take this out, and just let multiple buttons be next to each other
- designate separate ways to specify buttons that go on the left vs on the right

TomWoodward (changes requested):
https://github.com/openstax/rex-web/pull/485#pullrequestreview-360545983

janKozie1:
Maybe something like this? @TomWoodward

src/app/content/highlights/components/ConfirmationModal.tsx
```diff
@@ -1,28 +1,34 @@
...
 }
 
 // tslint:disable-next-line:variable-name
-const ConfirmationModal = ({deny, confirm}: Props) => <Modal
-  onModalClose={deny}
-  heading='i18n:discard:heading'
-  subheading='i18n:discard:body'
-  footer={
-    <>
+const Footer = styled.div`
```
TomWoodward: define and export footer from `../../../components/Modal/styles'` with the flex and padding, you can extend it here with the justify content by doing 

```
const ConfirmationFooter = styled(Footer)`
  justify-content: space-between;
`;
```

src/app/components/Modal/index.tsx
```diff
@@ -40,9 +38,6 @@ const Modal = ({
             </FormattedMessage>
             {children}
           </Styled.Body>
```
TomWoodward: i like adding the footer as a child of the modal, but putting it inside the body is going to make styling the body annoying. it would be good to also pull out the body of the modal into its own child, like this:

```
<Modal onModalClose={deny} heading='i18n:discard:heading'>
  <Body heading='i18n:discard:body'>this is the body</Body>
  <Footer>cool footer</Footer>
</Modal>;
```

TomWoodward (changes requested):

janKozie1 (commented):
I've tried few things about the tests that are failing despite no visual difference, it seems like the route param is making them fail. I'm not sure why since the import is not mocked.
`route: routes.content`

TomWoodward:
one of your mocks is probably breaking the strict equality. you might have to be better about resetting them, or something else. i can take a look later today or tomorrow

TomWoodward:
the problem is that `routes.content` is a different reference in your tests than in the code you're testing. the order of operations is this:

1) test file includes are executed (including routes.content)
1) beforeEach is run, deleting module cache of things that just got imported by the test file
1) inner beforeEach is run, requiring contentLinkHandler
1) contentLinkHandler includes are executing (there is no longer a cache, so this creates new modules)
1) stuff happens
1) when expect is processed, the routes.content reference from the test file is not the same as the routes.content reference from the code file, so expectation fails.

in order to cope with this you need to either:
- find a way to not have `resetModules` in your before, but it might be required to make some of your mocks work
- require routes in the before, after calling `resetModules` instead of importing it

src/app/content/highlights/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+    </Body>
+    <ConfirmationFooter>
+      <FormattedMessage id='i18n:discard:button:discard'>
+          {(msg) => <Button
+            data-testid='discard-changes'
+            onClick={confirm}
+            variant='primary'
+            > {msg}
+          </Button>}
+        </FormattedMessage>
```
TomWoodward: indentation is weird

TomWoodward (changes requested):
this is looking good, please merge master and resolve conflicts

src/app/content/highlights/components/Card.tsx
```diff
@@ -127,8 +135,10 @@ const Card = (props: Props) => {
     loginLink={props.loginLink}
     highlight={props.highlight}
     locationFilterId={locationFilterId}
+    hasUnsavedHighlight={hasUnsavedHighlight}
```
PiotrKozlowski: I think it would be easier to maintain if EditCard would use own selector for obtaining this data

src/app/content/highlights/reducer.ts
```diff
@@ -76,6 +78,10 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
         return state;
       }
 
+      const hasUnsavedHighlight =
+        oldHighlight.annotation === action.payload.highlight.annotation
+        && state.hasUnsavedHighlight;
```
PiotrKozlowski: Value for `state.hasUnsavedHighlight` is determined by current value `state.hasUnsavedHighlight`. It is misleading for me, why is that?

PiotrKozlowski (commented):
Code looks good to me, I've left to minor comments

src/app/content/highlights/reducer.ts
janKozie1: Well it was `onlyColorHasChanged` before. This action triggers when either note is saved or color is changed hence why it needs to carry over. When color is changed, the action does not have the new annotation.

src/app/content/highlights/components/Card.tsx
janKozie1: I agree. `loginLink` and `user` are also only passed through the `Note` into the `EditNote`, so i assumed that there is a reason why `EditNote` is not being directly connected to the store.

PiotrKozlowski: I don't think there is any reason for that, maybe @TomWoodward knows?

TomWoodward: there isn't a particularly good reason, you're free to move selectors down to subcomponents if they're only used there

TomWoodward (approved):

janKozie1:
@TomWoodward I've done https://openstax.slack.com/archives/C8V1XET1R/p1583525039205600?thread_ts=1583516222.202800&cid=C8V1XET1R (about showing discard modal after clicking on the link to book details). 

Since the link is different from content link, i wasn't sure if there is a "good" way of navigating the user in case they choose to discard, but on the other hand it would be nice to keep things consistant.

TomWoodward:
i don't like the behavior in 4cd1bd80deef36ada0192b73e0fe891d007d7b85 where it is stopping new confirmations if a modal is already open, that works for now, but might be a problem for another use case. i think we should do one of the options from here https://openstax.slack.com/archives/C8V1XET1R/p1584725048477600 to change our click handling so it just doesn't try to run both confirmations at the same time @janKozie1

TomWoodward (changes requested):
@janKozie1 i don't see where you addressed https://github.com/openstax/rex-web/pull/485#issuecomment-601821240

janKozie1:
@TomWoodward I've reversed the changes from 4cd1bd80deef36ada0192b73e0fe891d007d7b85. I was trying to implement the "click once to focus, click second time to navigate" but I've ran out of ideas. Originally i thought that Highlighter attaches events to each span so i thought that the `stopPropagation` will be enough, but that's not the case because event first reaches the link then the onClick of HighlightManager.
Besides that, I've tried few approaches for checking if a link has a highlight inside it, then checking which highlight is currently focused & deciding to navigate based on those two things, but I think its failing because `EditCard` will blur the highlight, then in contentLinkHandler it will be undefined unless `defer`.
I've tried making some sketchy counters but that breaks when user clicks on the highlight then on the link which should navigate them, but it waits for the second click on the link.

So the next option you mentioned was `we can either keep the current behavior, but clean up some of that janky stuff that is happening, ` but I'm not exactly sure what does it mean. Stop highlights from being focused if navigation is happening?

janKozie1:
So I've been digging more and I'm pretty sure the issue boils down to 
given that a link inside a highlight is clicked:
1. Highlight gets blurred on mousedown due to `onClickOutside`
2. `contentLinkHandler` on anchor runs, but state has been updated so for it the highlight is never focused
3. Event propagates to the page node where Highlighter captures it and refocuses the highlight

First solution that comes to my mind is changing `onClickOutside` to run `onClick` instead of `mouseDown`, which I think should work.
Second one, making the `contentLinkHandler` somehow aware of the previously clicked highlight.

First one sounds better to me but i don't know how important it is for `onClickOutside` to run on mousedown

I haven't tested it yet, but i think it might be nice to change Highlighter to run on click because that way we can stop the event from propagating on second link click, which will prevent the highlight from re-focusing when navigation is happening

janKozie1:
So the first approach seems to be working fine. Changing the event in highlighter seems to be break tests in a lot of places. The `stopPropagation` prevents the `onClickOutside` firing when the navigation is going to happen.

TomWoodward:
@janKozie1 i read your comments but i'm not sure what the current state of this is. are you still working on it? or do you think you found an approach that works well now?

janKozie1:
@TomWoodward I think this one is working fine. Just wanted to know if its okay to change `onClickOutisde` to listen to click event rather than `mousedown` really.

TomWoodward:
@janKozie1 if you tested clicking on various things, especially other highlights to switch the focus from one to another, its probably fine.

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -37,39 +43,65 @@ const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageRefere
...
-      route: content,
-      state: {
-        ...locationState,
-        ...getBookPageUrlAndParams(book, page).state,
 
-      },
-    }, {hash, search}));
-  }
-};
+    const containedHighlight = getContainedHighlightNode(anchor);
```
TomWoodward: its possible that the user highlights part of a link, and they then click the link but not the highlight. what we want to check for is if the user actually clicked the highlight and in that case not process the link.

i had thought doing the stopPropagation from the highlight click handler would be good, but i see now that the highlighter event listener is on the container, while the link listeners are on the links, so the links get evaluated first.

first of all, this logic doesn't do well in the case that a highlight does not span the entire link, its possible to click on a link that has a highlight in it, but not actually click on the highlight, in the code that is written here, those clicks would just do nothing. similarly, its possible to have _two_ separate highlights in a link, and with this code you would never be able to click the link, because one of them would always be unfocused.

i'm not loving the way the link handler is aware of the highlights logic, but i don't really have a better idea right now. maybe the best we can do for now is to rewrite your `getContainedHighlightNode` to accept `event.target` and see if that is inside a highlight, and if its not focused, not handle the link

TomWoodward (changes requested):

janKozie1:
This seems to be working well too. I tested it with two highlight in one link too. If you approve of this changes, it can can be moved. Not sure if Malar should review it again? @TomWoodward

TomWoodward (approved):
please fix conflicts @janKozie1

TomWoodward:
@janKozie1 this has been approved by QA but has conflicts again

janKozie1:
@TomWoodward i've resolved the conflicts

src/app/content/components/BookBanner.tsx
```diff
@@ -239,7 +262,14 @@ export class BookBanner extends Component<PropTypes, {scrollTransition: boolean}
       data-analytics-region='book-banner-collapsed'
     >
       <TopBar>
-        <BookTitle href={bookUrl} variant='mini' colorSchema={book.theme}>
```
janKozie1, line 242: .

PR #486, opened 2020-02-06
Rename route params book.book to book.slug

janKozie1:
for: openstax/unified#901


TomWoodward:
@janKozie1 i was having a hard time explaining this in a comment so i pushed a commit, what do you think about trying to make the params look like this https://github.com/openstax/rex-web/pull/486/commits/93bb837373da27b77d7c42d5910c351aaefd0a05 ?

TomWoodward (changes requested):
https://github.com/openstax/rex-web/pull/486#issuecomment-584347853

janKozie1:
I think it's better this way. Is there a reason why `UuidParams` were changed to `VersionedUuidParams`?

TomWoodward:
@janKozie1 better which way? the change of `UuidParams` to `VersionedUuidParams` isn't totally necessary right now, but i had been thinking about allowing the use of the page uuid sometime in the future and it would prepare for that a little

janKozie1:
Seemed more logical to me. But i'm not sure how to implement it with `pathToRegexp`.
This will require changes as well @TomWoodward 
```js
const getMatchParams = (keys: Key[], match: RegExpExecArray) => {
  const [, ...values] = match;
  return keys.reduce((result, key, index) => {
    const value = values[index] ? decodeURIComponent(values[index]) : values[index];
    return {...result, [key.name]: value};
  }, {});
};
```

TomWoodward:
yea since pathToRegex expects all params to be on one level you could use a library like this https://www.npmjs.com/package/dot-object to convert between parameters like 
```
{
  book: {
    slug: 'asdf'
  }
}
```

to 

```
{
  "book.slug": "asdf"
}
```

and vice versa, that way according to `pathToRegex` the param would be named `"book.slug"` and it would all work out

janKozie1:
I've decided to use an another [package](https://www.npmjs.com/package/flat) it can turn param objects back into flattened objects. Also, `pathToRegex` only seems to support '_' besides numbers and letters. [link](https://github.com/pillarjs/path-to-regexp/blob/7b4078db2cfc929eb287432be0ecd3798b0de820/src/index.ts#L63)

TomWoodward:
the package you picked looks more popular than the one i did so that is probably a good move. the one i posted did support using a custom delimiter though

src/app/navigation/utils.ts
```diff
@@ -16,12 +19,14 @@ export const locationChangeForRoute = <R extends AnyRoute>(
 ): locationChange is Required<LocationChange<Match<R>>> =>
   !!locationChange.match && locationChange.match.route.name === route.name;
 
+export const getUrlRegexParams = <T extends {}>(obj: T): T => flatten(obj, {delimiter});
```
janKozie1: This one does too,  `pathToRegex` doesn't though [(for generating tokens)](https://github.com/pillarjs/path-to-regexp/blob/7b4078db2cfc929eb287432be0ecd3798b0de820/src/index.ts#L141)

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -144,10 +142,11 @@ const resolvePage = async(
     throw new Error('Page not found');
   }
 
+  const loadingPage = select.loadingPage(state);
   const pageState = select.page(state);
   if (pageState && pageState.id === pageId) {
     return pageState;
-  } else if (match.params.page !== select.loadingPage(state)) {
+  } else if (!(loadingPage && loadingPage.slug === match.params.page.slug)) {
```
TomWoodward: can't you use isEqual here as we did above for the book param?

src/app/content/reducer.ts
```diff
@@ -82,7 +82,7 @@ function reduceContent(state: State, action: AnyAction) {
       }
 
       // book is the same, page is different
-      if (state.book && state.page && action.payload.match.params.page !== getPageSlug(state.book, state.page)) {
+      if (state.book && state.page && action.payload.match.params.page.slug !== getPageSlug(state.book, state.page)) {
```
TomWoodward: eventually we're gonna make it so you can have different page arguments than just slug, you can future proof by comparing `state.params.page` and `action.payload.match.params.page` with isEqual. incidentally, the comparison up on line 75 can be simplified to avoid the `omitPage` just as you did in `resolveContent`

src/app/content/selectors.ts
```diff
@@ -42,23 +42,24 @@ export const loading = createSelector(
...
-  (state) => state.params ? state.params.page : null
+  (state) => state.params ? state.params.page.slug : null
 );
 
 export const pageNode = createSelector(
   book,
   pageParam,
-  (selectedBook, slug) => selectedBook && slug ? findArchiveTreeNodeBySlug(selectedBook.tree, slug) : undefined
+  (selectedBook, slug) =>
+    selectedBook && slug ? findArchiveTreeNodeBySlug(selectedBook.tree, slug) : undefined
```
TomWoodward: is this different? it doesn't look different

TomWoodward (changes requested):
this is looking good in general, i left some comments, and there are typescript errors coming up in CI

janKozie1:
`getPageSlug` doesn't seem to be needed anymore and  the missing coverage comes from it's return never being checked. Should i remove it or add the missing test?

TomWoodward:
@janKozie1 looks like you can remove the function

TomWoodward:
there are still a bunch of typescript errors, you should be able to see them by running `yarn tsc`

src/app/content/types.ts
```diff
@@ -22,10 +22,7 @@ export interface Params {
 export interface State {
   tocOpen: boolean | null;
   params: Params | null;
-  loading: {
-    book?: Params['book']
-    page?: Params['page']
-  };
+  loading: Partial<Params>;
```
janKozie1: I think this might be better? If its just `Params | null` it implies that if its truthy then both `book` and `page` are loading. I guess ts didnt check it since the spread operator was used in reducer while setting the `loading`

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -116,7 +114,7 @@ const loadPage = async(
   bookLoader: ReturnType<AppServices['archiveLoader']['book']>,
   pageId: string
 ) => {
-  services.dispatch(requestPage(match.params.page));
+  services.dispatch(requestPage({slug: match.params.page.slug}));
```
TomWoodward: shouldn't this be 
```suggestion
  services.dispatch(requestPage(match.params.page));
```

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -133,7 +131,7 @@ const resolvePage = async(
   const state = getState();
   const pageId = match.state && match.state.pageUid
     ? match.state.pageUid
-    : getPageIdFromUrlParam(book, match.params.page);
+    : getPageIdFromUrlParam(book, match.params.page.slug);
```
TomWoodward: ideally this would take the whole page param as the second argument

src/app/content/reducer.ts
```diff
@@ -69,10 +68,8 @@ function reduceContent(state: State, action: AnyAction) {
         return initialState;
       }
 
-      const omitPage = omit(['page']);
-
       // book is different
-      if (!equals(omitPage(action.payload.match.params), omitPage(state.params))) {
+      if (!state.params || !isEqual(action.payload.match.params.book, state.params.book)) {
```
TomWoodward: you shouldn't need the first part with the `!state.params || `, if it hasn't been set then `state.params.book` will be undefined, which will never match `action.payload.match.params.book`

src/app/content/selectors.ts
```diff
@@ -52,7 +52,7 @@ export const loadingPage = createSelector(
 
 export const pageParam = createSelector(
   localState,
-  (state) => state.params ? state.params.page : null
+  (state) => state.params ? state.params.page.slug : null
```
TomWoodward: it looks like 

`findArchiveTreeNodeBySlug` and `getPageIdFromUrlParam` should be somehow combined or set up to share some logic that will find the page from the tree based on the page param.

we're trying to move away from the assumption that the page param is just a slug, so we shouldn't be introducing more places that rely on that

src/app/content/types.ts
TomWoodward: yes i think this is good

src/app/content/utils/urlUtils.ts
```diff
@@ -33,21 +33,21 @@ export const getBookPageUrlAndParams = (
...
-    return {book: book.slug};
+    return {slug: book.slug};
   } else {
     return {uuid: book.id, version: book.version};
   }
 };
 
 const getUrlParamForPageIdCache = new Map();
-export const getUrlParamForPageId = (book: Pick<Book, 'id' | 'tree' | 'title'>, pageId: string): string => {
+export const getUrlParamForPageId = (book: Pick<Book, 'id' | 'tree' | 'title'>, pageId: string): SlugParams => {
```
TomWoodward: return type should be `Params['page']` for future proofing

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
janKozie1: I've done it expecting that eventually routing for pages will be similar to routing for books, i.e pages can be requested with id.

src/app/content/reducer.ts
janKozie1: But `state.params` can be null. I dont think `state.params.page` and `state.params.book` are set separately, so i kept their type as `Params | null` compared to  `state.loading` being `Partial<Params>`

TomWoodward: oh right ok, that makes sense

TomWoodward (approved):

PR #487, opened 2020-02-07
Js dropdown for context menu

PiotrKozlowski:
@TomWoodward I have a problem with `reactUtils.spec.tsx` test coverage.
It seems like `initFocusEvent` is not a function so I have to use `initEvent` which does not allow me to set `relatatedTarget` which makes this one branch very hard to test.


src/app/components/Dropdown.tsx
```diff
@@ -58,6 +59,7 @@ const TabHiddenDropDown = styled(({toggle, children, className}: Props) => {
   const container = React.useRef<HTMLElement>(null);
 
   useOnClickOutside(container, open, () => setOpen(false));
```
TomWoodward: in my testing this `useOnClickOutside` doesn't appear to be necessary anymore, i wonder if there might be something you can click on that can't receive focus, that might require this extra handler.

if you can't think of a reason to keep it we might as well remove it

src/app/reactUtils.ts
```diff
@@ -12,3 +13,26 @@ export const useDrawFocus = <E extends HTMLElement = HTMLElement>() => {
...
+
+  if (isEnabled) {
+    el.addEventListener('focusout', handler as any);
+  }
+
+  return () => el.removeEventListener('focusout', handler as any);
+};
+
+export const useFocusLost = (ref: React.RefObject<HTMLElement>, isEnabled: boolean, cb: () => void) => {
+  React.useEffect(onFocusLostHandler(ref, isEnabled, cb), [ref, isEnabled, cb]);
```
TomWoodward: i think you want this to be like this
```suggestion
  React.useEffect(onFocusLostHandler(ref, isEnabled, cb), [isEnabled]);
```

TomWoodward:
@PiotrKozlowski you could try using the focusevent constructor, i think that is the new better way to create events. if this still doesn't let you set the relatedTarget you could always just use `Object.defineProperty` in the test or mock `element.addEventListener` so that you can intercept the handler and trigger it manually with a pojo

src/app/reactUtils.ts
PiotrKozlowski: So, there is a lint error about `react-hooks/exhaustive-deps` which always make me wonder if it is useful. From one perspective, if we know that those args will not change then there is no reason to add them to dependencies array, but from the other perspective they can be dynamically set in some other case in the future and then keeping them inside will be helpful.
Eventually, I've ended up adding those deps depending on the case. Here we have `ref` which actually can change so hook should be called one more time after this happens and we have `cb` which most likely will not change, but in this case, adding it to dependency array will not have any side effects so I guess it's okay. 
Things get complicated when you are using like 6 dependencies inside hook but you know they will not change. In this case I will most likely add only those which are needed for sure.

If you would like to have other guidelines for this lint error just inform me

TomWoodward: the callback will be different every time the hook is called. agree that the ref is probably safe to keep

PiotrKozlowski: Ahh, right, because we are creating new function instead of passing a reference. I've removed `cb` from dependencies array

src/app/reactUtils.ts
```diff
@@ -12,3 +13,26 @@ export const useDrawFocus = <E extends HTMLElement = HTMLElement>() => {
...
+  const handler = (ev: FocusEvent) => {
+    const relatedTarget = ev.relatedTarget as HTMLElement | null;
+
+    if (!relatedTarget || !elementDescendantOf(relatedTarget, ref.current!)) {
+      cb();
+    }
+  };
+
+  if (isEnabled) {
+    el.addEventListener('focusout', handler as any);
```
TomWoodward: can we type this handler correctly so we don't need to do `as any`?

PiotrKozlowski: When I remove `any` then it is throwing:
```
Type '(ev: FocusEvent) => void' is not assignable to type 'EventListener'
```
When I remove import for `import FocusEvent from '@openstax/types/lib.dom';` and use global `FocusEvent` then the handler is okay, but then its throwing an error about:
```
Property 'relatedTarget' does not exist on type 'FocusEvent'
```

PiotrKozlowski: it is working if I use `FocusEventInit`

TomWoodward:
@PiotrKozlowski check out the type wrangling i did in https://github.com/openstax/rex-web/pull/487/commits/18546c3e8b6ef0ad1011dad12de4f6d9bdcf6e33 , i guess the `addEventListener` type is not smart enough to narrow the handler to a specific event type, so you have to narrow it inside the handler

PiotrKozlowski:
@TomWoodward What about removing:
```
if (!(event instanceof assertWindow().FocusEvent)) {
      return;
    }
```
in favor of:
```
const relatedTarget = (event as FocusEvent).relatedTarget as HTMLElement | null;
```
where `FocusEvent` is imported from `@openstax/types/lib.dom`?

We now this will be this type because we are using `focusout` and we omit additional branch to test

TomWoodward:
@PiotrKozlowski i see what you mean about that extra branch being annoying to test, i moved the logic into a util function that does all the checking, this way when you're writing your handler you can just assume its the right type 55e75cf7

PiotrKozlowski:
@TomWoodward Code looks good, I just feel overwhelmed by the number of util functions we have in this project :D

PR #488, opened 2020-02-07
When creating new pending highlight, insert card in proper order

PiotrKozlowski:
When creating new pending highlight, insert card in proper order


src/app/content/components/Page/highlightManager.ts
```diff
@@ -110,6 +110,25 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
...
+    if (!highlighter) { return highlights; }
+
+    const ordered = highlights.filter((highlight) => !pending || highlight.id !== pending.id);
+    const prevHighlight = highlighter.getHighlightBefore(pending);
+    if (!prevHighlight) {
+      ordered.unshift(pending);
+      return ordered;
+    }
+
+    const indexToInsert = ordered.findIndex((highlight) => highlight.id === prevHighlight.id);
```
TomWoodward: if you change this condition to be like this you can delete the conditional above
```suggestion
    const indexToInsert = ordered.findIndex((highlight) => prevHighlight && highlight.id === prevHighlight.id);
```

PiotrKozlowski:
@TomWoodward I've planned to make separate pr with tests for `issue-581` branch when it gets approved, because I want to be sure that there will be no more changes to the logic

TomWoodward:
this seems like it would be pretty safe to add tests for, but i understand what you're saying. you can merge this whenver

PR #489, opened 2020-02-10
Fetch all remaining highlights before printing

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/810

I didn't want to make the `loadMore` hook to work differently depending on the action so i decided to make a separate hook which is almost identical to it. I've tried abstracting parts of it so less repetition.


src/app/content/highlights/hooks/loadMore.ts
```diff
@@ -91,20 +71,7 @@ const loadMore = async({getState, highlightClient, dispatch}: MiddlewareAPI & Ap
...
-    );
-    const locationFilterId = stripIdVersion(location.id);
-
-    return addSummaryHighlight(result, {
-      highlight,
-      locationFilterId,
-      pageId,
-    });
-  }, {} as ReturnType<typeof addSummaryHighlight>);
+  const formattedHighlights = formatReceivedHighlights(highlights, locationFilters);
```
PiotrKozlowski: I like how you extract those to util functions

PiotrKozlowski (commented):
I'm not sure about duplicating all of that code, however, it seems like it may be better than adding complexity to `loadMore`. In cases like this, I usually appreciate some kind of comment at the top of a new hook which says that it is similar to `loadMore` hook and why

src/app/content/highlights/hooks/loadMore.ts
```diff
@@ -1,22 +1,15 @@
...
 import { actionHook } from '../../../utils';
 import { book as bookSelector } from '../../selectors';
 import { loadMoreSummaryHighlights, receiveSummaryHighlights, setSummaryFilters } from '../actions';
 import { summaryPageSize } from '../constants';
 import * as select from '../selectors';
 import { SummaryHighlightsPagination } from '../types';
-import { getNextPageSources } from '../utils/paginationUtils';
-import { fetchHighlightsForSource, formatReceivedHighlights, incrementPage } from './utils';
+import { getNewSources } from './utils';
+import { fetchHighlightsForSource, incrementPage, loadMoreByFunction } from './utils';
```
PiotrKozlowski: Two imports from the same file

src/app/content/highlights/hooks/utils.ts
```diff
@@ -4,12 +4,22 @@ import {
...
+import * as select from '../selectors';
 import { HighlightLocationFilters, SummaryHighlightsPagination, } from '../types';
 import { addSummaryHighlight, getHighlightLocationFilterForPage } from '../utils';
+import { getNextPageSources } from '../utils/paginationUtils';
+
+type fetchFunctionBody = (args: { previousPagination: SummaryHighlightsPagination, getState: Store['getState'],
+    highlightClient: AppServices['highlightClient'],
+    highlights?: Highlight[]
+    sourcesFetched: string[]}
+) =>  Promise<{pagination: SummaryHighlightsPagination, highlights: Highlight[]}>;
```
PiotrKozlowski: Indentation + double spae before `Promise` - btw. if you are using vscode you can sett default indentation for a given project and auto fixing on `ctrl + s`

src/app/content/highlights/utils/paginationUtils.ts
```diff
@@ -35,13 +35,13 @@ const totalOfCountsPerSource: (counts: CountsPerSource) => number = flow(
 export const getNextPageSources = (
   remainingCounts: CountsPerSource,
   tree: ArchiveTree,
-  nextPageSize: number
+  nextPageSize?: number
```
TomWoodward: you don't want this to return too many sources, because you'll start getting errors from the api that the url is malformed. in the case of the print we do want to target loading all highlights so having a null/undefined `nextPageSize` makes sense, but we should also introduce a constant for `max_query_sources` or something and limit the number of potential sources to return here

src/app/content/highlights/hooks/utils.ts
```diff
@@ -0,0 +1,118 @@
...
+export const loadMoreByFunction = async(fetchingFunction: fetchFunctionBody, {
+  getState,
+  highlightClient,
+}: MiddlewareAPI & AppServices) => {
+  const state = getState();
+  const locationFilters = select.highlightLocationFilters(state);
+  const previousPagination = select.summaryPagination(state);
+  const sourcesFetched = Object.keys(select.loadedCountsPerSource(state));
+
+  const {pagination, highlights} = await fetchingFunction({
```
TomWoodward: do you really need to pass a fetching function into this? or can you just pass a page size of type `number | null`? null for everything

TomWoodward (changes requested):

src/app/content/highlights/utils/paginationUtils.spec.ts
```diff
@@ -1,32 +1,53 @@
...
   });
 
-  it('fails over to next page if more records are needed', () => {
-    expect(getNextPageSources({
-      page1: {[HighlightColorEnum.Green]: 5},
-      page2: {[HighlightColorEnum.Pink]: 10},
-    }, treeWithUnits, 10)).toEqual(['page1', 'page2']);
+  describe('under resource limit' , () => {
+    it('gets one page id if it has enough records left to fill a response', () => {
+      const {getNextPageSources} = require('./paginationUtils');
```
janKozie1: I wasn't able to mock the `maxResourcesPerFetch` for one test only any other way. I'm not sure what am i doing wrong.

src/app/content/highlights/hooks/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    return addSummaryHighlight(result, {
+      highlight,
+      locationFilterId,
+      pageId,
+    });
+}, {} as ReturnType<typeof addSummaryHighlight>);
+
+export const incrementPage = (pagination: NonNullable<SummaryHighlightsPagination>) =>
+({...pagination, page: pagination.page + 1});
```
TomWoodward: this should be indented 2

src/app/content/highlights/reducer.ts
```diff
@@ -155,6 +155,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.clearFocusedHighlight): {
       return omit('focused', state);
     }
+    case getType(actions.printSummaryHighlights):
```
TomWoodward: i think you have to either switch on `hasMoreResults` in the component or have two actions, one for handling the loading and the printing and one just to set the loading state

TomWoodward (changes requested):
i didn't look hard at the specs, but here is some feedback about the other stuff

src/app/content/highlights/constants.ts
```diff
@@ -2,6 +2,8 @@ import { HighlightColorEnum } from '@openstax/highlighter/dist/api';
 import books from '../../../config.books';
 export const highlightingFeatureFlag = 'enableHighlighting';
 
+export const maxHighlightsPerPage = 200;
+export const maxResourcesPerFetch = 10;
```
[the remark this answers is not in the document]
janKozie1: I did test it with 10 and it seems to work. Is 5 safer?

src/app/content/highlights/hooks/loadMore.ts
```diff
@@ -1,72 +1,52 @@
...
-    ? {sourceIds, page}
-    : null;
-
-  const highlights = args.highlights
-    ? [...args.highlights, ...data]
-    : data
-  ;
-
-  if (highlights.length < summaryPageSize) {
+  if (highlights.length < summaryPageSize || !args.pageSize) {
```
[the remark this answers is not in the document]
janKozie1: I agree with replacing `summaryPageSize` with `perPage` but i think `!args.pageSize` has to stay. Because usually after requesting to print highlights, there will be a pagination object coming from previous scroll event which has to be 'used up' before continuing with `maxHighlightsApiPageSize`.

janKozie1:
I'm not sure what to do about scenario like: 
1. User scrolls down and triggers `loadMore`
2. User click print before `loadMore` is done running which triggers next fetch with the same pagination.

Similar scenario possible if user clicks the print button again while the highlights are still fetching.
Either disable the button while `loading` or add separate place to store resources/paginations that are currently being fetched. Thoughts? @TomWoodward

PiotrKozlowski:
I'll vote for 
> disable the button while loading`

TomWoodward:
@PiotrKozlowski 's suggestion sounds fine to me. you would have to try pretty hard to be loading more from the pagination and press the print button at the same time, because the queries are fast and the filter bar isn't sticky.

janKozie1:
@TomWoodward updated

src/app/content/highlights/components/SummaryPopup/Filters.tsx
```diff
@@ -55,7 +48,7 @@ interface Props {
 }
 
 // tslint:disable-next-line:variable-name
-const Filters = ({className}: Props) => <div className={className}>
+const Filters = ({className}: Props) =>   <div className={className}>
```
TomWoodward: too many spaces

src/app/content/highlights/components/SummaryPopup/HighlightsPrintButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  shouldFetchMore: boolean;
+  isLoading: boolean;
+  print: typeof printSummaryHighlights;
+}
+
+// tslint:disable-next-line:variable-name
+const HighlightsPrintButton = ({className, print, shouldFetchMore, isLoading}: Props) => {
+  return <PrintButton
+    className={className}
+    onClick={() => print({shouldFetchMore})}
```
TomWoodward: there is no reason to pass `shouldFetchMore`, you can either:
- use `shouldFetchMore` here to either dispatch the action or just call window.print
- use `hasMoreResults` within the hook to see if more results are needed

src/app/content/highlights/hooks/utils.ts
```diff
@@ -0,0 +1,91 @@
...
+export const incrementPage = (pagination: NonNullable<SummaryHighlightsPagination>) =>
+  ({...pagination, page: pagination.page + 1});
+
+export const getNewSources = (state: AppState, omitSources: string[], pageSize?: number) => {
+  const book = bookSelector(state);
+  const remainingCounts = omit(omitSources, select.filteredCountsPerPage(state));
+  return book ? getNextPageSources(remainingCounts, book.tree, pageSize) : [];
+};
+
+export const extractDataFromHighlightClientResponse = (highlightsResponse: Highlights) => {
```
TomWoodward: can you update `src/app/content/highlights/hooks/locationChange.ts` to use this function to reduce duplication?

src/app/content/highlights/utils/paginationUtils.spec.ts
TomWoodward: you're doing it right. because regular imports are cached and you can't modify an exported constant by reference you have to reset and require it for each test you run. you can do it in a `beforeEach` to reduce duplication, but in general you're doing it right already.

TomWoodward (changes requested):

TomWoodward (changes requested):
i think this looks good, but please merge master and resolve conflicts

janKozie1:
I've resolved the conflicts, should be done now

TomWoodward (approved):
sending to qa

janKozie1:
I saw Malar's comment about figures not loading prior printing. I think math could be causing problems like this too. 

I thought about:
  1. Making a function which takes a ref to a node and will return a promise which resolves when all images in that node have been loaded.
  2. Preload the images
  3. Fetch all highlights but use the default pagination (which seems to be 'just enough' for the images to have time to load) and dispatch with each fetch until done.

Thoughts, @TomWoodward @PiotrKozlowski ?

PiotrKozlowski:
@janKozie1 There is a utility function `allImagesLoaded.ts` which does what you've described in `1)`. I think using this function should be enough.
cc @TomWoodward

janKozie1:
It works, but im not sure what to think about ref being passed with an action. I will write the remaining tests once this approach gets approved. (should i write them in `ShowMyHighlights` if so? The file is quite big, the printing method needs tests).

TomWoodward:
@janKozie1 i agree that it is a little weird having the ref passed to the action, and that doesn't help with the math being rendered problem. 

we have a thing in AppServices called a PromiseCollector, you can see in PageComponent how it adds the typesetMath promise to the PromiseCollector. you could do the same in ShowMyHighlights, and throw in a promise for allImagesLoaded when new highlights are received, and then from your hook you just have to run `services.promiseCollector.calm().then(() => window.print())`

i think this should work, but make sure your hook is not an async function or return a promise because if it is it'll be added to the collector automatically and you'll be stuck waiting for yourself to resolve forever

src/app/content/highlights/components/ShowMyHighlights.tsx
```diff
@@ -86,7 +89,17 @@ class ShowMyHighlights extends Component<ShowMyHighlightsProps, { showGoToTop: b
...
+    if (!prevProps.summaryIsLoading || !this.myHighlightsBodyRef.current) { return; }
+
+    const {summaryHighlights: prevHighlights} = prevProps;
+    const {summaryHighlights: currHighlights} = this.props;
+
+    if (!prevHighlights || !currHighlights) { return; }
+
+    if (!isEqual(lastProp(prevHighlights), lastProp(currHighlights))) {
+      this.props.services.promiseCollector.add(allImagesLoaded(this.myHighlightsBodyRef.current));
+    }
```
janKozie1: I tried something like this. It doesn't work because at the moment this runs, the images are not yet appended to the html, but `summaryHighlights` have already been updated.

janKozie1: On the other hand [this](https://github.com/openstax/rex-web/pull/489/commits/ca01867c476cd4e30c76af94892deb9f59df946d) does work (tested with few 34 pages worth of highlights images+maths on slow 3g emulation). I dont think that looping over all images and checking one prop is _that_ bad compared to dealing with issues like ^.

cc. @TomWoodward

TomWoodward (changes requested):
the print button isn't getting the pointer cursor that a button should be getting

TomWoodward:
@janKozie1 please look at the most recent commit i added, the trick was to use `useLayoutEffect` which is triggered synchronously after the dom update

janKozie1:
Thats great. I tested it and it works well even on slow 3g. I can't replicate the issue with print button though, is it still there?

janKozie1:
Ive updated the tests, i think it's alright now. @TomWoodward

TomWoodward (approved):

PR #492, opened 2020-02-11
Show help info alert on mobile

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/591


src/app/content/highlights/components/HighlightsPopUp.tsx
```diff
@@ -123,6 +124,7 @@ class HighlightsPopUp extends Component<Props> {
           </Styled.Header>
           {this.props.user ? <ShowMyHighlights /> : this.loginForHighlights()}
         </Styled.Modal>
+        {this.props.user && <HighlightsHelpInfo />}
```
TomWoodward: you already have a conditional `return null` inside here, it might be cleaner to just check for user there as well

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const [show, setShow] = React.useState(false);
+
+  const dismiss = () => {
+    Cookies.set(cookieId, 'true');
+    setShow(false);
+  };
+
+  React.useEffect(() => {
+    setTimeout(() => {
+      if (!Boolean(Cookies.get(cookieId)) && assertWindow().matchMedia(theme.breakpoints.mobileQuery).matches) {
```
TomWoodward: the only reason to do this `matchMedia` in js is to have analytics on when we show this like we did for the signup banner. can you make a secondary PR into this one that adds that analytics event?

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return <Wrapper data-analytics-region='Mobile MH help info'>
+    <span>
+    <FormattedMessage id='i18n:toolbar:highlights:popup:help-info'/>
+      <FormattedMessage id='i18n:toolbar:highlights:popup:help-info:link'>
+        {(msg: Element | string) => <a
+          data-testid='support-link'
+          data-analytics-label='support-link'
+          href={supportCenterLink}
+        >{msg}</a>}
+      </FormattedMessage>
```
TomWoodward: you could make the complete content of this element, anchor and text together, an html message. that has the added bonus of allowing you to have a different help article per language, which is not unreasonable. 

also composing sentences out of multiple messages gets real dicey sometimes with other languages having different phrasing or sentence structure.

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
```diff
@@ -0,0 +1,96 @@
...
+const slideInAnimation = css`
+  animation: ${800}ms ${slideInFromBottom} ease-out;
+`;
+
+// tslint:disable-next-line: variable-name
+const Wrapper = styled.div`
+  position: fixed;
+  bottom: 0;
+  z-index: ${theme.zIndex.highlightsHelpInfoMobile};
+  ${textRegularStyle}
```
TomWoodward: if you make this `bodyCopyRegularStyle` its the same thing but includes the anchor styling, you probably don't want the `text-decoration: none;`, which is the only thing you added that it doesn't have

TomWoodward (changes requested):

TomWoodward (changes requested):
@PiotrKozlowski in your commit 249eff1 you removed the matchMedia call, and now that we've added analytics we will be triggering show even on desktop when it is not actually shown

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
```diff
@@ -0,0 +1,98 @@
...
+  justify-content: space-between;
+  width: 100%;
+  height: 6rem;
+  padding: 0 2rem;
+  background-color: ${theme.color.neutral.formBackground};
+  border: 1px solid ${theme.color.neutral.formBorder};
+  ${slideInAnimation}
+
+  a {
+    text-decoration: none;
```
TomWoodward: i don't think this is good, but i see that it is like this in the design, wait to see what devika says https://openstax.slack.com/archives/C8V1XET1R/p1581603867240200

TomWoodward (changes requested):

PR #496, opened 2020-02-12
Focus dropdown after esc

PiotrKozlowski:
When clicking on a child of dropdown focus is moved to this child and when we click `Escape` it is hiding, but the focus is lost which disturbs how ESC events are working.
This pr is setting focus on the main dropdown component after it is closed with ESC.

PS. Actually probably using `useDrawFocus` hook would be sufficient. I don't remember on which pr it was implemented.

### DIFF
--- a/src/app/components/Dropdown.tsx
+++ b/src/app/components/Dropdown.tsx
@@ -63,7 +63,7 @@ const TabHiddenDropDown = styled(({toggle, children, className}: Props) => {
     if (container.current) { container.current.focus(); }
   });
 
-  return <div className={className} ref={container} tabIndex={0}>
+  return <div className={className} ref={container} tabIndex={-1}>
     <DropdownToggle component={toggle} onClick={() => setOpen(!open)} />
     {open && children}
   </div>;
--- a/src/app/components/Dropdown.tsx
+++ b/src/app/components/Dropdown.tsx
@@ -58,9 +58,12 @@ const TabHiddenDropDown = styled(({toggle, children, className}: Props) => {
   const container = React.useRef<HTMLElement>(null);
 
   useOnClickOutside(container, open, () => setOpen(false));
-  useOnEsc(container, open, () => setOpen(false));
+  useOnEsc(container, open, () => {
+    setOpen(false);
+    if (container.current) { container.current.focus(); }
+  });
 
-  return <div className={className} ref={container}>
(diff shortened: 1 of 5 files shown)


TomWoodward:
useDrawFocus is from https://github.com/openstax/rex-web/pull/408, are you saying that it would be better to close this and address this issue after everything is merged down?

PiotrKozlowski:
@TomWoodward I've just checked:
```
const container = useDrawFocus();
...
```
but it is not focusing after the close. We could do sth like:
```
const useDrawFocus = <E extends HTMLElement = HTMLElement>(deps: any[] = [])
...
}, [ref, ...deps]);
```
and then:
```
const container = useDrawFocus([open]);
```
It is working, but it this case when we are opening MH then last visible Dropdown is focused instanly.

I think we can either work more on useDrawFocus so we can pass some prop which will decide if the component should be focused at the first render or just go with this pr.

TomWoodward:
@PiotrKozlowski we could make useDrawFocus work for this, but it would end up applying to anytime the dropdown is closed, when we only need to really set the focus when it is closed with the esc key (otherwise the focus will have been set by whatever action they took).

i'm a little nervous about doing this "focus after pressing the esc key" logic in a case by case way for each usage, i feel like we're going to run into a situation where we don't have a reference to the thing we want to focus... but i don't have a better idea right now.

lets proceed with what you have here and see what happens.

src/app/components/Dropdown.tsx
```diff
@@ -58,9 +58,12 @@ const TabHiddenDropDown = styled(({toggle, children, className}: Props) => {
...
 
   useOnClickOutside(container, open, () => setOpen(false));
-  useOnEsc(container, open, () => setOpen(false));
+  useOnEsc(container, open, () => {
+    setOpen(false);
+    if (container.current) { container.current.focus(); }
+  });
 
-  return <div className={className} ref={container}>
+  return <div className={className} ref={container} tabIndex={0}>
```
TomWoodward: isn't this going to clutter up the keyboard navigation? maybe put it to `-1`? i also think it would be good to focus the toggle, instead of the wrapper, unless you had some problem doing that

PiotrKozlowski: Not sure how to pass `ref` to `toggle` component since we are using `React.cloneElement` wrapped with styled components and it would have to be somehow wrapped with `React.forwardRef`

PR #497, opened 2020-02-12
Add analytics tracking for showing highlights help info

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/591

### DIFF
--- a/src/app/content/highlights/components/HighlightsPopUp.spec.tsx
+++ b/src/app/content/highlights/components/HighlightsPopUp.spec.tsx
@@ -1,11 +1,13 @@
 import React from 'react';
 import { Provider } from 'react-redux';
 import renderer, { act } from 'react-test-renderer';
+import createTestServices from '../../../../test/createTestServices';
 import createTestStore from '../../../../test/createTestStore';
 import { renderToDom } from '../../../../test/reactutils';
 import { receiveFeatureFlags } from '../../../actions';
 import { receiveUser } from '../../../auth/actions';
 import { User } from '../../../auth/types';
+import * as Services from '../../../context/Services';
 import * as appGuards from '../../../guards';
 import MessageProvider from '../../../MessageProvider';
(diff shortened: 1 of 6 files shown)


TomWoodward:
linting is failing because of the hooks dependencies thing. its possible to shut it up by doing this: 

```diff
diff --git a/src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx b/src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
index f517f73d..d9d0cb51 100644
--- a/src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
+++ b/src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
@@ -77,13 +77,15 @@ const HighlightsHelpInfo = () => {
     setShow(false);
   };
 
-  React.useEffect(() => {
+  const onMount = React.useCallback(() => {
     setTimeout(() => {
       if (Boolean(Cookies.get(cookieId))) { return; }
       setShow(true);
       trackShowHelpInfo();
     }, timeBeforeShow);
-  }, []);
+  }, [trackShowHelpInfo]);
+
+  React.useEffect(onMount, []);
 
   if (!show || !user) { return null; }
```

but i'm **_deeply_** annoyed by that, i would also be ok with disabling that linter rule. i'm pretty sure `trackShowHelpInfo` is going to be a different reference on every render, and we definitely don't want to be re-running anything when that happens. i feel like the purpose of the dependencies array is to say "when these things change re-run the hook" not "here is a list of all the variables the hook uses", the latter does not seem useful to me. i feel like hooks are already pretty obnoxious to read and make sense of, and i'm not feeling very good about making them _more_ obnoxious just to shut the linter up. i think in this case you should just add the comment to disable the rule. i'm on the fence about just disabling the rule everywhere, maybe lets not go that far just yet.

PR #504, opened 2020-02-13
Fix bg colors for cards on Edge

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/756

~Check if switching from rgba to rgb will fix it~
Switching to `rgb` works but there is one more issue where Edge does not support css `clip-path` so it looks like this:
image

TODO:
- [x] Replace this clip path with something else

Maybe something like: https://stackoverflow.com/a/13148495


KarinaMendez2 (approved):
Cannot test Edge, but on chrome, firefox and safari looks great.

TomWoodward:
sent to qa

PR #506, opened 2020-02-14
Versioned canonical URL

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/837


src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -53,7 +53,7 @@ const resolveBook = async(
   const loader = archiveLoader.book(bookId, bookVersion);
   const state = getState();
   const bookState = select.book(state);
-  const book = bookState && bookState.id === bookId ? bookState : undefined;
+  const book = bookState && bookState.id === bookId && bookState.version === bookVersion ? bookState : undefined;
```
janKozie1: I wasn't sure what 
`if a version is specified in the url, and if the canonical is defaulting to point to the current book, then it should not try to load the book again but with the version from books.config.js, it should use the loaded version.` exactly means. Is this it?

src/app/content/utils/canonicalUrl.ts
```diff
@@ -18,13 +19,17 @@ export async function getCanonicalUrlParams(
   ].filter((id) => !!BOOKS[id]);
 
   for (const id of canonicals) {
-    const version = BOOKS[id].defaultVersion;
+    const version = bookVersion || BOOKS[id].defaultVersion;
```
TomWoodward: you don't want to always use this version, because if you're pointing to another book this version will be invalid. i think you should change the `canonicals` array on line 16 to have uuid AND version, instead of just uuid, that way when you set the default (line 18 `bookId, // use the current book as a last resort`) you'll put the book and id there

src/app/content/hooks/locationChange/resolveContent.ts
TomWoodward: your changes to `src/app/content/hooks/receiveContent.ts` and `src/app/content/utils/canonicalUrl.ts` are on the right track. i don't think you should have to make any changes to `src/app/content/hooks/locationChange/resolveContent.ts`

TomWoodward (changes requested):

src/app/content/utils/canonicalUrl.ts
```diff
@@ -9,22 +9,31 @@ export async function getCanonicalUrlParams(
...
-  for (const id of canonicals) {
-    const version = BOOKS[id].defaultVersion;
+  const canonicals = [...(CANONICAL_MAP[bookId] || []), bookId] // use the current book as a last resort
+    .filter((id) => !!BOOKS[id])
+    .map((id) => ({
+      id,
+      version: bookVersion && id === bookId
+        ? bookVersion
+        : BOOKS[id].defaultVersion,
+    }));
```
TomWoodward: you could rewrite this as 
```
const canonicals = [
  ...(CANONICAL_MAP[bookId] || []).map(id => ({id, version: BOOKS[id].defaultVersion})),
  {id: bookId, version: bookVersion}
];
```

and it wouldn't need that weird conditional in the map

src/app/content/utils/canonicalUrl.ts
```diff
@@ -9,22 +9,31 @@ export async function getCanonicalUrlParams(
...
+  for (const {id, version} of canonicals) {
 
     const canonicalBook = await getBook(id, version);
     const treeSection = findArchiveTreeNode(canonicalBook.tree, pageShortId);
     if (treeSection) {
       const pageInBook = assertDefined(treeSection.slug, 'Expected page to have slug.');
-      return {book: canonicalBook.slug, page: pageInBook};
+      const params = {book: canonicalBook.slug, page: pageInBook};
+
+      return version === BOOKS[id].defaultVersion || id !== bookId
```
TomWoodward: isnt' there a utility that does this?

TomWoodward (changes requested):

src/app/content/utils/canonicalUrl.ts
janKozie1: Wouldn't this lead to book-loading functions being possibly called with a book id that is not in `BOOKS`? Filter has to be called before map, but the current book also should be checked (i think there is no reason to possibly get a rejected promise?)

TomWoodward: lets simplify this whole situation and do 

```
const canonicals = [
    ...(CANONICAL_MAP[bookId] || []),
    ...(bookVersion === BOOKS[bookId].defaultVersion ? [bookId] : []), // use the current book as a last resort if it has the same version as in books config
  ].filter((id) => !!BOOKS[id]);
```

that way we don't have to worry about any non-production urls and its almost the same logic as before

TomWoodward (changes requested):

src/app/content/utils/canonicalUrl.ts
janKozie1: Wasn't the point of all this to allow to create a canonical with a version that is a different from the default one if its specified? This makes it skip the current book an because of that it won't be possible to create versioned canonicals, right?

TomWoodward: in the task i did write `it should use the loaded version` but really i just want to not have an error. since canonicals are always rendered with `openstax.org` urls, and versioned urls don't work on production, i'm realizing it doesn't make _the most_ sense to render that

src/app/content/utils/canonicalUrl.ts
janKozie1: `getBookPageUrlAndParams` does but it will throw if tree doesn't have the section which i don't think is desired behavior here.

src/app/content/utils/canonicalUrl.ts
```diff
@@ -9,19 +9,22 @@ export async function getCanonicalUrlParams(
...
-  pageShortId: string
+  pageShortId: string,
+  bookVersion?: string
 ) {
   const getBook = makeUnifiedBookLoader(archiveLoader, osWebLoader);
+
   const canonicals = [
     ...(CANONICAL_MAP[bookId] || []),
-    bookId, // use the current book as a last resort
+    ...(bookVersion === BOOKS[bookId].defaultVersion ? [bookId] : []),
```
TomWoodward: does this throw if you load a book not in the books config?

TomWoodward (changes requested):

src/app/content/utils/canonicalUrl.ts
janKozie1: I think so? I wasn't sure when is the app supposed to work in production when the id is not in `BOOKS`. I've updated it.

TomWoodward: in production it shouldn't be possible to load a book outside the BOOKS config, but it seemed like this code would throw in development when loading dynamic books

TomWoodward (approved):

PR #507, opened 2020-02-14
Change publish_date to latest revision date in Attribution copyrights

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/320

### DIFF
--- a/package.json
+++ b/package.json
@@ -125,7 +125,7 @@
     "@types/yargs": "^13.0.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-plugin-transform-dynamic-import": "^2.1.0",
-    "codecov": "^3.1.0",
+    "codecov": "^3.6.5",
     "date-fns": "^1.30.1",
     "express": "^4.16.4",
     "glob": "^7.1.3",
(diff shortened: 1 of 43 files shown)


TomWoodward (changes requested):
@PiotrKozlowski did you try this? as far as i can see the cms doesn't have a `revised` field. the card specifies that we should use the `Revision field in Archive`, which does have a `revised` field. seems like your changes to `src/app/content/utils.ts`, `src/test/mocks/osWebLoader.ts`, and `src/gateways/createOSWebLoader.ts` are not needed and the change in `src/app/content/types.ts` should be on `ArchiveBook`

PiotrKozlowski:
@TomWoodward I've moved `revised` to `ArchiveBook`. I've implemented this to `OsWebBook` because I was going from `console.log` in `Attribution.tsx` up until I've updated all files.
I'm not familiar yet with differences between `ArchiveBook` and `OSWebBook`

TomWoodward:
@PiotrKozlowski OSWeb is the cms, here is an example book https://rex-web.herokuapp.com/apps/cms/api/v2/pages/?type=books.Book&fields=cnx_id,authors,publish_date,cover_color&slug=calculus-volume-1 it generally has book metadata having to do with the display (like the theme color) or navigation (like the book slug in the url) so that we can be consistent with openstax.org (managed by the osweb cms)

archive has the actual book content for example https://rex-web.herokuapp.com/contents/8b89d172-2927-466f-8661-01abc7ccdba4@12.1 it has things like the table of contents, what license the book is published under, a history of changes, things like that

TomWoodward (approved):
passing to qa

PR #509, opened 2020-02-17
Add makeApiCallOrThrow util function

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/836

~~@TomWoodward I have a problem with tests for `create/remove/updateHighlight.ts`
For some reason, `makeApiCallOrThrow` is getting `undefined` instead of a promise.~~


src/app/content/highlights/hooks/createHighlight.ts
```diff
@@ -1,9 +1,9 @@
 import { ActionHookBody } from '../../../types';
-import { actionHook } from '../../../utils';
+import { actionHook, makeApiCallOrThrow } from '../../../utils';
 import { createHighlight } from '../actions';
 
 export const hookBody: ActionHookBody<typeof createHighlight> = ({highlightClient}) => async({payload}) => {
-  await highlightClient.addHighlight({highlight: payload});
+  await makeApiCallOrThrow(highlightClient.addHighlight({highlight: payload}));
```
TomWoodward: is it possible to add this to the fetch call in `src/gateways/createHighlightClient.ts` instead of in every place the api is called from?

src/app/utils.ts
```diff
@@ -212,3 +212,18 @@ export const onEscHandler = (element: React.RefObject<HTMLElement>, isEnabled: b
 export const useOnEsc = (element: React.RefObject<HTMLElement>, isEnabled: boolean, cb: () => void) => {
   React.useEffect(onEscHandler(element, isEnabled, cb), [element, isEnabled]);
 };
+
+export const makeApiCallOrThrow = async<T>(promise: Promise<T>): Promise<T> => {
+  return promise.catch((error) => {
+    if (error.status !== 422) {
+      throw new Error(error);
```
TomWoodward: i feel like you're this is going to potentially nest things that shouldn't be nested. maybe doing 
```
return Promise.reject(error);
```

would be better? i'm assuming what you're being passed in this handler is already an error, so nesting it in another error object might not be ideal

TomWoodward (changes requested):

TomWoodward (changes requested):
this isn't working. after a fetch the promise won't be rejected unless it literally failed to connect to the server. its only in subsequent sdk code that it inspects the response and rejects the promise. 

i think we have to expand the `src/gateways/createHighlightClient.ts` file to proxy all the calls to the client so that we can append the error handler to the returned promises after the sdk error handling

TomWoodward:
looking at the sdk code, there is not significant logic in there. so maybe we don't have to worry too much about doing our own thing with the fetch response.

if you want, you can continue doing as you are with the fetchApi handler, and just add some response inspection stuff. i have utilities i've used for this before in `src/helpers/fetch.ts`, but the `acceptStatus` function only takes one argument so you'd have to change it to take a list of acceptable statuses

PiotrKozlowski:
`acceptStatus` was hard to reuse because of `rejectResponse` process response as `text` and `formatError` type is expecting different arguments in this case so I've decided it will be simpler to just write this on it's own

TomWoodward (approved):
thats fine, what you have looks good now

PR #511, opened 2020-02-18
fix: math fonts look better

TomWoodward:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/553


KarinaMendez2 (approved):

src/app/utils.spec.tsx
```diff
@@ -196,7 +196,7 @@ describe('mergeRefs', () => {
 describe('remsToPx', () => {
 
   it('converts based on body font size', () => {
-    assertDocument().body.style.fontSize = '14px';
+    assertDocument().documentElement!.style.fontSize = '14px';
```
KarinaMendez2: can this 14px be a variable?

TomWoodward: i try not to use variables in tests unless it really saves a lot of space

PR #514, opened 2020-02-20
Different approach for scrolling

PiotrKozlowski:
The approach based on https://github.com/openstax/rex-web/pull/436#pullrequestreview-361290946

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/589


src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -77,6 +79,11 @@ const searchHighlightManager = (container: HTMLElement) => {
   };
 
   return {
+    scrollTo: (searchHighlight: Highlight) => () => {
+      allImagesLoaded(services.container).then(
```
TomWoodward: this `allImagesLoaded` should be done in general before triggering a scroll from the scrollTargetManager, not specifically for search highlights.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -43,13 +42,20 @@ export default class PageComponent extends Component<PagePropTypes> {
     return cleanContent.body.innerHTML;
   };
 
-  public getCurrentScrollTarget = (prevProps: PagePropTypes) => {
+  public getScrollTargets = (prevProps: PagePropTypes): ScrollTargets => {
```
TomWoodward: i don't think we should be adding logic to this component to handle scrolling, or identifying scroll targets. i can understand why the centralized one in navigation might not have worked, since each target type has different data they need. i also liked it better when a scroll target had one identifier and a type, rather than one object with a bunch of keys that need to be known everywhere.

since i guess its possible to have both a search result and a hash target, it seems like we need a way to somehow generate a list of the current scroll targets. i don't think we should have a bunch of logic in pageComponent for that, maybe if navigation also isn't a good place we can put something in each manager which will return either null or a scrolltarget.

and in the scrolltargetmanager we need to be looking for targets which were not in the previous prop, so that we only catch it when its added, otherwise we'll be scrolling to it every time we re-render.

PiotrKozlowski:
@TomWoodward I've pushed changes based on your comment https://github.com/openstax/rex-web/pull/514#discussion_r384061450
This is working, however, prioritizing is hard, because when there is highlight id and hash or then hash is in the props from the beginning and highlight id will be loaded later which means that 
```
const newScrollTargets = differenceWith(scrollTargets, prevScrollTargets, compareScrollTargets);
```
will never contain both scroll targets at the same time.
Any ideas about this?

TomWoodward:
@PiotrKozlowski will we ever have the hash and highlight id (or search result) in the url at the same time intentionally? if not then we probably don't have to try to handle that

PiotrKozlowski:
@TomWoodward There is a case when a user is searching for a keyword and then clicks on the figure link which makes `search=asd#figure-id`. Now user can refresh and in theory, we should be able to prioritize but with a current approach, we will scroll to the search highlight only

TomWoodward:
clicking a figure link is an awkward example because the browser handles it, but for instance clicking a link to another page that has a `#hash` on the end, you'd want that to navigate and scroll to the new target, but you wouldn't want it to clear the existing search.

i think the difference is that having `?search=query` should only indicate the presence of a search, not a selected result (which would be a scroll target) so if you click a search result you get a url like `?search=query#search-result-asieskjah` and then if you click a link with a hash it would change to `?search=query#hash` representing the target element. then if you refresh, there is no search scroll target in the url, only the element hash, so it would work as intended.

does that make sense @PiotrKozlowski i'm not sure how different what i'm describing is from what you already have, i think its pretty close

PiotrKozlowski:
@TomWoodward Please take a look if this is what I've ment

PiotrKozlowski:
@TomWoodward Let me confirm that I understand correctly things which you'd like me to do:
- remove current behavior for storing selected search results in the location state
- store search result data, target highlight id (when clicking on go to highlight) and target element id (for ex. figure id) in the URL as encoded JSON, for example:
`url#targetId` for target element id
`url?target={highlightId: string}` for target highlight id
`url?target={query: string, elementId: string, highlightIndex string}` - we will have to update it after clicking on search result
- `getScrollTarget` method from `highlightManager` and `searchHighlightManager` will return `scrollToFunction` if they found proper scroll target - similar as they do now
- `scrollTargetManager` will not change a lot
- After opening an URL with the selected result I'll query search results and select the proper one

TomWoodward:
@PiotrKozlowski the search query url param controls if the search should be loaded and shown in the sidebar, not necessarily implying any particular target, so i think that is fine to leave in a separate param. some examples:

`url#targetId` maybe you clicked a figure link, its just a target id

`url?target={"type":"highlight","id":"uuid"}#elementid` a selected highlight, the element id in the hash will be the reference element of the highlight data. the javascript can ignore this hash, but the browser will use it to set the default scroll near the highlight, which is nice. a navigation selector can be used to extract the "scroll target" which just returns the parsed json. other logic (for instance in the highlight manager) will look at the type to see if it should try to handle the target, or we could have a mapping (this type of target, use this manager to handle it)

`url?query=number` when the page loads the search will be run and the results sidebar displayed. i thought this was working already, but it seems to not be.

`url?query=somequery&target={"type":"search","index":1}#elementid` the previous logic will handle triggering the search, but when the results are received it should resolve the element/index combination into a particular result and select that. i wouldn't be opposed to storing the "Selected search result" as element/index instead of the entire result object in state so that the selectors work more nicely. it sucks a little that if any of the 3 pieces of this url are missing the scroll target won't work, but i think it might be ok.

i'm a little bit brainstorming here. my goal is to create a single place that defines the scroll target so all logic can look there and there can never be more than one defined. at the same time i'm conflicted about leveraging the browser hash behavior to our benefit because the page does load several seconds before javascript will figure the rest of the logic out. for instance the search result url could be `url?query=somequery#{"type":"search","index":1,"element":"elementid"}` putting the json in the hash, now there is definitively one place to find the scroll target, the hash, it'll either be an id or json, and its pretty easy to figure that out. but it loses the browser handling of the hash for the json types.

with regard to the "manager" code, i want to make a system where there are "types" of scroll targets, and the logic for how to find/scroll to those targets are contained in their manager with not external logic having specific logic of any type of target and having no crazy switching logic surrounding the type managers.

PiotrKozlowski:
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/rex-web/702

PR #515, opened 2020-02-20
Call Sentry if attaching highlight fails

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/875

TODO:
- [x] Add tests


src/app/content/components/utils/attachHighlight.ts
```diff
@@ -0,0 +1,18 @@
+import Highlighter, { Highlight, SerializedHighlight } from '@openstax/highlighter';
+import Sentry from '../../../../helpers/Sentry';
+
+const attachHighlight = (
+  highlight: Highlight | SerializedHighlight,
+  highlighter: Highlighter
+) => {
+  if (highlight instanceof SerializedHighlight) {
```
TomWoodward: you don't have to recurse here, it does it internally

TomWoodward (changes requested):

src/app/content/components/utils/attachHighlight.ts
PiotrKozlowski: @TomWoodward It does, but `SerializedHighlight` has no `isAttached` method. I could write this as: (is this better)?
```
  highlighter.highlight(highlight);
  const hl = highlight instanceof SerializedHighlight
    ? highlighter.getHighlight(highlight.id)
    : highlight;
  if (!hl || !hl.isAttached()) {
    Sentry.captureException(`Highlight with id: ${highlight.id} has not been attached.`);
  }
```

PiotrKozlowski: I've updated the code, but it looks like some tests from `highlightManager.spec.tsx` are failing
Probably some mocks needs updates

TomWoodward: src/app/content/components/Page/highlightManager.ts already calls highlighter.getHighlight(data.id); after it calls highlight, you could have your attachHighlight utility do that and return the highlight to save it the trouble

src/app/content/components/utils/attachHighlight.ts
```diff
@@ -0,0 +1,17 @@
...
+const attachHighlight = (
+  highlight: Highlight | SerializedHighlight,
+  highlighter: Highlighter
+) => {
+  highlighter.highlight(highlight);
+  const hl = highlight instanceof SerializedHighlight
+    ? highlighter.getHighlight(highlight.id)
+    : highlight;
+  if (!hl || !hl.isAttached()) {
+    Sentry.captureException(`Highlight with id: ${highlight.id} has not been attached.`);
```
TomWoodward: captureException takes an `Error object`, i think you want `Sentry.error()` which takes a string

src/app/content/components/Page/highlightManager.ts
```diff
@@ -87,7 +88,7 @@ const highlightData = (services: Services) => (data: HighlightData) => {
 
   const serialized = SerializedHighlight.fromApiResponse(data);
 
-  highlighter.highlight(serialized);
+  attachHighlight(serialized, highlighter);
 
   return highlighter.getHighlight(data.id);
```
TomWoodward: you could have `attachHighlight` return the highlight object so you don't have to fetch it a second time here

src/app/content/search/utils.ts
```diff
@@ -149,7 +150,7 @@ export const highlightResults = (
     const hitHighlights = hit.highlight.visibleContent.map((highlightText, index) => {
       const highlights = getHighlightRanges(element, highlightText).map((range) => {
         const highlight = new Highlight(range.nativeRange, {content: range.toString()});
-        highlighter.highlight(highlight);
+        attachHighlight(highlight, highlighter);
```
TomWoodward: in the case of search highlights, these results will not have meaningful ids, so the error message is not really helpful

TomWoodward (changes requested):

src/app/content/search/utils.ts
PiotrKozlowski: What error message should I pass then?

TomWoodward: we want to be able to identify what result the problem happened with, but search results don't have a unique id. so i think the best we can do is include the search query, book, and page. maybe the index of the result in the results for the page too?

PiotrKozlowski: ok, let's do this in a separate card then

TomWoodward (approved):

PR #516, opened 2020-02-20
Collect sentry error IDs

janKozie1:
for: https://github.com/openstax/unified/issues/871
I thought about adding how the error IDs are displayed after https://github.com/openstax/rex-web/pull/485 gets merged


TomWoodward (changes requested):
it looks like sentry has a `beforeSend` option: https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/options.ts#L113 this will allow us to capture event details for all events that are sent, not just the ones that we trigger manually. you'd assign that in `/helpers/Sentry.ts` when it does `Sentry.init`. you can also modify the `initializeWithMiddleware` function to give you access to `dispatch` right in the sentry helper, so you won't have to do all this rigging to get the errors into the state, it can be done in one place right where sentry sends them.

src/app/errors/types.ts
```diff
@@ -1,4 +1,5 @@
 export interface State {
   code?: number;
-  error?: Error;
+  error?: Error ;
```
TomWoodward: space

src/helpers/Sentry.ts
```diff
@@ -1,26 +1,48 @@
...
-        new Integrations.Dedupe(),
-      ],
-      release: `rex@${config.RELEASE_ID}`,
-    });
-    IS_INITIALIZED = true;
-    return createSentryMiddleware(Sentry);
+  initializeWithMiddleware(): Middleware {
+    return (store) => {
+      Sentry.init({
+        beforeSend: (event: Sentry.Event) => {
```
TomWoodward: i see you filtering out the `recordSentryMessage` actions from the breadcrumbs sent to sentry, is that because this `beforeSend` is called with each breadcrumb? i as thinking it would only be called with errors, if its also called with breadbrumbs we might want to do some more filtering so only error event ids are saved

TomWoodward (changes requested):

TomWoodward:
you've got some failing tests in src/helpers/Sentry.spec.ts

src/helpers/Sentry.ts
janKozie1: Well every action gets added as a breadcrumb in the [middleware](https://github.com/vidit-sh/redux-sentry-middleware/blob/4d05e62f843441aa5015221bb0cadb399c8f3894/src/main.js#L45) and based on `Sentry supports a concept called Breadcrumbs, which is a trail of events which happened prior to an issue.` (from their documentation) i assumed that it clears them after it saves (?) an issue.

janKozie1: I'm not sure how the actual error raport looks, but isnt it fine if there is an error with breadcrumb of actions that lead to it? We could add `&& event.level === 'error'` to the if in `beforeSend` so when users screenshot the error modal they will show the error id only which will have all the breadcrumbs saved wherever sentry saves them

TomWoodward: i want all sentry events to be captured by this logic, but breadcrumbs are not events. a breadcrumb is created for every redux action that is dispatched, and that will be way too much noise for the ids being shown in the error dialog

janKozie1: I think it works better now, but I'm not sure if actions should be included in breadcrumbs, and if yes then which ones.

TomWoodward: i'm confused, i don't think we have to filter breadcrumbs at all. the existing behavior has breadcrumbs for all redux actions, and we want to keep that.

but its important that we only store the ids for actual error events. for one because storing the ids for breadcrumbs would be super noisy, but also because the action to store the id would create an infinite loop.

TomWoodward: i just did some testing and it seems like the beforeSend function is not called with breadcrumbs, its only called with errors, so i don't think any filtering is necessary

janKozie1: And because of that im pretty sure only error ids are being saved

src/helpers/Sentry.ts
```diff
@@ -1,26 +1,51 @@
...
+const importantActions: Set<AnyAction['type']> = new Set([
+  recordError,
+].map(getType));
+
+const filterBreadcrumbActions = (action: AnyAction) => importantActions.has(action.type);
+
+export const onBeforeSend = (store: MiddlewareAPI) => (event: Sentry.Event) => {
+  const { event_id, exception} = event;
+
+  if (event_id && exception)  {
```
TomWoodward: pretty sure you don't need this filtering

src/helpers/Sentry.ts
```diff
@@ -1,26 +1,51 @@
...
+          new Integrations.ExtraErrorData(),
+          new Integrations.CaptureConsole(),
+          new Integrations.Dedupe(),
+        ],
+        release: `rex@${config.RELEASE_ID}`,
+      });
+      IS_INITIALIZED = true;
+
+      return createSentryMiddleware(Sentry, {
+        filterBreadcrumbActions,
```
TomWoodward: i don't think we want to filter breadcrumbs either

TomWoodward (changes requested):

TomWoodward:
this looks good, need to add the ids to the dialog before we can merge though

TomWoodward:
@janKozie1 please try to finish this one up before starting anything new. it seems very close to being done and would be nice to get it off the board

janKozie1:
@TomWoodward i've mentioned it before, I thought about waiting for https://github.com/openstax/rex-web/pull/485, should i not?

TomWoodward:
@janKozie1 ah ok, yea that makes sense before touching the dialog again

TomWoodward:
@janKozie1 openstax/rex-web#485 is merged now, you should be able to proceed with adding those ids to the UI

janKozie1:
@TomWoodward Is there going to be a design for it? Or just print the IDs in a column?

PiotrKozlowski (commented):
This pr looks good to me
I'm not sure about naming `recordSentryMessage` because you are actually storing event ids, not messages
There could be also some kind of header above the list of ids in the popup so users are not confused about what it is

TomWoodward:
i wasn't expecting the ids to be so long, and for some reason i thought the modal was bigger than it is.

i had wanted to comma separate the ids, instead of putting each on new lines, this doesn't have to be super readable, is just so that when we get screenshots of the error message we have something to follow up on.

additionally, since the modal is so small, maybe we should limit the number shown in the UI, maybe to whatever fits on two lines. leave them all in the redux state, we should be able to find the complete list in sentry there.

changing the font color would help a lot with helping users to ignore this, maybe the light gray that we use for disabled checkboxes would be good.

janKozie1:
Seems like they are always 32 characters long. I've made them little more visible than checkboxes (from opacity 0.4 to 0.6) because i felt like they draw less attention this way.

src/app/errors/components/ErrorModal.tsx
```diff
@@ -7,26 +7,36 @@ import { supportCenterLink } from '../../components/Footer';
...
 
 // tslint:disable-next-line:variable-name
 const BodyErrorText = styled.div`
   padding: ${modalPadding * 0.5}rem 0;
 `;
 
+// tslint:disable-next-line:variable-name
+const ErrorList = styled.div`
+  ${labelStyle}
+  font-size: 0.90rem;
```
PiotrKozlowski, line 24: `0.9rem;`?

PiotrKozlowski (approved):

TomWoodward (approved):

PR #524, opened 2020-02-25
Issue 817

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/817

- [x] Change <kbd>Buy book</kbd> from `<a>` to `<PlainButton>` 
- [x] Outline for focused filter when tabbing trough them is now visible


TomWoodward (changes requested):
@PiotrKozlowski i'm sorry this safari situation is a mess.

i didn't like the idea of changing this anchor into a button, because really its a link. after some research i found this https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html and learned that in _addition_ to the system setting you _also_ have to change a safari setting in order to actually tab to all things. this is crazy. after changing this setting the link focus works for me.

your change to the focus on the filters looks good. 

can you remove the changes to the buy link (i'm sorry) and then we can get qa to review this?

PiotrKozlowski:
@TomWoodward np, done.

TomWoodward (approved):
sending to qa

PR #525, opened 2020-02-25
Issue 816

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/816


src/app/content/highlights/components/SummaryPopup/Filters.tsx
```diff
@@ -71,27 +108,28 @@ const Filters = ({className}: Props) => <div className={className}>
...
-      ${theme.breakpoints.mobile(css`
-        margin-right: 4.8rem;
-      `)}
+    ${Dropdown} {
+      & > *:not(${DropdownToggle}) {
+        top: calc(100% - 0.1rem);
+        box-shadow: 0 0 0.6rem 0 rgba(0, 0, 0, 0.2);
+        max-height: calc(100vh - 26rem);
+        ${theme.breakpoints.mobile(css`
+          max-height: calc(100vh - 19.6rem);
```
TomWoodward: i assume `0.1rem` is the bottom border width, `26rem` is the top and bottom space around the modal for desktop and `19.6rem` is the top and bottom space for mobile, but all of those magic numbers need to be extracted or derived from the other style constants

src/app/content/highlights/components/SummaryPopup/Filters.tsx
```diff
@@ -28,17 +26,49 @@ const HighlightsPrintButton = styled(PrintButton)`
...
+const Toggle = styled(React.forwardRef<HTMLButtonElement, ToggleProps>(
   ({label, ...props}, ref) => <PlainButton ref={ref} {...props}>
     <div tabIndex={-1}>
       {label}
       <DownIcon />
     </div>
   </PlainButton>
 ))`
+  position: relative;
+  padding: 2rem 2.4rem;
```
TomWoodward: i think there are constants for some of these things you're hardcoding

src/app/content/highlights/components/SummaryPopup/FiltersList.tsx
```diff
@@ -125,8 +125,12 @@ export default styled(FiltersList)`
   flex-wrap: wrap;
   width: 100%;
   margin: 0;
-  padding: 0.4rem 0;
+  padding: 0 2.8rem 1rem 2.8rem;
+  margin-top: -1rem;
   list-style: none;
+  ${theme.breakpoints.mobile(css`
+    padding: 0 2.4rem 0.4rem 2.4rem;
```
TomWoodward: a little bit of inline numbers is understandable, but you've got a lot of them, i feel like a bunch should be derived from constants

src/app/content/highlights/components/SummaryPopup/__snapshots__/ChapterFilter.spec.tsx.snap
```diff
@@ -252,8 +251,7 @@ input:checked + .c10 .c11 {
 
 @media screen and (max-width:75em) {
   .c0.c0.c0 {
-    left: -1.6rem;
-    max-width: calc(100vw - 1.6rem);
+    max-width: calc(100vw - 1.6rem - 3.2rem);
```
TomWoodward: this rem subtraction can probably be done in js

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/Filters.tsx
PiotrKozlowski: It felt to me that hardcoding them would be easier to maintain, but this might not be the case. I've extracted the most required to the constants.

TomWoodward: if they're one off numbers i think its fine to hardcode them. but i felt like these were derivations on the modal padding and should probably be generated by some math involving those numbers, that way if we change the modal positioning we won't have to update the filters too

src/app/content/highlights/components/SummaryPopup/__snapshots__/ChapterFilter.spec.tsx.snap
```diff
@@ -252,8 +251,7 @@ input:checked + .c10 .c11 {
 
 @media screen and (max-width:75em) {
   .c0.c0.c0 {
-    left: -1.6rem;
-    max-width: calc(100vw - 1.6rem);
+    max-width: calc(100vw - 4.800000000000001rem);
```
TomWoodward: please round this or find some other way to make the floating point math not produce a crazy number

TomWoodward (changes requested):
the background color of the filter toggle and dropdown aren't the same. please double check the designs or ask devika if there is something we should change there to make it look less weird.

TomWoodward (changes requested):
updates look good. i realized i was wrong before, the background colors of the dropdown and toggle are the same, but there is some shadow visible between the two which messes with my eyes and makes them look different. we should make it so there is no shadow visible between the toggle and dropdown so they look more connected

TomWoodward (changes requested):
i can still see a line between the two

PiotrKozlowski:
@TomWoodward Can you provide a screenshot? I'm not seeing this line in Firefox and Chrome. Is it now smaller than previously?

TomWoodward:
@PiotrKozlowski yea its much lighter, but still there

PiotrKozlowski:
@TomWoodward Please check now, I've moved the shadow one more px to the top.

TomWoodward:
@PiotrKozlowski i've been playing with the shadow parameters and it really doesn't seem possible to completely eliminate the shadow on one side by moving it to the other side. brief googling (https://stackoverflow.com/questions/14067358/remove-right-side-of-box-shadow) suggests using clip-path along the bottom or removing the shadow from the button and adding the shadow to a `:before` element with 0 width along the right side of the area.

the `:before` thing is probably more portable, but i'll leave it to you

PiotrKozlowski:
@TomWoodward `clip-path` seems to work perfectly in this situation. I've also removed shadow from the top side.

TomWoodward (approved):
perfection

TomWoodward (approved):

PR #527, opened 2020-02-27
Sync logging out between tabs

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/782

TODO:
- [x] Fix tests


src/gateways/createHighlightClient.ts
```diff
@@ -7,6 +9,8 @@ const formatError = (response: Response) => {
     });
 };
 
+export class DoNotHandleMe extends Error {}
```
TomWoodward: please rename this `UnauthenticatedError`, move it to a more generic location, and move the `dispatch(receiveLoggedOut())` to the middleware in app/utils

src/app/utils.ts
```diff
@@ -27,6 +28,9 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
 
       if (matches(action)) {
         const catchError = (e: Error) => {
```
TomWoodward: could be worth extracting this `catchError` function just to reduce complexity in this function

TomWoodward (changes requested):

src/app/types.ts
```diff
@@ -70,3 +70,5 @@ export type ActionHookBody<C extends AnyActionCreator> = (helpers: MiddlewareAPI
 // helpers
 export type ArgumentTypes<F> = F extends (...args: infer A) => any ? A : never;
 export type FirstArgumentType<F> = F extends (first: infer A, ...args: any) => any ? A : never;
+
+export class UnauthenticatedError extends Error {}
```
TomWoodward: this isn't a typescript type, it shouldn't be in here

src/app/utils.ts
```diff
@@ -26,23 +28,28 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
...
       if (matches(action)) {
-        const catchError = (e: Error) => {
-          Sentry.captureException(e);
-          stateHelpers.dispatch(recordError(e));
-        };
         try {
           const promise = boundHook(action);
           if (promise) {
-            services.promiseCollector.add(promise.catch(catchError));
+            services.promiseCollector.add(promise.catch((e) => catchError(e, stateHelpers.dispatch)));
```
TomWoodward: you could curry this ahead of time so you don't have to pass it in two places

```
const catchError = makeCatchError(stateHelpers);
```

```suggestion
            services.promiseCollector.add(promise.catch(catchError));
```

TomWoodward (changes requested):
in the past there have been problems extending from the base Error class (https://stackoverflow.com/a/35858868) please make sure that this is not still a problem

src/gateways/createHighlightClient.spec.ts
```diff
@@ -1,7 +1,12 @@
 import { GetHighlightsSourceTypeEnum } from '@openstax/highlighter/dist/api';
+import { UnauthenticatedError } from '../app/types';
 import { resetModules } from '../test/utils';
 import createHighlightClient from './createHighlightClient';
 
+jest.mock('../app/utils', () => ({
+  ...jest.requireActual('../app/utils'),
+}));
+
```
janKozie1: Is this necessary? Won't `jest.mock('../app/utils')` work?

TomWoodward: actually this is mocking it with the actual module, so wouldn't just not mocking it do the same?

PiotrKozlowski:
@TomWoodward 
> In the past there have been problems extending from the base Error class (https://stackoverflow.com/a/35858868) please make sure that this is not still a problem

We also had this problem with `adaptarr` but it doesn't seem to be a problem when anymore. Probably because we are targeting `"target": "es2017"` and babel doesn't have to compile this code in any way?

src/app/utils.ts
```diff
@@ -26,10 +27,7 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
       const result = next(action);
 
       if (matches(action)) {
-        const catchError = (e: Error) => {
-          Sentry.captureException(e);
-          stateHelpers.dispatch(recordError(e));
-        };
+        const catchError = makeCatchError(stateHelpers.dispatch);
```
TomWoodward: can you hoist this up to line 22ish?

TomWoodward (changes requested):
i left one little comment but i'm sending this to qa

TomWoodward (approved):

PR #529, opened 2020-02-28
Wait for postProcess on mount

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/762

I can't replicate this issue locally, but I think this change might help


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -41,8 +41,8 @@ export default class PageComponent extends Component<PagePropTypes> {
     return cleanContent.body.innerHTML;
   };
 
-  public componentDidMount() {
-    this.postProcess();
+  public async componentDidMount() {
```
TomWoodward: this can't be async because it results in `componentDidUpdate` being triggered before `componentDidMount` finishes processing.

if you look in `componentDidUpdate` it waits for the promise to finish before proceeding. i _think_ creating the managers in the `componentDidMount` doesn't actually add any highlights, but if i'm wrong you could be onto something here because it'll definitely happen before math is done rendering.

what i think is more likely though is that the promise returned by `typesetMath` is actually resolving before it should be. i say that only because math typesetting has a history of being janky. we could try adding something to `typesetDocumentPromise` in `src/helpers/mathjax.ts` before it resolves to have it check the document for math and introduce an additional wait before resolving if we find some

TomWoodward (changes requested):

src/app/content/components/Page/PageComponent.tsx
PiotrKozlowski: I couldn't replicate the issue about `componentDidUpdate` being triggered before `componentDidMount`, but I've checked that as you said, highlights are not rendered after `componentDidMount` so this `ascyn/await` makes no sense.

I've checked that `resolve` is called after all math is rendered (or I just can't replicate correctly this issue?). Anyway, I've added a recursive call to `typesetDocumentPromise` if some nodes weren't processed.

We could also think about migrating to MathJax v3.

PiotrKozlowski: Well... I didn't push it previously... Not it is there.

TomWoodward: @PiotrKozlowski we have not had problems with math remaining unrendered, so i think calling it recursively might cause issues. i was thinking more like we just wait for an amount of time after the promise resolves to see if some stuff is still completing

PiotrKozlowski: @TomWoodward  I've pushed a change, let me know if this is what you've meant

src/helpers/mathjax.ts
```diff
@@ -95,9 +95,26 @@ function typesetDocument(root: Element, windowImpl: Window) {
   );
 }
 
+let recursiveCounter = 0;
```
TomWoodward: this isn't gonna work if its global, you should pass it as `remainingTries` to the function, default it to 5 and subtract one each time you recurse

TomWoodward (changes requested):

src/helpers/mathjax.ts
```diff
@@ -95,9 +95,24 @@ function typesetDocument(root: Element, windowImpl: Window) {
   );
 }
 
+const resolveOrWait = (root: Element, resolve: () => void, remainingTries = 5) => {
+  if (
+    remainingTries
```
TomWoodward: i would make this 
```suggestion
    remainingTries > 0
```
for reduced possibility of infinite recursion (negative numbers are truthy)

TomWoodward (changes requested):
i left a comment, but i'm moving this to review anyway

TomWoodward (approved):

PR #530, opened 2020-02-28
Allow external content links in dev environment

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/868


src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -119,7 +119,7 @@ export const prevNextBookPage = (
   };
 };
 
-export const archiveTreeSectionIsBook = (section: LinkedArchiveTreeNode) => !section.parent;
+export const archiveTreeSectionIsBook = (section: LinkedArchiveTreeNode | undefined) => !section || !section.parent;
```
PiotrKozlowski: This means that `undefined` is a `Book`
Shouldn't it be `=> section && !section.parent`?

src/gateways/createArchiveLoader.ts
```diff
@@ -27,13 +28,19 @@ export default (url: string) => {
   };
 
   const extrasCache = new Map();
-  const getBookIdsForPage: (pageId: string) => Promise<string[]> = (pageId) => {
+  const getBookIdsForPage: (pageId: string) => Promise<Array<{id: string, bookVersion: string | undefined}>> =
```
PiotrKozlowski: I would write this type like this:
`Promise<{id: string, bookVersion?: string}[]>`
 because it is shorter

src/test/mocks/archiveLoader.ts
```diff
@@ -64,7 +64,9 @@ export default () => {
     return pages && pages[pageId];
   });
 
-  const getBookIdsForPage = jest.fn((_pageId: string) => Promise.resolve([] as string[]));
+  const getBookIdsForPage = jest.fn((_pageId: string) =>
+    Promise.resolve([] as Array<{id: string, bookVersion: string}>)
```
PiotrKozlowski: In `createArchiveLoader.ts` `bookVersion` can be also `undefined` so this mock probably should also reflect that

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -152,26 +152,54 @@ const resolvePage = async(
...
+      const archiveBook = await services.archiveLoader.book(id, bookVersion).load();
+
+      if (archiveBook && archiveTreeSectionIsBook(archiveBook.tree)) {
+        return {osWebBook, archiveBook};
+      }
+    }
+  }
+
+  return undefined;
+
```
PiotrKozlowski: empty line

src/app/content/hooks/locationChange/resolveContent.spec.ts
```diff
@@ -53,11 +80,15 @@ describe('locationChange', () => {
...
 
-    hook = require('./resolveContent').default;
+    const resolveContent = require('./resolveContent');
+    hook = resolveContent.default;
+    resolveExternalBookReference = resolveContent.resolveExternalBookReference;
   });
 
   describe('in development', () => {
-    jest.doMock('../../../../config', () => ({...mockConfig, APP_ENV: 'development'}));
+    beforeAll(() => {
```
PiotrKozlowski: Probably `beforeEach` or not needed at after all?

KarinaMendez2: not beforeEach, and also not needed, it is this way in master and works.

src/gateways/createArchiveLoader.ts
KarinaMendez2: Lint modified it, that's why it is written this way.

src/app/content/hooks/locationChange/resolveContent.spec.ts
PiotrKozlowski: It works without this in "in development", but if it is here for consistency I think it may be here

TomWoodward: `beforeAll` runs once before the block of tests, where `beforeEach` runs before each test. in this case `beforeAll` is fine, but it would still work with `beforeEach`

src/gateways/createArchiveLoader.ts
TomWoodward: `Array<{id: string, bookVersion: string | undefined}>` is the correct way

TomWoodward:
looks good other than piotr's comment https://github.com/openstax/rex-web/pull/530/files#r387729552, pretty sure hes right about that

PiotrKozlowski (approved):

TomWoodward:
@KarinaMendez2 there are conflicts that need resolving here

PR #535, opened 2020-03-03
Fix broken relative links

janKozie1:
for: https://github.com/openstax/unified/issues/894


src/app/content/utils/urlUtils.ts
```diff
@@ -91,6 +93,9 @@ export const toRelativeUrl = (from: string, to: string) => {
   // remove the last piece of the "to" so that it is always output
   const commonParts = getCommonParts(parsedFrom, parsedTo.slice(0, -1));
 
-  return '../'.repeat(parsedFrom.length - commonParts.length - 1)
+  const lastFromFragment = last(parsedFrom);
+  const hasTrailingSlash = lastFromFragment && last(parsedTo) !== lastFromFragment && /\/$/.test(from);
+
+  return '../'.repeat(parsedFrom.length - commonParts.length - 1 + (hasTrailingSlash ?  1 : 0))
```
PiotrKozlowski: This would be a little bit easier to read if it was something like this:
```
const hasTrailingSlash = /\/$/.test(from);
const valueToSubtract = hasTrailingSlash && (last(parsedTo) !== last(parsedFrom)
  ? 0
  : 1
return '../'.repeat(parsedFrom.length - commonParts.length - valueToSubtract)
...
```
`

TomWoodward (changes requested):
i see at the beginning of the function it calls `trimTrailingSlash`, would it work if you simply removed those calls?

src/app/content/utils.spec.ts
```diff
@@ -198,9 +198,4 @@ describe('toRelativeUrl', () => {
     const url = toRelativeUrl(PAGE_URL, PAGE_URL + '/');
     expect(url).toMatchInlineSnapshot(`"page1"`);
   });
-
-  it('ignores trailing slashes on the source', () => {
-    const url = toRelativeUrl(PAGE_URL + '/', PAGE_URL);
-    expect(url).toMatchInlineSnapshot(`"page1"`);
-  });
```
janKozie1: I don't think this is right. If the given url was `'book/a/page/b/'` and if `'b'` was returned then after clicking the relative url, the browser would navigate to `'book/a/page/b/b'`

src/app/content/utils/urlUtils.ts
```diff
@@ -80,15 +80,15 @@ const getCommonParts = (firstPath: string[], secondPath: string[]) => {
   return result;
 };
 
-const trimTrailingSlash = (path: string) => path.replace(/([^/]{1})\/+$/, '$1');
-
 export const toRelativeUrl = (from: string, to: string) => {
-  const parsedFrom = trimTrailingSlash(from).split('/');
-  const parsedTo = trimTrailingSlash(to).split('/');
+  const parsedFrom = from.split('/');
+  const parsedTo = to.replace(/\/$/, '').split('/');
```
TomWoodward: i don't think we want to be trimming the trailing slashes off of anything. if you leave this as it is passed in, and change all the logic below back to the way it started, i believe everything works the way we need it to

TomWoodward (changes requested):

src/app/content/utils/urlUtils.ts
janKozie1: Original logic assumed that the last part of `parsedTo` will be actually a part of url not an empty string. Because of that, there would need to be a check to how many parts to subtract or the part that repeats '../' would need to be adjusted

janKozie1: That is of course if there always should be a part of the actual url returned. I assumed that links like '../' are incorrect

janKozie1:
Not sure if this is what you requested, but i think it doesn't work because sometimes "" ends up as the last part of the `parsedTo`

src/app/content/utils/urlUtils.ts
```diff
@@ -80,15 +80,15 @@ const getCommonParts = (firstPath: string[], secondPath: string[]) => {
...
+  const parsedFrom = from.split('/');
+  const parsedTo = to.split('/');
 
   // remove the last piece of the "to" so that it is always output
   const commonParts = getCommonParts(parsedFrom, parsedTo.slice(0, -1));
 
-  return '../'.repeat(parsedFrom.length - commonParts.length - 1)
+  const valueToSubtract = commonParts.length + 1;
+
+  return '../'.repeat(parsedFrom.length - valueToSubtract)
```
TomWoodward: this logic is functionally the same as it was before, why change it?

TomWoodward (changes requested):

TomWoodward:
@janKozie1 can you give me an example of what you are describing in https://github.com/openstax/rex-web/pull/535#issuecomment-598893743

janKozie1:
@TomWoodward 
If `from === '/book/a/page/b'` and `to ===  '/book/a/page/b/'`
then `parsedFrom.length === 5` and `commonParts.length === 5`
because the part that is being "cut out" from the `to` is `''`

TomWoodward:
@janKozie1 i don't understand your example. if the urls are `from === '/book/a/page/b'` and `to === '/book/a/page/b/'` then the correct relative link would be `./` right? what are you getting?

janKozie1:
@TomWoodward with the original logic and just the trimming removed, the `value` in  `'../'.repeat(value)` will be invalid since common parts length is equal to from parts length and we are subtracting 1 more.

TomWoodward:
the extra `-1` in the logic is to stop `"page1"` from being `"../page1"`, which are technically the same, but the first one looks nicer

i think you should be able to put a `Math.max(0, ...)` around the length logic and it will work for either case

src/app/content/utils/urlUtils.spec.ts
```diff
@@ -173,11 +178,6 @@ describe('toRelativeUrl', () => {
 
   it('ignores trailing slashes on the target', () => {
```
TomWoodward: delete these tests about ignoring the trailing slash, that is no longer desired behavior

TomWoodward (changes requested):

src/app/content/utils/urlUtils.spec.ts
janKozie1: I've removed them

TomWoodward (approved):

PR #537, opened 2020-03-05
Add aria labels to MH - chapter and section titles

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/848


PiotrKozlowski (approved):

TomWoodward (approved):
aria labels are supposed to be human readable labels for screen readers to generate text for things that would otherwise be impossible for it to derive meaning from, for instance buttons with only icons inside.

if you can't come up with something useful to put in an `aria-label` and you just need some kind of something on it to help the automated tests, use a `data-testid` instead

TomWoodward (changes requested):
previous comment

TomWoodward (approved):

PR #540, opened 2020-03-06
Fix resources dynamic archive

TomWoodward:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/897

when archive was set dynamically, such as on `/books/4dd6fbe2-bbc4-4e67-8e5e-760bf0803bc5@9.1/pages/1-6-your-safety/?archive=https://archive-content03.cnx.org` the resources would try to load via the original proxy, and might not have existed there. this change makes the resources use the react archive url if its set.

### DIFF
--- a/src/app/content/components/Page/contentDOMTransformations.ts
+++ b/src/app/content/components/Page/contentDOMTransformations.ts
@@ -1,5 +1,6 @@
-import { Document, HTMLButtonElement, HTMLElement } from '@openstax/types/lib.dom';
+import { Document, HTMLButtonElement, HTMLElement, HTMLImageElement } from '@openstax/types/lib.dom';
 import { IntlShape } from 'react-intl';
+import { REACT_APP_ARCHIVE_URL } from '../../../../config';
 import { assertNotNull } from '../../../utils';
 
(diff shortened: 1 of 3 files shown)


src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -117,3 +119,9 @@ function wrapSolutions(rootEl: HTMLElement, intl: IntlShape) {
     `;
   });
 }
+
+function prefixResources(rootEl: HTMLElement) {
+  rootEl.querySelectorAll<HTMLImageElement>('img[src^="/resources/"]').forEach(
+    (el) => el.src = REACT_APP_ARCHIVE_URL + el.getAttribute('src')
+  );
+}
```
janKozie1: Can only images be resources? If not, maybe the function's name could reflect that it only picks images?

janKozie1 (approved):

src/app/content/components/Page/contentDOMTransformations.ts
TomWoodward: i think other things can load from resources, but i couldn't find any while i was working on this

PR #541, opened 2020-03-09
Safari bug - mh scroll

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/833

### DIFF
--- a/src/app/components/AllOrNone.tsx
+++ b/src/app/components/AllOrNone.tsx
@@ -24,8 +24,7 @@ export default styled(AllOrNone)`
   &,
   ${ButtonLink} {
     font-size: 1.4rem;
-    margin: 0;
-    overflow: hidden;
+    overflow: visible;
   }
 
   height: 2rem;
--- a/src/app/components/AllOrNone.tsx
+++ b/src/app/components/AllOrNone.tsx
@@ -24,6 +24,8 @@ export default styled(AllOrNone)`
   &,
   ${ButtonLink} {
     font-size: 1.4rem;
+    margin: 0;
+    overflow: hidden;
   }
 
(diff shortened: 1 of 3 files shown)


src/app/components/AllOrNone.tsx
```diff
@@ -24,6 +24,8 @@ export default styled(AllOrNone)`
   &,
   ${ButtonLink} {
     font-size: 1.4rem;
+    margin: 0;
```
PiotrKozlowski: In the `ChapterFilter` line 115 there is also margin for `AllOrNone` which happens to have bigger priority so this `margin: 0` is making a difference.
I think you can remove this line, it looks good without this.

src/app/components/AllOrNone.tsx
```diff
@@ -24,6 +24,8 @@ export default styled(AllOrNone)`
...
+    overflow: hidden;
```
PiotrKozlowski: Can this be `overflow: visible;`?
I've spotted this issue also on my pr: https://github.com/openstax/rex-web/pull/525 in the Edge browser and there only `overflow: visible` works

PiotrKozlowski (changes requested):

src/app/components/AllOrNone.tsx
KarinaMendez2: image

this overflow is what brakes the buttons in safari, `overflow: hidden` fixes it

KarinaMendez2: I guess we can make it visible too

PiotrKozlowski (approved):

PR #542, opened 2020-03-09
Show context menu - MH popup

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/861


PiotrKozlowski:
This is how it displays for me (Firfox Ubuntu) when `overflow` is set to `hidden`
image

and when `overflow` is set to `visible`
image

I also think `margin` for the last element is not necessary because when the context menu will be opened and overflow will be set to visible it should automatically expand. If it is not then this probably is some weird cross-browser issue and you can ignore this comment

PiotrKozlowski (approved):
Some snaps need to be updated, code looks good

PR #551, opened 2020-03-18
Show no senior contributing authors

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/921

### DIFF
--- a/src/app/content/components/Attribution.spec.tsx
+++ b/src/app/content/components/Attribution.spec.tsx
@@ -150,5 +150,29 @@ describe('Attribution', () => {
 
       expect(message).toEqual('BUG: Could not find publication date');
     });
+
+    it('displays first two contributing authors when no senior authors were found', async() => {
+      const newState = (cloneDeep({
+        content: {
+          ...initialState,
+          book: formatBookData(book, { ...mockCmsBook,
+            authors: [
+              {value: {name: 'Jhon Doe', senior_author: false}},
+              {value: {name: 'Jonny Doe', senior_author: false}},
+            ],
+          }),
+        },
+      }) as any) as AppState;
+
+      store = createTestStore(newState);
+      const { node } = renderToDom(render());
+      const details = node;
+      details.setAttribute('open', '');
+
+      // wait for dom events to do their thing
+      await new Promise((resolve) => setTimeout(resolve, 1));
+
+      expect(details.children[1].innerHTML).toMatch(`Authors: Jhon Doe, Jonny Doe`);
+    });
   });
 });
--- a/src/app/content/components/Attribution.spec.tsx
+++ b/src/app/content/components/Attribution.spec.tsx
@@ -165,7 +165,7 @@ describe('Attribution', () => {
       }) as any) as AppState;
 
       store = createTestStore(newState);
-      const { node: details } = await renderToDom(render());
+      const { node: details } = renderToDom(render());
       details.setAttribute('open', '');
 
       expect(details.children[1].innerHTML).toMatch(`Authors: Jhon Doe, Jonny Doe`);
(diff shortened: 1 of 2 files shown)


src/app/content/components/Attribution.spec.tsx
```diff
@@ -150,5 +150,29 @@ describe('Attribution', () => {
...
+              {value: {name: 'Jonny Doe', senior_author: false}},
+            ],
+          }),
+        },
+      }) as any) as AppState;
+
+      store = createTestStore(newState);
+      const { node } = renderToDom(render());
+      const details = node;
+      details.setAttribute('open', '');
```
PiotrKozlowski: Reassign `node` to `details` seems unnecessary. If you want to use different name you can use:
```
const { node: details } = renderToDom(render());
details.setAttribute('open', '');
```

src/app/content/components/Attribution.spec.tsx
```diff
@@ -150,5 +150,29 @@ describe('Attribution', () => {
...
+
+      // wait for dom events to do their thing
+      await new Promise((resolve) => setTimeout(resolve, 1));
```
PiotrKozlowski: It should work also with `0` instead of `1`
I'm not sure, but maybe there is a way to just await at `await renderToDom(render());` or `renderToDom(await render());`?

src/app/content/components/Attribution.tsx
```diff
@@ -161,8 +161,14 @@ class Attribution extends Component<Props> {
 
     const seniorAuthors = book.authors.filter((author) => author.value.senior_author);
 
+    const authorsToDisplay = () => {
+      return seniorAuthors.length > 0 ?
+        seniorAuthors.map(({value: {name}}) => name).join(', ')
+        : book.authors.slice(0, 2).map(({value: {name}}) => name).join(', ');
+    };
```
PiotrKozlowski: I think this doesn't have to be a function and can be transformed into:
```
 const authorsToDisplay = seniorAuthors.length > 0 ?
      seniorAuthors.map(({value: {name}}) => name).join(', ')
      : book.authors.slice(0, 2).map(({value: {name}}) => name).join(', ')
;
```

src/app/content/components/Attribution.spec.tsx
```diff
@@ -165,13 +165,9 @@ describe('Attribution', () => {
       }) as any) as AppState;
 
       store = createTestStore(newState);
-      const { node } = renderToDom(render());
-      const details = node;
+      const { node: details } = await renderToDom(render());
```
PiotrKozlowski: Actually `yarn test:unit` works for me even without this `await` statement so you are probably okay with removing this

PiotrKozlowski (approved):
I've left one comment, besides this everything looks good

src/app/content/components/Attribution.tsx
```diff
@@ -161,8 +161,12 @@ class Attribution extends Component<Props> {
 
     const seniorAuthors = book.authors.filter((author) => author.value.senior_author);
 
+    const authorsToDisplay = seniorAuthors.length > 0 ?
+      seniorAuthors.map(({value: {name}}) => name).join(', ')
+      : book.authors.slice(0, 2).map(({value: {name}}) => name).join(', ');
```
TomWoodward: you could leave the map below where it was and not have it here twice. the logic would reduce to 
```
const authorsToDisplay = seniorAuthors.length > 0
  ? seniorAuthors
  : book.authors.slice(0, 2)
```

TomWoodward (changes requested):

TomWoodward (approved):

PR #555, opened 2020-03-19
Allow routing by page uuid

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/863


src/app/content/utils/urlUtils.ts
```diff
@@ -25,19 +25,16 @@ export const getBookPageUrlAndParams = (
...
 };
 
 export const getUrlParamsForBook = (
   book: Pick<Book, 'id' | 'tree' | 'title' | 'version'> & Partial<{slug: string}>
 ): Params['book'] => {
   if ('slug' in book && book.slug && BOOKS[book.id]) {
-    return {slug: book.slug};
+    return book.version === BOOKS[book.id].defaultVersion
+      ? {slug: book.slug}
+      : {slug: book.slug, version: book.version};
```
janKozie1: This is not exactly related to this PR but I've noticed that content links dont get properly versioned, so i included it here

janKozie1: I also thought maybe about making a map that `toRelativeUrl` could use because if you navigate to a book with an uuid, and that books has a slug, then the urls will be '../../slug/page/pageThatChanges'. It works but just thought it might be nice.

src/app/content/routes.ts
```diff
@@ -1,15 +1,36 @@
...
+  const paramKeys: Array<keyof typeof params> = Object.keys(params);
+  const keyToInject = paramKeys[0];
+
+  return params[keyToInject].reduce((output, value) => {
+    const injected = baseUrl.replace(new RegExp(`(?<=:)${keyToInject}`), value);
+    return paramKeys.length === 1
+      ? [...output, injected]
+      : [...output, ...injectParamsToBaseUrl(injected, omit([keyToInject], params))];
+  }, [] as string[]);
+};
```
janKozie1: I'm not sure how to make it work if it's in `urlUtils` since I think the `react-loadable` makes the file load separately?

KarinaMendez2: this is probably a @TomWoodward question

KarinaMendez2 (dismissed):

src/app/content/utils/urlUtils.ts
```diff
@@ -47,20 +46,29 @@ export const getUrlParamForPageId = (book: Pick<Book, 'id' | 'tree' | 'title'>,
...
 
   const treeSection = findArchiveTreeNode(book.tree, pageId);
   if (!treeSection) {
     throw new Error(`BUG: could not find page "${pageId}" in ${book.title}`);
   }
-  const result = assertDefined(treeSection.slug, `could not find page slug for "${pageId}" in ${book.title}`);
-  getUrlParamForPageIdCache.set(cacheKey, result);
 
-  return {slug: result};
+  if (APP_ENV === 'production' || treeSection.slug) {
```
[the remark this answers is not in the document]
janKozie1: The issue says `if a page slug is undefined in the book tree and the environment is not production`, so i assumed that generating the params with slug is the preferred way & it should still throw if there is no slug in production.

src/app/navigation/utils.ts
```diff
@@ -19,7 +19,7 @@ export const locationChangeForRoute = <R extends AnyRoute>(
 ): locationChange is Required<LocationChange<Match<R>>> =>
   !!locationChange.match && locationChange.match.route.name === route.name;
 
-export const getUrlRegexParams = <T extends {}>(obj: T): T => flatten(obj, {delimiter});
+export const getUrlRegexParams = <T extends {}>(obj: T): {} => flatten(obj, {delimiter});
```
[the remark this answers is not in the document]
janKozie1: I think it was wrongly typed. It implied that flattened object will be the same as the original one. I changed it to `object` now since I've read that `{}` also allows primitives

src/app/content/utils/urlUtils.ts
```diff
@@ -90,3 +99,11 @@ export const toRelativeUrl = (from: string, to: string) => {
   return '../'.repeat(Math.max(0, parsedFrom.length - commonParts.length - 1))
     + parsedTo.slice(commonParts.length).join('/');
 };
+
+export const findPathForParams = (params: object, paths: string[]) => {
```
TomWoodward, line 103: move this, and the guard, into `navigation/utils` and `navigation/guards`

this file is for content urls, but these two functions apply to any routing

src/app/content/routes.ts
```diff
@@ -1,15 +1,37 @@
...
+ * Recursively creates combinations of supplied replacements
+ * for the base parameter in an url
+ */
+
+const injectParamsToBaseUrl = (baseUrl: string, params: {[key: string]: string[]}): string[] => {
+  const paramKeys: Array<keyof typeof params> = Object.keys(params);
+  const keyToInject = paramKeys[0];
+
+  return params[keyToInject].reduce((output, value) => {
+    const injected = baseUrl.replace(new RegExp(`(?<=:)${keyToInject}`), value);
```
TomWoodward, line 24: does this really need the regex? can't you just do
```suggestion
    const injected = baseUrl.replace(`:${keyToInject}`, value);
```

src/app/content/routes.ts
```diff
@@ -1,15 +1,37 @@
...
+    return paramKeys.length === 1
```
TomWoodward, line 25: i would lean towards having `injectParamsToBaseUrl` return `[baseUrl]` if params is empty at the top of its logic, and removing this conditional to simplify the logic to
```
return [...output, ...injectParamsToBaseUrl(injected, omit([keyToInject], params))]
```

src/app/content/routes.ts
```diff
@@ -1,15 +1,37 @@
...
-const CONTENT_PATH = '/books/:book_slug/pages/:page_slug';
-const UUID_CONTENT_PATH = `/books/:book_uuid(${MATCH_UUID})@:book_version/pages/:page_slug`;
-const VERSIONED_CONTENT_PATH = '/books/:book_slug@:book_version/pages/:page_slug';
+/*
+ * Recursively creates combinations of supplied replacements
+ * for the base parameter in an url
+ */
+
+const injectParamsToBaseUrl = (baseUrl: string, params: {[key: string]: string[]}): string[] => {
+  const paramKeys: Array<keyof typeof params> = Object.keys(params);
```
TomWoodward, line 20: this is unused, you might as well do
```suggestion
  const keyToInject = Object.keys(params)[0];
```

src/app/content/routes.ts
TomWoodward: `react-loadable` isn't the problem. the problem is that there is a dependency loop between the `routes` and `urlUtils` file, everything works fine as long as the logic is in functions that only execute after everything has been mapped up, but with logic in the root of the file your modules will be `undefined`.

but this function would be better in `navigation/utils` anyway, so that should work better for you

src/app/content/utils/urlUtils.ts
TomWoodward: if you navigate to a book with a uuid and it renders content links with the slug that might be a bug, but lets not worry about this for now.

src/app/developer/components/__snapshots__/Home.spec.tsx.snap
```diff
@@ -359,9 +360,18 @@ Array [
...
             /books/:book_uuid([\\da-z]{8}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{12})@:book_version/pages/:page_slug
+            <br />
+            /books/:book_slug@:book_version/pages/:page_uuid([\\da-z]{8}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{12})
+            <br />
             /books/:book_slug@:book_version/pages/:page_slug
+            <br />
+            /books/:book_slug/pages/:page_uuid([\\da-z]{8}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{4}-[\\da-z]{12})
+            <br />
             /books/:book_slug/pages/:page_slug
+            <br />
```
TomWoodward, line 374: bout time someone did this

TomWoodward (changes requested):

TomWoodward (approved):

TomWoodward:
@janKozie1 there is a test failing in the contentLinkHandler, it seems like maybe a logical conflict with another pr that was merged

TomWoodward (approved):

PR #556, opened 2020-03-23
Let Content Instance Persist in Links

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/917


src/app/content/components/ContentLink.tsx
```diff
@@ -2,8 +2,7 @@ import flow from 'lodash/fp/flow';
 import React from 'react';
 import { connect } from 'react-redux';
 import styled from 'styled-components/macro';
-import { linkStyle } from '../../components/Typography';
-import { push } from '../../navigation/actions';
+import { linkStyle } from '../../components/Typography'; import { push } from '../../navigation/actions';
```
PiotrKozlowski: upps :D

src/app/content/routes.ts
```diff
@@ -24,6 +25,10 @@ export const content: Route<Params, State> = {
     loading: () => null,
     modules: ['Content'],
   }),
+  getSearch: (_params: Params): string => REACT_APP_ARCHIVE_URL
```
PiotrKozlowski: I'm not sure if this should be called `getSearch` since this is returning search url only for `REACT_APP_ARCHIVE_URL`
It feels like we should have just a new constant like `ARCHIVE_SEARCH_URL` for this
and then where you have
``` 
const search = content.getSearch && content.getSearch(reference.params);
const query = search ? `?${search}` : '';
```
you could just:
```
const query = REACT_APP_ARCHIVE_URL && ARCHIVE_SEARCH_URL;
```

also `_params` is not used here

src/app/navigation/middleware.ts
```diff
@@ -17,7 +17,8 @@ export default (routes: AnyRoute[], history: History): Middleware => ({dispatch}
     history[action.payload.method]({
       hash: action.payload.hash,
       pathname: matchUrl(action.payload),
-      search: action.payload.search,
+      search: matchSearch(action.payload, action.payload.search),
+      // search: action.payload.search,
```
PiotrKozlowski: -

src/app/navigation/utils.ts
```diff
@@ -48,9 +50,23 @@ export const findRouteMatch = (routes: AnyRoute[], location: Location): AnyMatch
...
+  return querystring.stringify({
+    ...previous,
+    ...route,
+  });
+};
+
 export const matchUrl = (action: AnyMatch) => hasParams(action)
   ? action.route.getUrl(action.params)
-  : action.route.getUrl();
+  : (action.route.getUrl as any)();
```
PiotrKozlowski: Why this stopped working without `as any` and why did it work previously (it requires `params)`? :D

src/app/content/routes.ts
TomWoodward: `search` is what the query string is called. https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/search

PiotrKozlowski: yes, but this function is not returning search query
it is returning search query if `ARCHIVE_SEARCH_URL` is a truthy value and empty string if it is not and the name suggests that it is getting search query from a current URL / provided params

src/app/content/components/Page.spec.tsx
```diff
@@ -95,6 +99,7 @@ describe('Page', () => {
         text
         <a href="">link with empty href</a>
         <a href="#hash">hash link</a>
+        <a href="?archive=some-content">archive link</a>
```
TomWoodward: instead of adding this link, you should be looking to see that "when the archive url config is set, the query parameter is added to the links on lines 93 and 95"

TomWoodward (changes requested):

src/app/content/routes.ts
PiotrKozlowski: Okay, I get it. There are different routes and they can implement their own handler for `getSearch`
That makes more sense now, but I still don't know why it is returning an empty string if `REACT_APP_ARCHIVE_URL` is falsy? Shouldn't it return search query form the state?

PiotrKozlowski (changes requested):
There are some errors after `yarn tsc`
I have one question here: https://github.com/openstax/rex-web/pull/556#discussion_r399343861 - this probably is non-blocking thing
I think this `as any` from: https://github.com/openstax/rex-web/pull/556#discussion_r397169280 is not required

src/app/content/routes.ts
KarinaMendez2: No, when content needs to get pulled from `archive`, we set this variable so that all the links included in the content, when they're clicked, they carry the `archive` part with them if there is one, if not, an empty string is `not adding the archive part to the links`

PiotrKozlowski: Nice, thanks for explaining

src/app/navigation/utils.ts
```diff
@@ -66,7 +66,7 @@ export const matchSearch = (action: AnyMatch, search: string | undefined) => {
 
 export const matchUrl = (action: AnyMatch) => hasParams(action)
   ? action.route.getUrl(action.params)
-  : (action.route.getUrl as any)();
+  : (action.route.getUrl)();
```
PiotrKozlowski:
```suggestion
  : action.route.getUrl();
```

PiotrKozlowski (approved):

TomWoodward (approved):

PR #558, opened 2020-03-25
Do not fetch highlights when clicking on figure link

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/949


src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -45,10 +45,18 @@ const hookBody = (services: MiddlewareAPI & AppServices) => async(action?: AnyAc
...
   const pageFocusIn = action && action.type === getType(receivePageFocus) && action.payload;
 
-  if (!authenticated || !book || !page || typeof(window) === 'undefined' || (loaded && !pageFocusIn)) {
+  if (
+    !authenticated
+    || !book
+    || !page
+    || typeof(window) === 'undefined'
+    || (loaded && !pageFocusIn)
+    || (!pageFocusIn && recentlyLoadedFor === page.id)
```
janKozie1: I might be wrong, but couldn't this be replaced by checking if `state.highlights` contains any highlight that has `sourceId === page.id`? I think it could be nice to avoid adding additional state if it's possible to retrive it from other locations.
 If I'm not missing anything i think the first highlight in `state.highlights` should have that source id so searching for it won't be hard on performance.

PiotrKozlowski: I think this approach would not work if there are no highlights on a page

Adding new state like this is implemented in `requestBook` and `requestPage` which is called before `receiveBook` / `receivePage` to avoid re-fetching same book / page when it is already loading (`resolveContent.ts`) - it works a little bit different because this state is cleared when book / page is received and in our case, we just want to have information about which highlights are in the state

I was thinking about implementing additional action like `requestHighlights(page.id)` but this would have to be called before every `receiveHighlights` action + it would be a mistake if `receiveHighlights` would for some reason throw an error

src/app/content/highlights/reducer.ts
```diff
@@ -22,6 +22,7 @@ const defaultColors = highlightStyles.map(({label}) => label);
 export const initialState: State = {
   highlights: null,
   myHighlightsOpen: false,
+  recentlyLoadedFor: null,
```
TomWoodward, line 25: lets restructure the state into 
```
interface State {
  currentPage: {
    pageId: string;
    highlights: null;
  };
  summary: { 
    open: boolean;
    ....existing stuff
  }
}
```

TomWoodward (changes requested):

src/app/content/highlights/reducer.ts
PiotrKozlowski: done

PiotrKozlowski:
@TomWoodward There is image snapshot failing. I don't know how to update it

PiotrKozlowski:
@TomWoodward There is an image snapshot failing but when I run locally `yarn run test:screenshots -u` then all screenshots are updated and almost all of them have CTA popup. How can I check why this snapshot is failing and how to update it?

TomWoodward:
the pulse insights survey was accidentally showing up here, i got them to modify its settings so it only shows on certain hostnames and the tests are passing again

TomWoodward (approved):

TomWoodward:
there are some typescript errors in `src/app/content/highlights/reducer.spec.ts`, sending to QA now anyway though

src/app/content/highlights/reducer.ts
```diff
@@ -35,8 +38,22 @@ export const initialState: State = {
...
+      const actionPageId = action.payload.location.state && action.payload.location.state.pageUid;
+      return {
+        currentPage: {
+          ...initialState.currentPage,
+          highlights: currentPageId && actionPageId === currentPageId
+            ? state.currentPage.highlights
+            : initialState.currentPage.highlights,
+          pageId: currentPageId && actionPageId === currentPageId
+            ? state.currentPage.pageId
+            : initialState.currentPage.pageId,
```
TomWoodward, line 51: these two use the same condition, so you could push it up and do 
```
return {
        currentPage: currentPageId && actionPageId === currentPageId
            ? {...state.currentPage, hasUnsavedHighlight: false}
            : initialState.currentPage
```

TomWoodward (changes requested):

src/app/content/highlights/reducer.ts
PiotrKozlowski: Thanks, I've pushed a change

TomWoodward (approved):

PR #564, opened 2020-04-02
Display Buy Book button conditionaly and link it to specifc book page

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/941


src/app/content/selectors.ts
```diff
@@ -20,6 +21,12 @@ export const book = createSelector(
   (state) => state.book
 );
 
+export const buyBookLink = createSelector(
+  book,
+  // ts is throwing an error because amazon_link is undefined in ArchiveBook
+  (data) => data && (data as BookWithOSWebData).amazon_link
```
TomWoodward, line 27: yea, there is a guard. you could change this to 
```suggestion
  (data) => hasOSWebData(data) ? data.amazon_link : undefined
```
or something like that

src/app/content/types.ts
```diff
@@ -44,6 +44,7 @@ export interface BookWithOSWebData extends ArchiveBook {
   theme: 'blue' | 'green' | 'gray' | 'yellow' | 'deep-green' | 'light-blue' | 'orange' | 'red';
   slug: string;
   publish_date: string;
+  amazon_link?: string;
```
TomWoodward, line 47: is this really potentially missing? i thought it would always be there but might be `''`

src/gateways/createOSWebLoader.ts
```diff
@@ -14,6 +14,7 @@ export interface OSWebBook {
   }>;
   cover_color: BookWithOSWebData['theme'];
   cnx_id: string;
+  amazon_link?: string;
```
TomWoodward, line 17: same

TomWoodward (changes requested):

TomWoodward (approved):

TomWoodward:
there is a typescript error in a spec, and the browser specs seem to think that content pages are rendering an error, please look into those CI issues

TomWoodward (approved):

PR #567, opened 2020-04-03
Adding Physics (high school) to rex

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/965


PiotrKozlowski (approved):

PiotrKozlowski:
In https://cnx.org/contents/zOZP3vRI@10.1:QP7lQWBk@7/Introduction I can see only version `10.1` but I guess I should be checking somewhere else?

KarinaMendez2:
image

@PiotrKozlowski

PiotrKozlowski:
@KarinaMendez2 Thanks

When I went to the same link again I can see:
obraz

After clicking it new version showed up in the History tab

TomWoodward:
@KarinaMendez2 @PiotrKozlowski to get the most recent version of the book you can go to the archive url with no version specified and it will redirect you. eg:

https://archive.cnx.org/contents/cce64fde-f448-43b8-ae88-27705cceb0da

this is basically what this script does https://github.com/openstax/rex-web/blob/master/script/update-content-versions.ts which updates all the books to the most recent versions. so you could also just add it with an empty version and run the script to get the newest one.

src/gateways/createArchiveLoader.ts
```diff
@@ -9,6 +11,30 @@ interface Extras {
...
+const cleanArchiveResponse = (archiveContent: any) => {
+  const chapterSlug = 'chapter-';
+  const chapterTitle = 'Chapter';
+  const appendixSlug = 'appendix-';
+  const appendixTitle = 'Appendix ';
+
+  if (archiveContent.tree) {
+    archiveContent.tree.contents.map((item: any) => {
+      if (item.slug.includes(chapterSlug) || item.slug.includes(appendixSlug)) {
+        item.slug = item.slug.replace(chapterSlug, '').replace(appendixSlug, '');
```
TomWoodward, line 23: if you make these regex like `/^chapter-/` and `/^appendix-/` they'll only match at the beginning of the string, so that is safer, and you can run the replace without a conditional, it'll just do nothing if it doesn't find anything

src/gateways/createArchiveLoader.ts
```diff
@@ -9,6 +11,30 @@ interface Extras {
...
+      }
+
+      const splitTitle = splitTitleParts(item.title);
+
+      if (!isNullOrUndefined(splitTitle[0]) && (splitTitle[0].includes(chapterTitle) ||
+          splitTitle[0].includes(appendixTitle))) {
+        item.title = item.title.replace(chapterTitle, '').replace(appendixTitle, '');
```
TomWoodward, line 30: this makes me nervous, since we cant limit it to just the start of the string, what if the rest of the title has the bad word in it, when we only want to replace it in the `.os-number` section. you might have to parse this with DOMParser so that you can manipulate just that one section and then render it back to text

src/gateways/createArchiveLoader.ts
```diff
@@ -23,7 +49,7 @@ export default (url: string) => {
     return archiveFetch<ArchiveContent>(`${url}/contents/${id}`)
       .then((response) => {
         cache.set(id, response);
-        return response;
+        return cleanArchiveResponse(response);
```
TomWoodward, line 52: do the call to `cleanArchiveResponse` before the result is cached on the previous line, otherwise a second call to the same url will not have the cleaned result

TomWoodward (changes requested):

TomWoodward:
it would be easier to test  if you move your cleanArchiveResponse to a separate file in helpers

src/app/content/utils.ts
```diff
@@ -9,6 +9,30 @@ export { getBookPageUrlAndParams, getPageIdFromUrlParam, getUrlParamForPageId, t
...
+  const appendixSlug = /^appendix-/;
+  const chapterTitle = 'Chapter';
+  const appendixTitle = 'Appendix';
+
+  if (archiveContent.tree) {
+    archiveContent.tree.contents.map((item: any) => {
+      item.slug = item.slug.replace(chapterSlug, '').replace(appendixSlug, '');
+
+      const domNode = new DOMParser().parseFromString(item.title, 'text/html');
+      const numNode = domNode.querySelector('.os-number');
```
PiotrKozlowski, line 23: This probably can be replaces with `getArchiveTreeSectionNumber` util

TomWoodward: its done this way so it can be modified below, the util just returns a string

src/app/content/utils.ts
```diff
@@ -9,6 +9,30 @@ export { getBookPageUrlAndParams, getPageIdFromUrlParam, getUrlParamForPageId, t
 export { stripIdVersion } from './utils/idUtils';
 export { scrollSidebarSectionIntoView } from './utils/domUtils';
 
+export const cleanArchiveResponse = (archiveContent: any) => {
```
TomWoodward, line 12: there is an `ArchiveContent` type that fits here

src/app/content/utils.ts
```diff
@@ -9,6 +9,30 @@ export { getBookPageUrlAndParams, getPageIdFromUrlParam, getUrlParamForPageId, t
...
+
+      const domNode = new DOMParser().parseFromString(item.title, 'text/html');
+      const numNode = domNode.querySelector('.os-number');
+
+      if (numNode) {
+        const cleanNumNode = numNode.textContent.replace(chapterTitle, '').replace(appendixTitle, '');
+        item.title = item.title.replace(numNode.textContent, cleanNumNode);
+      }
+
+    });
```
TomWoodward, line 30: all this logic looks good, but you need to recurse by calling `cleanArchiveResponse` on each item

src/app/content/utils.ts
```diff
@@ -9,6 +9,30 @@ export { getBookPageUrlAndParams, getPageIdFromUrlParam, getUrlParamForPageId, t
...
 export { scrollSidebarSectionIntoView } from './utils/domUtils';
 
+export const cleanArchiveResponse = (archiveContent: any) => {
+  const chapterSlug = /^chapter-/;
+  const appendixSlug = /^appendix-/;
+  const chapterTitle = 'Chapter';
+  const appendixTitle = 'Appendix';
+
+  if (archiveContent.tree) {
+    archiveContent.tree.contents.map((item: any) => {
```
TomWoodward, line 19: if you change the type above you can probably just remove this, but if not the correct type is `ArchiveTree | ArchiveTreeSection`

TomWoodward (changes requested):

TomWoodward (approved):

PR #569, opened 2020-04-06
Format head titles

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/939


src/app/content/utils/seoUtils.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { Book, Page } from '../types';
+import { makeArchiveSection, makeArchiveTree } from './archiveTreeUtils.spec';
```
janKozie1, line 2: Pretty sure that importing an another test file will make it impossible to run `seoUtils` alone ( and i'm not sure if it's a good idea to make test files depend on each other). I'm not sure where to move the utils but i dont think this is a good idea. Maybe @TomWoodward could help.

src/app/content/utils/seoUtils.ts
```diff
@@ -0,0 +1,20 @@
...
+import { archiveTreeSectionIsChapter, findArchiveTreeNode, getArchiveTreeSectionNumber } from './archiveTreeUtils';
+
+export const createTitle = (page: Page, book: Book): string => {
+  let node = assertDefined(findArchiveTreeNode(book.tree, page.id), `couldn't find node for a page id: ${page.id}`);
+  const domNode = new DOMParser().parseFromString(node.title, 'text/html');
+  const osText = domNode.querySelector('.os-text');
+  const nodeTitle = osText ? osText.textContent : page.title;
+  let title = `${nodeTitle} - ${book.title} | OpenStax`;
+  while (node && node.parent) {
+    node = node.parent;
```
janKozie1, line 12: I'm not exactly sure of this but don't we want to start searching for the number from the node itself not its parent?

janKozie1 (changes requested):

src/app/content/utils/seoUtils.ts
```diff
@@ -0,0 +1,20 @@
+import { assertDefined } from '../../utils';
+import { Book, Page } from '../types';
+import { archiveTreeSectionIsChapter, findArchiveTreeNode, getArchiveTreeSectionNumber } from './archiveTreeUtils';
+
+export const createTitle = (page: Page, book: Book): string => {
+  let node = assertDefined(findArchiveTreeNode(book.tree, page.id), `couldn't find node for a page id: ${page.id}`);
+  const domNode = new DOMParser().parseFromString(node.title, 'text/html');
+  const osText = domNode.querySelector('.os-text');
+  const nodeTitle = osText ? osText.textContent : page.title;
```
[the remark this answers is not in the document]
PiotrKozlowski: I did it this way because I'm not sure if it is correct to assume that `node.title` will always have `os-text` but I think it might be the case here. cc @TomWoodward

src/app/content/utils/seoUtils.spec.ts
PiotrKozlowski: running `seoUtils` alone works but I agree that tests (or actually any files) should not depend on each other like this. I wasn't sure if `test/utils.ts` is a proper place for this.
I could also just duplicate it or create a mock `page` and `chapter` instead of using a function

src/app/content/utils/seoUtils.ts
PiotrKozlowski: No, we want `Ch. [#][Page Title] - [Book] | OpenStax` if `page` is inside a `chapter` and `[Page Title] - [Book] | OpenStax` in other cases

src/app/content/utils/seoUtils.ts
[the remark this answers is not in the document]
TomWoodward: `splitTitleParts` already defaults to the whole string if `.os-title` is not found, you want to just use that. 

if you want you can make an alias that takes the title portion of the return value, like `getArchiveTreeSectionNumber` does

src/app/content/utils/seoUtils.spec.ts
TomWoodward: these are mocks right? you can move them to `src/test/mocks/`

janKozie1 (approved):

src/app/content/utils/seoUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+  findArchiveTreeNode,
+  getArchiveTreeNodeTitle,
+  getArchiveTreeSectionNumber,
+} from './archiveTreeUtils';
+
+export const createTitle = (page: Page, book: Book): string => {
+  let node = assertDefined(findArchiveTreeNode(book.tree, page.id), `couldn't find node for a page id: ${page.id}`);
+  const nodeTitle = getArchiveTreeNodeTitle(node);
+  const title = `${nodeTitle} - ${book.title} | OpenStax`;
+  while (node && node.parent) {
```
TomWoodward, line 14: according to the requirements, we should include the chapter prefix only if the page doesn't have a section number of its own

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
PiotrKozlowski: This wasn't clear, because AC says:
> - Pages within a chapter container have this template for the tag: Ch. [#][Page Title] - [Book] | OpenStax
> - Pages NOT within a chapter container should have this template for the tag: [Page Title] - [Book] | OpenStax 

and sections with numbers may also be in the chapter.

TomWoodward: the spreadsheet is more clear, i'll have kim update the AC

src/app/content/utils/seoUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+  findArchiveTreeNode,
+  getArchiveTreeSectionNumber,
+  splitTitleParts,
+} from './archiveTreeUtils';
+
+export const createTitle = (page: Page, book: Book): string => {
+  let node = assertDefined(findArchiveTreeNode(book.tree, page.id), `couldn't find node for a page id: ${page.id}`);
+  const [nodeNumber, nodeTitle] = splitTitleParts(node.title);
+  const title = `${nodeTitle} - ${book.title} | OpenStax`;
+  if (nodeNumber) { return title; }
```
TomWoodward, line 14: if you hit this condition it won't have any number in the title

TomWoodward (changes requested):

TomWoodward (approved):

PR #577, opened 2020-04-14
fix error script

TomWoodward:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1015


script/domVisitor.ts
```diff
@@ -54,14 +57,20 @@ async function visitPages(page: puppeteer.Page, bookPages: string[], audit: Audi
...
+      await page.goto(`${rootUrl}${pageUrl}${queryString ? `?${queryString}` : ''}`);
+      await page.waitForSelector('body[data-rex-loaded="true"]');
+      await calmHooks(page);
+
+      const matches = await page.evaluate(audit);
+
+      if (matches.length > 0) {
+        bar.interrupt(`- (${matches.length}) ${basename(pageUrl)}#${matches[0]}`);
+      }
+    } catch (e) {
```
janKozie1, line 70: Is it possible to use the syntax without `(e)` here?

janKozie1 (approved):

script/domVisitor.ts
TomWoodward: no, its required even if you don't use it

PR #583, opened 2020-04-17
Fix canonicals script

janKozie1:
It needs changes in tests, but wanted to make sure it's the right direction

for: https://github.com/openstax/unified/issues/1027


TomWoodward (approved):
yea, this is perfect

PR #590, opened 2020-04-22
Add tests for issue 581

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/581

Tests for pr: #481 

TODO:
 - [ ] Fix tests for highlighter
 - [ ] 100% coverage


src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -40,8 +40,8 @@ interface Props {
 }
 
 // tslint:disable-next-line:variable-name
-const DisplayNote = React.forwardRef<HTMLElement, Props>((
-  {note, isFocused, onBlur, onEdit, onRemove, onFocus, onHeightChange, className}: Props,
+const DisplayNote = React.forwardRef<HTMLElement, DisplayNoteProps>((
+  {note, isFocused, onBlur, onEdit, onRemove, onFocus, onHeightChange, className},
```
janKozie1, line 44: I've checked it again and I've noticed that the generic type adds `children` to the supplied type. This is my bad, sorry.

PiotrKozlowski: I think this is actually better, even if we do not use `children`
btw. I've applied your suggestions

TomWoodward:
@PiotrKozlowski is this ready for review? the changes look fine, i think.

PR #594, opened 2020-05-07
Animation example for toc after closing search sidebar

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/914


PiotrKozlowski (approved):

TomWoodward (changes requested):
`sidebarTransitionTime` is set to 300ms, you changed the search sidebar animation to 2s and now aren't using the variable and it is still set to 300ms, so there will be a mismatch in the animation of the content and sidebar that are meant to animate together.

why did you choose a longer animation time? if you did this on purpose maybe it would be better to change the variable also. if you didn't have a reason you could just replace your `2s` with the variable and go back to the quicker animation

TomWoodward (approved):

PR #628, opened 2020-05-13
Url checker script

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1049


TomWoodward (changes requested):
i'm getting all 404s when trying to check if a new book versions urls 404 on production, i think this is because the new version's page urls are being generated with the version in them, because it differs from the config. 

you might have to use a lower level url generation function to get the default url and skip the extra logic that is supposed to make it work right when loading the content

script/domVisitor.ts
```diff
@@ -102,28 +102,7 @@ run().then(null, (err) => {
...
-  const bookInfo = bookId
-    ? [{id: bookId, version: bookVersion || assertDefined(bookConfig[bookId], '').defaultVersion}]
-    : Object.entries(bookConfig).map(([id, {defaultVersion}]) => ({id, version: defaultVersion}))
-  ;
-
-  return await Promise.all(bookInfo.map(({id, version}) => bookLoader(id, version)));
-}
-
-function findBookPages(book: Book) {
+export function findBookPages(book: Book) {
```
janKozie1, line 105:
```suggestion
function findBookPages(book: Book) {
```

script/urlChecker.ts
```diff
@@ -0,0 +1,82 @@
...
+
+async function checkUrls() {
+  const url = assertDefined(rootUrl, 'please define a rootUrl parameter, format: http://host:port');
+  const books = await findBooks({
+    archiveUrl,
+    bookId,
+    bookVersion,
+    rootUrl: url,
+  });
+  const isProduction = !/release-\d+/g.test(url);
```
TomWoodward, line 71: if you're gonna put this behavior behind a conditional, maybe add an argument that does what it says it does, like `--withoutBookVersions` or something

script/urlChecker.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  for (const book of books) {
+    const pages = findTreePages(book.tree);
+    await checkPages(book.slug, pages.map(getUrl(book, isProduction)));
```
TomWoodward, line 75: please include the (new) logic from the other script where you track if any of these have any errors and `exit(1)`

TomWoodward (changes requested):

script/urlChecker.ts
```diff
@@ -0,0 +1,93 @@
...
+      if (response.status === 404) {
+        notFound.push(pageUrl);
+      }
+    } catch (e) {
+      anyFailures = true;
+      bar.interrupt(`- (error loading) ${pageUrl}`);
+    }
+
+    bar.tick();
+  }
```
PiotrKozlowski, line 48: This will be checking one url at the time. You can use `Promise.all` to make it faster.
I don't know how many urls there is to check, but if there is like hundreds of them then we can split them into blocks for ~50
```
const promises: () => Promise<void>[] = []

  for (const pageUrl of pages) {
    promises.push(() => {
      fetch(`${rootUrl}${pageUrl}`)
      .then(r => {
        if (r.status === 404) { notFound.push(pageUrl); }
        bar.tick();
      })
      .catch(e => {
        anyFailures = true;
        bar.interrupt(`- (error loading) ${pageUrl}`);
        bar.tick();
      })
    })
  }
  
  await Promise.all(promises.map(p => p()))
```

script/urlChecker.ts
```diff
@@ -0,0 +1,93 @@
...
+    bookVersion,
+    rootUrl: url,
+  });
+
+  let anyFailures = false;
+
+  for (const book of books) {
+    const pages = findTreePages(book.tree);
+    anyFailures = await checkPages(book.slug, pages.map(getUrl(book)));
+  }
```
PiotrKozlowski, line 83: This also could be speed up be using `Promise.all`

janKozie1: I think doing so would be problematic with how the progress bars are printed. 

Besides that if lets say that each book has 200 pages and we chunk the pages by 50, books by 3.
I don't think there would be any benefit from running 50 requests for 3 books at the same time than running 150 requests for one book at the same time.

It would be slightly faster if the the last chunk would have only few pages in it, but I'm not sure if it's worth implementing.

script/urlChecker.ts
```diff
@@ -22,6 +22,9 @@ const {
   useUnversionedUrls?: boolean;
 };
 
+const chunkArray = <T>(array: T[], max: number) =>
```
PiotrKozlowski, line 25: There is some kind of `chunk` function. Maybe it is worth it to create new util and reuse it
https://github.com/openstax/rex-web/blob/master/src/app/content/highlights/components/SummaryPopup/ChapterFilter.tsx#L55

janKozie1: Lodash also has its chunk function so i guess it's better to use that.

TomWoodward: yea, the chapter filter has extra logic to try to get two even columns, lodash chunk was the right thing to use here

script/urlChecker.ts
```diff
@@ -0,0 +1,97 @@
...
+    await Promise.all(pageChunk.map(visitPage));
+  }
+
+  if (notFound.length) {
+    console.log(`404'd: ${notFound.length === pages.length // tslint:disable-line:no-console
+      ? 'all'
+      : `\n${notFound.join('\n')}\n`}`);
+  }
+
+  return anyFailures;
```
TomWoodward, line 60: the other script was just looking for errors, but the purpose of this script is to check for 404ing urls, so we should `return anyFailures || notFound.length > 0` here

script/urlChecker.ts
```diff
@@ -0,0 +1,97 @@
...
+    bookId,
+    bookVersion,
+    rootUrl: url,
+  });
+
+  let anyFailures = false;
+
+  for (const book of books) {
+    const pages = findTreePages(book.tree);
+    anyFailures = await checkPages(book.slug, pages.map(getUrl(book)));
```
TomWoodward, line 86: you need the ` || anyFailures` on the end of this or it'll get lost on any but the last book

TomWoodward (changes requested):

TomWoodward (approved):

PR #630, opened 2020-05-14
Fix search result highlighting failure related to math

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1096

So far I've found two distinct cases for when the highlighting fails:
1. Text that is initially passed [here](https://github.com/openstax/rex-web/blob/437a3a34f139812c6af396e4d06e02f3964283ea/src/app/content/search/utils.ts#L112) starts with a space and is preceded by math

2.  ~~Text is inside visually hidden node eg. exercise solutions as it is ignored by [rangy](https://github.com/timdown/rangy/wiki/Text-Range-Module#visible-text)~~

More about 1:
The reason is mainly [this line](https://github.com/timdown/rangy/blob/1e55169d2e4d1d9458c2a87119addf47a8265276/src/modules/rangy-textrange.js#L885)
Because of it every text that has a leading space and is right after math is never found. 
Some tags inside mathml fit [this requirement](https://github.com/timdown/rangy/blob/1e55169d2e4d1d9458c2a87119addf47a8265276/src/modules/rangy-textrange.js#L675) which injects a new line to the string that rangy is using to compare with the query.

Simply removing the space works

~~More about 2:
Not sure what can be done here since rangy doesn't provide options to disable this behavior.
It would be possible to hide the text some other way (like  display:block, height:0px, overflow:hidden) but I'm not sure if it's okay to do so.~~

Note: fix for the second case was moved [here](https://github.com/openstax/rex-web/pull/639).


janKozie1:
Example of the second case:
(check it on https://rex-web-fix-highlightin-cobqxc.herokuapp.com/books/calculus-volume-1/pages/1-1-review-of-functions because the first case should be removed there)

Search for 'number' in Calculus Volume 1 - 1.1 Review of Functions
Then wait for the results to be highlighted
Then click "Show solution" for example 1.11
Then two whole paragraphs will be highlighted because the text was hidden.


Search for 'dollars' in Calculus Volume 1 - 1.1 Review of Functions
Then wait for the results to be highlighted
Then click "Show solution" for example 1.9
Then the whole solution is highlighted.
Because there are two search results in one hidden paragraph:

1. Since the sale price is … off the original price, if an item is …
2. Since the coupon entitles an individual to … off the price of any item, if an item is … 

The highlighting fails (1. would cause the whole paragraph to be highlighted, second one would actually show the error modal)

cc. @kjdav

PiotrKozlowski (approved):
Tested for word `defines` in `Calculs Volume 1

PR #631, opened 2020-05-15
missing task for update-content script

TomWoodward:
for: openstax/unified#1087


PiotrKozlowski (approved):
I guess it works since I can see multiple pull requests created

PR #632, opened 2020-05-18
Add Study guides button to the toolbar

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1032

TODO:
- [x] Update an icon - waiting for access to Invision
- [x] Add analytics tracking - separate pr (https://github.com/openstax/rex-web/pull/633)
- [x] Show study guides only if books have a study guide - separate pr (https://github.com/openstax/rex-web/pull/634)

PR #634, opened 2020-05-18
Issue 1032 fetch data

PiotrKozlowski:
Part of https://github.com/openstax/rex-web/pull/632

for: https://github.com/openstax/unified/issues/1032

TODO:
- [x] Update somehow `highlightClient` and remove `as any` from `receiveBook.ts`
 Open conversation: https://openstax.slack.com/archives/C8V1XET1R/p1589899704450800
- [ ] Update highlighter when https://github.com/openstax/highlighter/pull/37#pullrequestreview-415380747 get merged


src/app/content/components/Toolbar/StudyGuidesButton.spec.tsx
```diff
@@ -33,7 +34,7 @@ describe('study guides button', () => {
     expect(component.toJSON()).toMatchSnapshot();
   });
 
-  it('render if feature flag is enabled', () => {
+  it('does not render if feature flag is enabled but books does not have study guides', () => {
```
PiotrKozlowski, line 37: `but book does not have study guide` - remove two `s`

src/app/content/components/Toolbar/StudyGuidesButton.spec.tsx
```diff
@@ -47,10 +48,26 @@ describe('study guides button', () => {
     expect(component.toJSON()).toMatchSnapshot();
   });
 
+  it('render if feature flag is enabled and books has study guides', () => {
```
PiotrKozlowski, line 51: remove two `s`

src/app/content/studyGuides/types.ts
```diff
@@ -0,0 +1,9 @@
+import { HighlightsSummary } from '@openstax/highlighter/dist/api';
+
+// Temporary until we expand this state with more data
+// tslint:disable-next-line: no-empty-interface
+export interface StudyGuides extends HighlightsSummary {}
+
+export interface State {
+  studyGuides: StudyGuides | null;
+}
```
janKozie1, line 9: This makes the state be `content.studyGuides.studyGuides`, it could probably be changed.

src/app/content/studyGuides/hooks/receiveBook.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+import { MiddlewareAPI, Store } from '../../../types';
+import { receiveBook } from '../../actions';
+import { formatBookData } from '../../utils';
+import { receiveStudyGuides } from '../actions';
+
+const mockConfig = {BOOKS: {
+ [book.id]: {defaultVersion: book.version},
+} as {[key: string]: {defaultVersion: string}}};
+
+jest.doMock('../../../../config', () => mockConfig);
```
janKozie1, line 16: I'm not sure if this is necessary here. I don't think the current tests depend on it.

src/app/content/studyGuides/types.ts
PiotrKozlowski: I've changed it to `summary` for now, but it will be changed later anyway so I think we should not focus on it too much

TomWoodward:
i haven't tested it, but https://github.com/openstax/highlighter/pull/37 should fix the client for you

src/app/content/studyGuides/hooks/receiveBook.ts
```diff
@@ -0,0 +1,22 @@
+import { GetHighlightsSummarySourceTypeEnum } from '@openstax/highlighter/dist/api';
+import { ActionHookBody, AppServices, MiddlewareAPI } from '../../../types';
+import { actionHook } from '../../../utils';
+import { receiveBook } from '../../actions';
+import { receiveStudyGuides } from '../actions';
+
+export const hookBody: ActionHookBody<typeof receiveBook> = (
```
TomWoodward, line 7: better to hook into locationChange and check the book state. in production receiveBook will be called at build time, not on page load, which would mean the summary could get very stale between releases

src/app/content/studyGuides/hooks/receiveBook.ts
```diff
@@ -0,0 +1,22 @@
...
+  services: MiddlewareAPI & AppServices
+) => async(action) => {
+  if (!services.getState().content.studyGuidesEnabled) { return; }
+
+  const studyGuidesSummary = await services.highlightClient.getHighlightsSummary({
```
TomWoodward, line 12: you're gonna want to use all those highlighting helpers to paginate through these results

src/app/content/studyGuides/selectors.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const localState = createSelector(
+  parentSelectors.localState,
+  (parentState) => parentState.studyGuides
+);
+
+export const studyGuidesSummary = createSelector(
+  localState,
+  (state) => state.summary
+);
```
TomWoodward, line 12: should move the enabled selector into here, and out of the base content selectors, same with the state

src/app/content/components/Toolbar/StudyGuidesButton.tsx
```diff
@@ -34,7 +35,9 @@ const StudyGuidesText = styled.span`
 // tslint:disable-next-line:variable-name
 const StudyGuidesButton = () => {
   const isEnabled = useSelector(studyGuidesEnabled);
-  if (!isEnabled) { return null; }
+  const hasStudyGuides = useSelector(studyGuidesSummary);
+
+  if (!isEnabled || !hasStudyGuides) { return null; }
```
TomWoodward, line 40: it wouldn't suck to have this extra check for if there is any data to happen inside the `studyGuidesEnabled` selector, but i don't have super strong feelings on this one. if we had more places checking to see if study guides were turned on i would care more

src/app/content/studyGuides/hooks/receiveBook.ts
PiotrKozlowski: Hmm, but `locationChange` is dispatched before `book`

src/app/content/studyGuides/hooks/receiveBook.ts
PiotrKozlowski: Should this be done on this pr / for this card?
These helpers need some work to be reusable and I thought I should only apply basic logic for now

src/app/content/studyGuides/hooks/receiveBook.ts
PiotrKozlowski: Ok, I see, I have to hook it to the content locationChange

TomWoodward: yea if you call it from there it makes sure the book exists

src/app/content/studyGuides/hooks/receiveBook.ts
TomWoodward: you're right this doesn't matter because its just the summary, we'll address the utils when we do the modal content

TomWoodward:
and resolve conflicts

PR #637, opened 2020-05-19
Regression discard changes after clicking figure link

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1088

https://github.com/openstax/unified/issues/949#issuecomment-629286725

1. When editing a note click on a figure link and then click <kbd>Discard changes</kbd> in the popup. Note should be closed and the screen should be scrolled to the figure.

2. search sidebar is open -> unsaved notes in a page -> close search sidebar -> the unsaved notes should not be closed.


src/app/content/highlights/reducer.spec.ts
```diff
@@ -33,11 +33,12 @@ describe('highlight reducer', () => {
     expect(state.currentPage.highlights).toEqual(initialState.currentPage.highlights);
   });
 
-  it('locationChange - reset hasUnsavedHighlight', () => {
+  it('locationChange - reset hasUnsavedHighlight and focused', () => {
     const state = reducer(
-      {...initialState, currentPage: {...initialState.currentPage, hasUnsavedHighlight: true}},
+      {...initialState, currentPage: {...initialState.currentPage, hasUnsavedHighlight: true, focused: 'asd'}},
```
janKozie1, line 38: I think this doesn't doesn't cover the bug scenario. It needs the `pageUId` to be set to `asdf` so that it goes to the `omit('focused', {...state.currentPage, hasUnsavedHighlight: false})` that was added as a fix.

janKozie1 (changes requested):

TomWoodward:
@PiotrKozlowski please stop requesting review from everybody. if you don't have a preference leave it blank and github will assign whoever has the least pending reviews assigned to them.

janKozie1 (approved):

PiotrKozlowski:
When I open deployment for this pr and search for something then all user highlights are hidden. It works however locally

TomWoodward:
@PiotrKozlowski are you able to figure out why this is happening?

> When I open deployment for this pr and search for something then all user highlights are hidden. It works however locally

PiotrKozlowski:
@TomWoodward I'll try

PiotrKozlowski:
When I commented out the line `services.highlighter.eraseAll();` in `searchHighlightManager.ts` then user highlights are not removed, but ofc this line is required to clear search highlights.
In theory `services.highlighter` is different than `highlighter` for user highlights. I'll keep digging how these two are connected.
For now I'll revert previous commit

PiotrKozlowski:
Searching a highlight is working properly on deployment for https://github.com/openstax/rex-web/pull/465
I think it is somehow related to actions in `componentDidUpdate` in `PageComponent` but I didn't find the exact place in the code which cause that.

janKozie1:
After some digging turns out the bug is caused by [this line](https://github.com/openstax/rex-web/blob/63d3d01da88a2a90b7ec1e86237487eff51941c9/src/app/content/highlights/components/CardWrapper.tsx#L117). Since closing the search results clears the `query` in `search` state, the keys change which doesn't trigger a rerender but rather an unmount & mount which resets the internal state back to `data.annotation` 

Changing back to `highlight.id` as key fixes it, but i don't know how crucial that part is for #481 

I'm guessing you didn't see it happening on #465 because it wasn't updated to include that PR, i can see it happening there too now.

cc. @PiotrKozlowski @TomWoodward

TomWoodward:
@PiotrKozlowski re:
> In theory services.highlighter is different than highlighter for user highlights
the user highlights and search highlights do have different highlighter instances, you can see in each manager that they construct their own highlighters. this shouldn't matter because each one keeps a map of highlight ids that they own and will ignore others.

seems like @janKozie1 is onto something in https://github.com/openstax/rex-web/pull/637#issuecomment-635456247

TomWoodward:
about 
> Changing back to highlight.id as key fixes it, but i don't know how crucial that part is for #481
including the window width in the key is bad, that is causing the flickering effect. in reality the highlights don't care how wide the window is, all they care about is if the window is bigger or smaller than the breakpoint where they can all show to the side of the content.

the problem is that there are a couple breakpoints for that depending on the state of the sidebars, so its best to leave the logic in css hiding and showing the cards, but when the card is hidden it has no height, so what we _really_ want is for `onHeightChange` to run on window resize. changing the key is causing the entire component to get unmounted and remounted, which is causing the animation issues

PiotrKozlowski:
Actually issue which @janKozie1 described in https://github.com/openstax/rex-web/pull/637#issuecomment-635456247 is a different issue.

I'll create a pr with changing key for cards to `highlight.id` and fire `onHeightChange` when one of the crucial props changes. This will fix a problem that Jan found, but it will not affect the issue which I found. I'm not sure what effect on the performance this will have because it will create separate window resize listeners for each card.

janKozie1 (approved):

PR #638, opened 2020-05-22
codeowners file

TomWoodward:
pytest files are still owned by the wrong team. https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners says the order is backwards, but i already tried the other direction, so i'm sure there is something else wrong, lets see what happens when we put it back

for: openstax/unified#1100


.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
-pytest-selenium/* @openstax/unified-integration-testing
-
 * @openstax/rex
+
+pytest-selenium/* @openstax/unified-integration-testing
```
PiotrKozlowski, line 3: On https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
I've read:
```
# The `docs/*` pattern will match files like
# `docs/getting-started.md` but not further nested files like
# `docs/build-app/troubleshooting.md`.
```
I think it should be changed to: `pytest-selenium/ @openstax/unified-integration-testing`

I also can see that `unified-integration-testing` is nested inside `unified` team but I don't think it matters

PiotrKozlowski (approved):

PR #639, opened 2020-05-26
Fix search result highlighting for visually hidden text 

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1131

Currently, highlighting a search result inside an exercise solution fails, because rangy ignores text inside elements with `dispaly:none`. 

To fix it, the solutions can be hidden visually with different css properities.


PiotrKozlowski (commented):
This is working, however I'm not sure how this will work for screen readers. Maybe if you would add `visibility: hidden' it would be better?

Also, scrolling to the solution element is not the best, but I guess it can be fixed in a separate card: https://drive.google.com/open?id=1dFffaFluL5mNoJiIqo9fcDPjHL8qVxd4

janKozie1:
According to rangy's docs https://github.com/timdown/rangy/wiki/Text-Range-Module#visible-text, `visibility:hidden` will be treated the same way as `display:none` which causes the bug. I've checked it and it indeed fails.

According to this https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element and https://www.digitala11y.com/aria-expanded-state/ the `aria-expanded` should be rather toggled between `"true"` and `"false"` rather than added / removed.

I think too that scrolling deserves it's own card.

janKozie1:
Also, https://webaim.org/techniques/css/invisiblecontent/ mentions that elements with `height:0` will be ignored by _most_ screenreaders so adding the `aria-expanded='false'` should make it certain that it will be hidden

PiotrKozlowski (approved):

PR #640, opened 2020-05-27
Hide Confirmation components using css

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1020

`DisplayNote` is using `useOnClickOutside(element, isFocused, onBlur);` to blur when user clicks outside. This hook is using `elementDescendantOf` which is iterating over parents for clicked element and is comparing them to the passed `element`.
It will not work for elements which are rendered conditionally because at some point their `parentElement` becomes `null` when they are deattached from the DOM.

This pr is changing how `Confirmation` components are hidden - it uses css over rendering them conditionally.


src/app/content/highlights/components/Confirmation.tsx
```diff
@@ -18,7 +18,9 @@ export const Overlay = styled.div`
   ${cardBorder}
   transition: background 200ms;
   position: absolute;
-  display: flex;
+  ${({ isOpen }: { isOpen: boolean}) => {
```
KarinaMendez2, line 21:
```suggestion
display: ${(props: {isOpen: boolean}) => props.isOpen ? 'flex' : 'none'};
```

KarinaMendez2: image

z-indexes might be irregular

KarinaMendez2:
image

z-indexes might be a little off

PR #641, opened 2020-05-27
fix paths for update script

TomWoodward:
for: openstax/unified#1087


PiotrKozlowski (commented):
I think I've missed these names previously. Is there any way to check if this script works locally?

PiotrKozlowski (approved):
I accept in that case and hopes it will work :D

PR #643, opened 2020-05-28
Hide os-part-text in ToC

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1101


PiotrKozlowski (changes requested):
I think change is not fully fixing the issue. The card does not mention it but I think it is related to updated content which you can see only after opening a link like this: https://localhost:3000/books/us-history@latest/pages/f-united-states-population-chart/?archive=https://archive-qa.cnx.org

image

janKozie1:
The tests are failing because there is invalid HTML in fixtures.

I'm not sure if this approach is okay. I thought it could be better than removing it on rerenders in ToC

cc. @PiotrKozlowski  @TomWoodward

src/app/content/utils.ts
```diff
@@ -20,17 +21,33 @@ export const getContentPageReferences = (content: string) =>
...
     });
 
+const parseBookTree = (archiveBook: ArchiveBook) => {
+  archiveBook.tree.contents = archiveBook.tree.contents.map((subtree) => {
+    const domNode = new DOMParser().parseFromString(`<div id="container">${subtree.title}</div>`, 'text/html');
+    const container = domNode.getElementById('container');
+    const extra = container.querySelector('.os-part-text');
+
+    if (extra) {
+      assertNotNull(extra.parentElement, '.os-part-text must have a parent').removeChild(extra);
```
PiotrKozlowski, line 31: You should be able to call `extra.remove()`?

src/app/content/utils.ts
```diff
@@ -20,17 +21,33 @@ export const getContentPageReferences = (content: string) =>
       };
     });
 
+const parseBookTree = (archiveBook: ArchiveBook) => {
+  archiveBook.tree.contents = archiveBook.tree.contents.map((subtree) => {
+    const domNode = new DOMParser().parseFromString(`<div id="container">${subtree.title}</div>`, 'text/html');
```
PiotrKozlowski, line 26: Can we use somehow use `getArchiveTreeSectionTitle` or any other util to parse this title? Or maybe create new util for this?

PiotrKozlowski (commented):
These failing snapshots seems okay to be updated

src/app/content/utils.ts
TomWoodward, line 31: we'll also need to insert ` | ` into the divider if the part text is `Appendix`

TomWoodward (changes requested):

src/app/content/utils.ts
janKozie1: I've felt like the parsing of book tree should be ran once per book. I don't think the additional data in `.os-part-text` is necessary in anyway right now.

I'm not sure about `createTitle` in `seoUtils`. Should it have access to that data?

cc. @TomWoodward

PiotrKozlowski: I guess we don't need this in any other place, but Tom is better to answer that

One more thing: please move new DOMParser() outside of `.map`

TomWoodward (approved):

TomWoodward:
@PiotrKozlowski if you approve we can move this to review

PiotrKozlowski (approved):

PR #646, opened 2020-05-29
Change text on HM modal 

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1103


src/app/content/highlights/components/HighlightStyles.tsx
```diff
@@ -129,28 +123,8 @@ export const ImageWrapper = styled.div`
   width: ${(desktopPopupWidth - popupBodyPadding) / 2}rem;
 `;
 
-// tslint:disable-next-line: variable-name
-export const StickyNoteBullet = styled.div`
```
janKozie1, line 133: Is there a reason why it was left out?

KarinaMendez2: Why were the bullets removed? @TomWoodward

KarinaMendez2 (approved):

src/app/content/highlights/components/HighlightStyles.tsx
TomWoodward: i think the "bullet" you removed is actually the point on the sticky note, which we want to keep

TomWoodward (changes requested):

janKozie1:
Should the styles for a tooltip be removed? They are not used anywhere after this PR gets merged.

TomWoodward:
yes the tooltip styles can be removed

TomWoodward:
also i had asked karina to re-sort the messages file so you are now getting conflicts, sorry about that

TomWoodward (approved):

PR #647, opened 2020-06-01
Use contains method to find descendant element

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1020

Use built-in method because clicking on a child may fire function which a child's parent and then this function will not be able to find a parent of target element since it is unattached.


src/app/domUtils.ts
```diff
@@ -113,14 +113,7 @@ export const scrollIntoView = (elem: HTMLElement) => {
...
 export const elementDescendantOf = (element: Element, ancestor: Element): boolean => {
-  if (element === ancestor) {
-    return true;
-  }
-  if (!element.parentNode || !(element.parentNode instanceof assertWindow().Element)) {
-    return false;
-  }
-
-  return elementDescendantOf(element.parentNode, ancestor);
+  return ancestor.contains(element);
```
TomWoodward, line 116: should probably just refactor out the whole `elementDescendantOf` method and replace with this

TomWoodward (changes requested):

src/app/domUtils.ts
TomWoodward: also out of curiosity what case was broken?

PiotrKozlowski: `DisplayNote` is using `useOnClickOutside(element, isFocused, onBlur);` to blur when user clicks outside. This hook is using `elementDescendantOf` which is iterating over parents for clicked element and is comparing them to the passed element.
It will not work for elements which are rendered conditionally because at some point their `parentElement` becomes null when they are deattached from the DOM.

There is also another version of this pr which uses CSS to hide `Confirmation` components, but Malar said it was breaking some automation tests: https://github.com/openstax/rex-web/pull/640

TomWoodward (approved):

TomWoodward:
there is a linting error 
```
src/app/content/highlights/components/utils/onClickOutside.spec.ts(3,1): error TS6133: 'domUtils' is declared but its value is never read.
```

PR #660, opened 2020-06-03
Do not close EditCard on click outside (on the content / figure link, etc.)

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1022


janKozie1:
After clicking a figure link with unsaved changes nothing happens. I think this is coupled with yet another regression bug but on other deployments it at least scrolls to the figure.

PiotrKozlowski:
@janKozie1 you were most likely clicking a figure link when you already had a hash to this figure in the url. If you will clear url and then click on a figure link with unsaved changes, a confirmation popup should show up. 

However after yesterdays master merges case for clicking a figure link when editing card (without changes) will not work as expected - the card will close, because this was required for https://github.com/openstax/rex-web/pull/637

I'll ask Malar about this.

janKozie1 (approved):

PR #662, opened 2020-06-04
Verify production urls script

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1050

I'm not sure what should be the output.

Besides that I'm not sure if the script will have access to files in `release` or do i need to `cd` into it first.


concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -ex
+
+bucket=sandbox-unified-web-primary
```
TomWoodward, line 4: this can be pulled in from a param, params are used like this https://github.com/openstax/unified-deployment/blob/master/concourse/update-release-candidate-environments/task.yml and the s3 bucket is param is called `prod-unified-s3-bucket`

concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,15 @@
...
+production_url=https://openstax.org
+
+work_dir=$(pwd)
+
+release_id=$(curl "${production_url}/rex/release.json" | jq .id -r)
```
TomWoodward, line 9: use environment.json

concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,15 @@
...
+files=$(aws s3api list-objects --bucket "$bucket" --prefix "rex/releases/$release_id/books/" | jq -r '.Contents[] | .Key')
+
+while IFS= read -r full_path; do
```
TomWoodward, line 12: does a standard for loop not work here?

concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,15 @@
...
+  book_path=$(grep -oP '/books.*' <<< "$full_path")
+  [ ! -f "$work_dir$book_path" ] && exit 1
```
TomWoodward, line 14: maybe store an exit state so that we can echo out each failing path

TomWoodward, line 14: the task is going to drop you in a build directory that has subdirs for each input and output resource. you have an input called `release`, so you'll have a release folder with all the stuff in it. so from where you are the path will be like `release/books/..`

concourse/verify-production-urls/task.yml
```diff
@@ -0,0 +1,10 @@
+---
+platform: linux
+
+inputs:
+- name: release
+outputs:
```
TomWoodward, line 6: you don't need this output

TomWoodward (changes requested):

concourse/verify-production-urls/task.yml
```diff
@@ -0,0 +1,11 @@
+---
+platform: linux
+
+inputs:
+- name: release
```
TomWoodward, line 5: i think you also need to have `rex-web` as an input just because that is where the script lives that you run down there on line 11

concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,18 @@
...
+work_dir=$(pwd)
+
+release_id=$(curl "${production_url}/rex/environment.json" | jq .release_id -r)
+files=$(aws s3api list-objects --bucket "$PROD_UNIFIED_S3_BUCKET" --prefix "rex/releases/$release_id/books/" | jq -r '.Contents[] | .Key')
+
+missing_files=""
+
+for full_path in $files; do
+  book_path="/release/$(grep -oP 'books.*' <<< "$full_path")"
+  [ ! -f "$work_dir$book_path" ] && missing_files="$missing_files$book_path'\n'"
```
TomWoodward, line 15: you don't have to collect `missing_files`, you can `echo` them as you go. but you do have to track if anything fails so you can `exit` with the right code. the way you have it now the script will always register as passing even if it finds issues. check out `concourse/execute-test-runs/script.bash ` for an example of how it stores a `failed` variable and then does `exit "$failed"` to properly succeed or fail based on multiple commands

TomWoodward (changes requested):

TomWoodward (changes requested):

concourse/verify-production-urls/script.bash
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -ex
+
+production_url=https://openstax.org
+
+work_dir=$(pwd)
+failed=0
+PROD_UNIFIED_S3_BUCKET=sandbox-unified-web-primary
```
TomWoodward, line 8: this snuck in again

TomWoodward (changes requested):

concourse/verify-production-urls/script.bash
janKozie1: sorry

TomWoodward (approved):

PR #665, opened 2020-06-05
Issue 1104

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1104

This pr is not adding pagination, but it is adding loadMore action which is dispatched when a user scrolls to the bottom of results.

I've left some comments in the code:
```
// Temporary import from /highlights directory until we make all this logic reusable and move it to content/
```
to indicated that we should update these imports when we add another pr with logic for pagination. We have to extract some types, utils, and selectors to the `/content/` folder.

TODO:
- [x] Add tests for missing coverage
- [x] https://github.com/openstax/rex-web/pull/665#discussion_r436785272
- [x] https://github.com/openstax/rex-web/pull/665#discussion_r436785515


src/app/content/studyGuides/components/StudyGuides.tsx
```diff
@@ -0,0 +1,62 @@
...
+import styled from 'styled-components';
+import { typesetMath } from '../../../../helpers/mathjax';
+import Loader from '../../../components/Loader';
+import { useServices } from '../../../context/Services';
+import theme from '../../../theme';
+import { assertWindow } from '../../../utils';
+import allImagesLoaded from '../../components/utils/allImagesLoaded';
+// Temporary import from /highlights directory until we make all this logic reusable and move it to content/
+import * as Styled from '../../highlights/components/ShowMyHighlightsStyles';
+import { SectionHighlights } from '../../highlights/components/SummaryPopup/SectionHighlights';
```
PiotrKozlowski, line 13: I think the best option would be to move `SummaryPopup` into `content/components` because most of these components like "filters", "chapter selectors", "go to top button", etc are reusable between "my highlights" and "study guides", but I wanted to wait for a review to make sure you don't have anything against it.

src/app/content/studyGuides/components/StudyGuides.tsx
```diff
@@ -0,0 +1,62 @@
...
+          key={highlightData.location.id}
+          highlightDataInSection={highlightData}
+          forStudyGuides={true}
+        />;
+      })}
+    </Styled.Highlights>}
+  </React.Fragment>;
+};
+
+// Why these styles are not applied?
```
PiotrKozlowski, line 46: No idea why this is not working

src/app/content/studyGuides/components/StudyGuides.tsx
TomWoodward: we've been moving some popup stuff to app/components, you can put anything not content specific there, like the back to top button, or generic filter stuff.

the "Chapter filter" specifically, is the only piece that is content specific, so that should not go in app/components, but content/components would be fine

src/app/content/studyGuides/components/StudyGuides.tsx
```diff
@@ -0,0 +1,62 @@
+import { HTMLElement } from '@openstax/types/lib.dom';
+import React from 'react';
+import { useSelector } from 'react-redux';
+import styled from 'styled-components';
```
TomWoodward, line 4: make sure you use `styled-components/macro`

src/app/content/studyGuides/components/StudyGuides.tsx
TomWoodward: https://styled-components.com/docs/basics#styling-any-component

you have to do something with the className that it passes into StudyGuides

TomWoodward (changes requested):

src/app/content/studyGuides/components/StudyGuides.tsx
PiotrKozlowski: I've ended up with moving, `SectionHighlights`, `GoToTopButton`, some `constants`, and move styles from `ShowMyHighlightsStyles` to associated components

TomWoodward (approved):

PR #672, opened 2020-06-08
FE selenium test: Updated test for order print copy feature

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1135
Update: The Order print button should be displayed in rex book page if amazon link is set in cms or if osweb book details page has 'order print copy' with amazon details


pytest-selenium/tests/test_book.py
```diff
@@ -28,19 +31,39 @@ def test_book_title_links_to_books_detail_page(selenium, base_url, book_slug, pa
...
+        with pytest.raises(Exception) as exc_info:
+            assert not rex.order_print_copy
 
-    # AND: The first tab has rex page open
-    browser.switch_to_window(0)
-    assert browser.current_url == rex_url
+        exception_raised = exc_info.type
+        assert "NoSuchElementException" in str(
+            exception_raised
+        ), "amazon print option present in rex but not present in osweb"
```
gregfitch, line 69: Instead of verifying that any exception is raised, use `with pytest.raises(NoSuchElementException):`
That way the test doesn't cover up other exceptions that may occur when checking the print copy status and it also removes the need for lines 66-68 (the line 69 statement would be attached to the first assert statement).

gregfitch (changes requested):

Malar-Natarajan:
@gregfitch this is ready for re-review

gregfitch (approved):

PR #683, opened 2020-06-11
Add custom dimension to pageview

steveburkett:
Add dimension3 to pageview events in google analytics

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/903

Not embedded
image

Embedded inside an iframe
image


src/gateways/googleAnalyticsClient.ts
```diff
@@ -28,7 +28,14 @@ interface SetCommand {
   };
 }
 
-type Command = SetCommand | SendCommand;
+interface SetCommandCustomDimension {
+  name: 'set';
+  payload: {
+    dimension3: string | undefined;
```
TomWoodward, line 34: you can just change SetCommand so that `userId` and `dimensionId` are optional like: 

```
interface SetCommand {
  name: 'set';
  payload: {
    userId?: string | undefined;
    dimension3?: string | undefined;
  };
}
```

i think the way you have it will work, but the pattern we're following here with the `Command` type is a [discriminated-union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) which is easier to follow if there is a top level discriminator key (`name` in this case).

src/gateways/googleAnalyticsClient.ts
```diff
@@ -110,6 +121,19 @@ class GoogleAnalyticsClient {
     this.flushPendingCommands();
   }
 
+  private referringHostName() {
```
TomWoodward, line 124: this logic probably shouldn't be in a gateway, lets move it to its own util or helper and give it its own unit tests

i can't decide if it would be clean to call that helper from in here, or pass the value into `setCustomDimensionForSession` as an argument, you can do whatever feels good to you

TomWoodward (changes requested):

TomWoodward (approved):

PR #684, opened 2020-06-12
Paginate study guides

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1126

Todo:
- [x] ~~Try merging `loadAll` into ~~`createSummaryHighlightsLoader`~~ `createLoader`~~
- [x] Organize shared utils better
- [x] Better naming for things with "highlights" that study guides also use
- [x] Fix existing tests
- [x] Write new tests
- [ ] Fix loading overlay if it was broken here


src/app/content/studyGuides/components/StudyGuides.tsx
```diff
@@ -16,8 +16,8 @@ import StudyGuidesListElement from './StudyGuidesListElement';
 
 // tslint:disable-next-line: variable-name
 const StudyGuides = ({ className }: { className: string }) => {
-  const orderedHighlights = useSelector(selectors.orderedStudyGuidesHighlights);
-  const isLoading = useSelector(selectors.studyGuidesIsLoading);
+  const orderedHighlights = useSelector(selectors.orderedSummaryStudyGuidesHighlights);
```
KarinaMendez2, line 19: This could have a reduced name, maybe like just `formattedSummary` or something

src/app/content/studyGuides/reducer.spec.ts
```diff
@@ -1,15 +1,19 @@
...
 
 describe('study guides reducer', () => {
   it('receive study guides', () => {
-    const summary = { asd: 'asd' } as any as HighlightsSummary;
-    const state = reducer(undefined, actions.receiveStudyGuides(summary));
+    const summary = {
+      location: {
+        page: [],
+      },
+    } as SummaryHighlights;
```
KarinaMendez2, line 13: if this is about studyguides, `Highlights` shouldn't be mentioned anywhere, it's a bit confusing

src/app/content/studyGuides/reducer.ts
```diff
@@ -10,29 +11,46 @@ import { State } from './types';
...
+        ...state,
+        summary: {
+          ...state.summary,
+          highlights: merge(state.summary.highlights || {}, action.payload),
+          loading: false,
+          pagination: action.meta,
+        },
       };
     }
     default:
```
KarinaMendez2, line 56: same in this file

KarinaMendez2 (changes requested):

src/app/content/studyGuides/reducer.spec.ts
janKozie1: I've treated `StudyGuides` as special kind of `Highlights` since in the end, they still use things like
`highlightClient.getHighlights` 
`GetHighlightsSetsEnum`
`GetHighlightsSourceTypeEnum`
`GetHighlightsColorsEnum`

TomWoodward: @janKozie1 i think that is a good frame of mind for this stuff

src/app/content/highlights/components/Card.tsx
```diff
@@ -52,7 +52,7 @@ const Card = (props: CardProps) => {
   const annotation = props.data && props.data.annotation;
   const element = React.useRef<HTMLElement>(null);
   const [editing, setEditing] = React.useState<boolean>(!annotation);
-  const locationFilters = useSelector(selectHighlights.highlightLocationFilters);
+  const locationFilters = useSelector(contentSelect.highlightLocationFilters);
```
TomWoodward, line 55: this selector i think should still be in `selectHighlights`. its currently true that both MH and SG want to have the same chapter filter options, but it wouldn't be unreasonable for that to change and for instance MH gets pages and SG doesn't, then we'd want the locationFIlters to be different for each one. in that case the pagination logic would still work i think, just with different ids passed in.

actually we will probably want to have them be different before launch because SG will not have content in things like the index or preface, so we will probably not have non-chapter options in there the way MH does.

for now, since the underlying logic of the locationFilter selector will be the same for both, just define the selectors separately but delegate the implementation to some shared util function

src/app/content/studyGuides/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { GetHighlightsColorsEnum, HighlightColorEnum } from '@openstax/highlighter/dist/api';
+
+export const allColors = [
+  GetHighlightsColorsEnum.Blue,
+  GetHighlightsColorsEnum.Green,
+  GetHighlightsColorsEnum.Pink,
```
TomWoodward, line 6: there is no pink for study guides

src/app/content/studyGuides/hooks/locationChange.ts
```diff
@@ -1,75 +1,45 @@
...
   });
 
-  const formattedHighlights = formatReceivedHighlights(
-    assertDefined(highlightsResponse.data, 'expected api data response to be defined'),
-    locationFilters);
-
-  services.dispatch(receiveStudyGuidesHighlights(formattedHighlights));
-
-  services.dispatch(receiveStudyGuides(studyGuidesSummary));
+  services.dispatch(receiveStudyGuides(highlights));
```
TomWoodward, line 42: we don't want to actually load these until they open the dialog. (which it looks like you're actually doing already, so maybe just remove this)

src/app/content/studyGuides/hooks/openStudyGuides.ts
```diff
@@ -0,0 +1,40 @@
...
+  dispatch,
+  getState,
+  highlightClient,
+}) => async() => {
+  const state = getState();
+  const book = bookSelector(state);
+  const totalCounts = select.totalCountsPerPageOrEmpty(state);
+  const locationFilters = highlightLocationFilters(state);
+
+  const summaryNeedsInitialization = () => select.summaryStudyGuidesHighlights(state) === null
```
TomWoodward: this doesn't have to be a function

src/app/content/studyGuides/actions.ts
```diff
@@ -1,11 +1,16 @@
...
 import { createStandardAction } from 'typesafe-actions';
-import { StudyGuidesHighlights } from './types';
+import { CountsPerSource, HighlightData, SummaryHighlights, SummaryHighlightsPagination } from '../types';
 
-export const receiveStudyGuides = createStandardAction('Content/StudyGuides/receive')<HighlightsSummary>();
+export const receiveSummaryStudyGuides = createStandardAction(
+  'Content/StudyGuides/Summary/receive'
+)<SummaryHighlights, SummaryHighlightsPagination>();
 export const openStudyGuides = createStandardAction('Content/StudyGuides/Summary/open')<void>();
+export const initializeStudyGuidesSummary = createStandardAction('Content/StudyGuides/Summary/Initialize')<void>();
```
TomWoodward, line 8: this initialize action isn't being used, i think in MH this is dispatched when it mounts and all initialization is done then, but you're doing that with a hook into the open action

idk what the correct approach is, but something seems to need to be cleaned up here

src/app/content/studyGuides/hooks/openStudyGuides.ts
```diff
@@ -0,0 +1,40 @@
...
+    && select.summaryIsLoading(state) === false;
+
+  if (summaryNeedsInitialization()) {
```
TomWoodward, line 24: here we can use the same loadMore function as the loadMore hook right?

src/app/content/types.ts
```diff
@@ -115,3 +116,32 @@ export interface ArchivePage {
 }
 
 export type ArchiveContent = ArchivePage | ArchiveBook;
+
+export type HighlightData = Highlight;
```
TomWoodward, line 120: lets either move these to a file like `highlight-data-types.ts` that is shared between the two or just move them back into the highlighting types file.

they don't make a lot of sense here outside of the context of any highlighting stuff, and they don't apply to other content logic

src/app/content/utils/highlightLoadingUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+  GetHighlightsSourceTypeEnum,
+  Highlight,
+  Highlights,
+} from '@openstax/highlighter/dist/api';
+import omit from 'lodash/fp/omit';
+import { AppServices } from '../../types';
+import { assertDefined } from '../../utils';
+import { maxHighlightsApiPageSize } from '../constants';
+import { addSummaryHighlight, getHighlightLocationFilterForPage } from '../highlights/utils';
+import { getNextPageSources, incrementPage } from '../highlights/utils/paginationUtils';
```
TomWoodward, line 12: it doesn't really make any sense to move logic into this file outside the highlighting folder and then reach into the highlighting folder for a bunch of dependencies. either we should pull out all the common stuff or just leave it in highlighting and mentally justify it by saying study guides are a type of highlight

src/app/content/utils/highlightSharedUtils.ts
```diff
@@ -0,0 +1,7 @@
+import mapValues from 'lodash/fp/mapValues';
+import pickBy from 'lodash/fp/pickBy';
+import { isDefined } from '../../guards';
+import { CountsPerSource } from '../types';
+
+export const extractTotalCounts = (countsPerSource: CountsPerSource) =>
```
TomWoodward, line 6: again, its cool that this is here, but you're still reaching into highlights for a bunch of other utils, we need to either go one way or the other

TomWoodward (dismissed):

src/app/content/studyGuides/hooks/openStudyGuides.ts
janKozie1: I think it should actually dispatch for loadMore so it also sets the loading state

src/app/content/studyGuides/selectors.ts
```diff
@@ -19,51 +24,78 @@ export const studyGuidesSummary = createSelector(
...
-  (_state) => false
+export const summaryStudyGuidesPagination = createSelector(
+  studyGuidesSummary,
+  (summary) => summary.pagination
 );
 
-export const hasMoreResults = createSelector(
-  studyGuidesPagination,
-  (pagination) => Boolean(pagination)
+export const studyGuidesLocationFilters = createSelector(
```
PiotrKozlowski, line 53: This selector is equal to `highlightLocationFilters`, we could move it to the the `content/selectors` and reference it as `parentSelectors.locationFilters`, but this would mean that we should also move `getHighlightLocationFilters` into `content/utils`
I personally was thinking about moving all type, interfaces, and utils which are used between MH and SG into `/content/utils/` and `/content/types` but it might be not worth of time

PiotrKozlowski (commented):
Look good overall.
@TomWoodward It would be nice to have more data in SG for Anatomy and Physiology to test in properly from the user side

src/app/content/studyGuides/selectors.ts
janKozie1: See https://github.com/openstax/rex-web/pull/684#discussion_r441657225

KarinaMendez2 (approved):
looks good

PR #693, opened 2020-06-18
record error boundary catches in state

TomWoodward:
there are two things here:
1) we hadn't been putting the errors caught by the error boundary in state, so they wouldn't have broken the pre-render or scanner scripts. we might have been releasing pre-rendered error pages to prod without realizing it
2) validate that there are no module links in the page content (visible in the `[1]See Appendix A for a discussion of powers of 10.` content on this page https://rex-web.herokuapp.com/books/physics@13.8/pages/1-3-the-language-of-physics-physical-quantities-and-units/

because previously the modal error display was shown whenever there was an error in state, i had to rearrange that so that the error boundary would not pop the modal

for: openstax/unified#1170


src/app/navigation/components/NavigationProvider.spec.tsx
```diff
@@ -1,16 +1,10 @@
...
+import createTestStore from '../../../test/createTestStore';
 import NavigationProvider from './NavigationProvider';
 
-const initialState = {
-  hash: '',
-  pathname: '',
-  search: '',
-  state: {},
-};
+const initialState = new URL('https://localhost');
```
janKozie1, line 7: Why is `new URL` used as `initialState`? 
Its properties are not enumerable and It passes TS validation because its `any`. If it's only for avoiding validation then it could be done in a clearer way.

src/app/errors/actions.ts
```diff
@@ -1,5 +1,8 @@
 import { createStandardAction } from 'typesafe-actions';
 
+export const showErrorDialog = createStandardAction('Errors/showDialog')<void>();
+export const hideErrorDialog = createStandardAction('Errors/hideDialog')<void>();
+
```
janKozie1, line 5: `clearCurrentError` is no longer being used

janKozie1 (changes requested):
The errors in inline error display are not center aligned as other elements of it, don't really know if it matters .

src/app/navigation/components/NavigationProvider.spec.tsx
TomWoodward: URL has the same props as Location, so it was just easy to mock that way. i didn't realize i was getting an `any` type though, and when i fix it to `Location` i realize i need to add something to match the reducer

janKozie1: Spread operator only copies objects' own enumerable properties, which instances of URL inherit from its prototype. 
That is unless I'm understanding it wrong, but i've checked the state in tests and there is nothing there besides `query` and `pathname` which could lead to weird bugs.

TomWoodward: i updated to assign onto the url, i think that will work better

janKozie1: I mean, it works up until you spread it again in `createTestStore` which does the same things as before. You would need to assign it with the pathname too.
I thought if one was to change the prototype of an object after assigning it then it wouldn't be able to tell which ones were inherited but changing prototype with `setPrototypeOf` drops inherited properties and I can't find information why.

TomWoodward: ok yea whatever i just hardcoded the values, in this case that is all i needed since i'm not using any of the modification functionality

janKozie1 (approved):

PR #695, opened 2020-06-18
Add color key - study guides

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1107


src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,131 @@
...
+                  shape='circle'
+                  style={color}
+                  size='small'
+                  aria-label={color.label + ' key term'}
+                  component={<label />}
+                />
+                <FormattedMessage id={'i18n:studyguides:popup:color-key:terms:' + color.label}>
+                  {(key: string) =>
+                    <KeyTermText>{key}</KeyTermText>
+                  }
```
PiotrKozlowski, line 120: This could be in one line

src/helpers/analyticsEvents/studyGuides/openClose.ts
```diff
@@ -2,8 +2,10 @@ import { createSelector } from 'reselect';
 import * as selectNavigation from '../../../app/navigation/selectors';
 import { AnalyticsEvent } from '../event';
 
-const open = 'REX Study guides (open SG popup)';
-const close = 'REX Study guides (close SG popup)';
+const openSG = 'REX Study guides (open SG popup)';
+const openColorKey = 'REX Study guides (open color key)';
+const closeSG = 'REX Study guides (close SG popup)';
+const closeColorKey = 'REX Study guides (close color key)';
```
PiotrKozlowski, line 8: I think it would be nice to have a separate event (file) for color key

PiotrKozlowski (commented):
Will you be adding this into SG modal on this pr or there will be a separate one?

KarinaMendez2:
@PiotrKozlowski this will be added when the filters section is done inside the SG modal and merged, otherwise I have nowhere to put it yet.

PiotrKozlowski:
It looks a little bit different for me than on specs:
- circles are a little bit "eggish"
- it seems like padding is too small
- probably width of the component has to be adjusted so it doesn't break to the newline
image

image

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <ColorKey />
+        </MessageProvider>
+      </Services.Provider>
+    </Provider>);
+
+    /*open color key*/
+    renderer.act(() => {
+      const button = component.root.findByType(ColorKeyButtonWrapper);
+      button.props.onClick();
+      expect(spyTrack).toHaveBeenCalled();
```
janKozie1, line 41: Generally, i think, assertions shouldn't be in `renderer.act`. It's only meant to prepare components for assertions

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <ColorKeyText>{msg}</ColorKeyText>
+        {open && <ColorKeyDescription>
+          {highlightStyles.filter((style) => style.label !== HighlightColorEnum.Pink)
+            .map((color, index) =>
+              <KeyTermWrapper key={index}>
+                <ColorIndicator
+                  shape='circle'
+                  style={color}
+                  size='small'
+                  aria-label={color.label + ' key term'}
```
janKozie1, line 125: Is there a reason why this doesn't have it's own line in `messages.json`?

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  return <FormattedMessage id='i18n:studyguides:popup:color-key'>
+    {(msg: Element | string) =>
+      <ColorKeyButtonWrapper onClick={toggleColorKey} aria-label={msg}>
+        <ColorKeyIcon src={colorKeyIcon}/>
+        <ColorKeyText>{msg}</ColorKeyText>
+        {open && <ColorKeyDescription>
+          {highlightStyles.filter((style) => style.label !== HighlightColorEnum.Pink)
+            .map((color, index) =>
```
janKozie1, line 119: Pretty sure this can be taken from `studyGuides/constants`

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,140 @@
...
+`;
+
+// tslint:disable-next-line:variable-name
+const ColorKey = () => {
+  const [open, setOpen] = React.useState<boolean>(false);
+  const trackOpenClose = useAnalyticsEvent('openCloseColorKey');
+
+  const toggleColorKey = () => {
+    setOpen((state) => !state);
+    trackOpenClose(open);
```
janKozie1, line 109: State updates are not synchronous. This will make the analytics event send `close color key` when they are opened and `open color key` when they are closed.

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,140 @@
...
+  };
+
+  return <FormattedMessage id='i18n:studyguides:popup:color-key'>
+    {(msg: Element | string) =>
+      <ColorKeyButtonWrapper onClick={toggleColorKey} aria-label={msg}>
+        <ColorKeyIcon src={colorKeyIcon}/>
```
janKozie1, line 115: What's the reason everything is inside one button?

PiotrKozlowski (changes requested):
https://github.com/openstax/rex-web/pull/695#issuecomment-647603800

src/app/content/studyGuides/components/ColorKey.tsx
KarinaMendez2: I got that inside the `openCloseColorKey` in analytics...

src/app/content/studyGuides/components/ColorKey.tsx
KarinaMendez2: That was my solution, since I've had to be moving around this button, and will move again when print is added, putting everything inside the same button is how I got it to stay in the same place.

KarinaMendez2:
@PiotrKozlowski have you checked again? 
image

src/app/content/studyGuides/components/ColorKey.spec.tsx
KarinaMendez2: don't understand your comment

src/app/content/studyGuides/components/ColorKey.tsx
janKozie1: What do you mean? I'm logging the object in `track` right before the payload gets created and i'm seeing reverse event categories. Am i missing something?

src/app/content/studyGuides/components/ColorKey.spec.tsx
janKozie1: Update state / do things that will update state in `renderer.act`, make assertions outside of it (if possible).

PiotrKozlowski:
@KarinaMendez2 
Firefox:
image

Chrome looks good besides strange outlines on focus:
weird outlines

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <ColorKeyText>{msg}</ColorKeyText>
+        {open && <ColorKeyDescription>
```
PiotrKozlowski, line 117: Having `<ColorKeyDescription>` inside this button is causing a strange focus outline in Chrome (https://github.com/openstax/rex-web/pull/695#issuecomment-653063852) and I think it might be problematic when we will be implementing selecting with `TAB`

PiotrKozlowski (changes requested):

src/app/content/studyGuides/components/ColorKey.tsx
KarinaMendez2: I'll ask what the behavior should be, thanks, can you check it out on mobile?

src/helpers/analyticsEvents/studyGuides/openCloseColorKey.ts
```diff
@@ -0,0 +1,24 @@
...
+);
+
+export const track = (
+  {pathname}: ReturnType<typeof selector>,
+  open?: boolean
+): AnalyticsEvent | void => {
+  return {
+    getGoogleAnalyticsPayload: () => ({
+      eventAction: 'button',
+      eventCategory: open ? openText : closeText, // async call
```
KarinaMendez2, line 20: @janKozie1

src/app/content/studyGuides/components/ColorKey.tsx
KarinaMendez2: You're right, the logic was fine but at some point my brain changed it to reverse events, fixed.

src/helpers/analyticsEvents/studyGuides/openCloseColorKey.ts
```diff
@@ -0,0 +1,24 @@
...
+);
+
+export const track = (
+  {pathname}: ReturnType<typeof selector>,
+  open?: boolean
+): AnalyticsEvent | void => {
+  return {
+    getGoogleAnalyticsPayload: () => ({
+      eventAction: 'button',
+      eventCategory: open ? closeText : openText, // async call
```
KarinaMendez2, line 20: @janKozie1

janKozie1: I'm not really sure if the comment makes things clearer. I feel like the ternary should be like it was before, just change how the event is dispatched in the component. 
I think the best would be to just add an `useEffect` that fires the event when the `open` changes, but that's just my opinion.

src/app/content/studyGuides/components/ColorKey.tsx
PiotrKozlowski: image

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -34,18 +46,18 @@ describe('Study Guides button and PopUp', () => {
       </Services.Provider>
     </Provider>);
 
-    /*open color key*/
+    /*open and close sccolor key*/
```
PiotrKozlowski, line 49: typo

src/helpers/analyticsEvents/studyGuides/openCloseColorKey.ts
PiotrKozlowski: Mine would be:
```
setOpen(state => {
  const newVal = !state;
  trackOpenClose(newVal);
  return newVal;
})
```

KarinaMendez2:
Since every button I click has the outline, I'm not sure if I should remove it from Color Key @PiotrKozlowski

PiotrKozlowski:
@KarinaMendez2 I've meant move the dropdown to be next to the button so when user click on the <kbd>Color key</kbd>  then dropdown will not get the focus

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const [open, setOpen] = React.useState<boolean>(false);
+  const trackOpenClose = useAnalyticsEvent('openCloseColorKey');
+
+  const toggleColorKey = () => {
+    setOpen((state) => !state);
+  };
+
+  React.useEffect(() => {
+    trackOpenClose(open);
+  }, [open, trackOpenClose]);
```
janKozie1, line 128: This will also run on initial render. You can put the effect in a component which only renders `<ColorKeyDescription>`
like
```js
useEffect(() => {
    trackOpenClose(true)
    return () => trackOpenClose(false)
}, [trackOpenClose]
```
or use Piotr's approach if you like it more.

PiotrKozlowski:
@KarinaMendez2 
This is still how it looks for Firefox - are we okay with this?
image

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,158 @@
...
+  return <ColorKeyWrapper>
+    <FormattedMessage id='i18n:studyguides:popup:color-key'>
+      {(msg: Element | string) =>
+        <ColorKeyButtonWrapper onClick={toggleColorKey} aria-label={msg}>
+          <ColorKeyIcon src={colorKeyIcon}/>
+          <ColorKeyText>{msg}</ColorKeyText>
+        </ColorKeyButtonWrapper>
+      }
+    </FormattedMessage>
+      {open && <ColorKeyDescription>
```
TomWoodward, line 136: indentation is messed up from here down

src/app/messages/en/messages.json
```diff
@@ -86,6 +86,12 @@
   "i18n:search-results:bar:query:results": "{total, plural, one {search result for } other {search results for }}",
   "i18n:search-results:bar:search-icon:value": "Search",
   "i18n:search-results:bar:toggle-text:mobile": "Search results",
+  "i18n:studyguides:popup:color-key": "Color key",
+  "i18n:studyguides:popup:color-key:terms:aria-label": "{color} key term",
```
TomWoodward, line 90: i don't think this message is right, is it in the spec? we might need malar or suhani to come up with something

src/helpers/analyticsEvents/studyGuides/openCloseColorKey.ts
```diff
@@ -0,0 +1,24 @@
...
+  (pathname) => ({pathname})
+);
+
+export const track = (
+  {pathname}: ReturnType<typeof selector>,
+  open: boolean
+): AnalyticsEvent | void => {
+  return {
+    getGoogleAnalyticsPayload: () => ({
+      eventAction: 'button',
```
TomWoodward, line 19: this action doesn't seem helpful here.

i'm thinking we can make the category `REX Study guides (color key)` and the action either `open` or `close`

TomWoodward (changes requested):

janKozie1 (commented):
Is it desired for it to stay open after clicking outside of it?

PiotrKozlowski:
There is this question from Jan:
> Is it desired for it to stay open after clicking outside of it?

and I have two more:
Should this pr handle `esc` key?
Should this pr include adding outline for focusing the labels with keyboard?
These are not described in the behavior docs so I'm willing to accept this pr without this

KarinaMendez2:
Response:

Is it desired for it to stay open after clicking outside of it? `yes`
Should this pr handle esc key? `no`
Should this pr include adding outline for focusing the labels with keyboard? `no`

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    renderer.act(() => {
+      const button = component.root.findByType(ColorKeyButtonWrapper);
+      button.props.onClick();
+    });
+
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('Tracks open close to GA', async() => {
```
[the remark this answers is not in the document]
KarinaMendez2: It needs it for the `await` from line `56`

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    </Provider>);
+
+    /*open and close color key*/
+    const button = component.root.findByType(ColorKeyButtonWrapper);
+    renderer.act(() => {
+      button.props.onClick();
+      expect(spyTrack).toHaveBeenCalledWith({pathname: '/'}, true);
+    });
+
+    await renderer.act(async() => {
```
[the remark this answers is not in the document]
KarinaMendez2: If it doesn't have the `async()` it gives: ` Warning: Do not await the result of calling act(...) with sync logic, it is not a Promise.`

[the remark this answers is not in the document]
PiotrKozlowski: I think Jan wanted you to remove all `async` / `await` statements from this test 
This is something which I very often forget to remove after trying to find a solution

PiotrKozlowski:
@KarinaMendez2 Thanks, I think now there is only one thing from Jan's suggestions which needs to be resolved: https://github.com/openstax/rex-web/pull/695#discussion_r450877500

src/app/content/studyGuides/components/ColorKey.spec.tsx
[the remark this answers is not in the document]
KarinaMendez2: I need the await for the `mockClear()` or else the statement below for `false` never gets called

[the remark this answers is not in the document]
janKozie1: Do you need to `await` it?

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+      spyTrack.mockClear();
+      button.props.onClick();
+    });
+
+    renderer.act(() => {
+      expect(spyTrack).toHaveBeenCalledWith({pathname: '/'}, false);
+    });
```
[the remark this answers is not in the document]
janKozie1: It doesn't need to be wrapped in anything. It's just an assertion on the component, right?

src/app/content/studyGuides/components/ColorKey.tsx
```diff
@@ -0,0 +1,157 @@
...
+  position: relative;
+  overflow: visible;
+  ${theme.breakpoints.mobile(css`
+    margin-right: ${mobilePaddingSides}rem;
+  `)}
+`;
+
+// tslint:disable-next-line:variable-name
+const ColorKey = () => {
+  const [open, setOpen] = React.useState<boolean>(false);
```
janKozie1, line 116: This shouldn't need a type argument

janKozie1 (approved):
Looks good to me

PiotrKozlowski (approved):

TomWoodward (approved):

src/app/content/studyGuides/components/ColorKey.spec.tsx
```diff
@@ -50,19 +48,47 @@ describe('Study Guides button and PopUp', () => {
...
     renderer.act(() => {
-      ReactTestUtils.Simulate.click(component);
-      button.props.onClick();
+      button.props.onClick(); /* closing from Color key button*/
     });
+
     /*mock close call to GA*/
     services.analytics.openCloseColorKey.track({pathname: '/'}, false);
-    expect(spyClickOutside).toHaveBeenCalledWith({current: null}, true, expect.any(Function));
+    expect(() => component.root.findByType(ColorKeyDescription)).toThrow();
```
janKozie1: I believe that if the event should no longer be dispatched on `close` then: 
 - it's better to assert that it actually doesn't get sent after closing
 - rename it so it's name indicates that it's only for opening  sg

KarinaMendez2: It will be implemented eventually, that's why I didn't modify much.

KarinaMendez2: Line 63 is just for code coverage on sending false, since we don't want the GA event right now but will need later.

PR #702, opened 2020-06-25
Different approach for scrolling v2

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/589

This approach is based on this comment: https://github.com/openstax/rex-web/pull/514#issuecomment-648297519

TODO:
- [x] Wait for Tom's opinion on the approach
- [x] Add tests


src/app/content/highlights/components/SummaryPopup/ContextMenu.tsx
```diff
@@ -88,23 +102,39 @@ const HighlightDropdownMenu = React.forwardRef((props, ref) => {
...
   const deleteMessage = 'i18n:highlighting:dropdown:delete';
+  const book = useSelector(bookSelector);
+
+  const page = React.useMemo(() => book
+    ? findArchiveTreeNode(book.tree, sourceId)
+    : null
+  , [book, sourceId]);
+
+  const linkToHighlight = React.useMemo(() => {
+    const search = queryString.stringify({ target: JSON.stringify({ id, type: 'highlight' }) });
```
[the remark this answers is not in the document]
PiotrKozlowski: Hmm, that can be hard since everything which is a string can be passed to url params

src/app/content/highlights/components/SummaryPopup/ContextMenu.tsx
```diff
@@ -88,23 +102,39 @@ const HighlightDropdownMenu = React.forwardRef((props, ref) => {
...
   onDelete: () => void;
   onEdit: () => void;
   onColorChange: (color: HighlightColorEnum) => void;
 }
 
 // tslint:disable-next-line:variable-name
 const ContextMenu = ({
-  color,
-  hasAnnotation,
+  highlight: {
```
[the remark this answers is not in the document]
PiotrKozlowski: Keeping `ContextMenu` as presentational components seem okay to me

src/app/content/highlights/utils/index.ts
```diff
@@ -14,3 +16,14 @@ export {
   updateSummaryHighlightsDependOnFilters,
   getSortedSummaryHighlights
 } from './summaryHighlightsUtils';
+
+export const getHighlightScrollTarget = (object: {[key: string]: any}, hash: string): HighlightScrollTarget | null => {
+  if (hash && object.type === 'highlight' && typeof object.id === 'string') {
```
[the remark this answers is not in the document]
PiotrKozlowski: It is supposed to work with any object because this is what `state.query` in navigation is like
I could change it to accept only an object with specific properties, but then I would have to make another util to check if any object has these properties

src/app/content/routes.ts
```diff
@@ -19,7 +18,7 @@ interface State {
   bookUid: string;
   bookVersion: string;
   pageUid: string;
-  search?: {query: string | null, selectedResult: SelectedResult | null};
+  search?: {query: string | null};
```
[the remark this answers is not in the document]
PiotrKozlowski: I'm almost sure that this is no longer required because I no longer use location.state to retrieve data from, I just didn't check it since it's still WIP. Please leave this conversation unresolved, so I can go back to this later

src/app/content/search/selectors.ts
```diff
@@ -27,6 +29,11 @@ export const selectedResult = createSelector(
   (state) => state.selectedResult
 );
 
+export const hits = createSelector(
+  localState,
+  (state) => state.results ? state.results.hits.hits : null
```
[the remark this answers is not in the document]
PiotrKozlowski: Not sure, but I usually don't try to modify existing code which is not directly related to the pr because it's often leading to the rabbit hole where you spent time on trying to understand another part of the repo (which in some cases might be a good thing after all)

src/app/content/search/components/SearchResultsSidebar/SearchResultContainers.tsx
```diff
@@ -10,7 +10,8 @@ import { Book, Page } from '../../../types';
 import { stripIdVersion } from '../../../utils/idUtils';
 import { closeSearchResultsMobile, selectSearchResult } from '../../actions';
 import { isSearchResultChapter } from '../../guards';
-import { SearchResultChapter, SearchResultContainer, SearchResultPage, SelectedResult } from '../../types';
+import { SearchResultChapter, SearchResultContainer,
+  SearchResultPage, SearchScrollTarget, SelectedResult } from '../../types';
```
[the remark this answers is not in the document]
PiotrKozlowski: I've seen that already in a few other locations, for ex.: https://github.com/openstax/rex-web/blob/master/src/app/navigation/utils.ts and I start to like it

src/app/navigation/utils.ts
```diff
@@ -109,3 +110,15 @@ export const findPathForParams = (params: object, paths: string[]) => {
       paramsInPath.every(({name}) => paramKeys.includes(name.toString()));
   });
 };
+
+export const getTargetFromQuery = (query: OutputParams): {[key: string]: any} | null => {
```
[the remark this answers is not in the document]
PiotrKozlowski: I think it might be against Tom's suggestion from: https://github.com/openstax/rex-web/pull/514#pullrequestreview-435226251
> i don't like having navigation be aware of the specific types of scroll targets. I want the highlight or search logic to be able to tell if the target does or does not apply to them, but having the specific selectors in navigation seems bad to me

but I've already implemented this because I've realized that was part of my another approach after I've done this

Please don't close this conversation so Tom can take a look on it

I think having selectors for `searchScrollTarget` and `highlightScrollTarget` in the `navigation/selectors.ts` would be okay

PiotrKozlowski:
@janKozie1 
> For some results it doesn't scroll to then exactly if that result was navigated to

It seems like there is race condition for scrolling to the selected result and scrolling to the element id, however `scrollTargetManager` is called before `searchHighlightManager.update` so I'm not sure why this is happening in this reversed order and why only sometimes, not always

src/app/navigation/utils.ts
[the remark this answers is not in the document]
janKozie1: I think there can't be more than one at a given time so i believe that the scroll target should be somewhere, with no distinction if it is a highlight or search scroll target. Determine what kind of a scroll target it is when it needs to be handled.

src/app/content/highlights/components/SummaryPopup/ContextMenu.tsx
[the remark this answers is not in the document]
janKozie1: I'm not sure if I understand it. I thought the point was to store serialized scroll targets in the url, then handle them if they fit given criteria. I feel like the script should not care if whatever is in the url param doesn't fit highlight/search scroll target schema

I feel like there should be only one place where the string that will be passed to url param was to be created. Just like there should be one place where the string gets validated and the scroll target object gets created.

janKozie1:
It seems to be always happening to me:
https://drive.google.com/file/d/1HyVUX5YvNNrmYprNO0sjK6AD1q8oEz5F/view?usp=sharing  (the video is mostly uneventful, just me waiting for the "kill page" dialog to open)

PiotrKozlowski:
@janKozie1 Could you please check if any of your highlights in this book can cause this?
https://drive.google.com/file/d/18YLc2DdQpG5qNOLbEWt3C5iID7K66giS/view?usp=sharing

src/app/navigation/utils.ts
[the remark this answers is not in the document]
PiotrKozlowski: For now, I've created `scrollTarget` selector in `navigation` and just check if this is proper scroll target directly in the logic which needs one of its types

janKozie1:
Logging out and logging back in fixed it for whatever reason, my bad. @PiotrKozlowski

src/app/content/highlights/components/SummaryPopup/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    ? findArchiveTreeNode(book.tree, highlight.sourceId)
+    : undefined
+  , [book, highlight.sourceId]);
+
+  React.useEffect(() => {
+    setLink(createHighlightLink(highlight, page, book));
+  }, [highlight, page, book]);
+
+  return link;
+};
```
janKozie1, line 33: Doesn't this essentaily double the amount of re-renders for whatever component would need it? First the component would rerender, then the effect would run, then it would rerender again because the state is set here too.

Reducing amount of rerenders is not necessarily a priority but i feel like this is also against the principle of only "one source of truth"

Besides that, I'm not sure if highly specialized hooks are a good practice. I feel like there should be only a function that gets memoized, unless it's not expensive. I don't really have an another argument against one time use hooks so maybe my opinion is invalid.

src/app/navigation/utils.ts
```diff
@@ -109,3 +114,26 @@ export const findPathForParams = (params: object, paths: string[]) => {
       paramsInPath.every(({name}) => paramKeys.includes(name.toString()));
   });
 };
+
+export const getScrollTargetFromQuery = (
```
janKozie1, line 118: I like this approach and i feel like it should be the same for generating a query from scroll target.

src/app/navigation/utils.ts
```diff
@@ -109,3 +114,26 @@ export const findPathForParams = (params: object, paths: string[]) => {
...
+  query: OutputParams,
+  hash: string
+): HighlightScrollTarget | SearchScrollTarget | null => {
+  if (!hash || !query.target || Array.isArray(query.target)) { return null; }
+  try {
+    const parsed = JSON.parse(decodeURIComponent(query.target));
+    if (
+      parsed instanceof Object
+      && (hasSearchScrollTargetParams(parsed) || hasHighlightScrollTargetParams(parsed))
```
janKozie1, line 127: Is the only difference between a scroll target and scroll target params the presence of `elementId` that is coming from the hash? Maybe it would be enough to just merge the params with hash as soon as possible and validate it as a scroll target?

src/app/content/highlights/components/SummaryPopup/utils.ts
PiotrKozlowski: I feel like `useSelector` is doing almost the same thing, but in this case rewriting it to the `React.useMemo` made it 3 times smaller so I guess this should be our approach

TomWoodward: but why are you using state in this case anyway? highlight, page, and book will all be set from the start so i'm not seeing why any change tracking is necessary

PiotrKozlowski: Only the current page is set from the start, but we need `Page` for specific `highlight`. I wanted to memoize results because in order to create a `linktToHighlight` we have to obtain a page from the book which then is passed to `getBookPageUrlAndParams` and doing so requires flattening of book tree and there is no reason to do this for each highlight on every rerender. Initially, I've used `hook` approach, but `useMemo` is okay in this case

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -77,40 +87,60 @@ export default class PageComponent extends Component<PagePropTypes, PageState> {
...
 
     this.searchHighlightManager.update(prevProps.searchHighlights, this.props.searchHighlights, {
       forceRedraw: highlgihtsAddedOrRemoved,
-      onSelect: this.onHighlightSelect,
+      onSelect: this.onSearchHighlightSelect,
     });
   }
 
-  public onHighlightSelect: OptionsCallback = ({current, selectedHighlight}) => {
+  public onHighlightSelect = (scrollTarget: HighlightScrollTarget | null, highlight: Highlight | null) => {
```
TomWoodward, line 102: i don't feel like we should need to have `onHighlightSelect` or `onSearchHighlightSelect` in PageComponent. can't all that logic be in the managers? 

PageComponent is handling the errors, but i don't think it has to have case by case logic or awareness, you could pass in functions like `clearError`, `setError(subjectId: string, messageKey: string)` into the managers for them to be able to communicate that back.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -49,6 +56,9 @@ const onClickHighlight = (services: HighlightManagerServices, highlight: Highlig
   }
 
   services.getProp().focus(highlight.id);
+  if (services.getProp().scrollTarget) {
+    services.history.replace({hash: '', search: ''});
```
TomWoodward, line 60: we shouldn't be clearing the active search when they click on a highlight

if your goal here is to clear their selected search result, i don't feel like we have to go out of our way to do that until they set a new one, that is how browser behavior already works.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -163,10 +178,22 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
...
 
       highlighter.clearFocus();
-      const focusedId = getProp().focused;
+      const { scrollTarget, focus, focused: focusedId, highlightsLoaded, loggedOut, page } = getProp();
+      const highlightScrollTarget = scrollTarget && scrollTarget.type === 'highlight' ? scrollTarget : null;
       const focused = focusedId && highlighter.getHighlight(focusedId);
-      if (focused) {
-        focused.focus();
+      const scrollTargetHighlight = highlightScrollTarget && highlighter.getHighlight(highlightScrollTarget.id);
+      const toFocus = scrollTargetHighlight || focused;
```
TomWoodward, line 185: i don't think we can set the priority like this, we need to do it in a way where when one of these is changed we process it

src/app/content/highlights/components/Card.tsx
```diff
@@ -120,7 +124,15 @@ const Card = (props: CardProps) => {
     className: props.className,
     highlight: props.highlight,
     isFocused: props.isFocused,
-    onBlur: props.blur,
+    onBlur: () => {
+      if (props.scrollTarget) {
+        services.history.replace({
```
TomWoodward, line 129: again, i don't think we have to clear the scroll target until they set a new one or navigate to a new page (when it should be cleared on its own)

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -81,7 +83,9 @@ interface HighlightListElementProps {
 const HighlightListElement = ({ highlight, locationFilterId, pageId }: HighlightListElementProps) => {
   const [isEditing, setIsEditing] = React.useState(false);
   const [isDeleting, setIsDeleting] = React.useState(false);
+  const book = useSelector(bookSelector);
   const dispatch = useDispatch();
+  const linkToHighlight = React.useMemo(() => createHighlightLink(highlight, book), [highlight, book]);
```
TomWoodward, line 88: can you do this inside `ContextMenu` ?

src/app/content/highlights/components/SummaryPopup/utils.ts
```diff
@@ -0,0 +1,11 @@
+import { Highlight } from '@openstax/highlighter/dist/api';
+import { generateTargetQueryFromScrollTarget } from '../../../../navigation/utils';
+import { Book } from '../../../types';
+import { findArchiveTreeNode } from '../../../utils/archiveTreeUtils';
+import { getBookPageUrlAndParams } from '../../../utils/urlUtils';
+
+export const createHighlightLink = ({ sourceId, id, anchor: elementId }: Highlight, book?: Book) => {
+  const page = book ? findArchiveTreeNode(book.tree, sourceId) : undefined;
+  const target = generateTargetQueryFromScrollTarget({ id, type: 'highlight', elementId });
+  return `${page && book ? getBookPageUrlAndParams(book, page).url : ''}?${target}`;
```
TomWoodward, line 10: instead of generating the actual url here, you could generate the props to set on the ContentLink component, that might be a little simpler

src/app/content/search/hooks/index.ts
```diff
@@ -29,20 +31,31 @@ export const requestSearchHook: ActionHookBody<typeof requestSearch> = (services
     searchStrategy: 's1',
   });
 
+  const params = queryString.parse(services.history.location.search);
+  params.query = payload;
+  services.history.replace({
+    hash: services.history.location.hash,
+    search: queryString.stringify(params),
+  });
```
TomWoodward, line 39: why do we need this block up here? won't all the url stuff be handled by the navigation after receiveSearchResults?

src/app/navigation/utils.ts
```diff
@@ -109,3 +112,40 @@ export const findPathForParams = (params: object, paths: string[]) => {
...
+): HighlightScrollTarget | SearchScrollTarget | null => {
+  if (!hash || !query.target || Array.isArray(query.target)) { return null; }
+  try {
+    const parsed = JSON.parse(decodeURIComponent(query.target));
+    if (parsed instanceof Object) {
+      parsed.elementId = hash.replace('#', '');
+      if (isScrollTarget(parsed)) { return parsed; }
+    }
+    return null;
+  } catch {
```
TomWoodward, line 137: it'd be nice if this catch was only around the JSON.parse or limited to only the deserialization error

src/app/navigation/utils.ts
```diff
@@ -109,3 +112,40 @@ export const findPathForParams = (params: object, paths: string[]) => {
       paramsInPath.every(({name}) => paramKeys.includes(name.toString()));
   });
 };
+
+const isScrollTarget = (
+  object: { [key: string]: any }
+): object is HighlightScrollTarget | SearchScrollTarget => {
```
TomWoodward, line 118: lets create a generic ScrollTarget type which has no subtype knowledge, something like
```
interface {
  elementId?: string;
  type: string;
}
```
and at this point only assert that it matches this type. in other places other logic can have other guards for seeing if it matches their expected format

src/app/navigation/utils.ts
```diff
@@ -109,3 +112,40 @@ export const findPathForParams = (params: object, paths: string[]) => {
...
+    }
+    return null;
+  } catch {
+    return null;
+  }
+};
+
+export const generateTargetQueryFromScrollTarget = (scrollTarget: HighlightScrollTarget | SearchScrollTarget) => {
+  let data: object;
+  if (scrollTarget.type === 'search') {
```
TomWoodward, line 144: this type of branch and bound logic is not good, we should handle this separately in the different domains.

src/app/notifications/components/SearchOrHighlightFailure/index.tsx
```diff
@@ -27,13 +28,14 @@ export const syncState = (prevState: typeof initialState) => {
 };
 
 // Appears when search searchHighlightManager in PageComponent.tsx fails to handle
-// selecting a search result. It's meant to not be dismissable before shouldAutoDismissAfter elapses
+// selecting a search result or when HighlightScrollTarget can't be find on a page.
```
TomWoodward, line 31: is there anything in here that is actually dependent on the subject being a search result or highlight? we could rename it FlashMessageError and i would feel a little better

TomWoodward:
@PiotrKozlowski i submitted some comments, in some places i didn't spend the time to 100% figure out what was going on, so definitely comment back if anything doesn't make sense.

in general i think this approach is looking good

src/app/content/highlights/components/SummaryPopup/utils.ts
PiotrKozlowski: using state is not necessary, this is just something which I'm used to do since in the other project I'm writing quite a few hooks for async methods and then using a state is the only option

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
PiotrKozlowski: I've moved it from the ContextMenu as Jan suggested to leave it as presentional compoment

src/app/content/components/ContentLink.tsx
```diff
@@ -71,6 +77,15 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
       if (onClick) {
         onClick();
       }
+
+      const searchParams = queryString.parse(assertWindow().location.search);
+      if (searchScrollTarget) {
+        searchParams.target = JSON.stringify(omit('elementId', searchScrollTarget));
+      }
+      const options = searchScrollTarget
+        ? { hash: searchScrollTarget.elementId, search: queryString.stringify(searchParams) }
```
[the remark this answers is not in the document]
PiotrKozlowski: I've modified it to always pass `search.query` and if `scrollTarget` is passed as prop (direct prop on the component, not from the state) then it will also set `target` and `hash` from this `scrollTarget`

src/app/content/highlights/components/SummaryPopup/utils.ts
PiotrKozlowski: I feel like `<ContentLink>` is already too heavy and it is hard to manage

src/app/content/search/hooks/index.ts
PiotrKozlowski: looks likes that were some leftovers after my testing

src/app/content/search/hooks/index.ts
```diff
@@ -36,86 +38,89 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
+  const searchResultHit = meta && meta.searchScrollTarget && findSearchResultHit(results, meta.searchScrollTarget);
+  const selectedResult = searchResultHit && meta.searchScrollTarget
+    ? {result: searchResultHit, highlight: meta.searchScrollTarget.index}
+    : getFirstResult(book, payload);
 
-  if (!selectedResult) {
+  if (
+    // selectedResult may equal to null if api did not return any results
+    !selectedResult
+    // We are clearing selected result when requesting a new search so in the theory this should never happen
```
PiotrKozlowski, line 55: this check was added previously so maybe there is a reasone for it

src/app/notifications/components/__snapshots__/Notifications.spec.tsx.snap
```diff
@@ -26,7 +26,7 @@ exports[`Notifications matches snapshot 1`] = `
         className="Button-ayg7nk-0 gkTzYh"
         onClick={[Function]}
       >
-        Update
+        Reload
```
PiotrKozlowski, line 29: Where this change come from?

src/app/navigation/utils.spec.ts
```diff
@@ -241,3 +244,47 @@ describe('matchSearch', () => {
...
+describe('isScrollTarget', () => {
+  it('return true if passed object is vali scroll target', () => {
+    expect(isScrollTarget({ type: 'some-type', elementId: 'elem' })).toEqual(true);
+  });
+
+  it('return false if passed object does not have elementId or type or has wrong type', () => {
+    expect(isScrollTarget({ elementId: '' })).toEqual(false);
+    expect(isScrollTarget({ notType: 'some-type', elementId: 'elem' })).toEqual(false);
+    expect(isScrollTarget({ type: 1, elementId: 'elem' })).toEqual(false);
+    expect(isScrollTarget({ type: 'type', elementId: 1 })).toEqual(false);
```
janKozie1, line 257: You could decouple those assertions into separate tests which would give you clearer tests descriptions

src/app/notifications/components/FlashMessageError/FlashMessageError.spec.tsx
```diff
@@ -148,7 +178,12 @@ describe('SearchFailure', () => {
 
     renderer.act(() => {
       component.update(<MessageProvider>
-        <SearchFailure dismiss={dismiss} mobileToolbarOpen={false} selectedHighlight={{} as any}  />
+        <FlashMessageError
+          messageKey='i18n:notification:search-failure'
+          dismiss={dismiss}
+          mobileToolbarOpen={false}
+          uniqueId={{} as any}
```
janKozie1, line 185: You could pass a different string instead

src/app/content/components/ContentLink.tsx
```diff
@@ -31,6 +33,7 @@ interface Props {
   currentPath: string;
   hasUnsavedHighlight: boolean;
   search: RouteState<typeof content>['search'];
+  scrollTarget?: Exclude<ReturnType<typeof selectNavigation.scrollTarget>, null>;
```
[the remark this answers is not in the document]
PiotrKozlowski: If we allow passing `null` `scrollTarget` it would seem like we do something with it, but in reality, we would omit it as we would do with `undefined`

src/app/content/search/utils.ts
```diff
@@ -171,3 +166,15 @@ export const highlightResults = (
...
+  results: SearchResultHit[],
+  target: SearchScrollTarget
+): SearchResultHit | undefined => {
+  return results.find((result) => result.source.elementId === target.elementId);
+};
+
+export const isSearchScrollTarget = (target: ScrollTarget): target is SearchScrollTarget => {
+  if (target.type === 'search' && typeof target.index === 'number') { return true; }
+  return false;
+};
```
janKozie1, line 180: Similar comment to the other guard here. `search/guards.ts` exists already in this case

src/app/content/search/hooks/index.ts
```diff
@@ -36,86 +38,89 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-        } : undefined
-      )
-    );
-  } else if (savedSearch && savedSearch.selectedResult && !isEqual(savedSearch.selectedResult, selectedResult)) {
-    services.dispatch(selectSearchResult(savedSearch.selectedResult));
-  } else if ((!savedSearch || !savedSearch.query) && query) {
-    services.dispatch(clearSearch());
+export const syncSearch: RouteHookBody<typeof content> = (services) => async() => {
+  const state = services.getState();
+  const navigationQuery = selectNavigation.localState(state).query.query;
```
janKozie1, line 103: There is a selector that selects the `query` but it's also is called `query` so you end up with the same amount of `query`. Not sure which one is better.

src/app/content/components/Page/highlightManager.spec.tsx
```diff
@@ -53,8 +55,12 @@ describe('highlightManager', () => {
       focused: undefined,
       hasUnsavedHighlight: false,
       highlights: [],
+      highlightsLoaded: true,
+      loggedOut: false,
       page,
+      scrollTarget: null,
     };
+    prevProp = prop;
```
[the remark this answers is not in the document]
PiotrKozlowski: It guesses it is more descriptive, but it actually should be a shallow copy, not a reference to the prop, so if there will be a need we will be able to update it like `prevProp.sth = 'asd'` without changing the `prop`

src/app/content/components/Page/highlightManager.spec.tsx
```diff
@@ -199,12 +205,142 @@ describe('highlightManager', () => {
...
+    Highlighter.mock.instances[0].getHighlights.mockReturnValue(mockHighlights);
+    Highlighter.mock.instances[0].getHighlight.mockImplementation((id: string) => keyBy('id', mockHighlights)[id]);
+
+    update(prevProp);
+
+    expect(highlightFocus).toHaveBeenCalledTimes(1);
+    expect(focus).toHaveBeenCalledWith(mockHighlights[1].id);
+  });
+
+  it('calls options.clearError if highlightsLoaded === true', () => {
```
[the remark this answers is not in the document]
PiotrKozlowski: If user is logged out and opens an URL with highlight scroll target in it we want to wait for a page to be loaded and only then display an error. This is why I check if `highlightsLoaded` or `(loggedOut && page)` - `highlightsLoaded` is `null` for not logged users.
`const stateEstablished = (highlightsLoaded || (loggedOut && page));`

This test case is mostly for coverage to test if it works with both ways for `stateEstablished`

src/app/content/highlights/components/SummaryPopup/utils.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    expect(createHighlightLink(mockHighlight, book))
+      .toEqual(`mocked/BookPageUrl/And/Params/?${target}`);
+  });
+
+  it('creates link when book is not passed for highlight', () => {
+    const data = {
+      target: JSON.stringify({ id: mockHighlight.id, type: 'highlight', elementId: mockHighlight.anchor }),
+    };
+    const target = queryString.stringify(data) + `#${mockHighlight.anchor}`;
+    expect(createHighlightLink(mockHighlight)).toEqual(`?${target}`);
```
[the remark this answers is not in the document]
PiotrKozlowski: Maybe you can use it for highlights on the current page

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -28,13 +28,20 @@ if (typeof(document) !== 'undefined') {
 const parser = new DOMParser();
 
 interface PageState {
-  hasSearchError: boolean;
-  selectedSearchResultId: null | string;
+  flashMessageError: boolean;
+  flashMessageErrorId: string | null;
+  flashMessageErrorKey: string | null;
+  flashMessageErrorType: 'highlight' | 'search' | null;
 }
```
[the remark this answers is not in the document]
PiotrKozlowski: If there is a wrong highlight scroll target in the URL then `highlightManager.update(...)` will call `setError` every time when it is updated - for example when clicking on a highlight. Which means that we will be showing over and over the same error message. By having `flashMessageError: boolean` we are managing showing error message, and we use other values to check if we should reset it.

This case could be solved by passing the current error id to the `highlightManager` but I think we should avoid this kind of connection between PageComponent and managers and let PageComponent decide what to do with an error.

PiotrKozlowski:
CI `yarn test:browser` fails for 
but locally it fails for me with:
```
(node:19716) UnhandledPromiseRejectionWarning: Error: Caught error after test environment was torn down

Protocol error (Input.dispatchMouseEvent): Target closed.
    at Promise (/home/piotr/Dokumenty/github/rex-web/node_modules/puppeteer/lib/cjs/common/Connection.js:208:63)
    at new Promise (<anonymous>)
    at CDPSession.send (/home/piotr/Dokumenty/github/rex-web/node_modules/puppeteer/lib/cjs/common/Connection.js:207:16)
    at Mouse.up (/home/piotr/Dokumenty/github/rex-web/node_modules/puppeteer/lib/cjs/common/Input.js:396:28)
    at Mouse.click (/home/piotr/Dokumenty/github/rex-web/node_modules/puppeteer/lib/cjs/common/Input.js:369:22)
    at ElementHandle.click (/home/piotr/Dokumenty/github/rex-web/node_modules/puppeteer/lib/cjs/common/JSHandle.js:373:32)
    at processTicksAndRejections (internal/process/task_queues.js:86:5)
(node:19716) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:19716) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
 FAIL  src/app/content/search/components/SearchResultsSidebar/index.browserspec.tsx (151.003s)
  ● clears search input without affecting search results sidebar

    expect(received).toBe(expected) // Object.is equality

    Expected: ""
    Received: "moon"

      46 |   if (searchInput) {
      47 |     const inputValue = await (await searchInput.getProperty('value')).jsonValue();
    > 48 |     expect(inputValue).toBe('');
         |                        ^
      49 |   }
      50 | });
      51 | 

      at Object.toBe (src/app/content/search/components/SearchResultsSidebar/index.browserspec.tsx:48:24)

  ● closes the search results sidebar without affecting search input

    Timeout - Async callback was not invoked within the 120000ms timeout specified by jest.setTimeout.

      50 | });
      51 | 
    > 52 | it('closes the search results sidebar without affecting search input', async() => {
         | ^
      53 |   setDesktopViewport(page);
      54 |   await navigate(page, TEST_PAGE_URL);
      55 |   await finishRender(page);

      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Object.it (src/app/content/search/components/SearchResultsSidebar/index.browserspec.tsx:52:1)
```

When I change `{hidden: true}` to `{hidden: false}` and add run only this test: `it.only('closes the search...` then it passes
However pushing this change to the pr didn't help

TomWoodward:
@PiotrKozlowski it looks like its breaking on the nudge overlay again

src/app/content/components/Page/highlightManager.ts
```diff
@@ -163,10 +174,37 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
...
+        && !prevProps.focused
+        && scrollTargetHighlight
+      ) {
+        toFocus = scrollTargetHighlight;
+      }
+
+      if (toFocus) {
+        toFocus.focus();
+        if (toFocus.id !== focusedId) {
+          focus(toFocus.id);
```
TomWoodward, line 206: would it be simpler to have a redux hook watching for `locationChange` containing a scrolltarget and dispatch focus from there? i feel like there is a lot of stuff going on right in here and moving that out might reduce a dimension of it

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
TomWoodward: i suppose "presentational" is a subjective thing, i think it would have been fine to have the link generation in there, but whatever you don't have to put it back

src/app/content/highlights/components/SummaryPopup/utils.ts
TomWoodward: i don't understand, i don't think you'd have to change contentlink to use it for this, it already does all but the fist line of this function

you could rename this function `createHighlightLinkProps`, have it return the book, page, and scrolltarget, and just apply those to contentlink

src/app/content/routes.ts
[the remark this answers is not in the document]
TomWoodward: yea i think this can be removed right?

src/app/content/search/hooks/index.ts
TomWoodward: i think the previous check was to make it not reload the search when clicking a figure link on a page when you have search results shown

TomWoodward: you seem to have removed the bit that looks at the target page, so we should make sure we didn't remove any important checks

src/app/content/search/reducer.ts
```diff
@@ -34,11 +33,6 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(actions.clearSearch): {
       return initialState;
     }
-    case getType(locationChange): {
-      return action.payload.action === 'PUSH' && !action.payload.location.state.search
```
TomWoodward, line 38: why would we not want to clear the search data when navigating with a PUSH and not having a new search? you'd have to change this to look in the query instead of the state, but it seems probably still valid

src/app/navigation/utils.ts
```diff
@@ -109,3 +111,32 @@ export const findPathForParams = (params: object, paths: string[]) => {
...
+): ScrollTarget | null => {
+  if (!hash || !query.target || Array.isArray(query.target)) { return null; }
+  let parsed: any;
+  try {
+    parsed = JSON.parse(decodeURIComponent(query.target));
+  } catch {
+    return null;
+  }
+  if (isPlainObject(parsed)) {
+    (parsed as {[key: string]: any}).elementId = hash.replace('#', '');
```
TomWoodward, line 138: this deserialization logic is sort of related to the serialization logic that exists in the ContentLink component, it would be nice if there were helpers for these that lived near one another

src/app/content/components/Page/highlightManager.ts
PiotrKozlowski: It would have to be triggered after `highlightsLoaded` is truthy and it would split focusing logic into one more place which can be tricky to debug in the future and we would only get rid of:
```js
} else if (
  !pendingHighlight
  && !prevProps.focused
  && scrollTargetHighlight
) {
  toFocus = scrollTargetHighlight;
}
```
and
```js
if (toFocus.id !== focusedId) {
  focus(toFocus.id);
}
```
I think I can just create a helper like `const toFocus = getHighlightToFocus(...)`

src/app/content/highlights/components/SummaryPopup/utils.ts
PiotrKozlowski: `createHighlightLink` requires `Book` to be passed to it - I think there is no way to get `Book` for `Highlight` synchronously, is there?
Anyway, since `<ContentLink>` requires `Book` and `Page` to be passed in and already accepts `scrollTarget` I think it has everything to create a link to the highlight  

However, this function (`createHighlightLink`) is used to pass a link to the `<ContextMenu>` and renders a link like this:
```jsx
<DropdownItem
  data-testid='go-to-highlight'
  message='i18n:highlighting:dropdown:go-to-highlight'
  prefix={<StyledLinkIcon/>}
  href={linkToHighlight}
  target='_blank'
/>
```
Would you like me to change this `<DropdownItem>` to custom styled `<ContentLink>`?

src/app/content/search/hooks/index.ts
PiotrKozlowski: I've removed `page.id === stripIdVersion(targetPage.id)` check because page id is stored in selected result so this check should be enough
`isEqual(select.selectedResult(state), selectedResult)`

src/app/content/search/reducer.ts
PiotrKozlowski: When you click on a figure link then there will be PUSH action dispatched with no query and just hash
This would mean that we will remove the search results which doesn't seem good
But maybe we should clear this state when `book` id does not match with selected results id?
```ts
      return state.selectedResult
        && action.payload.action === 'PUSH'
        && action.payload.location.state.pageUid !== stripIdVersion(state.selectedResult.result.source.pageId)
          ? initialState
          : state;
    }
```

PiotrKozlowski: Actually, this is clearing results when clicking on a search result that is on another page so this is also not good

janKozie1:
- Race condition between search highlight manager and scroll manager reacting to hash / target change
- Error triggering when removing a highlight that used to be a scroll target
-

janKozie1:
To fix:
- [ ] `query?=` sometimes being added to the url (seems to happen after closing search results and navigating)
- [ ] showing error that highlight wasn't found after deleting scroll target
- [ ] race condition between two different things trying to scroll at the same time 

Other changes:
- [ ] make highlight manager not aware of the concept of a "scroll target" at all
- [ ] remove potentially unused props from `contentLink`
- [ ] make the guards work with any data
- [ ] make scroll manager do all the scrolling
- [ ] move highlight focusing logic to locationChange hook

src/app/content/components/ContentLink.tsx
```diff
@@ -92,14 +98,12 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
 
 // tslint:disable-next-line:variable-name
 export const ConnectedContentLink = connect(
-  (state: AppState, ownProps: {search?: Partial<RouteState<typeof content>['search']>}) => ({
```
janKozie1, line 95: I think this is no longer used, but tests test for this functionality.

PiotrKozlowski:
@janKozie1 
> make highlight manager not aware of the concept of a "scroll target" at all

and

> make scroll manager do all the scrolling

We've tried that approach before and then switched to splitting it into managers after Tom's review
since there can be only one scroll target in the URL there is no reason to handle scrolling in the different place

---

I also see that now there is no error message shown when opening a link with highlight scroll target in incognito mode

> I feel like there is no point in implementing the error message before this or this

What was wrong with the previous implementation?

---

When there is a highlight scroll target and you click on a highlight it will always scroll to the selected highlight - even if it is fully visible - this is not how it works on the master
**video:** https://drive.google.com/file/d/1DrwP9aN7SopqtLswrNCpL306HdbSR7Bf/view?usp=sharing

janKozie1:
> What was wrong with the previous implementation?
I don't think there is anything wrong with it, it's just that those PRs need to make the toasts controllable from the global level and I felt like it would be odd to have part of them stored in PageComponent and part of them in redux state.

> https://drive.google.com/file/d/1DrwP9aN7SopqtLswrNCpL306HdbSR7Bf/view?usp=sharing
I see, i imagine it should be fixable. I could work on it if you want me to.

> We've tried that approach before and then switched to splitting it into managers
Oh okay, well then I think my changes can be reverted.

PiotrKozlowski:
> I see, i imagine it should be fixable. I could work on it if you want me to.

Yes, please do. You could also fix conflicts with master

> Oh okay, well then I think my changes can be reverted.

I don't know if this is necessary, however, Tom approved the previous approach and I was thinking that we should just fix existing issues / improve code readability and merge it

Did you discuss these changes with Phil maybe?

PiotrKozlowski:
I've added fixes with 100% test coverage for Jan's version however as I pointed out in https://github.com/openstax/rex-web/pull/702#issuecomment-688300699 I'm not convinced to use an approach based on handling all scrolling logic in `scrollTargetManager` because we still need logic in both `highlightManager` and `searchHighlightManager` which will return `scrollTarget` and then we have to pass this scroll target to the `scrollTargetManager` which means that 3 different functions (`highlightManager`, `searchHighlightManager`, `scrollTargetManager`) has to be used together and `<PageComponent>` has to control everything

Previous implementation (https://github.com/openstax/rex-web/pull/702/commits/099b93746c95b1043c7147a806880da2ea31b43f) was, in fact, a little bit more complex in terms of handling scrolling in 3 different places instead of one but had one big advantage - neither `scrollTargetManager` or `<PageComponent>` didn't have to know anything about scrolling targets and since we can have only one valid scroll target in the URL (`hash`, `highlightScrollTarget` or `SearchScrollTarget`) we can safely handle scrolling in separate managers

janKozie1:
> which means that 3 different functions (highlightManager, searchHighlightManager, scrollTargetManager) has to be used together

I'm not sure if i understand. Nor `highlightManager` depends on scrollTarget of `searchHighlightManager` nor vice versa (besides, they already depend on each other even if it did). I'm not sure either why `PageComponent` controlling the behavior of its managers is a bad thing, they depend on its methods and props already.

> neither scrollTargetManager or <PageComponent> didn't have to know anything about scrolling targets

why is `scrollTargetManager` knowing about scrolling targets a bad thing? I agree that it's not that great that PageComponent knows about them.

I went with this approach because scrolling in highlight or search manager felt like a side effect from what they are actually doing. Nor the first nor the latter care about scrolling. I named the parameter they return a `scrollTarget` but it might as well be called 'highlightedNode` which makes sense to me that it would get returned from them. On the other hand, scrollTargetManager cares about scrolling so i thought it would be best if it decided what to do.

Besides that, i know there was an issue with highlightManager's scroll not actually applying correctly (it was scrolling only to the container of the highlight). I feel like it was caused by it and scrollTargetManager trying to scroll, but I'm not sure.

It's unfortunate that highlightManager needs to be involved in checking the scrollTarget, simply because something has to decide that it needs to be ignored after the first scroll. I imagined flipping a boolean in page's state could work too, or doing something in `componentDidMount`. I know that those solutions are not really ideal.

The current fix still makes it scroll to the scroll target highlight when it gets refocused.

PiotrKozlowski:
> I'm not sure either why PageComponent controlling the behavior of its managers is a bad thing, they depend on its methods and props already.

It's more about this part, where you have to have info from 2 other managers to call properly another manager:
```
const scrollTarget = isHighlightScrollTarget(urlScrollTarget)
? highlightUpdate.scrollTarget
: isSearchScrollTarget(urlScrollTarget)
  ? searchUpdate.scrollTarget
  : undefined;

await this.scrollTargetManager({...scrollManagerProps, htmlNode: scrollTarget});
```

For me, it's just more intuitive that `highlightManager` will be responsible for managing everything related to highlights
I would, however, propose changing the name of `scrollTargetManager` to sth like `scrollToTopOrHashManager` because this was initially a behavior that this module implemented

> Besides that, i know there was an issue with highlightManager's scroll not actually applying correctly (it was scrolling only to the container of the highlight). I feel like it was caused by it and scrollTargetManager trying to scroll, but I'm not sure.

It could happen in case of promises race, we should investigate and fix that

> It's unfortunate that highlightManager needs to be involved in checking the scrollTarget, simply because something has to decide that it needs to be ignored after the first scroll. 

I agree that this is unfortunate, but if you will handle everything in `scrollTargetManager` then how would you know that the `highlight` that you are scrolling to really exists and is already mounted? You would have to pass `highlightManager` and also some other props to check if `highlights` were already loaded

TomWoodward:
i agree with @PiotrKozlowski's comment about the scrollTargetManger having a bad name
> I would, however, propose changing the name of scrollTargetManager to sth like scrollToTopOrHashManager because this was initially a behavior that this module implemented

re:
> I'm not sure either why PageComponent controlling the behavior of its managers is a bad thing, they depend on its methods and props already.

there is an inversion of control between the page component and the managers. yes, the managers depend on the component's props, but they control the props by defining their own selectors that the component composes. the idea is that if the component has logic related to the managers' functionality, it will become overloaded by doing too many things. instead the component should only do the minimum amount of logic possible to simply hook up the managers, and nothing else. additionally, as much as possible, the managers should maintain a consistent api (like they all accept props via a selector they define) so that the composition of all the managers can be as generic as possible

PiotrKozlowski:
I've created a pr with the previous approach
https://github.com/openstax/rex-web/pull/864

PiotrKozlowski:
We will go with https://github.com/openstax/rex-web/pull/864

PR #716, opened 2020-07-07
Chapter filter for logged out state

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1154


src/app/content/studyGuides/components/Filters.spec.tsx
```diff
@@ -49,4 +50,29 @@ describe('Filters', () => {
...
+      </MessageProvider>
+    </Provider>);
+
+    let message: string | undefined;
+    try {
+      component.root.findByType(ConnectedFilterList);
+    } catch (e) {
+      message = e.message;
+    }
+    expect(message).toEqual('No instances found with node type: "Connect(FiltersList)"');
```
PiotrKozlowski, line 75: This works, however, it may be a problem if a message will change in the future due to some dependencies upgrade
We've been using `expect(() => component.root.findByType(ConnectedFilterList)).toThrow();` approach in some of the other cases which seems fine also for this one

PiotrKozlowski (dismissed):

src/app/content/studyGuides/components/Filters.spec.tsx
TomWoodward: this strategy with the try/catch comparing the message is necessary for async logic, we do that in other places. in this case i think piotr's suggestion would be better

KarinaMendez2: I tried that before, it throws the error anyways.

PiotrKozlowski: @KarinaMendez2 Did you `expect` `() => component.root.findByType(ConnectedFilterList)` or only `component.root.findByType(ConnectedFilterList)` - the first will work, the second will throw an error

src/app/components/AllOrNone.tsx
```diff
@@ -10,15 +12,19 @@ interface Props {
...
-  <FormattedMessage id='i18n:highlighting:filters:all'>
-    {(msg: Element | string) => <ButtonLink decorated onClick={onAll}>{msg}</ButtonLink>}
-  </FormattedMessage>
-  <span>|</span>
-  <FormattedMessage id='i18n:highlighting:filters:none'>
-    {(msg: Element | string) => <ButtonLink decorated onClick={onNone}>{msg}</ButtonLink>}
-  </FormattedMessage>
-</div>;
+const AllOrNone = ({className, onAll, onNone}: Props) => {
+  const userLoggedOut = useSelector(loggedOut);
```
TomWoodward, line 16: this component is shared in places outside the scope of what you're trying to change, you can't add this logic here.

you could add a `disabled` prop, that would be ok

src/app/components/Button.tsx
```diff
@@ -109,8 +109,12 @@ export const PlainButton = styled.button`
 // tslint:disable-next-line:variable-name
 export const ButtonLink = styled(PlainButton)`
   outline: none;
-  ${textStyle}
-  ${(props: {decorated: boolean}) => props.decorated ? decoratedLinkStyle : linkStyle}
+  ${(props: {decorated: boolean, disabled: boolean}) =>
+    props.disabled ? `
+      cursor: not-allowed;
```
TomWoodward, line 114: are you sure you want to remove the link styling? is there design for this? i would think that you'd leave the existing styling and just add the disabled cursor.

you should do the cursor in a separate `${}` of the `ButtonLink` styles, there is no reason to make this logic so crazy.

src/app/components/Button.tsx
```diff
@@ -109,8 +109,12 @@ export const PlainButton = styled.button`
...
+    `
+    : props.decorated ? {textStyle, decoratedLinkStyle} : {textStyle, linkStyle}
```
TomWoodward, line 116: is this valid? i don't think this is valid

src/app/components/Checkbox.tsx
```diff
@@ -32,7 +32,7 @@ const CustomCheckbox = styled(
   }
 
   input:checked + & {
-    background-color: ${theme.color.primary.orange.darkest};
+    background-color: ${(props) => props.disabled ? theme.color.primary.gray.base : theme.color.primary.orange.darkest};
```
TomWoodward, line 35: don't you want to set the disabled style regardless of whether its checked or not?

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -67,6 +69,8 @@ const ChapterFilter = ({
   selectedLocationFilters,
   setFilters,
 }: Props) => {
+  const userLoggedOut = useSelector(loggedOut);
```
TomWoodward, line 72: you can't add this auth logic here

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -31,13 +32,16 @@ const ConnectedFilterList = connect(
   })
 )(FiltersList);
 
-export default () =>
-  <Filters>
+export default () => {
+  const userLoggedOut = useSelector(loggedOut);
```
TomWoodward, line 36: this is the correct place for the auth logic, from here just pass `disabled` props down where they are needed

TomWoodward (changes requested):

src/app/components/Button.tsx
KarinaMendez2: it worked haha

src/app/components/Checkbox.tsx
KarinaMendez2: no, because this sets the background of the checked box gray when disabled, which is only if it's in logged out state, that's what's in the design

TomWoodward: yea for the background, but i think the border needs to be changed when disabled as well

src/app/components/AllOrNone.tsx
```diff
@@ -7,18 +7,21 @@ interface Props {
...
-  <FormattedMessage id='i18n:highlighting:filters:all'>
-    {(msg: Element | string) => <ButtonLink decorated onClick={onAll}>{msg}</ButtonLink>}
-  </FormattedMessage>
-  <span>|</span>
-  <FormattedMessage id='i18n:highlighting:filters:none'>
-    {(msg: Element | string) => <ButtonLink decorated onClick={onNone}>{msg}</ButtonLink>}
-  </FormattedMessage>
-</div>;
+const AllOrNone = ({className, onAll, onNone, disabled}: Props) => {
+  return <div className={className}>
```
TomWoodward, line 15: why did you change this?

src/app/components/Checkbox.tsx
```diff
@@ -18,7 +18,7 @@ interface Props {
 }
 
 // tslint:disable-next-line:variable-name
-const CustomCheckbox = styled(
+export const CustomCheckbox = styled(
```
TomWoodward, line 21: why?

src/app/components/Typography/index.ts
```diff
@@ -52,3 +53,16 @@ export const labelStyle = css`
   line-height: 1.6rem;
   font-weight: normal;
 `;
+
+export const disabledStyle = css`
```
TomWoodward, line 57: can you add this to `linkStyle` and `decoratedLinkStyle` instead of making a new thing? i'm pretty sure those css blocks can have functions in them that look at props

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -9,7 +10,7 @@ import { setSummaryFilters } from '../actions';
 import * as selectors from '../selectors';
 
 // tslint:disable-next-line:variable-name
-const ConnectedChapterFilter = connect(
+export const ConnectedChapterFilter = connect(
```
TomWoodward, line 13: are you using this?

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -21,7 +22,7 @@ const ConnectedChapterFilter = connect(
 )(ChapterFilter);
 
 // tslint:disable-next-line:variable-name
-const ConnectedFilterList = connect(
+export const ConnectedFilterList = connect(
```
TomWoodward, line 25: or this?

TomWoodward (changes requested):

TomWoodward (approved):

PR #721, opened 2020-07-07
Gh actions selenium

TomWoodward:
for: openstax/unified#1193

### DIFF
--- a/.github/workflows/ci-selenium.yml
+++ b/.github/workflows/ci-selenium.yml
@@ -8,7 +8,13 @@ jobs:
     runs-on: ubuntu-latest
     container: openstax/rex-web:CI
     steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.4.0
+      with:
+        access_token: ${{ github.token }}
     - uses: actions/checkout@v2
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
     - name: find deployment
       env:
         GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/ci-selenium.yml
+++ b/.github/workflows/ci-selenium.yml
@@ -1,9 +1,10 @@
-name: Selenium Smoke Test
+name: CI
 
 on: [pull_request]
 
 jobs:
   selenium:
+    name: Selenium Smoke Test
     runs-on: ubuntu-latest
     container: openstax/rex-web:CI
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-selenium.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    container: openstax/rex-web:CI
+    steps:
+    - uses: actions/checkout@v2
+    - name: find deployment
+      env:
+        GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      run: |
+        . concourse/common/get_review_environment
+        review_url=$(get_review_environment)
```
PiotrKozlowski, line 16: I'm not really familiar with bash. Is this `${get_review_environment}` a returned valud from `. concourse/common/get_revview_environment`?

PiotrKozlowski (commented):
In logs: https://github.com/openstax/rex-web/pull/721/checks?check_run_id=847303711
I can see:
`jq: error (at <stdin>:4): Cannot index object with number`
I don't know however on which `jq` call this is throwing an error, but I can see that in the line 29 in `get_review_environment` file `.[0].state` is not in `''` like it is in the other statements so my guess would be to check if wrapping it like `jq -r '.[0].state'` will help and if not then add some console logs to figure out why this is throwing an error

.github/workflows/ci-selenium.yml
TomWoodward: `get_review_environment` is a function defined in `concourse/common/get_review_environment`. the `. <filename>` sources the file so you can call the function in it.

the `jq: error (at <stdin>:4): Cannot index object with number` error happens when the github api is rate limited. i'm going to do some testing today about that

TomWoodward:
@PiotrKozlowski this is ready for review now, sorry i didn't make it as a draft to begin with

concourse/common/get_review_environment
```diff
@@ -7,33 +7,42 @@ then
 fi
 
 function github {
-  curl -s -H "Authentication: token $GITHUB_ACCESS_TOKEN" "https://api.github.com/$1"
+  response=$(curl -s -H "Authentication: token $GITHUB_ACCESS_TOKEN" "https://api.github.com/$1")
+  echo "https://api.github.com/$1">&2
+  echo "$response">&2
+  echo "$response"
```
PiotrKozlowski, line 13: do we want to leave all of these 'echo`es in this file?

PiotrKozlowski (approved):

concourse/common/get_review_environment
TomWoodward: yea they are captured in the build logs and have been helpful for debugging

PR #723, opened 2020-07-09
Do not use regexp to set attributes for links

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1025

There are ~two~ three issues causing this to load so long.

TODO:
- [x] This regexp: `.replace(/<a(.*?href="\.\.\/.*?)>/g, '<a target="_blank"$1>')` is consuming to much processing power. Since it is related to the regexp above I've just moved this logic to the `transformContent` function
- [x] The second issue is with `loadContentReferences` which is blocking rendering for some time - caused by running `flatteArchiveTree` on each reference (there are 3190 references on Index page from the issue)
- [x] `contentLinks.reduceReferences` was using `html.replace` to update hrefs which also was costly. I've rewritten it to use dom processing
- [x] Add tests

~~I'm not sure about the caching method for `flattenArchiveTree` because it is messing with some tests cases and it may be hard to investigate in the future for devs when they write new tests~~
Edit: I've just disabled caching of this function for testing env


src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -209,3 +210,18 @@ function moveFootnotes(document: Document, rootEl: HTMLElement, intl: IntlShape)
     sup.appendChild(link);
   }
 }
+
+const setLinksAttributes = (rootEl: HTMLElement) => {
+  for (const a of Array.from(rootEl.querySelectorAll('a'))) {
+    const href = a.getAttribute('href') || '';
```
janKozie1, line 216: You could `continue` in case that it's missing but i doubt it will change much

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -11,8 +11,13 @@ import {
 } from '../types';
 import { getIdVersion, stripIdVersion } from './idUtils';
 
+// This is exported so it can be cleared for tests cases
+export const CACHED_FLATTENED_TREES = new Map<string, Array<LinkedArchiveTree | LinkedArchiveTreeSection>>();
 export function flattenArchiveTree(tree: LinkedArchiveTree): Array<LinkedArchiveTree | LinkedArchiveTreeSection> {
-  return [tree, ...flatten(tree.contents.map((section) =>
+  if (CACHED_FLATTENED_TREES.has(tree.id)) {
+    return CACHED_FLATTENED_TREES.get(tree.id)!;
```
janKozie1, line 18: I think the preferred way of doing it is `assertDefined` but it shouldn't make a difference

janKozie1 (commented):
I think it looks good, only left two comments that don't really need to be addressed.

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -209,3 +210,18 @@ function moveFootnotes(document: Document, rootEl: HTMLElement, intl: IntlShape)
     sup.appendChild(link);
   }
 }
+
+const setLinksAttributes = (rootEl: HTMLElement) => {
+  for (const a of Array.from(rootEl.querySelectorAll('a'))) {
```
TomWoodward, line 215: is NodeList not iterable? i'm not sure you really need the `Array.from` here

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -209,3 +210,18 @@ function moveFootnotes(document: Document, rootEl: HTMLElement, intl: IntlShape)
...
+    const href = a.getAttribute('href') || '';
+    if (href.startsWith('https://') || href.startsWith('http://') || href.startsWith('//')) {
+      // target blank and add `rel` to links that begin with: http:// https:// //
+      a.setAttribute('target', '_blank');
+      a.setAttribute('rel', 'noopener nofollow');
+    } else if (href.startsWith('../')) {
```
TomWoodward, line 221: relative content won't always start with this, eg page urls just have the page slug in the href

TomWoodward (changes requested):

src/app/content/components/Page/contentDOMTransformations.ts
TomWoodward: oh but this is actually correct, the comment is just misleading.

the actual goal is to `// target blank and allow indexing links to other books`

TomWoodward (approved):
non-blocking comments

src/app/content/components/Page/contentDOMTransformations.ts
TomWoodward: the desire is for links outside the book to be `target: blank` but we would be fine to allow indexing to all relative links

TomWoodward:
@PiotrKozlowski this has some linting errors

src/app/content/utils/archiveTreeUtils.ts
PiotrKozlowski: I'm always feeling weird when doing `assertDefined()` in cases when I've just checked for it, but there is one advantage: it is harder to overlook when changing the code

src/app/content/components/Page/contentDOMTransformations.ts
PiotrKozlowski: It can't be used in for of loop because it lacks Symbol.iterator() - according to my linting plugin, but it has `.forEach` method built it. I'll use it

PiotrKozlowski:
I've disabled caching for `flattenArchiveTree` when `NODE_ENV === 'test'` and now there is not problem with clearing it in tests

PR #727, opened 2020-07-14
Fix scrolling in safari: search sidebar header

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/869


src/app/content/search/components/SearchResultsSidebar/SearchResultsBarWrapper.tsx
```diff
@@ -125,6 +125,18 @@ export class SearchResultsBarWrapper extends Component<ResultsSidebarProps> {
...
     this.scrollToSelectedPage();
+
+    const searchSidebar = this.searchSidebar.current;
+
+    if (!searchSidebar) {
+      return;
+    }
+
+    const sidebarChildren = searchSidebar.children;
+    if (sidebarChildren.length === 2) {
```
TomWoodward, line 136: did you try setting `Styled.NavOl` to fill the remaining height with flexbox or something and setting that to overflow:auto?

TomWoodward (dismissed):

PiotrKozlowski (commented):
Is there any reason why `SearchResultsBar` has `overflow: auto` which makes it scrollable instead of making `NavOl` scrollable?
sth like:
```
.NavOl {
overflow: scroll;
height: 100%;
}

.SearchResultsBar {
overflow: visible;
}
```

KarinaMendez2:
@PiotrKozlowski 

1. There's a function called `fixSafariScrolling` which affects both TOC and Search sidebars, that's the one that sets their overflows to auto.

2. By setting `height: 100%` on the NavOl, this gets the height from the parent, which is `SearchResultsBar`, this height includes the `SearchQuery`, so this generates an additional scroll.

src/app/content/search/components/SearchResultsSidebar/SearchResultsBarWrapper.tsx
KarinaMendez2: no :dizzy_face: one sec

PiotrKozlowski:
@KarinaMendez2 
> There's a function called fixSafariScrolling which affects both TOC and Search sidebars, that's the one that sets their overflows to auto.

I think it may not be needed for search sidebar

> By setting height: 100% on the NavOl, this gets the height from the parent, which is SearchResultsBar, this height includes the SearchQuery, so this generates an additional scroll.

I wasn't specific about values but wanted to just suggest CSS solution which I've tested very rapidly and it seems to work. You would have to also remove `position: sticky` and related CSS from `SearchQueryWrapper`

Since `SearchQueryWrapper` is above `NavOl` and `NavOl` will be scrollable, and their parent `SearchResultsBar` will not be scrollable, everything should work as expected without any js

src/app/content/search/components/SearchResultsSidebar/SearchResultsBarWrapper.tsx
```diff
@@ -125,6 +125,12 @@ export class SearchResultsBarWrapper extends Component<ResultsSidebarProps> {
 
   public componentDidUpdate() {
     this.scrollToSelectedPage();
+
+    const searchSidebar = this.searchSidebar.current;
+
+    if (!searchSidebar) {
+      return;
+    }
```
PiotrKozlowski, line 133: Will this prevent anything? Since it is in `componentDidUpdate` everything should be already updated so this is most likelly not preventing anything from happening

src/app/content/search/components/SearchResultsSidebar/styled.tsx
```diff
@@ -128,6 +126,9 @@ export const SearchResultsBar = styled.div`
   }
 
   ${disablePrint}
+
+  display: flex;
+  flex-direction: column;
```
PiotrKozlowski, line 131: This is not changing anything in Firefox and Chrome, is it needed?

KarinaMendez2: Yes, like the card and PR say, it's for Safari.

KarinaMendez2: Yes, like the card and PR both say, this is a fix for Safari.

PiotrKozlowski: I wasn't expecting standard display being broken in Safari ¯\_(ツ)_/¯

KarinaMendez2: This happens when you remove `display: flex` and do what you suggested:

image

PiotrKozlowski (approved):

src/app/content/search/components/SearchResultsSidebar/styled.tsx
KarinaMendez2: Yeah and that would work perfectly, but the thing is that we don't know how big the search query will get, for example searching for a sentence, search query doesn't have a `max-height` so we don't really know what height to expect that's why I tried doing it with `js` but Tom suggested `flexbox`, just wanted to explain why I took this solution.

image

PiotrKozlowski: ohh, now it makes perfect sens, thanks

KarinaMendez2: Safari breaks every now and then, so we've had to come up with solutions and then remove them or put cleaner ones, but in this case I didn't want to remove anything if it wasn't needed.

PR #728, opened 2020-07-16
Upgrade puppeteer

TomWoodward:
this fixes the books check but breaks screenshot tests because of some behavior in the new chrome

i'm gonna either overhaul or remove the screenshot tests in a separate change

for: openstax/unified#1198


TomWoodward:
idk why this is broken in travis but it works on my local mac, local docker, and in github actions on #726, so i'm closing this and including there

PR #732, opened 2020-07-19
Alignment for questions inside exercise sections: generic styles

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/935


generic-styles/content-common.less
```diff
@@ -364,6 +364,16 @@ blockquote {
     }
   }
 
+  [data-type="problem"] {
+    display: flex;
+    align-items: baseline;
+
+    .os-problem-container {
+      display: grid;
```
PiotrKozlowski, line 372: it works for me without this line

KarinaMendez2: In every browser? For me in Chrome and Safari it doesn’t.

KarinaMendez2: Specifically the sentences that start with capital vocals, the number goes lower.

KarinaMendez2: image

PiotrKozlowski: I'm starting to hate that browsers lack compatibility 
Please add a comment above this line because it's weird to see just `display: grid` in that place

janKozie1, line 372: I'm not sure why `display: grid` fixes the issue, but I think it's worth noting that setting it makes its contents to be in rows, which could be undesirable if `.os-problem-container` had some elements that were supposed to fit in one line.

Currently, It's not the case, but I think it's not unreasonable for a `span` to be inside it.

TomWoodward:
@KarinaMendez2 was this `display: grid` fix copied from cnx.org or did you come up with it on your own?

KarinaMendez2:
@TomWoodward I came up with it

TomWoodward:
please reset your changes in the common file so there are no conflicts @KarinaMendez2

generic-styles/content-baked.less
```diff
@@ -314,6 +314,10 @@ h1.example-title .text {
     p:first-of-type {
       display: inline-block;
     }
+
+    .os-problem-container, .os-problem-container > p {
```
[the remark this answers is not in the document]
TomWoodward: this solution was copied from https://vendor.cnx.org/contents/zOZP3vRI@13.9:XwcQ_hAo@9/1-1-Physics-Definitions-and-Applications so we can have some confidence that it works with other variants in the content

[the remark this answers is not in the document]
KarinaMendez2: Yeah I did the wrong change before because I came up with the fix, when actually I should've checked the styles from `cnx` and copy them, so I fixed that based on the real thing, generic styles is copying styles from `cnx` to `rex`, so this should be fine, styles when highlighting has had issues before, depends on where you start highlighting but I think those aren't exactly related to this.

PR #735, opened 2020-07-21
Adding spacing to list items: generic styles

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/896


generic-styles/content-baked.less
```diff
@@ -26,6 +26,10 @@ ul, ol {
   color: @gray;
 }
 
+ol li {
+  margin-top: 1rem;
+}
```
PiotrKozlowski, line 31: This will add also margin-top to the first item which will result in kind of margin-top: 2rem for `ol`
It will also add this margin to any nested `li` - I think you should ask someone to specify behavior for this

KarinaMendez2:
More research needed.

PR #736, opened 2020-07-27
Issue 1138

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1138


src/app/content/components/NudgeStudyTools/index.tsx
```diff
@@ -55,6 +57,17 @@ const NudgeStudyTools = () => {
...
+  const ariaLabelKey = hasStudyGuides
+    ? 'i18n:nudge:study-tools:aria-label:with-study-guides'
+    : 'i18n:nudge:study-tools:aria-label:only-highlighting';
+
+  const messageKey = hasStudyGuides
+    ? 'i18n:nudge:study-tools:text:with-study-guides'
+    : 'i18n:nudge:study-tools:text:only-highlighting';
+
+  // tslint:disable-next-line: variable-name
+  const NudgeText = htmlMessage(messageKey, NudgeTextStyles);
```
janKozie1, line 69: Creating components inside components changes their reference each render (at least in functional components) so they are always re-mounted, which probably won't cause any problems here but should generally be avoided.

src/app/messages/en/messages.json
```diff
@@ -75,8 +75,10 @@
...
-  "i18n:nudge:study-tools:aria-label": "Image points to the Study Guides and My Highlights buttons in the toolbar. The study tools you need. 100% free. Boost your studies with expert-created study guides and highlighting. To close image, navigate to or click the x button.",
+  "i18n:nudge:study-tools:text:with-study-guides": "Boost your studies with expert-created study guides and highlighting. <strong>It's all free.</strong>",
+  "i18n:nudge:study-tools:text:only-highlighting": "Highlight, take notes, and make your own study guides. <strong>It's all free.</strong>",
+  "i18n:nudge:study-tools:aria-label:with-study-guides": "Image points to the Study Guides and My Highlights buttons in the toolbar. The study tools you need. 100% free. Boost your studies with expert-created study guides and highlighting. To close image, navigate to or click the x button.",
+  "i18n:nudge:study-tools:aria-label:only-highlighting": "Image points to the My Highlights buttons in the toolbar. The study tools you need. 100% free. Highlight, take notes, and make your own study guides. To close image, navigate to or click the x button.",
```
janKozie1, line 81: The text is different from design doc, but this seems more accurate. Was the doc not updated? Also shouldn't it be  `My highlights button`?

src/app/content/components/NudgeStudyTools/index.spec.tsx
```diff
@@ -112,6 +110,30 @@ describe('NudgeStudyTools', () => {
...
+        </MessageProvider>
+      </Services.Provider>
+    </Provider>);
+
+    expect(() => component.root.findByProps({
+      id: 'i18n:nudge:study-tools:aria-label:with-study-guides',
+    })).not.toThrow();
+    expect(() => component.root.findByProps({
+      id: 'i18n:nudge:study-tools:text:with-study-guides',
+    })).not.toThrow();
```
janKozie1, line 136: I think it would be nice if this also asserted that the previous text is no longer shown, probably just one of the assertions from above is enough.

src/app/content/components/NudgeStudyTools/utils.ts
```diff
@@ -42,7 +44,11 @@ export const getPositions = (target: HTMLElement, isMobile: boolean, windowWidth
   const spotlightHeight = height + (padding * 2);
   const spotlightWidth = width + (padding * 2);
   const arrowTopOffset = spotlightTopOffset + spotlightHeight + remsToPx(arrowTopMargin);
-  const arrowLeft = spotlightLeftOffset + remsToPx(arrowLeftMargin);
+  // right edge of arrow image should be on the middle of the spotlight (adjusted for a margin)
+  const centerPoint = spotlightLeftOffset + (spotlightWidth / 2);
+  const arrowWidth = remsToPx(isMobile ? arrowMobileWidth : arrowDesktopWidth);
+  const arrowLeft = centerPoint - arrowWidth + remsToPx(arrowLeftMargin);
```
janKozie1, line 50: I imagine that if you had a bounding box of the spotlight, then you could probably position all of the other elements with plain css, but I'm not sure if it's worth changing.  Up to you which approach you think is better.

janKozie1 (changes requested):
The requirements also mention that `Users will be tracked per account if logged in`. The previous nudge had the same requirement but it was removed, does that apply to this nudge as well?

PiotrKozlowski:
> The requirements also mention that Users will be tracked per account if logged in. The previous nudge had the same requirement but it was removed, does that apply to this nudge as well?

and https://github.com/openstax/rex-web/pull/736#discussion_r461609319

@janKozie1  https://openstax.slack.com/archives/C8V1XET1R/p1596028719376700

src/app/components/htmlMessage.tsx
```diff
@@ -9,8 +9,8 @@ type Type = <T extends any>(messageKey: string, Component: ComponentType<HTMLAtt
   ComponentType<Props & HTMLAttributes<T>>;
 
 // tslint:disable-next-line:variable-name
-const htmlMessage: Type = (messageKey, Component) => ({values, ...props}) =>
-  <FormattedHTMLMessage id={messageKey} values={values ? values : {}}>
+const htmlMessage: Type = (messageKey, Component) => ({values, id, ...props}) =>
+  <FormattedHTMLMessage id={id || messageKey} values={values ? values : {}}>
```
janKozie1, line 13: I'm not sure about this. You can pass `id` because the component `htmlMessage` returns is allowed to get any html attributes which i think implies that those attributes (including `id`) would get forwarded to the dom node. 
In this case it's true for all attributes except `id`

janKozie1 (changes requested):

janKozie1 (approved):

src/app/components/htmlMessage.tsx
```diff
@@ -5,14 +5,17 @@ import { assertString } from '../utils';
...
-const htmlMessage: Type = (messageKey, Component) => ({values, ...props}) =>
-  <FormattedHTMLMessage id={messageKey} values={values ? values : {}}>
+const htmlMessage: Type = (msgKey, Component) => ({values, ...props}) =>
+  <FormattedHTMLMessage id={msgKey} values={values ? values : {}}>
     {(msg: string | Element) =>
-      <Component dangerouslySetInnerHTML={{__html: assertString(msg, `${messageKey} must be a string`)}} {...props} />
+      <Component
+        dangerouslySetInnerHTML={{__html: assertString(msg, `${msgKey} must be a string`)}}
+        {...props}
+      />
```
janKozie1, line 18: Should those changes be reverted? They don't change anything

TomWoodward:
the test for GA data is breaking because it has a very brittle expectation, you could change that to check for an element of the array, or you could change it so the nudge isn't shown for that test, by setting the cookie or something

TomWoodward (changes requested):
these changes look fine, but one of the tests is breaking

TomWoodward (approved):

PR #738, opened 2020-07-28
Issue 1213

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1213


src/app/content/highlights/hooks/loadMore.spec.ts
```diff
@@ -130,6 +132,7 @@ describe('filtersChange', () => {
       }))
     ;
 
+    filters = store.getState().content.highlights.summary.filters;
```
janKozie1, line 135: Do the filters get changed somewhere?

src/app/content/highlights/reducer.spec.ts
```diff
@@ -406,6 +407,36 @@ describe('highlight reducer', () => {
...
+          ...initialState.summary,
+          filters: {
+            colors: [HighlightColorEnum.Green],
+            locationIds: ['id'],
+          },
+          loading: true,
+        },
+      };
+
+      const staleFilters = {} as any as SummaryFilters;
```
janKozie1, line 432: You could add `locationIds: [], colors: []` into the object and it wouldn't need to be casted.

janKozie1 (commented):
You could write a test with timeouts if you wanted to replicate the problem. 
Left some minor comments too.

janKozie1 (approved):

src/app/content/highlights/reducer.ts
```diff
@@ -237,6 +238,10 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
       };
     }
     case getType(actions.receiveSummaryHighlights): {
+      // Check if filters wasn't updated while we were loading response.
+      // It may happen if user with slow network connection change filters very fast.
+      if (action.meta.filters && !equals(state.summary.filters, action.meta.filters)) { return state; }
```
TomWoodward, line 243: you can use strict equality here, i don't think lodash equals is necessary

TomWoodward (changes requested):

TomWoodward (approved):

PR #741, opened 2020-07-29
Allow clicking through spotlight nudge

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1219


src/app/content/reducer.ts
```diff
@@ -98,6 +100,9 @@ function reduceContent(state: State, action: AnyAction) {
     case getType(actions.openNudgeStudyTools): {
       return {...state, showNudgeStudyTools: true };
     }
+
```
PiotrKozlowski, line 103: newline to needed

PiotrKozlowski (approved):
Cool! I wouldn't think of using `pointer-events`

TomWoodward:
is it possible to do the pointer events thing on only the spotlight? with this change i'm able to click through to any element behind the overlay anywhere on the page, which isn't totally ideal

janKozie1:
I don't think so since it was a child of the overlay. If only it had them disabled then the overlay would be getting clicks.
I remade the spotlighting a little and it should work now.

cc @TomWoodward

PR #745, opened 2020-07-30
Display newlines in study guides notes

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1220


src/app/content/studyGuides/components/StudyGuidesListElement.tsx
```diff
@@ -35,6 +35,7 @@ const HighlightAnnotation = styled.div`
   display: flex;
   padding: 1.2rem 0;
   color: ${theme.color.text.black};
+  white-space: pre-wrap;
```
PiotrKozlowski, line 38: In MH we are also using:
```
word-break: break-word;
overflow-wrap: break-word;
```
Should we add it here to make it consistent?

janKozie1: I thought about it but it felt little unrelated.

Pretty sure `overflow-wrap: break-word` is obsolete when `word-break` is set to `break-word`. I think its only relevant if a single word would span over the width of the container; it seems to break URLs without it. 

I've added the `word-break`

PiotrKozlowski (approved):

PR #748, opened 2020-07-31
Allow invalid links unless ?validateLinks is present

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1217


src/app/content/components/Page/validateDOMContent.ts
```diff
@@ -15,7 +16,17 @@ const validateLinks = (rootEl: HTMLElement) => {
   const urls = Array.from(rootEl.querySelectorAll('a[href^="/"]'))
     .map((element) => element.getAttribute('href'));
 
-  if (urls.length > 0) {
-    throw new Error(['found invalid links in content: ', ...urls].join('\n'));
+  if (urls.length === 0) {
+    return;
+  }
+
+  const throwOnInvalidLink = 'validateLinks' in query;
```
TomWoodward, line 23: perhaps you should switch on `config.UNLIMITED_CONTENT`, which is set based on the presence of this query param already in `src/config.development.js`

the advantage to using the config is that it is set up on page load so the query param doesn't need to be in every subsequent page, i think that might lead to issues if we depend on it being there

TomWoodward (changes requested):

TomWoodward (approved):

PR #752, opened 2020-08-05
Stop SG from loading the same data twice

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1231

### DIFF
--- a/src/app/content/studyGuides/components/ShowStudyGuides.tsx
+++ b/src/app/content/studyGuides/components/ShowStudyGuides.tsx
@@ -4,7 +4,6 @@ import { useDispatch, useSelector } from 'react-redux';
 import styled, { css } from 'styled-components/macro';
 import GoToTopButton from '../../../components/GoToTopButton';
 import theme from '../../../theme';
-import { assertNotNull } from '../../../utils';
 import FiltersList from '../../components/popUp/FiltersList';
 import { loadMoreDistanceFromBottom } from '../../constants';
 import { PopupBody } from '../../styles/PopupStyles';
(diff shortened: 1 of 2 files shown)


src/app/content/studyGuides/components/ShowStudyGuides.tsx
```diff
@@ -44,26 +44,30 @@ const ShowStudyGuides = () => {
...
     }
-    return () => refElement ? refElement.removeEventListener('scroll', fetchMoreHighlights) : undefined;
-  }, [fetchMoreHighlights]);
+  };
 
   return (
     <StudyGuidesBody
       ref={ref}
+      onScroll={() => {
+        const refElement = assertNotNull(ref.current, 'Scroll event on a non existing DOM node');
```
janKozie1, line 66: Can this actually happen? From what I've read refs are set when component mounts and set back to null when it unmounts. 

Also, was there a reason why the event handler was added to the node instead of using React's `onScroll`?

src/app/content/studyGuides/components/ShowStudyGuides.tsx
```diff
@@ -44,26 +44,30 @@ const ShowStudyGuides = () => {
...
+
+        updateBackToTop(refElement);
```
PiotrKozlowski, line 68: I think you can just `setShowGoToTop(refElement.scrollTop > 0)`

PiotrKozlowski (approved):
I've left non-blocking comments

src/app/content/studyGuides/components/ShowStudyGuides.tsx
TomWoodward: yea the one line alias doesn't seem necessary to me

PR #754, opened 2020-08-05
Fixing unit test warnings

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1216


src/app/content/components/Page.spec.tsx
```diff
@@ -925,14 +928,21 @@ describe('Page', () => {
...
     // search result is the same. This makes it think that a new highlight was
     // added and will force reselection
 
+    const helpers = {
+      ...createTestServices(),
+      dispatch: store.dispatch,
+      getState: store.getState,
+    };
+    const hook = (require('../highlights/hooks/createHighlight').hookBody)(helpers);
+    await hook(createHighlight({locationStrategies: [{type: 'TextPositionSelector'}]} as any, {} as any));
```
janKozie1, line 943: Nor this nor the previous version of the test check what they were meant to check.

When a new highlight gets added (and it passes [this validation](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/highlightManager.ts#L154) (it didnt in the previous version of the test)), [highlightsManager.update](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/PageComponent.tsx#L96) returns true, which makes [onHighlightSelect](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/PageComponent.tsx#L104) run again, even though no new search result was selected. This makes the search error appear again. 
To prevent it [this id was added](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/PageComponent.tsx#L114) and since it's generated to be the same unless selected result, query or page changed, it would prevent the error from being shown again.

So ideally, if the id was replaced with a random string each time `onHighlightSelect` runs, the test would fail (but it doesn't since no new highlights are actually added to the store)


In retrospec similar condition to [this one](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/searchHighlightManager.ts#L73) could be wrappinig the `onSelect` callback and it should solve the same problem that the id generation is solving.

src/app/content/components/Page.spec.tsx
```diff
@@ -941,7 +951,9 @@ describe('Page', () => {
     // after images are loaded
     await Promise.resolve();
 
+    root.querySelector('[data-testid=banner-body]');
```
janKozie1, line 954: Unused

janKozie1 (changes requested):

TomWoodward:
@janKozie1 i'm trying to wrap my head around this test, can you explain to me why this doesn't fail? i tried to set it up so that it selects a different search result, in which case the highlight thing shouldn't be necessary, but it doesn't find the error again

now that i'm thinking about it, its not even clear to me what is causing the selection to fail in the first case

https://github.com/openstax/rex-web/commit/ccec1379a92fd875269c9b04c498fb549f10925c

janKozie1:
I think that the last warning is coming from [this call ](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page/highlightManager.ts#L177)
It happens between [these two](https://github.com/openstax/rex-web/blob/e96ea33a943ba0280e7073612fcb353a12980238/src/app/content/components/Page.spec.tsx#L940) `await`s. 

I don't have any ideas how to fix it properly yet

src/app/index.spec.tsx
```diff
@@ -45,6 +45,32 @@ describe('create app', () => {
...
+    const windowBackup = window;
+    jest.unmock('../helpers/Sentry');
+
+    beforeEach(() => {
+      (global as any).window = windowBackup;
+    });
+
+    afterEach(() => {
+      delete (global as any).window;
+    });
```
janKozie1, line 58: What does this accomplish? The window is there in the first place, right? Why does it need to be cleaned after each test?

src/app/content/components/Page.spec.tsx
```diff
@@ -34,6 +34,7 @@ import PageNotFound from './Page/PageNotFound';
 import allImagesLoaded from './utils/allImagesLoaded';
 
 jest.mock('./utils/allImagesLoaded', () => jest.fn());
+jest.mock('./utils/attachHighlight', () => jest.fn());
```
janKozie1, line 37: Is this still needed? I thought I removed that part that caused it to be called

src/app/index.spec.tsx
TomWoodward: i think this was copied from below where it does the reverse, deleting the window for each test and then restoring it. i don't think its necessary here

src/app/content/components/Page.spec.tsx
KarinaMendez2: idk? I just merged master in and didn't change this file in my last commit.

TomWoodward: yea but this was added before when we were messing around karina, it might need to be removed now

src/app/index.spec.tsx
```diff
@@ -45,6 +45,24 @@ describe('create app', () => {
     expect(app.history.location.state).toEqual('asdf');
   });
 
+  describe('inside the browser', () => {
+    jest.unmock('../helpers/Sentry');
+    (global as any).window = window;
```
TomWoodward, line 50: this isn't doing anything

src/app/index.spec.tsx
```diff
@@ -45,6 +45,24 @@ describe('create app', () => {
...
+
+    it('adds sentry middleware when enabled', () => {
+      jest.mock('../config', () => ({
```
TomWoodward, line 53: it'd be more explicit to use `doMock` here, `mock` has associated magic that you don't want in this case

src/app/index.spec.tsx
```diff
@@ -45,6 +45,24 @@ describe('create app', () => {
     expect(app.history.location.state).toEqual('asdf');
   });
 
+  describe('inside the browser', () => {
+    jest.unmock('../helpers/Sentry');
```
TomWoodward, line 49: instead of restoring this mock it would probably be better to create a new mock with the spy you need so you don't accidentally trigger any of that underlying error capture functionality

TomWoodward (changes requested):

janKozie1 (approved):

src/app/index.spec.tsx
```diff
@@ -45,6 +45,19 @@ describe('create app', () => {
     expect(app.history.location.state).toEqual('asdf');
   });
 
+  it('adds sentry middleware when enabled', () => {
+    const actualSentry = jest.requireActual('../helpers/Sentry').default;
+    const mockedSentry = require('../helpers/Sentry').default;
+    mockedSentry.shouldCollectErrors = true;
+
+    const initializeWithMiddleware = jest.spyOn(mockedSentry, 'initializeWithMiddleware');
+    initializeWithMiddleware.mockReturnValue(actualSentry.initializeWithMiddleware());
```
TomWoodward, line 54: you should use a mocked middleware here instead of the real one. in this case you may be able to get away with just setting it to `jest.fn()`

TomWoodward (changes requested):

src/app/index.spec.tsx
KarinaMendez2: I tried a bunch of stuff, and just `jest.fn()` definitely doesn't work

TomWoodward (approved):

PR #755, opened 2020-08-06
Issue 535

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/535


src/app/content/components/BookBanner.spec.tsx
```diff
@@ -88,6 +89,67 @@ describe('BookBanner', () => {
       const component = renderer.create(<BarWrapper colorSchema='blue' up={true} />);
       expect(component.toJSON()).toMatchSnapshot();
     });
+
+    it('defaults tab indexes on banner links', () => {
```
PiotrKozlowski, line 93: `default` - without `s`?

src/app/content/components/BookBanner.tsx
```diff
@@ -248,6 +255,7 @@ export class BookBanner extends Component<PropTypes, {scrollTransition: boolean}
           onClick={(e: React.MouseEvent<HTMLAnchorElement>) => {
             this.handleLinkClick(e, bookUrl);
           }}
+          tabIndex={this.state.tabbableBanner !== 'big' ? -1 : undefined}
```
PiotrKozlowski, line 258: I think it would be easier to read if this would be `===` check:
`this.state.tabbableBanner === 'mini' ? -1 : undefined`

src/app/content/components/BookBanner.tsx
```diff
@@ -270,6 +278,7 @@ export class BookBanner extends Component<PropTypes, {scrollTransition: boolean}
           onClick={(e: React.MouseEvent<HTMLAnchorElement>) => {
             this.handleLinkClick(e, bookUrl);
           }}
+          tabIndex={this.state.tabbableBanner !== 'mini' ? -1 : undefined}
```
PiotrKozlowski, line 281: I think it would be easier to read if this would be `===` check:
`this.state.tabbableBanner === 'big' ? -1 : undefined`

src/app/content/components/BookBanner.tsx
```diff
@@ -57,6 +57,11 @@ export interface PropTypes {
   hasUnsavedHighlight?: boolean;
 }
 
+interface BookBannerState {
```
PiotrKozlowski, line 60: This interface could be placed directly above BookBanner class

src/app/content/components/BookBanner.spec.tsx
```diff
@@ -88,6 +89,67 @@ describe('BookBanner', () => {
...
+
+      const createNodeMock = (element: any) => {
+        const analyticsRegion = element.props['data-analytics-region'];
+
+        if (analyticsRegion === 'book-banner-expanded') {
+          return expandedBannerNode;
+        }
+        if (analyticsRegion === 'book-banner-collapsed') {
+          return collapsedBannerNode;
+        }
```
PiotrKozlowski, line 115: I think you could add test ids to these elements just so when we update the analytics region these tests will not fail

src/app/content/components/Toolbar/index.tsx
```diff
@@ -86,9 +88,12 @@ class Toolbar extends React.Component<Props, State> {
 
     const showBackToSearchResults = !this.props.searchSidebarOpen && this.props.hasSearchResults;
 
+    const hideFromFocus = this.props.tocOpen === true
+      || (this.props.tocOpen === null && !this.props.searchSidebarOpen);
```
PiotrKozlowski, line 92: I think that's weird that behavior differs depends on toc state because when toc is open user has to tab through all buttons to get to the Toc and when it is closed it is in another tab order - before search input.
I think we can ask if that should be fixed in this pr or in a separate one

PiotrKozlowski:
Most of my comments are purely stylistic and non-blocking, but I think you should ask someone about https://github.com/openstax/rex-web/pull/755#discussion_r467051932

src/app/content/components/BookBanner.tsx
janKozie1: I moved the props too since they are far away from it as well

src/app/content/components/BookBanner.spec.tsx
janKozie1: It's used as a verb here

src/app/content/components/Toolbar/index.tsx
janKozie1: https://openstax.slack.com/archives/C8V1XET1R/p1596809983200700

PiotrKozlowski (approved):

src/app/content/components/BookBanner.tsx
```diff
@@ -191,6 +197,7 @@ export class BookBanner extends Component<PropTypes, {scrollTransition: boolean}
       this.setState({
         scrollTransition: miniRect.top === 0 &&
           this.bigBanner.current.offsetTop + this.bigBanner.current.clientHeight > window.scrollY,
+        tabbableBanner: miniRect.top === 0 ? 'mini' : 'big',
```
TomWoodward, line 200: did you have to add a new property here? can't you use the existing `scrollTransition` to switch the tabbing around?

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
janKozie1: `scrollTransition` seems to be true only when the banners are transitioning, i think? It's set back to false after scrolling down further

TomWoodward (approved):

PR #756, opened 2020-08-06
use approved-books.json

TomWoodward:
for: openstax/unified#1235


TomWoodward:
as of today CMs are not maintaining the approved-books json, that'll happen eventually, but i don't think there is any issue with putting this in place ahead of time as long as we keep that in mind

philschatz (approved):

concourse/update-content/script.bash
```diff
@@ -43,21 +26,40 @@ for book_id in $book_ids; do
...
 
   book_title=$(node script/entry.js book-info "$book_id" --field title)
-
   curl -s -X POST -H "Authorization: token $GITHUB_ACCESS_TOKEN" "https://api.github.com/repos/openstax/rex-web/pulls" --data-binary @- << JSON
     {
       "title": "$book_title updates",
       "head": "$branch",
-      "base": "master",
-      "draft": true
+      "base": "master"
```
philschatz, line 62: Could/Should this look up the default branch?

concourse/update-content/script.bash
```diff
@@ -13,25 +13,8 @@ set -ex
...
-  Object.keys(newBooks).forEach((key) => {
-    if (newBooks[key].defaultVersion !== oldBooks[key].defaultVersion) {
-      console.log(key);
-    }
-  });
-script
-)")
-
-rm src/config.books.new.js
+approved_books=$(curl -sL https://github.com/openstax/content-manager-approved-books/raw/master/approved-books.json)
```
philschatz, line 16: Could/Should this look up the default branch? `default_branch` in https://api.github.com/repos/openstax/content-manager-approved-books

concourse/update-content/script.bash
philschatz, line 62: Could/Should this look up the default branch? `default_branch` in https://api.github.com/repos/openstax/rex-web

philschatz (commented):
Just noting that now might be a good time to use the default_branch rather than hardcoding `master`

PR #757, opened 2020-08-06
Highlights on smaller resolutions

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/940


src/app/theme.ts
```diff
@@ -88,6 +88,7 @@ const color = {
 
 const mobileBreak = 75;
 const mobileQuery = `(max-width: ${mobileBreak}em)`;
+const touchDeviceQuery = `not all and (pointer: fine), not all and (hover: hover)`;
```
PiotrKozlowski, line 91: Could this be `(hover: none) and (pointer: coarse)`? I didn't test, just looked on it here: https://medium.com/@ferie/detect-a-touch-device-with-only-css-9f8e30fa1134

janKozie1: I don't think so, since `pointer` can be also `none`.

I assumed that only devices that are grouped with mouse here https://www.w3.org/TR/mediaqueries-4/#mf-interaction can make highlights.

PiotrKozlowski (approved):

PR #762, opened 2020-08-12
Update search error logic

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1246

This will need to change when https://github.com/openstax/unified/issues/1232 and / or https://github.com/openstax/unified/issues/905 are implemented


src/app/content/components/Page.spec.tsx
```diff
@@ -925,14 +977,7 @@ describe('Page', () => {
...
     expect(root.querySelector('[data-testid=banner-body]')).toBeFalsy();
 
-    const highlightData = jest.spyOn(highlightUtils, 'highlightData').mockReturnValueOnce(() => undefined);
-
-    // normally, search result selection handler would noop if the
-    // search result is the same. This makes it think that a new highlight was
-    // added and will force reselection
-
     renderer.act(() => {
-      store.dispatch(createHighlight({} as any, {} as any));
```
janKozie1, line 935: I've decided to make a new test for this in `./searchHighlightManager.spec.ts`. 
Since this and the approach with `receiveHighlights([])` would boil down to the same warning described here https://github.com/openstax/rex-web/pull/754#issuecomment-670648157 

I think that's the best choice out of mocking things or manually calling `searchManager.update` from here

philschatz (approved):

philschatz (approved):
It looks good but there is still a failing CI test

PR #764, opened 2020-08-12
FE selenium test: No results message

Malar-Natarajan:
1. Updated few page objects in MH page to differentiate between dropdown filter & filter tags
2. C593151: Select None from both filters in highlights summary page
3. C593153: remove all filter names using the x in highlights summary page

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1249


pytest-selenium/regions/my_highlights.py
```diff
@@ -397,11 +400,11 @@ def color_filters(self) -> MyHighlights.FilterBar.ColorFilters:
 
             """
             filter_menu_root = self.find_element(
-                *self._color_filters_locator)
+                *self._color_dropdown_locator)
             return self.ChapterFilters(self, filter_menu_root)
```
gregfitch, line 404: the property return type should be `ChapterFilters` instead of `bool` and you may want to rename the function as it is no longer returning a yes/no value

pytest-selenium/regions/my_highlights.py
```diff
@@ -418,14 +421,14 @@ def print(self):
             button = self.find_element(*self._print_locator)
             Utilities.click_option(self.driver, element=button)
```
gregfitch, line 422: similar issue, the function is no longer returning a value so the return and doctext should be changed to show that

pytest-selenium/regions/my_highlights.py
```diff
@@ -434,7 +437,7 @@ class ChapterFilters(FilterSelection):
             @property
             def chapters(self) \
                     -> List[MyHighlights.FilterBar.ChapterFilters.Chapter]:
-                r"""Access the book chapter filters.
+                """Access the book chapter filters.
```
gregfitch, line 440: Is Black complaining about the raw string? The doctext should be raw because it has a line continuation in the `rtype`.

Malar-Natarajan: No. I just did not knew what raw string was

pytest-selenium/regions/my_highlights.py
Malar-Natarajan: @gregfitch can we have a call to go over this and the below comment. I dont see any need to change the return type

gregfitch (approved):

PR #765, opened 2020-08-12
Initial styles for using this guide banner

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1239


src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -60,16 +62,27 @@ const ConnectedPrintButton = connect(
...
+      <FilterDropdown
+        label='i18n:highlighting:filters:chapters'
+        ariaLabelId='i18n:studyguides:popup:filters:filter-by:aria-label'
+      >
+        <ConnectedChapterFilter disabled={userLoggedOut}/>
+      </FilterDropdown>
+      <ColorKey />
+      <UsingThisGuideButton onClick={toggleUsingThisGuide} open={UTGopen}/>
+      <ConnectedPrintButton studyGuidesButton />
+      {!userLoggedOut && <ConnectedFilterList />}
```
PiotrKozlowski, line 84: FiltersList should be pushed down when `<UsingThisGuideBanner/>` is open

PiotrKozlowski (commented):
https://github.com/openstax/rex-web/commit/0978bf1ebbecb8359d1ac61fc5bbf353ddf8388b is merged so you can use a structure for Filters.tsx from there

This is a draft pr - so I assume I should wait with a full review when it is marked as ready

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+// tslint:disable-next-line:variable-name
+const UsingThisGuideText = styled.span`
+  ${toolbarDefaultText}; 
+`;
+
+// tslint:disable-next-line:variable-name
+const UsingThisGuideWrapper = styled.div`
+  margin-right: 4.8rem;
+  margin-bottom: ${(props) => props.isOpen ? -(UTGButton.marginBottom + UTGButton.negativeMarginBottom) : '0'}rem;
```
janKozie1, line 44: The button "jumps" a little because of the changing margin. I think you should rather aim for making the button longer vertically while keeping the contents in the same place rather than pushing it down.

KarinaMendez2: I put the negative margin because even putting the button longer, it does not cover the `border-bottom` from the filters bar, so the black background on the button never reaches the banner, can you take a look into that?

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+export const UsingThisGuideButtonWrapper = styled(PlainButton)`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-left: auto;
+  position: relative;
+  overflow: visible;
+  color: ${(props) => props.isOpen ? theme.color.white : theme.color.primary.gray.base};
+  background: ${(props) => props.isOpen ? 'black' : 'white'};
+  padding:  ${(props) => props.isOpen ? '0 1.6rem' : '0'};
```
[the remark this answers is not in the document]
KarinaMendez2: I think since it already has `margin-right: 4.8rem` the extra padding takes it even further away from the print button

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -80,6 +91,13 @@ const ConnectedPrintButton = connect(
 
 export default () => {
   const userLoggedOut = useSelector(loggedOut);
+  const [UTGopen, setUTGopen] = React.useState(false);
+
+  const toggleUsingThisGuide = () => {
+    setUTGopen((state) => !state);
+  };
+
+  const closeUsingThisGuide = React.useCallback(() => { setUTGopen(false); }, []);
```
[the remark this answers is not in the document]
KarinaMendez2: I don't understand this one :thinking_face: In color key I used it for making GA events easier to handle, so it will be the same case here

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideBanner.tsx
```diff
@@ -0,0 +1,100 @@
...
+  ${bannerStyles}
+  padding: 0 4.2rem 4.9rem;
+  ${theme.breakpoints.mobile(css`
+    display: none;
+  `)}
+`;
+
+// tslint:disable-next-line: variable-name
+const MobileBanner = styled.img`
+  ${bannerStyles}
```
PiotrKozlowski, line 37: I think mobile banner should have max-width set to the image original width

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideButton.tsx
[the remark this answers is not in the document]
PiotrKozlowski: Currently, when you open the banner, the button will change size which makes text "jump". I think in this case you should set padding to be always the same and adjust margin-right

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideButton.tsx
PiotrKozlowski: Maybe you could try with adding a missing background with `:after` selector

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -31,6 +34,14 @@ const StyledColorFilter = styled(ColorFilter)`
   min-width: 29rem;
 `;
 
+// tslint:disable-next-line: variable-name
+const RightButtonsWrapper = styled.div`
+  margin-left: auto;
+  display: flex;
+  align-items: center;
+  height: ${barHeight}rem;
```
PiotrKozlowski, line 42: This should be different for the mobile version, however I think it is not necessary to use static height

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuide.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+import createTestStore from '../../../../../test/createTestStore';
+import createTestServices from '../../../../../test/createTestServices';
+import MessageProvider from '../../../../MessageProvider';
+import * as Services from '../../../../context/Services';
+import UsingThisGuideButton from './UsingThisGuideButton';
+import UsingThisGuideBanner from './UsingThisGuideBanner';
+
+describe('Using this guide', () => {
+  let store: Store;
+  let services: ReturnType<typeof createTestServices>;
```
PiotrKozlowski, line 15: `<UsingThisGuideButton/>` and `<UsingThisGuideBanner/>` are not using `store` / `services`
It can be added later when we'll add analytics event

src/app/content/studyGuides/components/Filters.tsx
PiotrKozlowski: This is causing chapter / color filter to not be aligned to the top on mobile resolution:
image

src/app/theme.ts
```diff
@@ -97,8 +97,10 @@ const color = {
 };
 
 const mobileSmallBreak = 30; // 480px
+const mobileLargerBreak = 50; // 800px
```
PiotrKozlowski, line 100: maybe name it `mobileMadiumBreak` so we will have `small`, `medium` and "normal"

PiotrKozlowski (dismissed):
I think it's almost ready, I've left some minor comments

Also, this is how it looks on 1190px screen:
image

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideBanner.tsx
```diff
@@ -0,0 +1,102 @@
...
+import mobileBanner from './assets/banner_mobile.png';
+import { filters } from '../../../styles/PopupConstants';
+import { FormattedMessage } from 'react-intl';
+import { H2, h4MobileStyle } from '../../../../components/Typography/headings';
+import { Times } from 'styled-icons/fa-solid/Times';
+
+const bannerStyles = css`
+  width: 100%;
+  margin: auto;
+`;
```
janKozie1, line 14: It's only used it one place now so you can probably move it there.
Also i don't think `margin:auto` changes anything

janKozie1 (changes requested):

PiotrKozlowski:
No more comments from me, waiting for Jan

PiotrKozlowski (commented):
No more comments from me, waiting for Jan

janKozie1 (changes requested):

janKozie1 (approved):

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideBanner.tsx
```diff
@@ -0,0 +1,113 @@
...
+    </HeaderWrapper>
+    <BodyWrapper>
+      <FormattedMessage id='i18n:studyguides:popup:using-this-guide:alt'>
+        {(msg: Element | string) => <picture>
+          <source media={theme.breakpoints.mobileMediumQuery} srcSet={mobileBanner} />
+          <DesktopBanner src={desktopBanner} alt={msg} tabIndex={0}/>
+        </picture>}
+      </FormattedMessage>
+    </BodyWrapper>
+    <CloseIconButton onClick={props.onClick}>
```
philschatz, line 107: Either this needs an `aria-label="Close the How-to-use-this-guide description"` so the user knows what the button does or we could just add `aria-hidden="true"` to hide the button altogether from the screenreader. That would also require us to make the button not tabbable (clickable is ok).

The idea being that since this is not a real dialog, a non-sighted user may want to keep the description available to refer to.

The aria-label in the next file is related to this comment

src/app/content/studyGuides/components/UsingThisGuide/UsingThisGuideButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+interface Props {
+  onClick: () => void;
+  open: boolean;
+}
+
+// tslint:disable-next-line:variable-name
+const UsingThisGuideButton = (props: Props) => {
+  return <FormattedMessage id='i18n:studyguides:popup:using-this-guide'>
+      {(msg: Element | string) =>
+        <UsingThisGuideButtonWrapper aria-label={msg} onClick={props.onClick} isOpen={props.open}>
```
philschatz, line 68: Why is an aria-label added here? It seems to duplicate what is inside the UsingThisGuide banner.

PR #768, opened 2020-08-14
College Physics for AP® Courses updates

staxbot:


TomWoodward:
the approved-books.json has not caught up with reality yet, ignore this

PR #769, opened 2020-08-17
Add target="_blank" to links in MH and SG

PiotrKozlowski:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1229


src/app/content/studyGuides/components/__snapshots__/StudyGuides.spec.tsx.snap
```diff
@@ -71,7 +71,7 @@ exports[`StudyGuides properly display summary highlights 1`] = `
             className="summary-highlight-content StudyGuidesListElement__HighlightContent-sc-5qgyda-2 jwHLiF"
             dangerouslySetInnerHTML={
               Object {
-                "__html": undefined,
+                "__html": "undefined",
```
janKozie1, line 74: Should this be changed? It renders a `"undefined"` string but normally the `highlightedContent` can't be `undefined`

PiotrKozlowski: I saw that and I think it doesn't really matter? I could change mock to contain something which will render this mocked value

janKozie1: I don't think it matters either, just looks bit strange. There is probably nothing wrong with adding data to mocks that should be there

janKozie1 (approved):

philschatz (approved):
Nice use of memoizing. I am not familiar with `React.useMemo` . Does it drop least-recently-used items?

PiotrKozlowski:
Do you mean if the previous value is removed from the memory when deps changes?
I think yes, since they are stored in the array and overwritten on deps change https://github.com/facebook/react/blob/df14b5bcc163516fc0f1ad35e9b93732c66c1085/packages/react-dom/src/server/ReactPartialRendererHooks.js#L352

PR #773, opened 2020-08-18
Issue 1255

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1255
TODO:
- [x] Redirect also notFound.paths to openstax.org/error/404


src/app/utils.ts
```diff
@@ -45,6 +45,10 @@ const makeCatchError = (dispatch: Dispatch) => (e: Error) => {
   if (e instanceof UnauthenticatedError) {
     dispatch(receiveLoggedOut());
     return;
+  } else if (e instanceof BookNotFoundError) {
+    Sentry.captureException(e);
+    assertWindow().location.replace('https://openstax.org/error/404');
```
janKozie1, line 50: This probably should be a constant somewhere, maybe `errors/routes` ?

src/app/content/components/Page/PageNotFound.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+import PageNotFound, { ToCButton } from './PageNotFound';
+
+describe('PageNotFound', () => {
+  let store: Store;
+  let services: ReturnType<typeof createTestServices>;
+
+  beforeEach(() => {
+    store = createTestStore();
+    services = createTestServices();
+  });
```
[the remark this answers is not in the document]
PiotrKozlowski: it needs only `store` for `SidebarControl` button

src/app/content/components/Page/PageNotFound.tsx
```diff
@@ -0,0 +1,60 @@
...
+const PageNotFoundTitle = styled.h1`
+  font-weight: bold;
+  padding: 1rem 0;
+`;
+
+// tslint:disable-next-line: variable-name
+const PageNotFoundText = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
```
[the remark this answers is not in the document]
PiotrKozlowski: draft pr

src/app/content/components/Page/PageNotFound.tsx
```diff
@@ -0,0 +1,60 @@
...
+  text-align: center;
+  color: ${theme.color.primary.gray.base};
+  ${theme.breakpoints.mobile(css`
+    margin: 4rem 0 0 0;
+  `)}
+`;
+
+// tslint:disable-next-line: variable-name
+const PageNotFoundTitle = styled.h1`
+  font-weight: bold;
```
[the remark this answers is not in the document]
PiotrKozlowski: I think it's better to have these styles here anyway because we should not assume that styles will be inherited by other components

src/app/content/reducer.ts
```diff
@@ -74,6 +75,10 @@ function reduceContent(state: State, action: AnyAction) {
     case getType(actions.receivePage): {
       return reduceReceivePage(state, action);
     }
+    case getType(actions.receivePageNotFound): {
+      const loading = omit('page', state.loading);
+      return {...omit(['page', 'references'], state), loading, pageNotFound: action.payload};
```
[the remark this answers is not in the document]
PiotrKozlowski: I've changed it to `pageNotFoundId` and left `pageNotFound` selector
I know that in this app it's pretty common patter to call additional actions on `locationChange` but I find it very hard to maintain when we have so many side effects

src/app/errors/components/PageNotFound.spec.tsx
```diff
@@ -1,13 +1,68 @@
...
+    });
+
+    expect(dispatch).toHaveBeenCalledWith(openToc());
+
+    renderer.act(() => {
+      component.root.findByProps({ 'data-testid': 'toc-button' }).props.onClick();
+    });
+
+    expect(dispatch).toHaveBeenCalledTimes(2);
+    expect(dispatch).toHaveBeenCalledWith(openToc());
```
janKozie1, line 66: This wouldn't fail if the second click caused a dispatch of `closeToc`. Maybe it's okay to assert on the state, rather than what was called?

src/app/errors/components/PageNotFound.tsx
```diff
@@ -1,10 +1,52 @@
...
+import { SidebarControl } from '../../content/components/Toolbar/styled';
+import theme from '../../theme';
 
-export default () => {
-  return <FormattedMessage id='i18n:error:404'>
-    {(txt) => (
-        <h1>{txt}</h1>
-    )}
-  </FormattedMessage>;
-};
```
janKozie1, line 10: I'm not really sure if this should be replaced. The `/errors/404` used to render an empty page with just the heading and now it renders the message with toc button even thought the toc doesn't exist there.

src/app/errors/routes.ts
```diff
@@ -3,13 +3,14 @@ import { Route } from '../navigation/types';
 
 const CATCH_ALL = '/(.*)';
 
-export const notFound: Route<undefined> = {
+export const notFound: Route<undefined> & { getFullUrl: () => string } = {
```
janKozie1, line 6: I'm not sure if there is a benefit of altering this, can't the link be outside of it?

janKozie1 (changes requested):

src/app/content/reducer.ts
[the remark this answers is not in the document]
janKozie1: Well it wouldn't need dispatch additional actions, i imagine it would only require you to modify [this line](https://github.com/openstax/rex-web/blob/04108ef9bedbb19a22030ebc676009697e5fe2bb/src/app/content/reducer.ts#L94), unless I'm missing something.

I agree about the side effects making it harder to maintain although am not really sure if this counts as it.

src/app/errors/routes.ts
PiotrKozlowski: Maybe we could just create left only `getUrl` and modify it to contain full url?

src/app/errors/components/PageNotFound.tsx
PiotrKozlowski: I've moved this component back to `componenets/Page/` and deleted this one since now we are redirecting everything to `/error/404`

src/app/errors/routes.ts
```diff
@@ -1,16 +1,19 @@
...
-    webpack: /* istanbul ignore next */ () => [(require as any).resolveWeak('./components/PageNotFound')],
+    modules: ['LoaderCentered'],
+    webpack: /* istanbul ignore next */ () => [(require as any).resolveWeak('./components/LoaderCentered')],
   }),
-  getUrl: () => '/errors/404',
+  getFullUrl: () => 'https://openstax.org' + notFound.getUrl(),
+  getUrl: () => '/error/404',
   name: 'NotFound',
   paths: [CATCH_ALL],
+  redirect: () => assertWindow().location.replace(notFound.getFullUrl()),
```
PiotrKozlowski, line 18: It has to redirect to the full url, because otherwise, puppeteer from `src/index.browserspec.tsx` will fail

src/app/content/reducer.ts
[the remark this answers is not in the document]
PiotrKozlowski: I was thinking about dispatching clearPageNotFound on locationChange.
Using existing `locationChange` in the reducer to reset it seems better, however, it would still require us to maintain two places - or actually 5 places (there are 4 `return` statements in `locationChange` reducer) - instead of one place
I think moving this logic to `pageNotFound` selector makes it simpler

janKozie1 (commented):
I think it looks good but I'm not really sure if the routing should be modified. I thought about just redirecting if `BookNotFoundError`, displaying the 404 message inside content if no page, leave other 404 handling as it is.

Also, the acceptance criteria mention 
>If a user requests a URL associated w/a REX book, and they do not have a service worker running, redirect them to https://openstax.org/error/404.

Does it mean I should be getting redirected always while on localhost since sw is not enabled?

janKozie1 (approved):
> I think we don't want this page to be showing in any circumstances

That's true, I'm just worried why it was left unchanged for almost a year. 

> If a user requests a URL that is not associated w/a REX book, redirect them to https://openstax.org/error/404.

Sorry, I thought it means a `books/` route but without a correct book, not any route.

philschatz (approved):

PR #763, opened 2020-08-12
add .json to the archive urls

TomWoodward:
pretty sure this is the only change required to get some content from RAP loaded in REX

it should be possible now to use a url such as:
```
https://rex-web-json-dot-json-zdhcztu1.herokuapp.com/books/calculus-volume-1/pages/1-introduction?archive=https://openstax.org/apps/archive/20200101.123456
```

to load RAP content into rex

for: openstax/unified#1254

PR #777, opened 2020-08-27
temporarily disable links that go to another book and are not in /extras

philschatz:
As Content Managers publish using batch (instead of legacy) the `/extras` endpoint only shows one book instead of all the related books.

Also, the `/extras/{uuid}.json` is incorrect... the file extension was mistakenly added in #763 

for: openstax/unified#1271


src/app/content/components/Page/validateDOMContent.ts
```diff
@@ -25,6 +26,11 @@ const validateLinks = (rootEl: HTMLElement) => {
   if (config.UNLIMITED_CONTENT) {
     // tslint:disable:no-console
     console.warn(message);
+    Array.from(rootEl.querySelectorAll('a[href^="/"]')).forEach((element) => {
+      const href = assertNotNull(element.getAttribute('href'), 'href was null');
+      element.setAttribute('data-broken-href', href);
```
janKozie1, line 31: What's the purpose of this attribute?

janKozie1 (commented):
Does this fix the prerendering issue? When i tried prerendering locally, i get an error with 

` Error response from archive "http://localhost:8000/extras/549f2627-d46b-4bf4-96b0-f54d18a285ee.json" 404`

I might be setting it up wrong though.

src/app/content/components/Page/validateDOMContent.ts
philschatz: I wanted to leave a note for a future developer to know that the attribute was removed intentionally rather than being missing in the content. Since this PR is temporary it is unlikely that we'll forget why there are a few links that are missing an href attribute; should it be removed?

philschatz:
It does fix the prerendering issue; I forgot to commit an update to the `config.production.js` file which may have caused some confusion.

If a book contains a link to a page outside of the book and `/extras` does not include one of REX' books 2 things would happen:

- prerendering would fail
- users that navigate to the page in production would see an "Oops" dialog (because /extras is fetched & did not include a book in REX)

This PR addresses both problems by setting `UNLIMITED_CONTENT: true` in production. It also prevents users from landing on a 404 page by removing the `href` on the broken link.

src/app/content/components/Page/validateDOMContent.ts
janKozie1: Maybe an open PR that reverses those changes, and explains why they were made, would be more noticeable?
I'm not really sure which is better so it's probably fine to keep it this way.

janKozie1:
Going to https://rex-web-quiet-broken-li-wt0pzw.herokuapp.com/books/university-physics-volume-2/pages/1-3-thermal-expansion still shows the uh-oh because of 404 fetch. Is that expected? 

I can see a similar error when prerendering locally although im not sure if any of the required env variables are not causing it.
I believe it's caused by the request that [starts here](https://github.com/openstax/rex-web/blob/6f89f10f2db21758d38d4b225bcbdf30d610f075/src/app/content/hooks/locationChange/resolveContent.ts#L158).

Let me know if I'm missing something, thanks!

philschatz:
Aha! The problem is that the /extras URL now ends with `.json` but it should not because that results in 404 (see #763)

I will fix

janKozie1 (approved):
Looks good, not sure why the build tests are failing though.

PR #822, opened 2020-09-08
Handle highlight network failures

janKozie1:
for: https://github.com/openstax/unified/issues/905


janKozie1:
It needs to wait for some changes from issue 1246 (tests in page.spec.tsx), but besides that, it should be ready for a review

src/app/notifications/reducer.spec.ts
```diff
@@ -70,11 +72,61 @@ describe('notifications reducer', () => {
...
+    });
+
+    it('refreshes the timestamp if a toast with the same message appears', async() => {
+      const newState = reducer(initialState, actions.addToast('mytoast'));
+      const toast = newState.toastNotifications.find((notification) => notification.message === 'mytoast');
+
+      if (!toast) {
+        return expect(toast).toBeTruthy();
+      }
+      await new Promise((res) => setTimeout(res, 10));
```
PiotrKozlowski, line 104: Maybe you could try to mock the return value for `action.addToast(...)` to return another timestamp, this way we will avoid blocking a thread

src/app/notifications/types.ts
```diff
@@ -17,7 +17,17 @@ export interface AppMessageNotification {
 
 export type Messages = Message[];
 
-export type AnyNotification = ActionType<Pick<typeof actions, 'updateAvailable' | 'acceptCookies' >>
+export type ModalNotification = ActionType<Pick<typeof actions, 'updateAvailable' | 'acceptCookies' >>
   | AppMessageNotification;
 
-export type State = AnyNotification[];
+export interface ToastNotification {
+  message: string;
```
PiotrKozlowski, line 24: maybe `messageKey` to indicate more clearly that this values will be used for localization?

src/app/utils.ts
```diff
@@ -37,7 +35,8 @@ export const actionHook = <C extends AnyActionCreator>(actionCreator: C, body: A
           catchError(e);
         }
       }
-      return result;
+
+      return next(action);
```
PiotrKozlowski, line 39: just curious - is this change changing any behavior - it shouldn't right?
However, maybe it was written this way to make sure that `matches(action)` or `boundHook(action)` does not cause any side effects?

PiotrKozlowski (changes requested):
Looks good, I've left some minor comments

src/app/content/highlights/hooks/updateHighlight.ts
```diff
@@ -1,9 +1,35 @@
...
+    if (meta.failedToSave) { return; }
+
+    const oldHighlight = assertDefined(
+      findHighlight(highlightsLocalState(getState()), payload.id), 'Can\'t update a highlight that doesn\'t exist'
+    );
+
+    highlightClient.updateHighlight(payload).catch((error) => {
+      Sentry.captureException(error);
+
+      dispatch(addToast('i18n:notification:toast:highlights:update-failure'));
```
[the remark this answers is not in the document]
janKozie1: I'm going to wait until the final decision, since they are considering switching to one note for all errors.

src/app/utils.ts
janKozie1: The reason why I moved it is so that the action hooks get the access to store that was not yet updated (as long as they are sync which is why i removed async from `updateHighlight` for example).

from redux docs: `The last middleware in the chain will receive the real store's dispatch method as the next parameter`
`The store's reducing function will be called ... synchronously. Its return value will be considered the next state. It will be returned from getState() from now on`

I will double check if any of the existing hooks don't depend on the opposite thing happening

src/app/notifications/reducer.spec.ts
janKozie1: I'm just gonna reduce the timeout, the less mocks the better.

janKozie1:
@PiotrKozlowski 
I'm still working on animating the toasts so anything related to that is not ready. If you want to wait with your review until I finish, I'm okay with that.

src/app/notifications/reducer.spec.ts
PiotrKozlowski: Maybe this can be used by using `jest.useFakeTimers();`? 
The thing is that we should try to make our unit tests as sync as possible, and `setTimeout` set even to `5`ms will block a thread for **at least** 5 ms. That sounds like a small amount of time but if we keep doing so we can end up a bigger and bigger amount

PiotrKozlowski:
@janKozie1 ok, I'll wait

src/app/notifications/components/ToastNotifications/Toast.spec.tsx
```diff
@@ -148,18 +142,21 @@ describe('SearchFailure', () => {
...
     expect(bannerBody.props.isFadingOut).toBe(false);
     component.unmount();
   });
 
-  describe('syncState', () => {
-    it('doesn\'t allow to fade out if error is not ready to be dimissed', () => {
-      const state = { isFadingOut: false, shouldAutoDismiss: false};
-      expect(syncState(state)).toBe(state);
+  describe('syncAnimationState', () => {
+    it('doesn\'t allow to fade out if toast is not ready to be dimissed', () => {
```
PiotrKozlowski, line 154: I think it would be okay to add to this test also the test case when `prevState.shouldAutoDismiss` is set to `true`

src/app/notifications/components/ToastNotifications/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+interface Props {
+  dismiss: (toast: ToastNotification) => void;
+  toasts: ToastNotification[];
+  mobileToolbarOpen: boolean;
+}
+
+// tslint:disable-next-line:variable-name
+const ToastNotifications = (props: Props) => {
+  const realIndexes = new Map(orderBy(props.toasts, ['timestamp'], ['desc']).map((toast, index) => [toast, index]));
```
PiotrKozlowski, line 22: Should this be wrapped in `useMemo` so it's not recalculated on `mobileToolbaOpen` state change?

src/app/notifications/components/ToastNotifications/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export default connect(
+  (state: AppState) => ({
+    mobileToolbarOpen: mobileToolbarOpen(state),
+    toasts: select.toastNotifications(state),
+  }),
+  (dispatch) => ({
+    dismiss: flow(dismissNotification, dispatch),
+  })
+)(ToastNotifications);
```
PiotrKozlowski, line 47: I feel like we should try to switch to the "modern" way when we can and use hooks for this

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -4,7 +4,11 @@ import WeakMap from 'weak-map';
 import { APP_ENV } from '../../../../config';
 import { typesetMath } from '../../../../helpers/mathjax';
 import Loader from '../../../components/Loader';
+<<<<<<< HEAD
```
PiotrKozlowski, line 7: woops

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -28,6 +32,10 @@ if (typeof(document) !== 'undefined') {
 
 const parser = new DOMParser();
 
+<<<<<<< HEAD
```
PiotrKozlowski: woops

PiotrKozlowski (changes requested):
at `0:55` on [this video](https://drive.google.com/file/d/1XQ4NmwggXVcOCdeimJLBzOnEngwh3sGr/view?usp=sharing) I've managed to break these toasts

Other comments are pretty minor things

src/app/content/components/Page/PageComponent.tsx
janKozie1: :see_no_evil:

src/app/notifications/components/ToastNotifications/index.tsx
janKozie1: I'm not convinced that it would be faster since you won't get many errors to sort and memo comes with its own cost too

PiotrKozlowski (approved):

src/app/utils.ts
janKozie1: Turns out the pageView GA event depends on it, will need to rewrite the logic a bit.

PiotrKozlowski: Nice catch, please let me know when you finish fixing this so I can review

janKozie1:
@PiotrKozlowski it should be ready for your review, thanks!

src/app/content/highlights/reducer.ts
```diff
@@ -126,6 +128,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
           ...state.currentPage,
           hasUnsavedHighlight,
           highlights: newHighlights,
+          revertableHighlight: oldHighlight,
```
PiotrKozlowski, line 131: is `revertableHighlight` used for something?

PiotrKozlowski (commented):
I've left minor comments, but there is also this one:
I think that maybe we should stop using `hooks` for creating / updating / deleting highlight and instead go with a [redux-thunk](https://github.com/reduxjs/redux-thunk) approach
In this case, you wouldn't have to pass `preUpdateData` and call any revert actions - instead, you would first call the api and then depends on the response, dispatch an action for creating a highlight in case of success or adding a toast in case of failure

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -120,17 +141,16 @@ const EditCard = React.forwardRef<HTMLElement, EditCardProps>((props, ref) => {
...
+    const data  = assertDefined(props.data, 'Can\'t update highlight that doesn\'t exist');
 
-    dispatch(updateHighlight({
-      highlight: {
-        annotation: pendingAnnotation,
-        color: toSave.color as string as HighlightUpdateColorEnum,
-      },
-      id: toSave.id,
-    }, {
+    const addedNote = (data.annotation === undefined) ? true : false;
```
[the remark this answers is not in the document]
janKozie1: I left it since old code was written that way. I think it's a good idea to change it.

janKozie1:
While it would be nice to not need to send data require to revert changes, it would also mean that users need to wait for the request to finish to see the results of their actions so It would make the highlight-related UI at least a bit less responsive. 

It would also add a separate place where actions / network requests are handled which I don't think is a bad thing but would add more to this PR (including the types that would need to account for that change)

It could also prove to be problematic when users try to make highlights while the request is not finished yet (I'm mainly thinking about the discard note since `hasUnsavedChanges` depends on the action hitting the reducer).

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -24,6 +24,30 @@ import Confirmation from './Confirmation';
 import Note from './Note';
 import { isElementForOnClickOutside, useOnClickOutside } from './utils/onClickOutside';
 
+const generateUpdatePayload =
```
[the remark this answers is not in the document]
janKozie1: I've extracted it and wrote tests for it but it wouldn't make the tests more readable in `EditCard.spec.tsx`

PiotrKozlowski (approved):

PiotrKozlowski:
From the other hand, with redux-thunk we could do a function that dispatches create/update/delete action imidietally, then call the api and then dispatch revert action on failure.
Since this function would be called synchronously, we would have an access to the current state so we wouldn't have to pass anything new and we wouldn't have to chceck if this is reverting action.

But I agree that this is probably not the best moment to introduce a new approach

PR #824, opened 2020-09-09
Special licensing and attribution for HS Physics and Statistics HS

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1273


src/app/content/components/Attribution.tsx
```diff
@@ -125,6 +125,10 @@ class Attribution extends Component<Props> {
 
   public render() {
     const {book} = this.props;
+    // different attribution text for Physics and Statistics
+    const specialCases = ['cce64fde-f448-43b8-ae88-27705cceb0da', '394a1101-fd8f-4875-84fa-55f15b06ba66'];
```
PiotrKozlowski, line 129: Maybe name this `bookIdsWithTEAAttributionText` and then you can remove the comment above?

src/app/content/components/Attribution.tsx
```diff
@@ -125,6 +125,10 @@ class Attribution extends Component<Props> {
...
+    const attributionTextId = (book && specialCases.includes(book.id)) ? 'i18n:attribution:tea-text'
+      : 'i18n:attribution:default-text';
```
PiotrKozlowski, line 131: This could be formatted:
```
const attributionTextId = (book && specialCases.includes(book.id))
  ? 'i18n:attribution:tea-text'
   : 'i18n:attribution:default-text';
```

src/app/messages/en/attributionText.ts
```diff
@@ -61,3 +61,71 @@ export default `
...
+        Section URL: <a href="https://openstax.org{currentPath}">https://openstax.org{currentPath}</a>
+      </li>
+    </ul>
+  </li>
+</ul>
+
+<p>
+  © {bookLatestRevision, date, medium} Texas Education Agency (TEA). <strong>The OpenStax name, OpenStax logo,
+  OpenStax book covers, OpenStax CNX name, and OpenStax CNX logo are not subject to the Creative Commons license and may
+  not be reproduced without the prior and express written consent of Rice University.</strong>
```
PiotrKozlowski, line 129: Shouldn't this all be in one line?

src/app/content/components/Attribution.spec.tsx
```diff
@@ -177,5 +177,27 @@ describe('Attribution', () => {
...
+        return expect(document).toBeTruthy();
+      }
+
+      details.setAttribute('open', '');
+
+      store.dispatch(
+        actions.receiveBook({...formatBookData(book, mockCmsBook), id: 'cce64fde-f448-43b8-ae88-27705cceb0da'})
+      );
+
+      expect(details.children[1].innerHTML).toMatch(`you must attribute “Texas Education Agency (TEA)."`);
```
PiotrKozlowski, line 196: Maybe this could check for a match with `teaText` from `attributionText.ts`? I guess there might be some errors with comparing a `string` with `innerHTML` which is already parsed, but maybe this can be solved by using `DOMParser` to parse also `teaText` before comparing them?

src/app/messages/en/attributionText.ts
PiotrKozlowski: I see, most of the text is copied from `defaultText`
Maybe you could create a function like: ?
```
const generateAttributionText = (forTEA = false): string => {
   ...generate text here
}
export const defaultText = generateAttributionText(false);
export const teaText = generateAttributionText(true);
```

KarinaMendez2: ? no, max length situation

src/app/content/components/Attribution.spec.tsx
```diff
@@ -177,5 +177,27 @@ describe('Attribution', () => {
...
     });
+
+    it('renders special licensing and attribution for HS Physics', async() => {
+
+      const { node } = renderToDom(render());
+      const details = node;
+
+      if (!document) {
+        return expect(document).toBeTruthy();
+      }
```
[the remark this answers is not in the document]
KarinaMendez2: probably not, but used in previous tests cases so...

src/app/content/components/Attribution.spec.tsx
```diff
@@ -177,5 +177,27 @@ describe('Attribution', () => {
...
+
+      details.setAttribute('open', '');
+
+      store.dispatch(
+        actions.receiveBook({...formatBookData(book, mockCmsBook), id: 'cce64fde-f448-43b8-ae88-27705cceb0da'})
+      );
```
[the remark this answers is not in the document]
KarinaMendez2: I need a book with either Statistics or Physics ID on it JUST in this test, why would it be at the top?

src/app/messages/en/attributionText.ts
```diff
@@ -61,3 +61,71 @@ export default `
...
+  Changes were made to the original material, including updates to art, structure, and other content updates.
+</p>
+
+<strong>Attribution information</strong>
+<ul>
+  <li>
+    If you are redistributing all or part of this book in a print format,
+    then you must include on every physical page the following attribution:
+    <blockquote>
+      Access for free at https://openstax.org{introPageUrl}
```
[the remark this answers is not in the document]
KarinaMendez2: All text and other indications are on the card

src/app/messages/en/attributionText.ts
KarinaMendez2: I thought about it but it makes more conditions, want to keep it as simple as possible, this is a very special case at the moment, might be more later, we'll adapt on future requirements.

src/app/content/components/Attribution.spec.tsx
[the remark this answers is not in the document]
janKozie1: `top of the test case` as in line 182. Right now the test is written like
- render page with a book
- open attribution
- suddenly the book changes

src/app/content/components/Attribution.tsx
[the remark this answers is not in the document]
KarinaMendez2: it already checks for hasOSWebData(book), you think we should check twice?

src/app/content/components/Attribution.spec.tsx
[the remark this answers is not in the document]
KarinaMendez2: well when I put it at the top, the render grabs the other `mockCmsBook`, and it makes my test fail

src/app/content/components/Attribution.tsx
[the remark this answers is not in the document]
janKozie1: I mean you can move that to be the first thing it does. You don't have to find the correct attribution text template if it's not going to be rendered anyway.

src/app/content/components/Attribution.spec.tsx
KarinaMendez2: Actually it gets complicated because of the variables needed, I would have to pass in every single variable for DomParser to work, that's why it's easier to just check if TEA is included in the text.

src/app/content/components/Attribution.tsx
```diff
@@ -125,12 +129,13 @@ class Attribution extends Component<Props> {
 
   public render() {
     const {book} = this.props;
-    // different attribution text for Physics and Statistics
-    const specialCases = ['cce64fde-f448-43b8-ae88-27705cceb0da', '394a1101-fd8f-4875-84fa-55f15b06ba66'];
-    const attributionTextId = (book && specialCases.includes(book.id)) ? 'i18n:attribution:tea-text'
+    if (!hasOSWebData(book)) { return null; }
+
+    const attributionTextId = (book && this.bookIdsWithTEAAttributionText.includes(book.id))
```
PiotrKozlowski: you don't have to check for `book` anymore

src/app/content/components/Attribution.spec.tsx
```diff
@@ -179,21 +179,15 @@ describe('Attribution', () => {
...
-
-      details.setAttribute('open', '');
-
       store.dispatch(
         actions.receiveBook({...formatBookData(book, mockCmsBook), id: 'cce64fde-f448-43b8-ae88-27705cceb0da'})
       );
+      const { node: details } = renderToDom(render());
+      details.setAttribute('open', '');
+
+      const matchTEAText = 'you must attribute “Texas Education Agency (TEA)."';
```
PiotrKozlowski, line 188: You could also render this with `react-test-renderer` and then check for `<FormattedHTMLMessage>` with id `'i18n:attribution:tea-text'` and specifc values

^^ This is a non-blocking comment

src/app/content/components/Attribution.spec.tsx
```diff
@@ -177,5 +177,21 @@ describe('Attribution', () => {
 
       expect(details.children[1].innerHTML).toMatch(`Authors: Jhon Doe, Jonny Doe`);
     });
+
+    it('renders special licensing and attribution for HS Physics', async() => {
+      store.dispatch(
+        actions.receiveBook({...formatBookData(book, mockCmsBook), id: 'cce64fde-f448-43b8-ae88-27705cceb0da'})
+      );
+      const { node: details } = renderToDom(render());
+      details.setAttribute('open', '');
```
janKozie1, line 186: This doesn't actually make a difference for this test case, but i don't think there is a way to assert on the text being visible (instead of just being in the document)

janKozie1 (approved):
Don't have anything to add besides Piotr's comments

src/app/content/components/Attribution.spec.tsx
KarinaMendez2: yeah I think it being in the document is enough

PiotrKozlowski (approved):

PR #829, opened 2020-09-11
parseTitleNode also for nested sections

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1283

This removes recursively `.os-part-text` class from sections, however, this book update cause also another issue that we discuss over here: https://openstax.slack.com/archives/CH5514N57/p1599832840080900 

I've had to change functions that check for `unit`/`chapter` so it does not depend on the `.os-number` class, because books like American Gov has `.os-number`s inside of Unit titles.

To check if this pr works you can update Amercian Gov. to the `6.5` version in `config.books.json` - id: `9d8df601-4f12-4ac1-8224-b450bf739e5f`


src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -137,10 +137,11 @@ export const archiveTreeSectionIsUnit = (section: LinkedArchiveTreeNode) =>
   isArchiveTree(section)
   && !!section.parent
   && archiveTreeSectionIsBook(section.parent)
-  && getArchiveTreeSectionNumber(section) === null
+  && section.contents.every((node) => (node as LinkedArchiveTree).contents !== undefined)
```
janKozie1, line 140: Do the other types need changes? It feels like if they were correct you would need an another guard here instead of casting the type.

src/app/content/highlights/utils/paginationUtils.spec.ts
```diff
@@ -73,17 +73,14 @@ describe('filterCountsPerSourceByLocationFilter', () => {
   it('filters by pages outside chapter', () => {
     expect(filterCountsPerSourceByLocationFilter(
       new Map([
-        ['chapter1', findArchiveTreeNodeById(treeWithUnits, 'chapter1')!],
-        ['preface', findArchiveTreeNodeById(treeWithUnits, 'preface')!]]
```
janKozie1, line 77: Why was this removed? I'm not sure why does `treeWithUnits` not contain preface.

janKozie1 (commented):
It works well, just left some minor comments

src/app/content/utils/archiveTreeUtils.ts
PiotrKozlowski: `section.contents` may have items of type `ArchiveTree` or `ArchiveTreePage` and since `ArchiveTreeNode` does not have `contentes` then typescript is not allowing us to check `node.contents` even if `node` is an `object`
I was thinking about doing something like `section.contents.every(archiveTreeSectionIsChapter)` but this will check for a lot more of things than simply having `contents` and I don't know if this is worth it

From the other side, I think that maybe it's worth it to reconsider naming of types and interfaces that we use and simplify them somehow to use `book`, `page`, `unit` and `chapter` instead of `LinkedArchiveTree/Section/Node`
Maybe:
`LinkedArchiveBook` for `book`
`LinkedArchivePage` for `page`
`LinkedArchiveUnit` for `unit`
`LinkedArchiveChapter` for `chapter`

However, I didn't want to do that in this pr without a discussion with Tom, so I've decided to go with a cast here

src/app/content/highlights/utils/paginationUtils.spec.ts
PiotrKozlowski: you are right, I should have just move preface outside of a unit - I've pushed a change for this

src/app/content/utils/archiveTreeUtils.ts
janKozie1: But the type `LinkedArchiveTree` has contents typed as always being there. I think combining guards is a safe way of doing it. 
> if this is worth it

i feel like if the checking for what is desired, then guards are the right place where we should trade slightly more things to check for better type safety.

PiotrKozlowski: After some investigation:
`node` from `section.contents.every(node => ...)` in `archiveTreeSectionIsUnit` is of type `ArchiveTree` | `ArchiveTreeNode` so if I would like to use it as `section.contents.every(archiveTreeSectionIsChapter)` I would have to type `archiveTreeSectionIsChapter` to allow `ArchiveTree` and `ArchiveTreeNode` but then `!archiveTreeSectionIsBook(section)` will always return `true` because these types does not have `parent` - they are not `linked`
In other terms - If we would like to introduce any changes to this guards I would suggest that we talk about it, prepare a bigger solution and then talk with @TomWoodward about implementation. I'm not sure if this should be done in this pr

janKozie1 (approved):

src/app/content/utils/archiveTreeUtils.ts
PiotrKozlowski: I've replaced casts with `isArchiveTree` guard that also has cast but at least this is now in one place

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -137,10 +137,11 @@ export const archiveTreeSectionIsUnit = (section: LinkedArchiveTreeNode) =>
   isArchiveTree(section)
   && !!section.parent
```
TomWoodward, line 138: you don't need this check i think, `archiveTreeSectionIsBook` accepts undefined

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -137,10 +137,11 @@ export const archiveTreeSectionIsUnit = (section: LinkedArchiveTreeNode) =>
...
   && archiveTreeSectionIsBook(section.parent)
-  && getArchiveTreeSectionNumber(section) === null
+  && section.contents.every(isArchiveTree)
 ;
 export const archiveTreeSectionIsChapter = (section: LinkedArchiveTreeNode): section is LinkedArchiveTree =>
   isLinkedArchiveTree(section)
   && !archiveTreeSectionIsBook(section)
   && getArchiveTreeSectionNumber(section) !== null
```
TomWoodward, line 145: could we replace this with a check to see if the parent is a unit or a book? see thread https://openstax.slack.com/archives/CH5514N57/p1600276301142100?thread_ts=1599832840.080900&cid=CH5514N57

TomWoodward (changes requested):

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -137,10 +137,11 @@ export const archiveTreeSectionIsUnit = (section: LinkedArchiveTreeNode) =>
...
+  && section.contents.some((node) => !isArchiveTree(node))
```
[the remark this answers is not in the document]
PiotrKozlowski: `isLinkedArchiveTreeSection` check for `section.parent !== undefined` but `section.contents` here is of type `ArchiveTree` | `ArchiveTreeNode` and these does not have `parent` on them

src/app/content/utils/archiveTreeUtils.ts
PiotrKozlowski: In this case "Answer Key" from [Amerivan Gov](https://rex-web.herokuapp.com/books/american-government-2e/pages/chapter-1) will be treated as a chapter and will be displayed in the StudyGuides Chapter Filter

We could do this change but then we should change filters for study guides filters - and probably also for highlights filters
This will also have an impact on `<title>` tags

Should I do this?

TomWoodward: no, we need to make sure answer key doesn't show up as a chapter. in that example i'm not sure how to differentiate it.

lets keep the logic you have now and maybe i'll work with content to put some type flags in the tree so we don't have to gues

TomWoodward (approved):

PR #860, opened 2020-09-18
Issue 1294

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1294


src/app/content/highlights/components/Highlights.spec.tsx
```diff
@@ -399,6 +399,6 @@ describe('Highlights', () => {
       pageId,
     }));
 
-    expect(component.root.findAllByType(HighlightDeleteWrapper).length).toEqual(1);
+    expect(component.root.findAllByType(HighlightDeleteWrapper).length).toEqual(0);
```
janKozie1, line 402: This was wrong, right? One highlight gets deleted, other one cancelled so there shouldn't be any more left.

src/app/content/highlights/reducer.ts
```diff
@@ -136,11 +139,16 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
     case getType(actions.deleteHighlight): {
       const highlightToRemove = findHighlight(state, action.payload);
+      const highlights = state.currentPage.highlights;
 
-      if (!state.currentPage.highlights || !highlightToRemove) {
+      if (!highlightToRemove) {
         return state;
       }
 
+      const newHighlights = highlights
```
PiotrKozlowski, line 148: maybe name it `newCurrentPageHighlights`?

src/app/content/highlights/components/Highlights.spec.tsx
PiotrKozlowski: it seems like yes
since `state` does not have currentPage highlights clicking delete will dispatch an action but will not close the menu - this is there was one still open

PiotrKozlowski (approved):

PR #862, opened 2020-09-18
Adding target blank to special attribution and licensing links

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1295


PiotrKozlowski (changes requested):
nvm...

PiotrKozlowski (approved):

src/app/messages/en/attributionText.ts
```diff
@@ -65,12 +65,12 @@ export const defaultText = `
 export const teaText = `
 <p>
   Want to cite, share, or modify this book? This book is
-  <a href="https://creativecommons.org/licenses/by/4.0/">
+  <a target="_blank" rel="noopener" href="https://creativecommons.org/licenses/by/4.0/">
```
PiotrKozlowski, line 68: Should these links also have `nofollow`?
Btw. there is also `addTargetBlankToLinks.ts`, but in this case, I think it's better to hardcode

KarinaMendez2: It hasn't been required

PiotrKozlowski: Since you've added a `noopener` that is not in the scope of the issue I think you could also add `nofollow` if we want it

PR #864, opened 2020-09-21
Handle scrolling in multiple managers

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/589

The previous approach of [pr 702](https://github.com/openstax/rex-web/pull/702) because of this comment: https://github.com/openstax/rex-web/pull/702#issuecomment-694379906

TODO:
- [x] fix tests


TomWoodward (approved):
i haven't tested this but i like the code organization, please fix the tests

janKozie1:
i found few bugs
- [Search doesn't reinitialize sometimes ](https://drive.google.com/file/d/1JUWfaB9UEkbuDYfJZ5oysK2LD8bm17KL/view?usp=sharing)
- [Search doesn't scroll properly to the first selected result](https://drive.google.com/file/d/16xmCUmSRzJzqB_P0ouS9KHzS7WRCZ3Qs/view?usp=sharing) - seems to happen when navigating to a different page when clicking a search result too
- [Not scrolling to search highlight after clicking a figure link](https://drive.google.com/file/d/1L7dxBAByPp27n7Z0MaKlkb5b_Fr08TMy/view?usp=sharing) 
- [Opening an url to a highlight when logged out doesn't scroll to hash](url) 
- [Deleting a highlight that was scrolled to shows an error](https://drive.google.com/file/d/1OSUYsX1BSUIfmjSNFihJcCJSdC9lcFKJ/view?usp=sharing) 
- [After navigating with a highlight scroll target and unfocusing it, changing tabs does weird jumps](https://drive.google.com/file/d/16xmCUmSRzJzqB_P0ouS9KHzS7WRCZ3Qs/view?usp=sharing)

PiotrKozlowski:
@janKozie1 
> Search doesn't reinitialize sometimes 

May this be related to your highlights? It works on me locally and on deployment in Firefox and Chrome

> Search doesn't scroll properly to the first selected result

This seems to be linking to the wrong video

> Not scrolling to search highlight after clicking a figure link

Fixed

> Opening an url to a highlight when logged out doesn't scroll to hash

Wrong url for a video, but this seems to be working for me. I've tested with [this link](https://rex-web-handle-scrollin-le2nde.herokuapp.com/books/american-government-2e/pages/1-introduction?target=%7B%22id%22%3A%22da7aeb86-5362-4839-986c-b76ecb1d1415%22%2C%22type%22%3A%22highlight%22%2C%22elementId%22%3A%22fs-id1171472269687%22%7D#fs-id1171472269687) on both firefox and chrome

> Deleting a highlight that was scrolled to shows an error

Fixed

> After navigating with a highlight scroll target and unfocusing it, changing tabs causes weird jumps

Fixed

> Also, when go to highlight, the initial scroll seems to happen bit lower, then is corrected moment later. I'm guessing that's okay? It looks like it bounces off a bit

If this lower scroll happens before highlights are loaded and is fixed when they are mounted then I think it is okay

Now I have to add and fix tests...

janKozie1:
Yes, sorry, they were pointing to wrong videos. Those should work:
https://drive.google.com/file/d/1Z_sPU-MG7jUHJzQcdciDzydjLVqqXICN/view?usp=sharing
https://drive.google.com/file/d/1dZ-dzFnUpmWv5XADZUICxEQJdzYVsi1x/view?usp=sharing

> May this be related to your highlights? It works on me locally and on deployment in Firefox and Chrome

It seems to happen to me mostly on books containing math. Might be related to my PC.

src/app/navigation/reducer.ts
```diff
@@ -15,6 +19,15 @@ export default (location: Location): Reducer<State, AnyAction> => (state = addQu
...
       return addQuery(action.payload.location);
+    case getType(deleteHighlight): {
+      const scrollTarget = getScrollTargetFromQuery(state.query, state.hash);
+      if (scrollTarget && isHighlightScrollTarget(scrollTarget) && scrollTarget.id === action.payload) {
+        const window = assertWindow();
+        window.history.replaceState(null, '', window.location.origin + window.location.pathname);
+        return {...state, search: '', hash: '', query: {}};
+      }
+      return state;
+    }
```
janKozie1, line 30: Ideally, reducers should be pure functions of actions and state.

PiotrKozlowski: Are you concerned about using `window.history` here or do you mean that this part of code should be extracted to the separated function?

The thing is that we don't really have to clear the history state here - clearing search, hash and query is enough, however, then there is a scrollTarget in the url

I was thinking about dispatching `locationChange` in deleteHighlight hook, but this is async action so the error alert will be already displayed + we would have block somehow this locationChange so we don't load a page again

janKozie1: Pure functions can't have side effects and modifying `window.history` is one.

TomWoodward: yea putting this replaceState in the reducer is kinda bad, really it should be in a redux hook or something

PiotrKozlowski: the problems is a follows:
- we want to remove the scroll target from the URL when a user removes the highlight that it was pointing to
- we have to do this synchronously because if we not then highlightManager will know that there is a scrollTarget but it will not find the highlight and will throw the error
- to stop the error from happening we have to clear the navigation's state as in the example above
- there is still a scroll target in the URL
- we can't clear it in hook because it is running after the action updated the state so we no longer know what is the current scroll target
- we could parse `window.location.search` manually in the hook but this seems weird + it is splitting the code for clearing the scroll target into even more places

This could be solve by using `redux-thunk` approach that I mentioned in another pr https://github.com/openstax/rex-web/pull/822#issuecomment-694626803  

Any ideas?

TomWoodward: the solution is to use two actions, you'd have to do this with redux-thunk as well because thunk actions are not reduced.

you use one action, which i sometimes prefix with "request" like `requestDeleteHighlight` which may or may not be reduced at all, theoretically it could create some kind of loading state. that action will trigger a redux hook which does whatever side effect you need and then dispatch `receiveDeleteHighlight` indicating that the action is complete and you can reduce that action to update the state.

typically i'd do the api call in response to the `request` action and dispatch the result with the `receive` payload, but this does not fit with the current eager/optimistic design of the delete highlight interaction so in order to maintain your current behavior you might have to hook into `request` to do your history update and again into `receive` to do the api call

janKozie1:
I didn't look at the code, but I found two smaller issues:
- https://drive.google.com/file/d/1lJphSsDNGLplPsWPmVmXz7Y2DOpTr1bJ/view?usp=sharing
- https://drive.google.com/file/d/1u7Qas-cNmaXLTm-L2sRhsOUmc7Zz_qhI/view?usp=sharing - i imagine this one might be happening because of my PC.

PiotrKozlowski:
The first issue (https://drive.google.com/file/d/1lJphSsDNGLplPsWPmVmXz7Y2DOpTr1bJ/view?usp=sharing):
fixed

The second issue (https://drive.google.com/file/d/1u7Qas-cNmaXLTm-L2sRhsOUmc7Zz_qhI/view?usp=sharing ):
I can replicate this on the deployment only after opening a page for the first time, then it works correctly until I restart the browser :D Does it behave like this for you?

janKozie1:
> Does it behave like this for you?

It doesn't seem to follow any pattern for me, it just _sometimes_ happens.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -163,10 +193,34 @@ export default (container: HTMLElement, getProp: () => HighlightProp) => {
...
+        }
+
+        if (
+          scrollTargetHighlight
+          && toFocus.id === scrollTargetHighlight.id
+          && toFocus.id !== focusedId
+          && toFocus.id !== scrollTargetHighlightIdThatWasHandled) {
+          focus(toFocus.id);
+          (toFocus.elements[0] as HTMLElement).scrollIntoView();
+          scrollTargetHighlightIdThatWasHandled = scrollTargetHighlight.id;
```
janKozie1, line 219: I think it would be nice to have this logic in a separate function close to `setPendingHighlight`. I feel like there is a lot of noise in `manager.update`

src/app/content/components/PageWithMockedHighlighter.spec.tsx
```diff
@@ -0,0 +1,255 @@
...
+    dispatch = jest.spyOn(store, 'dispatch');
+
+    const testServices = createTestServices();
+
+    services = {
+      ...testServices,
+      dispatch: store.dispatch,
+      getState: store.getState,
+    };
+    archiveLoader = testServices.archiveLoader;
```
janKozie1, line 85: Is it worth having it as a separate variable if it's used in only one test?

src/app/content/highlights/components/Highlights.spec.tsx
```diff
@@ -400,9 +406,16 @@ describe('Highlights', () => {
       const [firstDeleteWrapper, secondDeleteWrapper] = component.root.findAllByType(HighlightDeleteWrapper);
       firstDeleteWrapper.props.onDelete();
       secondDeleteWrapper.props.onCancel();
+
+      requestRemoveHighlight.hookBody({...services, getState: store.getState, dispatch: store.dispatch})(
+        requestDeleteHighlight(hlBlue as Highlight, {
+          locationFilterId: pageId,
+          pageId,
+        }
+      ));
```
janKozie1, line 415: Why does it need to be called manually?

janKozie1 (commented):
I'm leaving some minor comments, I didnn't get through all the files.

src/app/content/components/PageWithMockedHighlighter.spec.tsx
PiotrKozlowski: Probably doesn't matter
It is all about types - I can't use `services.archiveLoader.mockPage` because it does not exist on type `(AppServices & MiddlewareAPI)['archiveLoader']` so I would have to cast it
I've just copied this part from `Page.spec.tsx`

src/app/content/highlights/components/Highlights.spec.tsx
PiotrKozlowski: it seems like a middleware with hooks is not passed in `createTestStore`
It kind of makes sense, right?

src/app/content/components/Page.spec.tsx
```diff
@@ -992,6 +999,113 @@ describe('Page', () => {
...
+      store.dispatch(selectSearchResult({result: hit2, highlight: 1}));
+    });
+
+    // page lifecycle hooks
+    await Promise.resolve();
+    // after images are loaded
+    await Promise.resolve();
+
+    expect(root.querySelector('[data-testid=banner-body]')).toBeTruthy();
+    expect(dispatch).toHaveBeenCalledWith({
```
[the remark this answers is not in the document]
PiotrKozlowski: I've added dispatch.mockClear() after the first one

src/app/content/search/hooks.spec.ts
```diff
@@ -93,132 +108,84 @@ describe('hooks', () => {
...
+
+      expect(dispatch).not.toHaveBeenCalledWith();
     });
 
-    it('dispatches selectSearchResult if query and page are the same', () => {
+    it('selects search result', () => {
+      const hit = makeSearchResultHit({book, page});
+      const searchScrollTarget = { type: 'search', index: 0 } as SearchScrollTarget;
+      store.dispatch(navigationLocationChange({
+        location: { hash: hit.source.elementId, search: `?query=asdf&target=${JSON.stringify(searchScrollTarget)}` },
```
janKozie1, line 157: This is the same as `mockSearchScrollTarget`, right? Maybe consider making both stringified and normal versions of it available in `syncSearch` tests

src/app/content/search/components/SearchResultsSidebar/index.browserspec.tsx
```diff
@@ -50,6 +51,10 @@ it('clears search input without affecting search results sidebar', async() => {
 });
 
 it('closes the search results sidebar without affecting search input', async() => {
+  await page.setCookie({
+    name: cookieNudge.date,
+    value: new Date().toString(),
+  });
```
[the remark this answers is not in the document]
PiotrKozlowski: seems like a wrongly resolved merge commit

src/app/content/components/Page.spec.tsx
```diff
@@ -928,7 +936,7 @@ describe('Page', () => {
     expect(root.querySelector('[data-testid=banner-body]')).toBeFalsy();
 
     renderer.act(() => {
-      store.dispatch(selectSearchResult({result: hit2, highlight: 0}));
+      store.dispatch(selectSearchResult({result: hit2, highlight: 1}));
```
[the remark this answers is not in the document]
PiotrKozlowski: I think I did this change before a merge with new `<Toasts>` component and I need the `id` of search result to be different

src/app/content/components/Page/highlightManager.ts
PiotrKozlowski: Is this not resolved? Did you mean something different than I did?

janKozie1 (approved):
I think it works great.

I was wondering if the search results list should be automatically closed on mobile devices when navigating with search scroll target?

PiotrKozlowski:
> I was wondering if the search results list should be automatically closed on mobile devices when navigating with search scroll target?

Good question

For now, Tom said we should leave it as it is: https://openstax.slack.com/archives/C8V1XET1R/p1601995388300300

janKozie1 (approved):

PR #878, opened 2020-09-30
Clicking figure/table links in MH and SG : right location

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/805

### DIFF
--- a/src/app/components/ContentExcerpt.tsx
+++ b/src/app/components/ContentExcerpt.tsx
@@ -1,0 +1,59 @@
+import { Highlight } from '@openstax/highlighter/dist/api';
+import { HTMLAnchorElement } from '@openstax/types/lib.dom';
+import React from 'react';
+import { useSelector } from 'react-redux';
+import styled from 'styled-components/macro';
+import { content } from '../content/routes';
+import { book } from '../content/selectors';
+import { BookWithOSWebData } from '../content/types';
+import { fromRelativeUrl, getUrlParamForPageId } from '../content/utils/urlUtils';
+import { bodyCopyRegularStyle } from './Typography';
+
+// tslint:disable-next-line:variable-name
+export const HighlightContent = styled.div`
+  ${bodyCopyRegularStyle}
+  overflow: auto;
+
+  * {
+    overflow: initial;
+  }
+`;
+
+interface Props {
+  highlight: Highlight;
+  content: any;
+  className: string;
+}
+
+// tslint:disable-next-line:variable-name
+const ContentExcerpt = (props: Props) => {
+  const currentBook = useSelector(book);
+
+  if (!currentBook) {
+    return null;
+  }
+
+  const absolutePath = content.getUrl({
+    book: {
+      slug: (currentBook as BookWithOSWebData).slug,
+    },
+    page: getUrlParamForPageId(currentBook, props.highlight.sourceId),
+  });
+
+  const parser = new DOMParser();
+  const domNode = parser.parseFromString(props.content, 'text/html');
+  domNode.querySelectorAll('a').forEach(
+    (a: HTMLAnchorElement) => {
+      const currentHref = a.getAttribute('href') || '';
+      return a.setAttribute('href', fromRelativeUrl(currentHref, absolutePath));
+    }
+  );
+
+  return <HighlightContent
+    className={props.className}
+    data-highlight-id={props.highlight.id}
+    dangerouslySetInnerHTML={{ __html: domNode.body.innerHTML }}
+  />;
+};
+
+export default ContentExcerpt;
--- a/src/app/components/ContentExcerpt.tsx
+++ b/src/app/components/ContentExcerpt.tsx
@@ -1,12 +1,12 @@
 import { Highlight } from '@openstax/highlighter/dist/api';
-import { HTMLAnchorElement } from '@openstax/types/lib.dom';
 import React from 'react';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components/macro';
-import { content } from '../content/routes';
 import { book } from '../content/selectors';
 import { BookWithOSWebData } from '../content/types';
-import { fromRelativeUrl, getUrlParamForPageId } from '../content/utils/urlUtils';
+import addTargetBlankToLinks from '../content/utils/addTargetBlankToLinks';
+import { findArchiveTreeNodeById } from '../content/utils/archiveTreeUtils';
+import { getBookPageUrlAndParams } from '../content/utils/urlUtils';
 import { bodyCopyRegularStyle } from './Typography';
 
 // tslint:disable-next-line:variable-name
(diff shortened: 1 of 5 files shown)


src/app/content/utils/urlUtils.ts
```diff
@@ -98,5 +98,18 @@ export const toRelativeUrl = (from: string, to: string) => {
...
+
+export const fromRelativeUrl = (from: string, to: string) => {
+  if (!window) { return null; }
+  const base = new URL( window.location.origin);
+
+  if (from.startsWith('#')) {
+    base.pathname = to;
+    base.hash = from;
+    return base;
+  }
```
PiotrKozlowski, line 112: I'm not sure if this check should be here or in the ContentExcerpt

src/app/content/utils/urlUtils.ts
```diff
@@ -98,5 +98,18 @@ export const toRelativeUrl = (from: string, to: string) => {
   const commonParts = getCommonParts(parsedFrom, parsedTo.slice(0, -1));
 
   return '../'.repeat(Math.max(0, parsedFrom.length - commonParts.length - 1))
-    + parsedTo.slice(commonParts.length).join('/');
+  + parsedTo.slice(commonParts.length).join('/');
+};
+
+export const fromRelativeUrl = (from: string, to: string) => {
+  if (!window) { return null; }
```
PiotrKozlowski, line 105: assertWindow()

PiotrKozlowski (commented):
I think the approach looks good. I've left some general comments

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+  if (!currentBook) {
+    return null;
+  }
+
+  const absolutePath = content.getUrl({
+    book: {
+      slug: (currentBook as BookWithOSWebData).slug,
+    },
+    page: getUrlParamForPageId(currentBook, props.highlight.sourceId),
+  });
```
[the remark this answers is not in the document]
TomWoodward: `getUrlParamForPageId` is internally memoized so i'm not sure if there is much benefit of adding `React.useMemo()` here

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+  ${bodyCopyRegularStyle}
+  overflow: auto;
+
+  * {
+    overflow: initial;
+  }
+`;
+
+interface Props {
+  highlight: Highlight;
```
TomWoodward, line 23: content excerpts will not always use a highlight, there is an example signature in the card

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+    },
+    page: getUrlParamForPageId(currentBook, props.highlight.sourceId),
+  });
+
+  const parser = new DOMParser();
+  const domNode = parser.parseFromString(props.content, 'text/html');
+  domNode.querySelectorAll('a').forEach(
+    (a: HTMLAnchorElement) => {
+      const currentHref = a.getAttribute('href') || '';
+      return a.setAttribute('href', fromRelativeUrl(currentHref, absolutePath));
```
TomWoodward, line 48: what you're calling `absolutePath` is the base url (the url that the source was taken from), we should make that the first argument of the function, and pass the relative url second

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+      slug: (currentBook as BookWithOSWebData).slug,
+    },
+    page: getUrlParamForPageId(currentBook, props.highlight.sourceId),
+  });
+
+  const parser = new DOMParser();
+  const domNode = parser.parseFromString(props.content, 'text/html');
+  domNode.querySelectorAll('a').forEach(
+    (a: HTMLAnchorElement) => {
+      const currentHref = a.getAttribute('href') || '';
```
TomWoodward, line 47: these need to be filtered to only relative urls

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+      return a.setAttribute('href', fromRelativeUrl(currentHref, absolutePath));
+    }
+  );
+
```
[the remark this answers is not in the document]
TomWoodward: yea part of the point of this component is to combine that functionality

src/app/content/utils/urlUtils.ts
```diff
@@ -98,5 +98,18 @@ export const toRelativeUrl = (from: string, to: string) => {
...
+  const base = new URL( window.location.origin);
```
TomWoodward, line 106: you shouldn't be using the window url for anything

src/app/content/utils/urlUtils.ts
TomWoodward: i'm not sure we need to mess around with the hash at all, ideally it would be preserved in the modified url

TomWoodward (changes requested):

src/app/content/utils/urlUtils.ts
```diff
@@ -98,5 +98,18 @@ export const toRelativeUrl = (from: string, to: string) => {
   const commonParts = getCommonParts(parsedFrom, parsedTo.slice(0, -1));
 
   return '../'.repeat(Math.max(0, parsedFrom.length - commonParts.length - 1))
-    + parsedTo.slice(commonParts.length).join('/');
+  + parsedTo.slice(commonParts.length).join('/');
+};
+
+export const fromRelativeUrl = (from: string, to: string) => {
```
janKozie1, line 104: I think, this should check where `to` and `from` intersect, and return `to` + (`from` - everything before intersection in `from`)

src/app/components/ContentExcerpt.tsx
KarinaMendez2: both HL and SG list element had `HighlightListElementProps {  highlight: Highlight; }` and they both do `data-highlight-id={highlight.id}`, is this different?

src/app/components/ContentExcerpt.tsx
KarinaMendez2: to know if it's a relative url I need to get the `href` first, right? Do I need to filter before grabbing the href?

src/app/components/ContentExcerpt.tsx
TomWoodward: content excerpts will not always use a highlight, you can accept a string for the excerpt

src/app/components/ContentExcerpt.tsx
TomWoodward: you could theoretically do it in the selector, or you could filter the set after, but you're not doing any filtering right now so you gotta add that somewhere

src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,59 @@
...
+// tslint:disable-next-line:variable-name
+const ContentExcerpt = (props: Props) => {
+  const currentBook = useSelector(book);
+
+  if (!currentBook) {
+    return null;
+  }
+
+  const absolutePath = content.getUrl({
```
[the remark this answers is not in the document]
TomWoodward: yea i think that requirement is safe to remove

src/app/content/utils/urlUtils.ts
TomWoodward: they won't literally string intersect because `to` is relative. you'll have to break the `from` into parts and remove pieces from the end for every `../` in the `to`, then add the remainder of the `to` to whatever is left. very similar logic to the `toRelativeUrl` function but backwards

KarinaMendez2:
Closing this PR due to new PR with a rebase workflow: https://github.com/openstax/rex-web/pull/888/

PR #888, opened 2020-10-01
Clicking on figure/table links

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/805


src/app/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,37 @@
...
+`;
+
+interface Props {
+  highlight: Highlight | string;
+  content: string;
+  className: string;
+}
+
+// tslint:disable-next-line:variable-name
+const ContentExcerpt = (props: Props) => {
```
janKozie1, line 24: I was thinking, maybe we could make `ContentExcerpt` accept a render prop as children instead?
You could then make the signature like

```js
<ContentExcerpt content={highlight.highlightedContent}>
    {(parsedContent) => <HighlightContent dangerouslySetInnerHTML={parsedContent}/> }
</ContentExcerpt>
```

I think it could make the component focus on parsing content and not doing anything where-the-content-is-coming-from specific.

KarinaMendez2: what about the other props? So you mean like render two components?

janKozie1: But does the `ContentExcerpt` care about other props? All it needs is html and data needed to parse that html, so book and page. All other props could be directly passed into the child.

KarinaMendez2: well data-highlight-id seams important, maybe className can always be the same but idk

janKozie1: But you can pass the other props directly into the child like

```js
<ContentExcerpt content={highlight.highlightedContent}>
    {(parsedContent) => <HighlightContent dangerouslySetInnerHTML={parsedContent} data-highlight-id=""/> }
</ContentExcerpt>
```

ContentExcerpt wouldn't be aware of what you do with the parsed content.

KarinaMendez2: ok, um what about when the excerpt isn't a highlight?

janKozie1: Well 

```js
<ContentExcerpt content={someHTML}>
    {(parsedContent) => <SomethingThatRendersParsedHTML otherProps={} dangerouslySetInnerHTML={parsedContent}/> }
</ContentExcerpt>
```

TomWoodward: @janKozie1 it seems like the purpose of your approach would be to pass an element that handles the display styling of the excerpt, where this component would only do the parsing.

i think your idea is reasonable, but i don't see why we couldn't accomplish the same thing at this point by simply extending `ContentExcerpt` as a styled component like:

const HighlightContent = styled(ContentExcerpt)`
  ...special highlight styles for the excerpt
`

which i feel is more straightforward and follows more common patterns.

if we ever run into a situation where this approach isn't working, we may consider coming back to your proposal here

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -3,11 +3,10 @@ import React from 'react';
 import { useDispatch } from 'react-redux';
 import styled, { css } from 'styled-components/macro';
 import { useAnalyticsEvent } from '../../../../../helpers/analytics';
-import { bodyCopyRegularStyle } from '../../../../components/Typography';
+import ContentExcerpt, { HighlightContent } from '../../../../components/ContentExcerpt';
```
TomWoodward, line 6: why do you need `HighlightContent` here? is it even used?

src/app/content/studyGuides/components/StudyGuidesListElement.tsx
```diff
@@ -1,11 +1,11 @@
 import { Highlight } from '@openstax/highlighter/dist/api';
 import React from 'react';
 import styled, { css } from 'styled-components/macro';
-import { bodyCopyRegularStyle, textRegularStyle } from '../../../components/Typography';
+import ContentExcerpt, { HighlightContent } from '../../../components/ContentExcerpt';
```
TomWoodward, line 4: same

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -154,10 +139,10 @@ const HighlightListElement = ({ highlight, locationFilterId, pageId }: Highlight
       onColorChange={updateColor}
     />}
     <HighlightContentWrapper color={highlight.color}>
-      <HighlightContent
+      <ContentExcerpt
         className='summary-highlight-content'
```
TomWoodward, line 143: i believe this class is what applies the content styling. you should move it into `ContentExcerpt` and maybe change the name

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
KarinaMendez2: is `summary-excerpt-styles` fine?

TomWoodward: why not `.content-excerpt`?

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,36 @@
...
+    {...excerptProps}
+  />;
+})`
+  ${bodyCopyRegularStyle}
+  overflow: auto;
+
+  * {
+    overflow: initial;
+  }
+` ;
```
TomWoodward, line 34: extra space

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,36 @@
...
+    ...excerptProps
+  } = props;
+
+  const fixedContent = React.useMemo(
+    () => addTargetBlankToLinks(props.content),
+    [props.content]
+  );
+
+  return <div
+    className='content-excerpt'
```
TomWoodward, line 23: youre losing this to the classname that styled is passing in

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.tsx
KarinaMendez2: fixed

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -0,0 +1,37 @@
...
+    () => addTargetBlankToLinks(props.content),
+    [props.content]
+  );
+
+  return <div
+    dangerouslySetInnerHTML={{ __html: fixedContent }}
+    {...excerptProps}
+  />;
+}).attrs({
+  className: 'content-excerpt',
```
TomWoodward, line 27: pretty sure this is going to break your ability to extend these styles, it would be more straightforward to do 

```
const ContentExcerpt = styled(({content, className, ...otherProps}: Props) => {
  ...
  return <div
    dangerouslySetInnerHTML={{ __html: fixedContent }}
    className={`content-excerpt ${className}`}
    {...otherProps}
  />;
})

**Review by TomWoodward — CHANGES_REQUESTED** [review]
we're gonna try a rebase workflow on this PR and see how it goes. once you finish addressing my comments please squash all these commits into one for "extract content excerpt logic into a common component" and begin adding your changes for the next stage of the issue.

when rebasing/squashing it is possible to lose changes if you do it wrong, so if you're nervous you may want to create another branch as backup or literally copy the files somewhere so you don't have to stress out about it.

File: src/app/content/utils/urlUtils.ts [hunk 8]
```diff
@@ -100,3 +100,8 @@ export const toRelativeUrl = (from: string, to: string) => {
   return '../'.repeat(Math.max(0, parsedFrom.length - commonParts.length - 1))
     + parsedTo.slice(commonParts.length).join('/');
 };
+
+export const fromRelativeUrl = (base: string, to: string) => {
+  const urlBase = new URL(to, `https://openstax.org${base}`);
```
TomWoodward, line 105: can you add a comment here that the hostname is actually irrelevant
```suggestion
  // this hostname is required by the URL constructor but we ignore it in our response, the value is irrelevant
  const urlBase = new URL(to, `https://openstax.org${base}`);
```

TomWoodward (changes requested):

script/entry.js
```diff
@@ -9,6 +9,9 @@ const extensions = ['.ts', '.tsx'];
 const JSDOM = require('jsdom').JSDOM;
 global.DOMParser = new JSDOM().window.DOMParser;
 
+const URL = require ('url');
```
TomWoodward, line 12: extra space around the parens

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,24 +1,44 @@
...
     ...excerptProps
   } = props;
 
+  const currentBook = useSelector(book);
+
+  if (!currentBook) {
+    return null;
+  }
+
+  const sourcePage = findArchiveTreeNodeById(currentBook.tree, sourcePageId) as unknown as Page;
```
TomWoodward, line 32: you shouldn't need this cast, it is bad

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,24 +1,44 @@
...
+
+  const excerptSource = getBookPageUrlAndParams(
+    currentBook,
+    sourcePage
+  );
+
   const fixedContent = React.useMemo(
-    () => addTargetBlankToLinks(props.content),
-    [props.content]
+    () => addTargetBlankAndFixRelativeURL(props.content, excerptSource.url),
```
TomWoodward, line 40: don't combine these into one function. leave the old one the way it was and create a new one for fixing the relative urls.

https://en.wikipedia.org/wiki/Single-responsibility_principle

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,28 +1,52 @@
...
+  }
+
+  const excerptSource = getBookPageUrlAndParams(
+    currentBook,
+    sourcePage
   );
 
+  const fixedContent = React.useMemo(() => flow(
+    addTargetBlankToLinks,
+    (newContent) => fixRelativeURLs(newContent, excerptSource.url)),
```
TomWoodward, line 45: add the call for this function here instead of on line 49, otherwise you're not memoizing the logic

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,28 +1,52 @@
...
+  const currentBook = useSelector(book);
+
+  if (!currentBook) {
+    return null;
+  }
+
+  const sourcePage = findArchiveTreeNodeById(currentBook.tree, sourcePageId);
+
+  if (!sourcePage) {
+    return null;
```
TomWoodward, line 35: you should throw instead of nooping here

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -121,7 +121,7 @@ const HighlightListElement = ({ highlight, locationFilterId, pageId }: Highlight
   };
 
   const confirmDelete = () => {
-    dispatch(requestDeleteHighlight(highlight, {
+    dispatch(deleteHighlight(highlight, {
```
TomWoodward: why are you changing this?

src/app/content/studyGuides/components/__snapshots__/StudyGuides.spec.tsx.snap
```diff
@@ -43,17 +43,7 @@ exports[`StudyGuides properly display summary highlights 1`] = `
...
-            className="content-excerpt ContentExcerpt-sc-16hg3lt-0 iKlVB"
-            dangerouslySetInnerHTML={
-              Object {
-                "__html": "<a href=\\"link\\" target=\\"_blank\\">Link</a>",
-              }
-            }
-            data-highlight-id="hl1"
-          />
-        </div>
+        />
```
TomWoodward, line 46: this seems bad

src/app/content/studyGuides/components/__snapshots__/StudyGuides.spec.tsx.snap
```diff
@@ -66,17 +56,7 @@ exports[`StudyGuides properly display summary highlights 1`] = `
...
-            className="content-excerpt ContentExcerpt-sc-16hg3lt-0 iKlVB"
-            dangerouslySetInnerHTML={
-              Object {
-                "__html": "content",
-              }
-            }
-            data-highlight-id="hl2"
-          />
-        </div>
+        />
```
TomWoodward, line 59: and this

src/app/content/utils/fixContentUrls.ts
```diff
@@ -0,0 +1,28 @@
+import { HTMLAnchorElement } from '@openstax/types/lib.dom';
```
TomWoodward, line 1: can you rename this file `contentManipulation` or something less url specific

src/app/content/utils/fixContentUrls.spec.ts
```diff
@@ -47,3 +47,19 @@ describe('addTargetBlankToLinks', () => {
...
+  it('converts relative to full path <a> tags' , () => {
+    const input = '<div class="test">'
+    + '<a href="some/link">First link</a>'
+    + '<a href="some/link">Second link</a>'
+    + '</div>';
+
+    const output = '<div class="test">'
+    + '<a href="some/link" target="_blank">First link</a>'
+    + '<a href="some/link" target="_blank">Second link</a>'
+    + '</div>';
```
TomWoodward, line 61: this test doesn't look like its testing what it says its testing

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { formatBookData } from '../utils';
+import ContentExcerpt from './ContentExcerpt';
+
+describe('ContentExcerpt', () => {
+  let store: Store;
+  let consoleError: jest.SpyInstance;
+
+  beforeEach(() => {
+    store = createTestStore();
+    consoleError = jest
```
TomWoodward, line 20: this is to ignore the error boundary warning? it would be better to add that to the block list in tests/setup so that any other errors are not hidden

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <MessageProvider>
+      <ContentExcerpt
+        content={content}
+        sourcePageId={sourcePage}
+        className='class1'
+      />
+    </MessageProvider>
+  </Provider>);
+
+  it('fixes urls in content and matches snapshot', () => {
```
TomWoodward, line 39: you're testing a pretty big integration here, it would be better to mock `addTargetBlankToLinks` and `fixRelativeURLs` and assert that they are called rather than testing what they do

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,26 +1,47 @@
...
+  const sourcePage = assertDefined(findArchiveTreeNodeById(currentBook.tree, sourcePageId), 'page not found in book');
+
+  const excerptSource = getBookPageUrlAndParams(
+    currentBook,
+    sourcePage
   );
 
+  const fixedContent = React.useMemo(() => flow(
+    addTargetBlankToLinks,
+    (newContent) => fixRelativeURLs(newContent, excerptSource.url))(props.content),
```
TomWoodward, line 42: can you reformat this so the `)(props.content),` is on the next line please

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,26 +1,47 @@
...
+    sourcePageId,
     ...excerptProps
   } = props;
 
-  const fixedContent = React.useMemo(
-    () => addTargetBlankToLinks(props.content),
-    [props.content]
+  const currentBook = useSelector(book);
+
+  if (!currentBook) {
```
TomWoodward, line 29: is there a legitimate situation where book would not be loaded? if not then we can throw here

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
TomWoodward: why are you changing this?

src/app/content/utils/contentManipulation.spec.ts
```diff
@@ -1,5 +1,5 @@
-import addTargetBlankToLinks from './addTargetBlankToLinks';
 
+import { addTargetBlankToLinks } from '../utils/contentManipulation';
```
TomWoodward, line 2: blank line above this?

src/app/content/utils/contentManipulation.ts
```diff
@@ -0,0 +1,28 @@
...
+const domParser = new DOMParser();
+
+export function addTargetBlankToLinks(htmlString: string): string  {
+  const domNode = domParser.parseFromString(htmlString, 'text/html');
+  domNode.querySelectorAll('a').forEach((a: HTMLAnchorElement) => a.setAttribute('target', '_blank'));
+  return domNode.body.innerHTML;
+}
+
+export const isAbsoluteUrl = (url: string) => {
+  const pattern = /^(http)?s?:\/\//i;
```
TomWoodward, line 13: this matches things it shouldn't you want to move the `s?:` into the parentheses with the `http`

you also don't want to treat urls beginning in a slash, like `/books/asdf/pages/qre` as relative, they aren't fully qualified urls because they don't have the domain or protocol, but they are absolute paths

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+    render('testbook1-testpage1-uuid', originalHtml);
+
+    expect(getUrlSpy).toHaveBeenCalled();
+    expect(addTargetBlankToLinksMock).toHaveBeenCalledWith(originalHtml);
+    expect(fixRelativeURLsMock)
+      .toHaveBeenCalledWith(htmlWithTargetBlank, '/book/book1/page/testbook1-testpage1-uuid');
+  });
+
+  it('returns null when book is not loaded', () => {
```
TomWoodward, line 55: the name of this test is wrong

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+    expect(() => {
+      render('testbook1-testpage1-uuid', '<a href="#hello"></a>');
+    }).toThrowError(new Error('book not loaded'));
+  });
+
+  it('returns null when page is not loaded', () => {
```
TomWoodward, line 61: and this one

src/test/setup.ts
```diff
@@ -27,12 +27,18 @@ const ignoreConsoleErrorMessages = [
...
+  /*
+   * complains about not having compenent that throw before render
+   * not being wrapped in <ErrorBoundary>
+  */
+  /(.*)Consider adding an error boundary to your tree to customize error handling behavior.(.*)/,
 ];
 
 const originalConsoleError = console.error;  // tslint:disable-line:no-console
-console.error = (msg: unknown) => {  // tslint:disable-line:no-console
+console.error = (msg: string) => {  // tslint:disable-line:no-console
```
TomWoodward, line 38: your types in here are messed up. its possible to console.error something that isn't a string, and the conditional below does a typecheck on it to see if it is a string. if it makes life easier i'm not opposed to changing the `ignoreConsoleErrorMessages` array to all be regexes

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,26 +1,43 @@
...
+  const excerptSource = getBookPageUrlAndParams(
+    currentBook,
+    sourcePage
   );
 
+  const fixedContent = React.useMemo(() => flow(
+    addTargetBlankToLinks,
+    (newContent) => fixRelativeURLs(newContent, excerptSource.url)
+  )(props.content),
+  [props.content, excerptSource.url]);
```
TomWoodward, line 39:
```suggestion
)(props.content), [props.content, excerptSource.url]);
```

TomWoodward (changes requested):
some minor comments, the logic is all good we're just doing cleanup now

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -1,26 +1,42 @@
...
+
+  const excerptSource = getBookPageUrlAndParams(
+    currentBook,
+    sourcePage
   );
 
+  const fixedContent = React.useMemo(() => flow(
+    addTargetBlankToLinks,
+    (newContent) => fixRelativeURLs(newContent, excerptSource.url)
+)(props.content), [props.content, excerptSource.url]);
```
TomWoodward, line 38: indentation is messed up

TomWoodward (changes requested):

TomWoodward (approved):

PR #893, opened 2020-10-08
issue 1162 calculate top and bottom offset and pass it as a props

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1162

this pr is enough to solve the reported issue, but it works better with https://github.com/openstax/highlighter/pull/39 

This is how it should work with pr mentioned above: https://drive.google.com/file/d/1BVR1cLvEeYw-RIL70PaT91aRvfkRfpMS/view?usp=sharing

The biggest difference is when you have a full paragraph of search results when there are highlights in this paragraph and you refresh a page - on `master`, search results will load, but highlights will fail to attach. PR above makes sure that those highlights are mounted and persisted in the DOM


janKozie1:
Is it supposed to be flying down like this? https://drive.google.com/file/d/1OgrUskut5nCo8a8nlYonY4VD4CnmcLFX/view?usp=sharing

PiotrKozlowski:
@janKozie1 no, I've pushed a fix for it and changed prop name to be more descriptive

janKozie1 (approved):

PR #897, opened 2020-10-15
Issue 1329

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1329 

TODO:
- [x] Icon for Practice is not ready yet - [slack conversation](https://openstax.slack.com/archives/C8V1XET1R/p1602773276497000)


src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { PlainButton, toolbarDefaultButton, toolbarDefaultText } from './styled';
+
+// tslint:disable-next-line:variable-name
+export const PracticeQuestionsWrapper = styled(PlainButton)`
+  ${toolbarDefaultButton}
+`;
+
+// tslint:disable-next-line:variable-name
+const PracticeQuestionsIcon = styled.img`
+  padding: 0.2rem;
```
KarinaMendez2, line 16: why not ` ${toolbarIconStyles}`?

src/app/content/components/Toolbar/PracticeQuestions.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+      </Services.Provider>
+    </Provider>);
+
+    expect(component.toJSON()).toMatchSnapshot();
+  });
+
+  it('clicking button works', () => {
+    store.dispatch(receiveFeatureFlags([practiceQuestionsFeatureFlag]));
+
+    const component = renderer.create(<Provider store={store}>
```
KarinaMendez2, line 53: what about putting all renderer.create into a render() function?

KarinaMendez2 (changes requested):
couple comments, everything else looks good

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
PiotrKozlowski: icon looks bigger than in the design with these styles

src/app/content/components/Toolbar/PracticeQuestions.spec.tsx
PiotrKozlowski: I think this is a good idea if all of the tests use the same JSX

PiotrKozlowski:
@KarinaMendez2 I've responded to your comments and also pushed a commit that will make sure we are removing the print button and changing styles of my highlights button only if practice questions are enabled
https://openstax.slack.com/archives/C8V1XET1R/p1603194876087500

KarinaMendez2 (approved):

PR #907, opened 2020-10-22
Issue 1328

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1328

I've added a simple mock summary data for Physics. If you have Redux dev tools then you can check that this is being loaded only for this book 

commit history is a mess, because this branch was based on another branch that is already mered into master


src/app/content/practiceQuestions/utils.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  it('returns location filters for practice questions', () => {
+    const page1 = (book as any).tree.contents[1].contents[0].contents[0];
+    const linkedPage1 = findArchiveTreeNodeById(book.tree, page1.id);
+    const page2 = (book as any).tree.contents[1].contents[1];
+    const linkedPage2 = findArchiveTreeNodeById(book.tree, page2.id);
+    const page3 = (book as any).tree.contents[1].contents[2];
+    const linkedPage3 = findArchiveTreeNodeById(book.tree, page3.id);
+    const page4 = (book as any).tree.contents[2].contents[0];
+    const linkedPage4 = findArchiveTreeNodeById(book.tree, page4.id);
```
janKozie1, line 35: Could this be replaced by `findArchiveTreeNode(book.tree, hardcoded-page-id)`? The structure is hardcoded anyway and I'm not sure if it makes it more readable what belongs to where.

janKozie1 (commented):
just minor things

src/app/content/practiceQuestions/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  if (!node) { return; }
+
+  const flatLocationFilters = flattenLocationFilters(locationFilters);
+  const flatTree = flattenArchiveTree(book.tree);
+  const nodeIndex = flatTree.findIndex((search) => search.id === node.id);
+
+  for (let i = nodeIndex + 1; i < flatTree.length; i++) {
+    if (flatLocationFilters.find((section) => section.id === stripIdVersion(flatTree[i].id))) {
+      return flatTree[i] as LinkedArchiveTreeSection;
```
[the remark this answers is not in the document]
PiotrKozlowski: `flatLocationFilters` is of type `LinkedArchiveTreeSection[]` and we return `flatTree[i]` only if we find it in there

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -0,0 +1,33 @@
+import Sentry from '../../../../helpers/Sentry';
+import { AppServices, MiddlewareAPI } from '../../../types';
+import { bookAndPage } from '../../selectors';
+import { receivePracticeQuestionsSummary } from '../actions';
+import { hasPracticeQuestions, practiceQuestionsEnabled } from '../selectors';
+import { PracticeQuestionsSummary } from '../types';
+
+// composed in /content/locationChange hook because it needs to happen after book load
+const loadSummary = async(services: MiddlewareAPI & AppServices): Promise<PracticeQuestionsSummary | undefined> => {
```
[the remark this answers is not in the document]
PiotrKozlowski: I like to know the type without having to hover the function name :D

src/app/content/practiceQuestions/utils.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+describe('getPracticeQuestionsLocationFilters', () => {
+  it('returns empty map if book or summary is not truthy', () => {
+    expect(getPracticeQuestionsLocationFilters(null, book)).toEqual(new Map());
+    expect(getPracticeQuestionsLocationFilters({ countsPerSource: {} }, undefined)).toEqual(new Map());
+  });
+
+  it('returns location filters for practice questions', () => {
+    const page1 = (book as any).tree.contents[1].contents[0].contents[0]; // testbook1-testpage2-uuid
+    const linkedPage1 = findArchiveTreeNodeById(book.tree, 'testbook1-testpage2-uuid');
```
janKozie1, line 29: You could just use `linkedPageX` everywhere, right? Just need to return with assertion if its undefined / assertDefined

src/app/content/practiceQuestions/utils.ts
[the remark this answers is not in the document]
janKozie1: Could you search in `flatLocationFilters` then? or is their order not guaranteed.

janKozie1 (approved):

src/app/content/practiceQuestions/utils.ts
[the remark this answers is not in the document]
PiotrKozlowski: Someone can pass `nodeId` that is not in `locationFilters`, for example when opening a modal on Preface page: https://projects.invisionapp.com/d/#/console/20557970/432268585/inspect

Then we have to find `nodeIndex` in the `tree` and find the first section which is in the `flatLocationFilters` behind that number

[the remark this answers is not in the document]
janKozie1: I thought of something more like 
```js
const remainingNodes = flatTree.slice(nodeIndex).map(node => stripIdVersion(node.id))
return flatLocationFilters.find(filter => remainingNodes.includes(filter.id))
```

src/gateways/createPracticeQuestionsLoader.ts
```diff
@@ -0,0 +1,15 @@
+import { PracticeQuestionsSummary } from '../app/content/practiceQuestions/types';
+import mockPhysics from '../test/fixtures/practice/cce64fde-f448-43b8-ae88-27705cceb0da.json';
```
TomWoodward, line 2: i wanted to import this with a lazy import using a variable path based on the uuid passed in. then even if the data is not production ready the code will be. the lazy import means we're not bundling all the data in the default js file, but will load only the necessary books data we need after page load. 

also my intent wasn't to have this be a test file, but to put it in something like `rex-web/data/practice/summary/cce64fde-f448-43b8-ae88-27705cceb0da.json` we will use the same wiring in the loader when we go to production, we'll just have to add additional build logic to populate the files

script/prerender/prerender.tsx
```diff
@@ -8,12 +8,14 @@ import {
   REACT_APP_ARCHIVE_URL,
   REACT_APP_HIGHLIGHTS_URL,
   REACT_APP_OS_WEB_API_URL,
+  REACT_APP_PRACTICE_QUESTIONS_URL,
```
TomWoodward, line 11: there is no need to pretend there will be a url here, we won't be using it for a while. as long as the loader exists it won't be hard to add it later

TomWoodward (changes requested):

src/gateways/createPracticeQuestionsLoader.ts
PiotrKozlowski: > i wanted to import this with a lazy import using a variable path based on the uuid passed in. then even if the data is not production ready the code will be. the lazy import means we're not bundling all the data in the default js file, but will load only the necessary books data we need after page load.

By "lazy loading" do you mean `React.lazy(() => import('.../bookUid.json'))`? If yes then I think it is not possible to pass `bookUid` dynamically because webpack have to know from the beginning what is the name of the file to handle it properly

We could, however, add `/data/practice/summary/cce64fde-f448-43b8-ae88-27705cceb0da.json` to the `static` folder that holds other assets and get it's content with `fetch`. Is this acceptable?

TomWoodward: i want you to use the 
```
const data = import(`/data/practice/summary${uuid}.json`);
```
syntax so that the data files will not be bundled with the rest of the js but will be loaded as needed

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -0,0 +1,33 @@
...
+  const state = services.getState();
+  const {book} = bookAndPage(state);
+  const isEnabled = practiceQuestionsEnabled(state);
+  const hasCurrentPQ = hasPracticeQuestions(state);
+
+  if (!isEnabled || !book || hasCurrentPQ) { return; }
+
+  try {
+    const summary = await services.practiceQuestionsLoader.getPracticeQuestionsBookSummary(book.id);
+    return summary;
```
TomWoodward, line 19: you don't need this temporary variable right?

src/app/content/practiceQuestions/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+export const getNextPageWithPracticeQuestions = (
+  nodeId: string, locationFilters: Map<string, LinkedArchiveTreeSection[]>, book: Book | undefined
+) => {
+  if (!book) { return; }
+
+  const node = findArchiveTreeNodeById(book.tree, nodeId);
+
+  if (!node) { return; }
+
+  const flatLocationFilters = flattenLocationFilters(locationFilters);
```
TomWoodward, line 48: doesn't this contain tree sections? can't you look for the index + 1 in here instead of re-flattening the tree and cross referencing them?

TomWoodward (changes requested):

src/app/content/practiceQuestions/utils.ts
PiotrKozlowski: It contain only tree sections that have practice questions so if `node` is a section like `Preface` we will not be able to find id it here

This is why I search for `nodeIndex` in a `tree` and then iterate over all sections after this one and check if any of them is in `locationFilters`

TomWoodward (approved):

PR #910, opened 2020-10-23
Issue 1334

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1334

To see this component just add `<ProgressBar total={5} activeIndex={3} />` somehwere in the modal


src/app/content/practiceQuestions/components/ProgressBar/ProgressBarItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+  isActive: boolean;
+  isDisabled: boolean;
+}
+
+// tslint:disable-next-line: variable-name
+const ProgressBarItem = ({ value, isActive, isDisabled }: ProgressBarItemProps) => (
+  <StyledItemWrapper isActiveOrDisabled={isDisabled || isActive}>
+    <StyledItem isActive={isActive} isDisabled={isDisabled}>
+      {value}
+    </StyledItem>
```
janKozie1, line 77: I think it would be nice to have separate named styled components for each type of item instead of this - i think something like
```js
{isActive 
    ? <ActiveItem>{value}<ActiveItem>
    : isDisabled 
       ? <DisabledItem>{value}</DisabledItem/>
        : <FilledItem>{value}</FilledItem>
 }
```
Makes it easier to read what styles which variant gets.

src/app/content/practiceQuestions/components/ProgressBar/ProgressBarItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+interface ProgressBarItemProps {
+  value: number;
+  isActive: boolean;
+  isDisabled: boolean;
+}
+
+// tslint:disable-next-line: variable-name
+const ProgressBarItem = ({ value, isActive, isDisabled }: ProgressBarItemProps) => (
+  <StyledItemWrapper isActiveOrDisabled={isDisabled || isActive}>
```
janKozie1, line 74: This prop could use a better name. Maybe something like `whiteLine`.

I think you wouldn't need it if you applied my other suggestion, you could use the parent selector on `<FilledItem>` to change the color of the line, although I'm not sure if it's more readable.

janKozie1 (changes requested):

janKozie1 (approved):

PR #912, opened 2020-10-23
Issue 1325

janKozie1:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1325


src/app/content/highlights/hooks/openMyHighlights.ts
```diff
@@ -1,17 +1,41 @@
...
     && select.summaryIsLoading(state) === false;
 
+  dispatch(push({
+    params: currentParams,
+    route: content,
+    state: book && page && !currentLocationState
+      ? getBookPageUrlAndParams(book, page).state
+      : currentLocationState,
+    }, {
+      search: getQueryForParam(modalQueryParameterName, modalUrlName),
```
PiotrKozlowski, line 35: You're not passing the third argument here for `getQueryForParam` so it overwrites other params that might be in the url
Same for SG

src/app/content/studyGuides/hooks/openStudyGuides.ts
```diff
@@ -1,21 +1,39 @@
...
   const defaultFilter = select.defaultLocationFilter(state);
+  const currentParams = assertNotNull(selectContent.contentParams(state), 'Opened modal before location was processed');
   const notLoggedIn = loggedOut(state);
-
-  const book = selectContent.book(state);
+  const currentLocationState = locationState(state);
+  const { book, page } = selectContent.bookAndPage(state);
   const firstChapter = book && findArchiveTreeNode(archiveTreeSectionIsChapter, book.tree);
 
+  services.dispatch(push({
```
PiotrKozlowski, line 27: I feel like we should have navigation util that creates push action for specific modal so we don't have to copy pase this part of a code for every modal

PiotrKozlowski (changes requested):
I think we should also clear the modal query after closing modal - probably with `replaceState` 

Edit. Actually, we probably want to do this with push action to keep the normal history behavior when we implement opening a modal if there is a query in the url

src/app/content/highlights/hooks/closeMyHighlights.ts
```diff
@@ -0,0 +1,22 @@
...
+export const hookBody: ActionHookBody<typeof closeMyHighlights> = ({
+  dispatch, getState,
+}) => () => {
+  const state = getState();
+  const existingQuery = query(state);
+
+  dispatch(push(getContentParams(state), {
+    search: removeParamFromQuery(modalQueryParameterName, existingQuery),
+  }));
+
```
PiotrKozlowski, line 19: newline

PiotrKozlowski (approved):

src/app/content/highlights/reducer.ts
```diff
@@ -41,6 +43,11 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(locationChange): {
       // Noops for locationChange dispatched when search query changes
       if (action.payload.action === 'REPLACE') { return state; }
+
+      const summaryShouldBeOpen =
+        getParamFromQuery(modalQueryParameterName, action.payload.location.search) === modalUrlName
+        && action.payload.action === 'PUSH';
```
TomWoodward, line 49: if modals are causing a location change do you even need this logic to close it? this reducer already closes the dialog on any location change. with the `push` condition in here i'm not sure its even doing anything

src/app/content/highlights/reducer.ts
```diff
@@ -41,6 +43,11 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
     case getType(locationChange): {
       // Noops for locationChange dispatched when search query changes
       if (action.payload.action === 'REPLACE') { return state; }
+
+      const summaryShouldBeOpen =
+        getParamFromQuery(modalQueryParameterName, action.payload.location.search) === modalUrlName
```
TomWoodward, line 48: i would support adding the parsed search query to the payload, we already add it to state, and it would simplify all these reducers

src/app/content/highlights/hooks/closeMyHighlights.ts
```diff
@@ -0,0 +1,21 @@
+import { push } from '../../../navigation/actions';
+import { query } from '../../../navigation/selectors';
```
TomWoodward, line 2: i think the code reads more nicely if you do `* as navigation` or `* as selectNavigation` here, so that in the code its `selectNavigation.query` instead of just `query`

not blocking

src/app/content/highlights/hooks/closeMyHighlights.ts
```diff
@@ -0,0 +1,21 @@
...
+import { removeParamFromQuery } from '../../../navigation/utils';
+import { ActionHookBody } from '../../../types';
+import { actionHook } from '../../../utils';
+import { modalQueryParameterName } from '../../constants';
+import { getContentParams } from '../../utils/urlUtils';
+import { closeMyHighlights } from '../actions';
+
+export const hookBody: ActionHookBody<typeof closeMyHighlights> = ({
```
TomWoodward, line 10: seems like we could have a generic `closeModal` hook body that we can just attach to different actions, i'm pretty sure you can get all the params you need out of the navigation state without any context specific selectors. 

also, if the opening if the modal always triggers a `push`, then we should be able to safely close it with a `pop` without worrying about all this other stuff right?

src/app/content/highlights/hooks/openMyHighlights.ts
```diff
@@ -1,17 +1,29 @@
...
   dispatch, getState,
 }) => async() => {
   const state = getState();
   const authenticated = user(state);
+  const existingQuery = query(state);
+
   const summaryNeedsInitialization = () => select.summaryHighlights(state) === null
     && select.summaryIsLoading(state) === false;
 
+  dispatch(push(getContentParams(state), {
```
TomWoodward, line 23: same idea, i believe you can get all your necessary information out of the navigation state, and could create a canned hook just for opening dialogs. you can have multiple hooks on a given action, or you can compose them manually if you need them to happen in a certain order. this might look something like

```
export const initializeMyHighlightsHook = actionHook(openMyHighlights, hookBody);
export const openMyHighlightsHook = actionHook(openMyHighlights, openModalHook(modalUrlName));
```

src/app/content/studyGuides/hooks/closeStudyGuides.ts
```diff
@@ -0,0 +1,19 @@
...
+import { actionHook } from '../../../utils';
+import { modalQueryParameterName } from '../../constants';
+import { getContentParams } from '../../utils/urlUtils';
+import { closeStudyGuides } from '../actions';
+
+export const hookBody: ActionHookBody<typeof closeStudyGuides> = (services) => () => {
+  const state = services.getState();
+  const existingQuery = query(state);
+
+  services.dispatch(push(getContentParams(state), {
```
TomWoodward, line 14: i checked and apparently the content params are actually not in the navigation state. if you go with the option of just calling `pop` it won't matter, but if you do need to maintain access to the route params so that you can construct a new route without the search, you could consider adding them to the navigation state. i think i only didn't do that because i didn't feel like having them in two places, but this seems like a good use case of wanting them to be in a consistent place

src/app/content/utils/urlUtils.ts
```diff
@@ -40,6 +44,29 @@ export const getUrlParamsForBook = (
   }
 };
 
+export const getContentParams = (appState: AppState) => {
```
TomWoodward, line 47: yea if we're deriving these on the fly it'd be nicer to just save the ones that got dispatched in the first place. again, you might not need to do that if you use `pop`, you may end up just removing this

TomWoodward (changes requested):

src/app/content/highlights/reducer.ts
janKozie1: This is for preventing them from closing on location change since it hits the reducer after `open<Modal>`

src/app/content/highlights/hooks/openMyHighlights.ts
janKozie1: I need the location state too since based on it highlights are cleared on location change. And since the location state is missing when navigated to a page via a link, I'm trying to generate it as normal links do.

src/app/content/highlights/hooks/index.ts
```diff
@@ -23,6 +26,8 @@ export default [
   openMyHighlightsHook,
   printHighlightsHook,
   loadMoreHook,
+  actionHook(closeMyHighlights, closeModal),
```
TomWoodward, line 29: if the url were to be replaced with a different modal name, you'd still want to close it, so i think this needs to work more similarly to the open one

src/app/content/highlights/hooks/sharedHooks.ts
```diff
@@ -0,0 +1,17 @@
+import { push } from '../../../navigation/actions';
+import * as navigation from '../../../navigation/selectors';
+import { getQueryForParam } from '../../../navigation/utils';
+import { AppServices, MiddlewareAPI } from '../../../types';
+import { modalQueryParameterName } from '../../constants';
+import { getContentParams } from '../../utils/urlUtils';
+
+export const openModal = (modalUrlName: string) => (services: MiddlewareAPI & AppServices) => () => {
```
TomWoodward, line 8: i don't think this has to be in `highlights` i think we can put it in `navigation`.

i see that you're using some pretty specific logic in your dispatch, but i think that could be replaced with generic logic if we normalize all the route params in the navigation state you could simply use those and replace the modal parameter

src/app/content/highlights/reducer.ts
TomWoodward: i don't understand, are you saying that there is something that is undoing a previous acton if you don't have this check here?

TomWoodward (changes requested):

src/app/content/highlights/hooks/index.ts
PiotrKozlowski: Since opening a new modal would dispatch `locationChange`, I think closing the current modal would be handled by the reducer:
```ts
      const summaryShouldBeOpen =
        getParamFromQuery(modalQueryParameterName, action.payload.location.search) === modalUrlName
        && action.payload.action === 'PUSH';

      return {...state, summary: {...state.summary, open: summaryShouldBeOpen}};
    }
```

src/app/content/highlights/reducer.ts
PiotrKozlowski: I think he meant that `openMyHighlights` is opening a modal, and then there is a `openModal` hook attached to this action that calls dispatches `push` action and then this `push` action is reduced in here as `locationChange` so we check if we should or should not close the modal

src/app/content/highlights/hooks/index.ts
PiotrKozlowski: However current implementation calls `history.goBack` when `closeModal` is dispatched and I think this will not work when we will implement adding a filters to the url - unless we will update the url with `replace` action

src/app/content/highlights/hooks/sharedHooks.ts
PiotrKozlowski: I've moved it to the `navigation/hooks` and created separate files for `openModal` and `closeModal` hooks

> i see that you're using some pretty specific logic in your dispatch, but i think that could be replaced with generic logic if we normalize all the route params in the navigation state you could simply use those and replace the modal parameter

Do you mean to add an action like `pushParam('modal', 'MH')` that would be transformed into proper history action or do you mean to change how we deal with url params entirly so we can dispatch an action that changes the single param and associated hook would update the url?

TomWoodward: i mean that when navigation events are pushed they have parameters that are embedded/extracted from the url, those parameters aren't currently stored in the navigation state, they are just available for whoever needs them to pick them up.

currently you are deriving your new prams by re-computing your content params from the state, but what you really want is  "whatever my current params are but without the modal part of the query"

we could accomplish this in a generic way by making the navigation reducer add route params to navigation state as it does with other location parameters like query. then, instead of re-computing your parameters from the content state, which is content specific logic, you could use generic logic, of just taking what is in your navigation state and removing the modal part of the query

src/app/content/highlights/reducer.ts
TomWoodward: why would we need the push filter? in that case the location search would include this modal name and this reducer would set `summaryShouldBeOpen` to `true`, that can be the thing that actually opens it, we don't need to reduce `openMyHighlights` (this brings us to supporting links and the forward button that open the dialogs)

src/app/content/highlights/reducer.ts
TomWoodward: this is still outstanding

src/app/content/highlights/hooks/index.ts
TomWoodward: i'm not sure what i was talking about here, i think i wanted to wrap the goBack in a check to see if the correct modal was currently in the url, but i'm not sure when it would matter

TomWoodward (changes requested):

src/app/content/highlights/reducer.ts
PiotrKozlowski: I'll remove reducing `openMyHighlights` and `closeMyHighlights` etc.
We filter here for `PUSH` because we wanted to handle opening a page with `modal` param (and support for forward button) on a separate card (it requires us to do more changes like: initialize load summary, handle modal=SG for book that does not have them, send GA events about opening modals, etc.)

PiotrKozlowski: Actually removing reducer for `openMyHighlights` / `PQ`, `SG` breaks like 20+ tests. These are tests like: open mh, check if modal is rendered - we would have to replace the calls with locationChange dispatch, but there are also tests like "click print highlights > start loading all of them > close modal while they are loading > make sure that we do not print them"
If we remove the reducer for `closeMyHighlights` then we would have to call `locationChange` without modal param instead of calling a close modal action because unit test will not know how to handle `.goBack` and also `createTestStore` does not implement middlewares do hooks are not supported by default.

I'm not sure if all of these changes are in the scope of this issue

src/app/content/highlights/hooks/index.ts
PiotrKozlowski: I'm not sure if we should use `goBack` here at all. Consider the case like changing the url when navigating through the questions:
1. Open PQ / modal=PQ
2. Go to first question / modal=PQ&question=1
3. Got next question / modal=PQ&question=2
4. Close the modal / .goBack will go to the question=1

src/app/navigation/hooks/openModalHook.ts
```diff
@@ -0,0 +1,18 @@
...
+import { AnyRoute, MatchWithParams } from '../types';
+import { getQueryForParam } from '../utils';
+
+export const openModal = (modalUrlName: string) => (services: MiddlewareAPI & AppServices) => () => {
+  const state = services.getState();
+  const existingQuery = navigation.query(state);
+  const match = navigation.match(state);
+
+  if (match && 'params' in match) {
+    services.dispatch(push(match as MatchWithParams<Omit<AnyRoute, 'undefined'>>, {
```
TomWoodward, line 14: would this type be less messed up if you separated location, match, and query in the state?

src/app/navigation/types.ts
```diff
@@ -5,6 +5,7 @@ import { routes } from '../';
 import { AnyAction, AppServices, MiddlewareAPI } from '../types';
 
 export type State = Location & {
+  match?: GenericMatch
```
TomWoodward, line 8: it might help to type this as `AnyMatch`, the two are slightly different

TomWoodward (changes requested):

src/app/navigation/actions.ts
```diff
@@ -1,8 +1,15 @@
+import queryString from 'query-string';
 import { createStandardAction } from 'typesafe-actions';
 import { AnyHistoryAction, AnyMatch, HistoryAction, LocationChange } from './types';
 
 export const callHistoryMethod = createStandardAction('Navigation/callHistoryMethod')<AnyHistoryAction>();
-export const locationChange = createStandardAction('Navigation/locationChange')<LocationChange>();
+export const locationChange = createStandardAction('Navigation/locationChange').map(
```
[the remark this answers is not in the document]
PiotrKozlowski: it is dispatched in 20 files, 41 times - mostly test cases
Should I update it everywhere instead of using this `.map`?

src/app/navigation/hooks/openModalHook.ts
PiotrKozlowski: typing it as `AnyMatch` worked fine

src/app/navigation/types.ts
PiotrKozlowski: This would require chaning more types since we are obtaing `match` from
```ts
export interface LocationChange<M = GenericMatch> {
  location: Location;
  match?: M;
  action: Action;
}
```

src/app/content/highlights/hooks/index.ts
PiotrKozlowski: Currently, we don't support it. I guess we can go back to this issue when we will implement this

src/app/content/highlights/reducer.ts
PiotrKozlowski: We can't because we are now opening a modal using `dispatch(openMyHighlights)` which is setting `open: true` and then there is a `openModalHook` that dispatches `push`/`locationChange` to update the history state and url. 

I think the proper solution would be to remove reducers for `openMyHighlights` and `closeMyHighlights` actions and handle opening modal using only `locationChange` action as you suggested, but this generates a tons of errors and I think we should do this in another card

src/app/navigation/actions.ts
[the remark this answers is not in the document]
TomWoodward: only one dispatch is outside of test files, but since the query and search are basically the same data, i suppose it seems a little weird to require passing both into the action. lets leave it the way you have it

src/app/content/highlights/reducer.ts
TomWoodward: ah so the purpose of this logic is really just to ignore the location change triggered by the open modal action where it adds the query param to the url. we would prefer to have that location change open the dialog, but we're deferring that to another card.

which card will that be done in?

TomWoodward (changes requested):

src/app/navigation/guards.ts
```diff
@@ -9,3 +10,7 @@ export const hasState = (payload: any & {state?: object}): payload is {state: ob
 
 export const pathTokenIsKey = (token: Token): token is Key =>
   isObject(token);
+
+export const isMatchWithParams = (payload: GenericMatch): payload is MatchWithParams<Omit<AnyRoute, 'undefined'>> => {
```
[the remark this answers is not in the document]
PiotrKozlowski: If I used `hasParams` then I'm getting this error after running `yarn start`
```
TypeScript error in /home/piotr/Dokumenty/github/rex-web/src/app/navigation/reducer.ts(18,67):
Type '(state: State | undefined, action: PayloadAction<"app/receiveFeatureFlags", string[]> | PayloadAction<"app/receivePageFocus", boolean> | PayloadAction<"Auth/receiveUser", User> | ... 61 more ... | PayloadAction<...>) => State | { ...; }' is not assignable to type 'Reducer<State, PayloadAction<"app/receiveFeatureFlags", string[]> | PayloadAction<"app/receivePageFocus", boolean> | PayloadAction<"Auth/receiveUser", User> | EmptyAction<"Auth/receiveLoggedOut"> | ... 60 more ... | PayloadAction<...>>'.
  Type 'State | { match: (MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }) | undefined; query: OutputParams; pathname: string; search: string; state: any; hash: string; key?: string | undefined; }' is not assignable to type 'State'.
    Type '{ match: (MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }) | undefined; query: OutputParams; pathname: string; search: string; state: any; hash: string; key?: string | undefined; }' is not assignable to type 'State'.
      Type '{ match: (MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }) | undefined; query: OutputParams; pathname: string; search: string; state: any; hash: string; key?: string | undefined; }' is not assignable to type '{ match?: MatchWithParams<Route<undefined, undefined> | Route<Params, State>> | undefined; query: OutputParams; }'.
        Types of property 'match' are incompatible.
          Type '(MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }) | undefined' is not assignable to type 'MatchWithParams<Route<undefined, undefined> | Route<Params, State>> | undefined'.
            Type 'MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }' is not assignable to type 'MatchWithParams<Route<undefined, undefined> | Route<Params, State>> | undefined'.
              Type 'MatchWithoutParams<Route<undefined, undefined> | Route<Params, State>> & { params: object; }' is not assignable to type 'MatchWithParams<Route<undefined, undefined> | Route<Params, State>>'.
                Types of property 'params' are incompatible.
                  Type 'object' is not assignable to type 'Params | undefined'.
                    Type '{}' is missing the following properties from type 'Params': book, page  TS2322

    16 | });
    17 | 
  > 18 | export default (location: Location): Reducer<State, AnyAction> => (state = addQuery(location), action) => {

```

TomWoodward:
@PiotrKozlowski i pushed a commit trying to simplify the navigation types so that the guards work better, let me know what you think

src/app/content/highlights/reducer.ts
PiotrKozlowski: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1389

PiotrKozlowski:
@TomWoodward I think it looks okay

src/app/content/highlights/reducer.ts
TomWoodward: i added a note to that card about removing the action type filtering

TomWoodward (approved):

PR #916, opened 2020-10-23
FE selenium test: Chapter filter collapses on opening color filter

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1365

1. https://openstax.testrail.com/index.php?/cases/view/593145
2. Added webdriver waits in Accounts.py since the tests were failing in my local machine consistently while trying to identify the new student workflow


pytest-selenium/pages/accounts.py
```diff
@@ -9,6 +9,8 @@
 from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.support import expected_conditions as expect
 from selenium.webdriver.support.ui import Select
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as expected
```
gregfitch, line 13: These should be merged: the two EC should be `expect` or `expected` but not both and the other should be `import Select, WebDriverWait`

pytest-selenium/pages/accounts.py
```diff
@@ -147,6 +149,8 @@ def _use_new_accounts_flow(self) -> bool:
         :rtype: bool
 
         """
+        WebDriverWait(self.driver, 1).until(
```
gregfitch, line 152: PyPOM Pages should have a WDW already limiting the need for a new instance as `self.wait.until()`

gregfitch: Obviously this wouldn't apply if you're looking for a specific limit on the wait time.

pytest-selenium/pages/accounts.py
gregfitch: If you use PyPOM's wait then you can remove `WebDriverWait`

gregfitch (changes requested):
The Expected Conditions imports should be consolidated. The other comments are suggestions.

gregfitch (approved):

PR #936, opened 2020-11-02
Issue 1333

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1333 

To test it go to the page that has some questions, for example: `/books/physics/pages/1-1-physics-definitions-and-applications`


src/app/content/practiceQuestions/components/IntroScreen.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import styled from 'styled-components/macro';
+import { nextQuestion } from '../actions';
+import * as pqSelectors from '../selectors';
+import PQButton from './PQButton';
+
+// tslint:disable-next-line: variable-name
+const IntroScreenWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
```
KarinaMendez2, line 12: I think if you add `display: block` to `IntroScreenMessage `, you could get away with just putting `width: 100%; and text-align: center;` to make it simpler

src/app/content/practiceQuestions/components/IntroScreen.tsx
```diff
@@ -0,0 +1,48 @@
...
+  width: 100%;
+
+  ${PQButton} {
+    margin: 2rem 0;
+  }
+`;
+
+// tslint:disable-next-line: variable-name
+const IntroScreenMessage = styled.span`
+  font-family: inherit;
```
KarinaMendez2, line 24: what about `textRegularSize` or something already defined?

src/app/content/practiceQuestions/components/IntroScreen.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+import createTestStore from '../../../../test/createTestStore';
+import * as Services from '../../../context/Services';
+import MessageProvider from '../../../MessageProvider';
+import { Store } from '../../../types';
+import { nextQuestion, setQuestions } from '../actions';
+import { PracticeQuestions } from '../types';
+import IntroScreen from './IntroScreen';
+import PQButton from './PQButton';
+
+describe('IntroScreen', () => {
```
KarinaMendez2, line 14: can you add `practice questions` to this description?

src/app/content/practiceQuestions/components/PQButton.tsx
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components/macro';
+import { PlainButton } from '../../../components/Button';
+import theme from '../../../theme';
+
+// tslint:disable-next-line: variable-name
+const PQButton = styled(PlainButton)`
+  background-color: ${theme.color.primary.orange.base};
+  color: ${theme.color.text.white};
+  font-size: 1.6rem;
```
KarinaMendez2, line 9: same here about `textRegularSize`

src/app/content/practiceQuestions/components/ShowPracticeQuestions.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+    ) as LinkedArchiveTreeSection;
+    jest.spyOn(content, 'getUrl')
+      .mockReturnValue('mockedUrl');
+  });
+
+  it('renders only a few components when there is no book and no selected section', () => {
+    const component = renderer.create(render());
+
+    expect(() => component.root.findByType(SectionTitle)).toThrow();
+    expect(() => component.root.findByType(ContentLink)).toThrow();
```
KarinaMendez2, line 57: it'd be good to expect what you're actually expecting it to throw in these tests if they're throwing something

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -1,24 +1,121 @@
...
+  padding: 2rem 3.2rem 2.5rem 3.2rem;
   ${theme.breakpoints.mobile(css`
     text-align: left;
     padding: 0;
   `)}
 `;
 
+// tslint:disable-next-line: variable-name
+export const SectionTitle = styled.h2`
+  font-size: 1.8rem;
```
KarinaMendez2, line 29: this is all `h4Style`

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -1,24 +1,121 @@
...
+      {section ? <SectionTitle dangerouslySetInnerHTML={{ __html: section.title }} /> : null}
+      <QuestionsWrapper>
+        <QuestionsHeader>
+          <FormattedMessage id='i18n:practice-questions:popup:questions'>
+            {(msg: string) => msg}
+          </FormattedMessage>
+        </QuestionsHeader>
+        <ProgressBar total={questionsCount} activeIndex={currentQuestionIndex} />
+        {
+          selectedSectionHasPracticeQuestions && currentQuestionIndex === null
```
KarinaMendez2, line 104: can't this be `!currentQuestionIndex `?

KarinaMendez2 (changes requested):

src/app/content/practiceQuestions/components/ShowPracticeQuestions.spec.tsx
PiotrKozlowski: I guess we could always assert the text but we're already using this syntax to make sure that the element just does not exist in the component

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: no, because `currentQuestionIndex` may be equal to `0` and `!0` is `true`

KarinaMendez2 (approved):

PR #941, opened 2020-11-12
Issue 1332

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1332


src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -102,11 +119,106 @@ const ChapterFilter = ({
...
+
+// tslint:disable-next-line: variable-name
+const ChapterFilterWithToggling = (props: ChapterFilterWithTogglingProps) => {
+  const [isOpenChapterId, setIsOpenChapterId] = React.useState<string | null>(
+    props.selectedSection ? stripIdVersion(props.selectedSection.parent.id) : null
+  );
+
+  return <div className={props.className} tabIndex={-1}>
+    <Row>
+      <Column>
```
TomWoodward, line 180: one column? is that intentional in the design? there was a reason we had to add the column grouping before for large books

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -102,11 +119,106 @@ const ChapterFilter = ({
...
+        {Array.from(props.locationFilters.entries()).map(([chapterId, data]) => {
+          return <StyledDetails key={chapterId} open={isOpenChapterId === chapterId}>
+            <StyledSummary onClick={(e: React.MouseEvent) => {
+              e.preventDefault();
+              setIsOpenChapterId((current) => current === chapterId ? null : chapterId);
+            }}>
+              <ChapterTitle dangerouslySetInnerHTML={{__html: data.chapter.title}} />
+              <DownIcon rotate={isOpenChapterId === chapterId} />
+            </StyledSummary>
+            <div>
```
TomWoodward, line 190: i still think we should make all this one component where the sub-selections are optional, if there are sub-selections its a dropdown, if not you select the top level element. is there a problem with that approach that i'm not seeing?

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -96,30 +107,33 @@ const ShowPracticeQuestions = () => {
...
+              ? <IntroScreen />
+              : null
+          }
+        </QuestionsWrapper>
         {
-          selectedSectionHasPracticeQuestions && currentQuestionIndex === null
-            ? <IntroScreen />
+          section && linkToTheSection
+            ? <StyledContentLink href={linkToTheSection} target='_blank' data-analytics-label='Read' >
+              <FormattedMessage id='i18n:practice-questions:popup:read'>
```
TomWoodward, line 129: whitespace is wrong. also, this component is becoming overloaded. there is too much data being selected and too many conditions in the rendering. you could make a separate component for this link that selects `linkToTheSection` and internally returns `null` when required.

`IntroScreen` probably should stay the way you have it because we'll have to swap it out for the actual questions so it won't be self contained logic

`section` i imagine will always be set, when could it not be?

src/app/content/practiceQuestions/utils.ts
```diff
@@ -21,23 +22,26 @@ export const getPracticeQuestionsLocationFilters = (
   for (const node of tree) {
     if (isLinkedArchiveTreeSection(node) && summary.countsPerSource[stripIdVersion(node.id)]) {
       const parentId = stripIdVersion(node.parent.id);
-      locationFilters.set(parentId, [...locationFilters.get(parentId) || [], node]);
+      const currentSections = locationFilters.has(parentId)
+        ? assertDefined(locationFilters.get(parentId), 'node has to be there').sections
```
TomWoodward, line 26: i agree that it is silly that typescript can't figure out the `has/get` relationship of maps, but since it can't i've been trying to write code more like

```typescript
const parentFilters = locationFilters.get(parentId);
const currentSections = parentFilters ? parentFilters.sections : [];
```

TomWoodward (changes requested):

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -23,6 +28,7 @@ const Row = styled.div`
 const Column = styled.div`
   display: flex;
   flex-direction: column;
+  overflow-x: hidden;
```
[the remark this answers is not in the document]
PiotrKozlowski: I couldn't find a way to get rid out of this:
image

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -1,13 +1,18 @@
...
 import styled, { css } from 'styled-components/macro';
 import AllOrNone from '../../../components/AllOrNone';
+import { PlainButton } from '../../../components/Button';
 import Checkbox from '../../../components/Checkbox';
 import { textStyle } from '../../../components/Typography/base';
 import { match, not } from '../../../fpUtils';
 import theme from '../../../theme';
 import ColorIndicator from '../../highlights/components/ColorIndicator';
 import { HighlightLocationFilters, SummaryFilters } from '../../highlights/types';
+import { PracticeQuestionsLocationFilters } from '../../practiceQuestions/types';
```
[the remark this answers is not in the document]
PiotrKozlowski: It kind of does, but we also have `HighlightLocationFilters, SummaryFilters` in here
I think I'll jsut create proper interfaces in this component

src/app/content/components/popUp/ChapterFilter.tsx
PiotrKozlowski: I think it is scrollable with one column and only one chapter can be expanded
https://projects.invisionapp.com/d/#/console/20557970/433309549/preview

src/app/content/components/popUp/ChapterFilter.tsx
PiotrKozlowski: Dealing with different types of `props` with the same names is tricky, but I think I can try to unify them

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: We are setting the `section` when we are opening a modal and we are resting it when we close it / on `locationChange` so if you open the modal short after navigating to the other page it will still be a `null`.

It is done this way because we don't want to remember user choices when he closed and then reopened the modal. But maybe we don't have to clear this data and then we can set selected section on `locationChange`?

src/app/content/components/popUp/Filters.tsx
```diff
@@ -11,12 +11,21 @@ import { disablePrint } from '../utils/disablePrint';
...
   width: ${filters.dropdownToggle.icon.width}rem;
   height: ${filters.dropdownToggle.icon.height}rem;
   margin-left: 0.8rem;
   padding-top: 0.2rem;
+  ${(props: { rotate: boolean }) => {
+    if (props.rotate) {
+      return `
+        transform: rotate(180deg);
+        padding-top: 0;
+        padding-bottom: 0.2rem;
```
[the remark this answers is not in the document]
PiotrKozlowski: It is moved from the lines 52 - 56 from this file:
```
      ${DownIcon} {
        transform: rotate(180deg);
        padding-top: 0;
        padding-bottom: 0.2rem;
      }
```
This was done this way to match design alignment when arrow is rotated

src/app/content/components/popUp/Filters.tsx
```diff
@@ -11,12 +11,21 @@ import { disablePrint } from '../utils/disablePrint';
 import FiltersList from './FiltersList';
 
 // tslint:disable-next-line:variable-name
-const DownIcon = styled(AngleDown)`
+export const DownIcon = styled(AngleDown)`
```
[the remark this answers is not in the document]
PiotrKozlowski: I've changed the name and added support for `up | down` for now

PiotrKozlowski:
@TomWoodward Please look at the updated code. If it looks okay to you then I will update the tests

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
TomWoodward: are you talking about the orange bars? what is that horizontal line at the top of the dropdown it looks like its a few px out of place

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
TomWoodward: i think we should try to align the two filters interfaces into one data format that has optional children, that way PQ and MH can use the same filter format

src/app/content/components/popUp/Filters.tsx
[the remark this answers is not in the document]
TomWoodward: ah ok i guess

src/app/content/components/popUp/ChapterFilter.tsx
TomWoodward: hmm... i guess it makes sense because you're gonna have height issues if you expand something in one column and not the other. i would really like to combine these two designs.

maybe if there is any expand/collapse then it uses one column but if there are no entries with children it does the multiple columns

it'll also need a prop for "multiselect" that shows the checkboxes vs the buttons

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
TomWoodward: perhaps it could be organized in such a way that the section is set after clicking the button but before the modal is shown? that way we know we can display the filters immediately

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
PiotrKozlowski: about the bottom one
I couldn't find why it appears

> it looks like its a few px out of place

I think this is because of how Firefox displays the scroll bars - when you hover on them they are getting wider and fill the rest of the space

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
PiotrKozlowski: I've used this type:
```ts
export type LocationFilters = Map<
  string,
  { section: LinkedArchiveTree | LinkedArchiveTreeSection, children?: LinkedArchiveTreeSection[] }
>;
```

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: I've moved this component to `<LinkToSection section={section} />`

and I've moved `setSelectedSection` action to the `practiceQuestionButton` component and removed `practiceQuestionOpenHook`

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
TomWoodward: the second thing i said was asking about the grey line at the top of the dropdown, it should line up with the top right? it looks like its 2px down

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
TomWoodward: looks good

src/app/content/components/popUp/Filters.tsx
```diff
@@ -73,13 +76,19 @@ const Toggle = styled(React.forwardRef<HTMLButtonElement, ToggleProps>(
   }
 `;
 
+interface FilterDropdownProps extends ControlledProps {
```
TomWoodward, line 79: pretty sure there is a helper to get the prop types out of a component so you could extend that instead of exporting the props interface separately

src/app/content/practiceQuestions/components/LinkToSection.tsx
```diff
@@ -0,0 +1,54 @@
...
+const LinkToSection = ({ section }: LinkToSectionProps) => {
+  const book = useSelector(contentSelectors.book);
+
+  const linkToTheSection = React.useMemo(() => {
+    return book && section ? getBookPageUrlAndParams(book, section).url : null;
+  }, [book, section]);
+
+  if (!section || !linkToTheSection) { return null; }
+
+  return <StyledContentLink href={linkToTheSection} target='_blank' data-analytics-label='Go to link' >
```
TomWoodward, line 46: why aren't you using the actual `ContentLink` component?

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -101,35 +88,29 @@ const ShowPracticeQuestions = () => {
...
-            <FormattedMessage id='i18n:practice-questions:popup:read'>
-              {(msg: string) => msg}
-            </FormattedMessage>
-            <span dangerouslySetInnerHTML={{ __html: section.title }} />
-          </StyledContentLink>
-          : null
-      }
+      <Filters />
+      <ShowPracitceQuestionsContent>
+        {section ? <SectionTitle dangerouslySetInnerHTML={{ __html: section.title }} /> : null}
```
TomWoodward, line 93: can we remove this condition now?

src/app/content/practiceQuestions/types.ts
```diff
@@ -30,3 +30,8 @@ export interface PracticeQuestion {
 }
 
 export type PracticeQuestions = PracticeQuestion[];
+
+export type PracticeQuestionsLocationFilters = Map<
```
TomWoodward, line 34: this shouldn't be used now right?

TomWoodward (changes requested):

src/app/content/practiceQuestions/types.ts
PiotrKozlowski: I'm using this in `getPracticeQuestionsLocationFilters`
The difference between this and `LocationFilters` is that a `section` has to be `LinkedArchiveTree` and it has to have `children`

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: `pqSelectors.selectedSection` returns either `null` or `LinkedArchiveTreeSection` because `selectedSection` is initially set to `null` so we could only get rid of it by creating a separate component and returning `null` internally

src/app/content/practiceQuestions/components/LinkToSection.tsx
PiotrKozlowski: We want to open this link in a new tab and `ContentLink` component doesn't handle that - I think it is because it calls `e.preventDefault` `onClick`

src/app/content/components/popUp/Filters.tsx
PiotrKozlowski: If `Dropdown` component would be a normal component then we could use `React.ComponentProps<typeof Dropdown>` but it is a `styled` component so this basically becomes "accept anything that normal HTML element can"

I've added
```ts
type TabTransparentDropdownProps = CommonDropdownProps & Props;
export type TabHiddenDropdownProps = CommonDropdownProps & (Props | Props & ControlledProps);
```
to `Dropdown.tsx` and used `TabHiddenDropdownProps` for this

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
```diff
@@ -26,15 +29,20 @@ const PracticeQuestionsText = styled.span`
...
 
   if (!isEnabled || !hasPracticeQs) { return null; }
 
   const openPracticeQuestionsSummary = () => {
-    dispatch(openPracticeQuestions());
-    trackOpenClose();
+    const section = book && page ? findArchiveTreeNodeById(book.tree, page.id) : null;
+    if (section && isLinkedArchiveTreeSection(section)) {
+      dispatch(setSelectedSection(section));
+      dispatch(openPracticeQuestions());
```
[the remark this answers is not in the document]
PiotrKozlowski: we would still need `setSelectedSection` for chapter filtering and we also do not pass anything for actions that are opening other modals
I think it doesn't really matter a lot because we will soon implement opening modals on location change anyway

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
TomWoodward: could we assert that it isn't null? the expectation is that it will always be set before the modal opens right?

src/app/content/practiceQuestions/components/LinkToSection.tsx
TomWoodward: but it has all this logic about the url handling already, if necessary we could add a prop for target _blank

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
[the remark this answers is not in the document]
TomWoodward: > I think it doesn't really matter a lot because we will soon implement opening modals on location change anyway

is a good argument for going back to the hook you deleted, if the modal was opened by a location change we'd need to select the default section somehow

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -51,62 +54,198 @@ const chunk = <T extends any>(sections: T[]) => {
   return [sections.slice(0, cutoff), sections.slice(cutoff)].filter((arr) => arr.length > 0);
 };
 
-interface Props {
+export type LocationFilters = Map<
```
TomWoodward, line 57: this is probably used enough that it belongs in a types file

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -51,62 +54,198 @@ const chunk = <T extends any>(sections: T[]) => {
...
-const ChapterFilter = ({
-  className,
-  disabled,
-  locationFilters,
-  locationFiltersWithContent,
-  selectedLocationFilters,
-  setFilters,
-}: Props) => {
+const ChapterFilter = (props: ChapterFilterProps) => {
+  const [isOpenChapterId, setIsOpenChapterId] = React.useState<string | null>(null);
```
TomWoodward, line 74: should `isOpenChapterId` be `openChapterId`, its not like a boolean or something right?

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -51,62 +54,198 @@ const chunk = <T extends any>(sections: T[]) => {
...
+  }
+`;
+
 export default styled(ChapterFilter)`
   ${textStyle}
   background: ${theme.color.white};
   font-size: 1.4rem;
-  padding: ${filters.dropdownContent.padding.topBottom}rem ${filters.dropdownContent.padding.sides}rem;
+  ${(props: ChapterFilterProps) => {
+    if ('locationFiltersWithContent' in props) {
```
TomWoodward, line 240: isn't this required?

src/app/content/highlights/types.ts
```diff
@@ -46,7 +47,7 @@ export interface State {
   };
 }
 
-export type HighlightLocationFilters = Map<string, LinkedArchiveTree | LinkedArchiveTreeSection>;
+export type HighlightLocationFilters = LocationFilters;
```
TomWoodward, line 50: i don't think we should need a `HighlightLocationFilters` defined, we can use use `LocationFilters`

src/app/content/practiceQuestions/components/Filters.tsx
```diff
@@ -0,0 +1,52 @@
...
+    const clickedSectionId = locationIds.pop();
+    if (!clickedSectionId) {
+      // user clicked on the selected section
+      setOpen(false);
+      return;
+    }
+    const search = book && locationIds.length
+      ? findArchiveTreeNodeById(book.tree, clickedSectionId)
+      : null;
+    const section = search && isLinkedArchiveTreeSection(search) ? search : null;
```
TomWoodward, line 27: you're doing a lot of work to find the section here, maybe we could change the I/O to work on the sections instead of just the ids? it would be more flexible

src/app/content/practiceQuestions/types.ts
TomWoodward: is that necessary? or can you just use `LocationFilters` everywhere?

TomWoodward (changes requested):

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: I think that handling `null` is safer than asserting - especially since we want to handle opening modals after entering a URL with `modal` param
An alternative that I can see would be to create a component for `ShowPracitceQuestionsContent` and pass there `section` as prop and return `null` internally if it is falsy

src/app/content/practiceQuestions/types.ts
PiotrKozlowski: I could use `LocationFilters` but then I would have to assert that `section` is `LinkedArchiveTree` and `children` is always present in a few places. I probably could create a separate type for this kind of filters

src/app/content/highlights/types.ts
PiotrKozlowski: I didn't want to produce the additional diff because of changing these names, but idc, it wasn't that much of the work

src/app/content/practiceQuestions/components/Filters.tsx
PiotrKozlowski: I've started working on it and it was a rabbit hole... :D I'm almost done with it, but it required a lot of changes

PiotrKozlowski:
@TomWoodward Please take a look at the rework I did for `ChapterFilter` and `ColorFilter`. I've also applied other requested changes.
If this looks okay then I will have to update a bunch of tests

jarosik10:
There are some outline problems with the focused details elements. These outlines are not visible in Firefox. It makes it hard to navigate using keyboard.

jarosik10:
Not sure if it is wrong, but when the dropdown is opened Escape exits the modal instead of closing the dropdown. In my opinion it seems inconvenient to use.

src/app/content/studyGuides/selectors.spec.ts
```diff
@@ -13,11 +14,12 @@ jest.mock('../selectors', () => ({
 describe('studyGuidesLocationFiltersWithContent', () => {
   it('filters', () => {
     mockBook.mockReturnValue({id: 'enabledbook', tree: treeWithoutUnits});
+    const section = findArchiveTreeNodeById(treeWithoutUnits, 'chapter1')!;
```
jarosik10, line 17: What is the purpose of using exclamation mark here?

PiotrKozlowski: `findArchiveTreeNodeById` is returning `LinkedArchiveTreeNode | undefined`
`!` is a typescript feature that exclude `undefined` type so we are saying to the typescript that in this case the `section` is for sure `LinkedArchiveTreeNode`

We are using this only in tests if we know for sure that a statement is true.
In the production code you can see that we are using `assertDefined` helper that will throw an error with a proper message if something is not right

src/app/content/components/popUp/ChapterFilter.spec.tsx
```diff
@@ -42,12 +43,12 @@ describe('ChapterFilter', () => {
...
-      assertDefined(findArchiveTreeNodeById(book.tree, 'testbook1-testpage1-uuid'), ''),
+      { section: assertDefined(findArchiveTreeNodeById(book.tree, 'testbook1-testpage1-uuid'), '') },
     ]])));
 
     const component = renderer.create(<Provider store={store}>
       <MessageProvider>
-        <ConnectedChapterFilter />
+        <ConnectedChapterFilter multiselect={true} />
       </MessageProvider>
     </Provider>);
```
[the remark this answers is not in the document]
PiotrKozlowski: I think that we are not using `beforEach` to create component in any other places so I would rather not implement a new design pattern, however, we tend to create a `render` function in before each and then call it in the tests as `const component =  renderer.create(render())`

Besides that, I would rather not update the existing code if it is really not necessary to do not produce a noise diff. This pr is already a big one

src/app/components/Dropdown.tsx
```diff
@@ -52,20 +52,30 @@ const visuallyHidden = css`
...
 // tslint:disable-next-line:variable-name
-const TabHiddenDropDown = styled(({toggle, children, className, onToggle}: Props) => {
-  const [open, setOpen] = React.useState<boolean>(false);
+const TabHiddenDropDown = styled((
+  {toggle, children, className, onToggle, ...props}: React.PropsWithChildren<Props | Props & ControlledProps>
+) => {
+  const { open: controlledOpen, setOpen: controlledSetOpen } = props as Props & ControlledProps;
+  const [open, setOpenState] = React.useState<boolean>(false);
+  const isOpen = controlledOpen !== undefined ? controlledOpen : open;
+  const setOpen = controlledSetOpen !== undefined ? controlledSetOpen : setOpenState;
```
jarosik10, line 73: Seems complicated for a logic of opening and closing the `Dropdown`.
Is `const [open, setOpenState] = React.useState<boolean>(false);` really necessary?
And the naming creates a confusion. One booleans have `is` prefix, but other don't.

PiotrKozlowski: Previously the state of this component was handled only internally, but know we need to be able to control the open/close state for this component from outside of the component and keep the internal solution for previous situations - this makes this component a controlled and uncontrolled at the same time

For simplicity of the whole component I've decided to use everywhere  `isOpen` and `setOpen` variables and lines 70 - 74 are making sure that these variables point to the proper state and set state handlers

PiotrKozlowski:
unit tests pass locally and here failed with:
```bash
FAIL src/gateways/googleAnalyticsClient.spec.ts
  ● GoogleAnalyticsClient › trackPageView › called before tracking ID set › saves the commands and sends them after tracking IDs set

    expect(received).toBeGreaterThanOrEqual(expected)

    Expected: >= 5
    Received:    4

      118 |         client.setTrackingIds(['foo']);
      119 |         expect(mockGa).toHaveBeenCalledWith('tfoo.set', 'queueTime', expect.any(Number));
    > 120 |         expect(mockGa.mock.calls[1][2]).toBeGreaterThanOrEqual(sleepMs);
          |                                         ^
      121 |         expect(mockGa).toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/some/path'});
      122 | 
      123 |         expect(client.getPendingCommands().length).toBe(0);

      at Object.toBeGreaterThanOrEqual (src/gateways/googleAnalyticsClient.spec.ts:120:41)
```

@TomWoodward This test is usually passing, but I'm not really sure why there is `await sleep(sleepMs);` call in it
Can we just remove it? It seems to work without it

src/app/content/components/popUp/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  string,
+  { section: LinkedArchiveTreeNode, children?: LinkedArchiveTreeSection[] }
+>;
+
+export type LocationFiltersWithChildren = Map<
+  string,
+  { section: LinkedArchiveTree, children: LinkedArchiveTreeSection[] }
+>;
+
+export interface FiltersChange<T> {
```
TomWoodward, line 13: i didn't catch why you changed the way the filters work to include the add/remove lists, was there some kind of race condition?

TomWoodward (approved):

src/app/content/components/popUp/types.ts
PiotrKozlowski: I don't remember exactly (last commit 1,5 month ago)
There was something about updating the ChapterFilter where I need to receive the `section` instead of the `id` and since previously everything was working on checkboxes we were dispatching all selected ids + the newly clicked one
Anyway, I remember that when I was trying to implement a solution I thought that it would be cleaner if we would have a standardized way of handling Color a Chapter filter updates

TomWoodward: well we used to dispatch the new selected set, now you've made it dispatch a diff, the only reason i can think of that being helpful is if they are received out of order or something

PR #947, opened 2020-11-19
Issue 1376

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1376


src/app/components/ScrollLock.tsx
```diff
@@ -36,8 +37,16 @@ const fadeIn = keyframes`
...
+interface OverlayProps {
+  className?: string;
+  [key: string]: any;
+}
+
 // tslint:disable-next-line:variable-name
-export const Overlay = styled.div`
+export const Overlay = styled(({ className, ...otherProps }: OverlayProps) => {
+  useDisableContentTabbing();
+  return <div className={className} {...otherProps} />;
```
TomWoodward, line 48: you don't have to split out the classname it could be applied the same way as `otherProps`

src/app/content/highlights/components/HighlightsPopUp.tsx
```diff
@@ -1,5 +1,6 @@
 import { HTMLElement } from '@openstax/types/lib.dom';
 import React from 'react';
+import { createPortal } from 'react-dom';
```
TomWoodward, line 3: i was hoping we would do the portal stuff in some kind of shared modal component and not have to do it in 4 places

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
```diff
@@ -73,7 +74,8 @@ const PracticeQuestionsPopup = () => {
         </Header>
         <ShowPracticeQuestions />
       </Modal>
-    </PopupWrapper>
+    </PopupWrapper>,
+    assertDocument().body
```
TomWoodward, line 78: also i'm not sure you're supposed to portal directly into the body, i think you might be supposed to create a mount component and add it to the body and portal into that, but i'm not totally sure, if you saw this in the docs then its fine

src/app/reactUtils.ts
```diff
@@ -191,3 +191,55 @@ export const useOnScrollTopOffset = () => {
 
   return topOffset;
 };
+
+const tabbableElementsSelector = [
```
TomWoodward, line 195: if you got this list from somewhere please link it in a comment

TomWoodward (changes requested):

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
PiotrKozlowski: I've neverd heard of this and docs just says:
```
render() {
  // React does *not* create a new div. It renders the children into `domNode`.
  // `domNode` is any valid DOM node, regardless of its location in the DOM.
  return ReactDOM.createPortal(
    this.props.children,
    domNode  );
}
```

There is also closed issue on Material-UI repo according to this
https://github.com/mui-org/material-ui/issues/21626

TomWoodward: /shrug good enough for material-ui good enough for me

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
```diff
@@ -32,6 +32,7 @@ const slideInAnimation = css`
 const Wrapper = styled.div`
   position: fixed;
   bottom: 0;
+  left: 0;
```
TomWoodward, line 35: why is this added here?

TomWoodward (approved):
i'm not sure about that one style change, but everything else looks good

src/app/content/highlights/components/SummaryPopup/HighlightsHelpInfo.tsx
PiotrKozlowski: `HighlightsHelpInfo` was previously outside of the `<Modal>` and now it is inside

PR #949, opened 2020-11-27
Issue 1336

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1336


src/app/content/practiceQuestions/reducer.ts
```diff
@@ -13,6 +13,7 @@ export const initialState: State = {
   isEnabled: false,
   open: false,
   questions: [],
+  questionsAndAnswers: new Map(),
```
TomWoodward, line 16: supposed to keep states as simple objects/arrays, you can convert to a map in the selector if you want

TomWoodward (changes requested):

src/app/content/practiceQuestions/components/PQButton.tsx
```diff
@@ -13,4 +12,26 @@ const PQButton = styled(PlainButton)`
   padding: 0 3rem;
 `;
 
+// tslint:disable-next-line: variable-name
+const PQButton = styled(PlainButton)`
+  ${sharedStyles}
+`;
+
+// tslint:disable-next-line: variable-name
+export const PQInput = styled.input`
```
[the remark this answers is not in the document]
PiotrKozlowski: Do we not want to use `input` for the submit buttons or do you mean replace `PQButton` with `<Button ...>`?

src/app/content/practiceQuestions/components/Question.tsx
```diff
@@ -40,25 +36,56 @@ const getChoiceLetter = (value: number) => {
...
   const [selectedAnswer, setSelectedAnswer] = React.useState<PracticeAnswer | null>(null);
+  const [showCorrect, setShowCorrect] = React.useState(false);
+  const question = useSelector(pqSelectors.question);
+  const section = useSelector(pqSelectors.selectedSection);
+  const questionsAndAnswers = useSelector(pqSelectors.questionsAndAnswers);
+  const dispatch = useDispatch();
+
+  React.useEffect(() => {
+    setSelectedAnswer(null);
+    setShowCorrect(false);
```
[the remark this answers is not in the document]
PiotrKozlowski: I think in theory it could be possible. Changing this to
```js
return () => {
    setSelectedAnswer(null);
    setShowCorrect(false);
}
```
should do the job, right?
It should update the state before processing the new question?

src/app/content/practiceQuestions/components/PQButton.tsx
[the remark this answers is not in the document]
TomWoodward: button can be rendered as an input with `<Button component={<input type="text" />}` you might have to reset some of the input styles but at least you'll be using the actual button styles instead of re-creating them

src/app/content/practiceQuestions/components/Question.tsx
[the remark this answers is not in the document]
TomWoodward: state updates are async while redux updates are synchronous so i'm not sure if there might still be a small race condition, but it'd be good to at least remove the `useEffect` from the equation because that will also cause additional delay

src/app/content/practiceQuestions/components/Answer.tsx
```diff
@@ -81,28 +99,92 @@ export const AnswerBlock = styled.div`
...
+  let variant: PracticeQuestionStyles['label'] | null = null;
+
+  if (isSubmitted && isSelected) {
+    variant = isCorrect ? 'correct' : 'incorrect';
+  } else if (isSubmitted && showCorrect && isCorrect) {
+    variant = 'correct';
+  }
+
+  return variant
+    ? <StyledAnswerResult variant={variant}>
```
TomWoodward, line 128: variant can be passed as `isCorrect ? 'correct' : 'incorrect'` without the pre-set variable right?

then the previous logic could be combined into a short circuit that returns `null`, remove the ternary in the return value and remove the `let variant`

src/app/content/practiceQuestions/components/Question.tsx
```diff
@@ -40,25 +36,51 @@ const getChoiceLetter = (value: number) => {
...
   const [selectedAnswer, setSelectedAnswer] = React.useState<PracticeAnswer | null>(null);
+  const [showCorrect, setShowCorrect] = React.useState(false);
+  const question = useSelector(pqSelectors.question);
+  const section = useSelector(pqSelectors.selectedSection);
+  const isSubmitted = useSelector(pqSelectors.isCurrentQuestionSubmitted);
+  const dispatch = useDispatch();
+
+  React.useEffect(() => {
+    return () => {
+      setSelectedAnswer(null);
```
TomWoodward, line 49: i think selectedAnswer is less bad because it will naturally not be found in the new question, we could add a helper
```
const selectedAnswer = question.answers.find(mach(selectedAnswerState));
```

and not bother even clearing it when question changes the helper will just start returning null.

similarly what if instead of a boolean we put the question `| null` as the value of showCorrect, and a helper

```
const showCorrect = showCorrectState === question;
```

that way the only thing that actually changes is the question and we dont have any race conditions

src/app/content/practiceQuestions/types.ts
```diff
@@ -31,8 +32,17 @@ export interface PracticeQuestion {
 
 export type PracticeQuestions = PracticeQuestion[];
 
+export interface QuestionsAndAnswers {
```
TomWoodward, line 35: would this be better named QuestionAnswers? since it doesnt actually include the question

TomWoodward (changes requested):

TomWoodward (changes requested):
i submitted some updates for things i was going to comment on, i think everything looks fine now except i would like you to look into using the common button component instead of reproducing those styles

src/app/content/practiceQuestions/components/PQButton.tsx
```diff
@@ -1,33 +1,19 @@
...
+import Button from '../../../components/Button';
+import { linkColor } from '../../../components/Typography';
 
 // tslint:disable-next-line: variable-name
-export const PQInput = styled.input`
-  ${sharedStyles}
-  border: none;
-  cursor: pointer;
+const PQButton = styled(Button)`
   ${(props: { disabled: boolean }) => props.disabled && `
```
TomWoodward, line 7: the button itself should have disabled styles, if it doesn't you can move these there

src/app/content/practiceQuestions/components/PQButton.tsx
```diff
@@ -1,33 +1,19 @@
...
     background-color: #f1f1f1;
     color: #c1c1c1;
     cursor: not-allowed;
+    &:hover {
+      background-color: #f1f1f1;
```
TomWoodward, line 12: can these colors come from the theme?

src/app/content/practiceQuestions/components/PQButton.tsx
```diff
@@ -1,33 +1,19 @@
...
+    }
   `}
   ${(props: { withoutBg?: boolean }) => props.withoutBg && `
```
TomWoodward, line 15: you could add this as a variant to the button component. call it `variant="transparent"` or `variant="dropout"` or something

src/app/content/practiceQuestions/components/QuestionNavigation/index.tsx
```diff
@@ -36,50 +36,65 @@ const QuestionNavigation = ({ question, selectedAnswer, ...props }: QuestionNavi
...
         {(msg: string) => (
-          <PQInput
-            type='submit'
-            value={msg}
-            default={true}
+          <PQButton
+            variant='primary'
+            size='large'
             disabled={selectedAnswer ? false : true}
+            component={<input type='submit' value={msg} />}
```
TomWoodward, line 58: where is the onSubmit handler for this form?

TomWoodward (changes requested):

src/app/content/practiceQuestions/components/QuestionNavigation/index.tsx
PiotrKozlowski: it is handled by `<form onSubmit={...} ...>` in `Question.tsx`

TomWoodward (approved):

PR #950, opened 2020-11-30
Modify relative urls for RAP

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1308

### DIFF
--- a/src/app/content/components/ContentExcerpt.spec.tsx
+++ b/src/app/content/components/ContentExcerpt.spec.tsx
@@ -14,14 +14,6 @@ import { findArchiveTreeNodeById } from '../utils/archiveTreeUtils';
 import * as contentManipulation from '../utils/contentManipulation';
 import ContentExcerpt from './ContentExcerpt';
 
-jest.mock('../../../config', () => {
-  const config = jest.requireActual('../../../config');
-  return {
-    ...config,
-    REACT_APP_ARCHIVE_URL: null,
-  };
-});
-
 describe('ContentExcerpt', () => {
   let store: Store;
 
--- a/src/app/content/components/ContentExcerpt.spec.tsx
+++ b/src/app/content/components/ContentExcerpt.spec.tsx
@@ -18,7 +18,7 @@ jest.mock('../../../config', () => {
   const config = jest.requireActual('../../../config');
   return {
     ...config,
-    REACT_APP_ARCHIVE_URL: '',
+    REACT_APP_ARCHIVE_URL: null,
   };
 });
 
@@ -44,7 +44,6 @@ describe('ContentExcerpt', () => {
   );
 
   it('fixes urls in content using addTargetBlank and resolveRelative', () => {
-    // routes.content.getSearch = jest.fn().mockReturnValue('archive=some-content');
     const originalHtml = '<a href="#hello"></a>';
     const htmlWithTargetBlank = '<a href="#hello" target="_blank"></a>';
     store.dispatch(receiveBook(formatBookData(book, mockCmsBook)));
@@ -55,15 +54,14 @@ describe('ContentExcerpt', () => {
     const addTargetBlankToLinksMock = jest.spyOn(contentManipulation, 'addTargetBlankToLinks')
       .mockReturnValueOnce(htmlWithTargetBlank);
 
-    const resolveRelativeURLsMock = jest.spyOn(contentManipulation, 'rebaseRelativeContentLinks').mockReturnValueOnce(
-      '<a href="/book/book1/page/testbook1-testpage1-uuid#hello" target="_blank"></a>'
-    );
+    const rebaseRelativeContentLinksMock = jest.spyOn(contentManipulation, 'rebaseRelativeContentLinks')
+      .mockReturnValueOnce('<a href="/book/book1/page/testbook1-testpage1-uuid#hello" target="_blank"></a>');
 
     render(mockSection, originalHtml);
 
     expect(getUrlSpy).toHaveBeenCalled();
     expect(addTargetBlankToLinksMock).toHaveBeenCalledWith(originalHtml);
-    expect(resolveRelativeURLsMock)
+    expect(rebaseRelativeContentLinksMock)
       .toHaveBeenCalledWith(htmlWithTargetBlank, '/book/book1/page/testbook1-testpage1-uuid');
   });
 
--- a/src/app/content/components/ContentExcerpt.spec.tsx
+++ b/src/app/content/components/ContentExcerpt.spec.tsx
@@ -14,6 +14,14 @@ import { findArchiveTreeNodeById } from '../utils/archiveTreeUtils';
 import * as contentManipulation from '../utils/contentManipulation';
 import ContentExcerpt from './ContentExcerpt';
 
+jest.mock('../../../config', () => {
+  const config = jest.requireActual('../../../config');
+  return {
+    ...config,
+    REACT_APP_ARCHIVE_URL: '',
+  };
+});
+
 describe('ContentExcerpt', () => {
   let store: Store;
 
@@ -35,7 +43,8 @@ describe('ContentExcerpt', () => {
     </Provider>
   );
 
-  it('fixes urls in content using addTargetBlank and fixRelative', () => {
+  it('fixes urls in content using addTargetBlank and resolveRelative', () => {
+    // routes.content.getSearch = jest.fn().mockReturnValue('archive=some-content');
     const originalHtml = '<a href="#hello"></a>';
     const htmlWithTargetBlank = '<a href="#hello" target="_blank"></a>';
     store.dispatch(receiveBook(formatBookData(book, mockCmsBook)));
@@ -46,7 +55,7 @@ describe('ContentExcerpt', () => {
     const addTargetBlankToLinksMock = jest.spyOn(contentManipulation, 'addTargetBlankToLinks')
       .mockReturnValueOnce(htmlWithTargetBlank);
 
-    const fixRelativeURLsMock = jest.spyOn(contentManipulation, 'fixRelativeURLs').mockReturnValueOnce(
+    const resolveRelativeURLsMock = jest.spyOn(contentManipulation, 'rebaseRelativeContentLinks').mockReturnValueOnce(
       '<a href="/book/book1/page/testbook1-testpage1-uuid#hello" target="_blank"></a>'
     );
 
@@ -54,7 +63,7 @@ describe('ContentExcerpt', () => {
 
     expect(getUrlSpy).toHaveBeenCalled();
     expect(addTargetBlankToLinksMock).toHaveBeenCalledWith(originalHtml);
-    expect(fixRelativeURLsMock)
+    expect(resolveRelativeURLsMock)
       .toHaveBeenCalledWith(htmlWithTargetBlank, '/book/book1/page/testbook1-testpage1-uuid');
   });
 
--- a/src/app/content/components/ContentExcerpt.tsx
+++ b/src/app/content/components/ContentExcerpt.tsx
@@ -7,7 +7,11 @@ import { assertDefined } from '../../utils';
 import { book } from '../selectors';
 import { LinkedArchiveTreeSection } from '../types';
 import { findArchiveTreeNodeById } from '../utils/archiveTreeUtils';
-import { addTargetBlankToLinks, fixRelativeURLs } from '../utils/contentManipulation';
+import {
+  addTargetBlankToLinks,
+  rebaseRelativeContentLinks,
+  rebaseRelativeResources,
+} from '../utils/contentManipulation';
 import { getBookPageUrlAndParams } from '../utils/urlUtils';
 
 interface Props {
@@ -37,7 +41,8 @@ const ContentExcerpt = styled((props: Props) => {
 
   const fixedContent = React.useMemo(() => flow(
     addTargetBlankToLinks,
-    (newContent) => fixRelativeURLs(newContent, excerptSource.url)
+    (newContent) => rebaseRelativeContentLinks(newContent, excerptSource.url),
+    (newContent) => rebaseRelativeResources(newContent, excerptSource.url)
   )(props.content), [props.content, excerptSource.url]);
 
   return <div
--- a/src/app/content/utils/contentManipulation.ts
+++ b/src/app/content/utils/contentManipulation.ts
@@ -15,13 +15,28 @@ export const isAbsoluteUrl = (url: string) => {
   return aux;
 };
 
-export function fixRelativeURLs(htmlString: string, sourceUrl?: string): string {
+export const rebaseRelativeContentLinks = (htmlString: string, sourceUrl?: string) => {
   const domNode = domParser.parseFromString(htmlString, 'text/html');
-  domNode.querySelectorAll('a').forEach((a: HTMLAnchorElement) => {
-    const hrefValue = a.getAttribute('href');
-    if (hrefValue && !isAbsoluteUrl(hrefValue) && sourceUrl) {
-      a.setAttribute('href', fromRelativeUrl(sourceUrl, hrefValue));
+  domNode.querySelectorAll('a').forEach((element: HTMLAnchorElement) => {
+    const hrefValue = element.getAttribute('href');
+    if (hrefValue && !isAbsoluteUrl(hrefValue) &&  sourceUrl) {
+      const resolvedUrl = fromRelativeUrl(sourceUrl, hrefValue);
+      element.setAttribute('href', resolvedUrl);
     }
   });
   return domNode.body.innerHTML;
-}
+};
+
+export const rebaseRelativeResources = (htmlString: string, sourceUrl?: string) => {
+  const domNode = domParser.parseFromString(htmlString, 'text/html');
+
+  domNode.querySelectorAll('img, iframe').forEach((element: HTMLAnchorElement) => {
+    const srcValue = element.getAttribute('src');
+    if (srcValue && !isAbsoluteUrl(srcValue) && sourceUrl ) {
+      const resolvedUrl = fromRelativeUrl(sourceUrl, srcValue);
+      element.setAttribute('src', resolvedUrl);
+    }
+  });
+
+  return domNode.body.innerHTML;
+};
--- a/src/app/content/utils/contentManipulation.ts
+++ b/src/app/content/utils/contentManipulation.ts
@@ -30,7 +30,7 @@ export const rebaseRelativeResources = (htmlString: string, sourceUrl: string) =
   const domNode = domParser.parseFromString(htmlString, 'text/html');
   domNode.querySelectorAll('img,iframe').forEach((element: HTMLAnchorElement) => {
     const srcValue = element.getAttribute('src');
-    if (srcValue && !isAbsoluteUrl(srcValue) && sourceUrl ) {
+    if (srcValue && !isAbsoluteUrl(srcValue) && sourceUrl) {
       element.setAttribute('src', fromRelativeUrl(sourceUrl, srcValue));
     }
   });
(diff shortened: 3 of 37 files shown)


src/app/content/utils/getCleanContent.ts
```diff
@@ -14,7 +16,10 @@ export default function getCleanContent(
 
   const pageContent = cachedPage ? cachedPage.content : '';
 
-  return transformer(pageContent)
+  const originalUrl = (book && page && archiveLoader.book(book.id, book.version).page(page.id).url())
+    || 'https://openstax.org/apps/archive/20201029.133542';
```
KarinaMendez2, line 20: this is what's making it work I think @TomWoodward

TomWoodward: this isn't a page url i'm surprised it would be working at all with this in here

src/app/content/components/Page/validateDOMContent.ts
```diff
@@ -14,7 +14,7 @@ const validateLinks = (rootEl: HTMLElement) => {
    * on other sites.
    */
 
-  const urls = Array.from(rootEl.querySelectorAll('a[href^="/"]'))
+  const urls = Array.from(rootEl.querySelectorAll('a[href^="../"]'))
```
TomWoodward, line 17: this isn't right

src/app/content/search/components/SearchResultsSidebar/__snapshots__/index.spec.tsx.snap
```diff
@@ -158,7 +158,7 @@ exports[`SearchResultsSidebar matches snapshot with results 1`] = `
       <a
         className="styled__SectionContentPreview-as3fh1-11 dMNXGN"
         data-testid="search-result"
-        href="books/book-slug-1/pages/test-page-1"
+        href="books/book-slug-1/pages/test-page-1?archive=https://openstax.org/apps/archive/20201029.133542"
```
TomWoodward, line 161: these shouldn't have changed, contentlink is adding it because it thinks it has to, your mocks probably need work

src/app/content/utils/getCleanContent.ts
```diff
@@ -24,4 +29,8 @@ export default function getCleanContent(
     // move (first-child) figure and table ids up to the parent div
     .replace(/(<div[^>]*)(>[^<]*<(?:figure|table)[^>]*?) (id=[^\s>]*)/g, '$1 $3$2 data-$3')
   ;
+
+  const fixRAPrelatives = (content: string) => fixRelativeURLs(replacements(content), ['a', 'img'], originalUrl);
+
+  return flow(fixRAPrelatives, transformer)(pageContent);
```
TomWoodward, line 35: as long as you're using flow you might as well put `replacements` on the beginning, its weird the way you have it

src/app/content/utils/getCleanContent.ts
```diff
@@ -14,7 +16,10 @@ export default function getCleanContent(
 
   const pageContent = cachedPage ? cachedPage.content : '';
 
-  return transformer(pageContent)
+  const originalUrl = (book && page && archiveLoader.book(book.id, book.version).page(page.id).url())
```
TomWoodward, line 19: rename this `contentUrl`

src/app/content/utils/getCleanContent.ts
```diff
@@ -24,4 +29,8 @@ export default function getCleanContent(
     // move (first-child) figure and table ids up to the parent div
     .replace(/(<div[^>]*)(>[^<]*<(?:figure|table)[^>]*?) (id=[^\s>]*)/g, '$1 $3$2 data-$3')
   ;
+
+  const fixRAPrelatives = (content: string) => fixRelativeURLs(replacements(content), ['a', 'img'], originalUrl);
```
TomWoodward, line 33: rename this like `resolveResourceUrls` or something

src/app/developer/components/__snapshots__/Home.spec.tsx.snap
```diff
@@ -56,7 +56,7 @@ Array [
               >
                 <a
                   className="ContentLink__StyledContentLink-ctb7jh-0 jAbAdk"
-                  href="books/book-slug-1/pages/1-introduction-to-science-and-the-realm-of-physics-physical-quantities-and-units"
+                  href="books/book-slug-1/pages/1-introduction-to-science-and-the-realm-of-physics-physical-quantities-and-units?archive=https://openstax.org/apps/archive/20201029.133542"
```
TomWoodward, line 59: shouldn't be changed

src/config.js
```diff
@@ -18,12 +18,12 @@ let config = {
...
   DEPLOYED_ENV: typeof(window) === 'undefined' ? 'server' : window.location.hostname,
   ACCOUNTS_URL: process.env.ACCOUNTS_URL || 'https://openstax.org',
-  ARCHIVE_URL: process.env.ARCHIVE_URL || 'https://openstax.org',
+  ARCHIVE_URL: process.env.ARCHIVE_URL || 'https://openstax.org/apps/archive/20201029.133542',
   OS_WEB_URL: process.env.OS_WEB_URL || 'https://openstax.org',
   SEARCH_URL: process.env.SEARCH_URL || 'https://openstax.org',
   HIGHLIGHTS_URL: process.env.HIGHLIGHTS_URL || 'https://openstax.org',
   REACT_APP_ACCOUNTS_URL: '/accounts',
-  REACT_APP_ARCHIVE_URL: '',
+  REACT_APP_ARCHIVE_URL: 'https://openstax.org/apps/archive/20201029.133542',
```
TomWoodward, line 26: this should be only the path, look at the other examples

src/config.js
```diff
@@ -18,12 +18,12 @@ let config = {
   IS_PRODUCTION: process.env.REACT_APP_ENV === 'production',
   DEPLOYED_ENV: typeof(window) === 'undefined' ? 'server' : window.location.hostname,
   ACCOUNTS_URL: process.env.ACCOUNTS_URL || 'https://openstax.org',
-  ARCHIVE_URL: process.env.ARCHIVE_URL || 'https://openstax.org',
+  ARCHIVE_URL: process.env.ARCHIVE_URL || 'https://openstax.org/apps/archive/20201029.133542',
```
TomWoodward, line 21: this shouldn't have the path, look at the other examples

src/gateways/createArchiveLoader.ts
```diff
@@ -66,6 +67,7 @@ export default (url: string) => {
         page: (pageId: string) => ({
           cached: () => pageCache.get(`${bookRef}:${pageId}`),
           load: () => pageLoader(`${bookRef}:${pageId}`),
+          url: () => `${url}/contents/${bookRef}:${pageId}`,
```
TomWoodward, line 70: i wanted you to dry this up, so `${bookRef}:${pageId}` is not in 3 places and `/contents` is not in two

src/setupProxy.js
```diff
@@ -105,9 +107,14 @@ function setupTestProxy(app) {
...
-    target: `${ARCHIVE_URL}${path}`,
-    prependPath: false,
+    target: ARCHIVE_URL,
+    prependPath: true,
     changeOrigin: true,
+    onProxyReq: (preq, req, res) => {
+      console.log(ARCHIVE_URL);
+      console.log(preq.path);
+      console.log(req.path);
+    }
```
TomWoodward, line 117: get rid of all this logging

src/test/mocks/archiveLoader.ts
```diff
@@ -76,6 +76,7 @@ export default () => {
       page: (pageId: string) => ({
         cached: () => cachedPage(bookId, bookVersion, pageId),
         load: () => loadPage(bookId, bookVersion, pageId),
+        url: () => `someurl.com/contents/${bookId}@${bookVersion}:${pageId}`,
```
TomWoodward, line 79: `someurl.com` shouldn't be on the front of this

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,12 +15,18 @@ export const isAbsoluteUrl = (url: string) => {
   return aux;
 };
 
-export function fixRelativeURLs(htmlString: string, sourceUrl?: string): string {
+export function fixRelativeURLs(htmlString: string, selectors: string[], sourceUrl?: string): string {
```
TomWoodward, line 18: `fix` is a little judgemental here, maybe `resolve` would be better

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,12 +15,18 @@ export const isAbsoluteUrl = (url: string) => {
...
   const domNode = domParser.parseFromString(htmlString, 'text/html');
-  domNode.querySelectorAll('a').forEach((a: HTMLAnchorElement) => {
-    const hrefValue = a.getAttribute('href');
-    if (hrefValue && !isAbsoluteUrl(hrefValue) && sourceUrl) {
-      a.setAttribute('href', fromRelativeUrl(sourceUrl, hrefValue));
+  domNode.querySelectorAll(selectors).forEach((element: HTMLAnchorElement) => {
```
TomWoodward, line 20: you're passing in a selector but each type of element has a different attribute that you're fixing, this seems kinda messy i think we can do better

is there anything wrong with having it always do both images and anchors? you could do them in different loops.

or you could check the element type inside this loop and see if its an anchor change the href and if its an image change the src.

i think i'd rather do the first thing unless there is some reason that causes issues

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -35,7 +43,8 @@ describe('ContentExcerpt', () => {
     </Provider>
   );
 
-  it('fixes urls in content using addTargetBlank and fixRelative', () => {
+  it('fixes urls in content using addTargetBlank and resolveRelative', () => {
+    // routes.content.getSearch = jest.fn().mockReturnValue('archive=some-content');
```
TomWoodward, line 47: commented code

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -46,15 +55,15 @@ describe('ContentExcerpt', () => {
     const addTargetBlankToLinksMock = jest.spyOn(contentManipulation, 'addTargetBlankToLinks')
       .mockReturnValueOnce(htmlWithTargetBlank);
 
-    const fixRelativeURLsMock = jest.spyOn(contentManipulation, 'fixRelativeURLs').mockReturnValueOnce(
+    const resolveRelativeURLsMock = jest.spyOn(contentManipulation, 'rebaseRelativeContentLinks').mockReturnValueOnce(
```
TomWoodward, line 58: names don't match

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -37,7 +41,8 @@ const ContentExcerpt = styled((props: Props) => {
 
   const fixedContent = React.useMemo(() => flow(
     addTargetBlankToLinks,
-    (newContent) => fixRelativeURLs(newContent, excerptSource.url)
+    (newContent) => rebaseRelativeContentLinks(newContent, excerptSource.url),
+    (newContent) => rebaseRelativeResources(newContent, excerptSource.url)
```
TomWoodward, line 45: technically i think this one might not be necessary here, because the excerpts should be taken from the content after the paths are already fixed.

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,13 +15,28 @@ export const isAbsoluteUrl = (url: string) => {
...
     }
   });
   return domNode.body.innerHTML;
-}
+};
+
+export const rebaseRelativeResources = (htmlString: string, sourceUrl?: string) => {
+  const domNode = domParser.parseFromString(htmlString, 'text/html');
+
+  domNode.querySelectorAll('img, iframe').forEach((element: HTMLAnchorElement) => {
```
TomWoodward, line 33: i think the space might be bad here

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,13 +15,28 @@ export const isAbsoluteUrl = (url: string) => {
...
   const domNode = domParser.parseFromString(htmlString, 'text/html');
-  domNode.querySelectorAll('a').forEach((a: HTMLAnchorElement) => {
-    const hrefValue = a.getAttribute('href');
-    if (hrefValue && !isAbsoluteUrl(hrefValue) && sourceUrl) {
-      a.setAttribute('href', fromRelativeUrl(sourceUrl, hrefValue));
+  domNode.querySelectorAll('a').forEach((element: HTMLAnchorElement) => {
+    const hrefValue = element.getAttribute('href');
+    if (hrefValue && !isAbsoluteUrl(hrefValue) &&  sourceUrl) {
+      const resolvedUrl = fromRelativeUrl(sourceUrl, hrefValue);
+      element.setAttribute('href', resolvedUrl);
```
TomWoodward, line 24: there used to be more logic here but now that its been moved you can revert most of these changes

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,13 +15,28 @@ export const isAbsoluteUrl = (url: string) => {
   return aux;
 };
 
-export function fixRelativeURLs(htmlString: string, sourceUrl?: string): string {
+export const rebaseRelativeContentLinks = (htmlString: string, sourceUrl?: string) => {
```
TomWoodward, line 18: source url should be required

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,13 +15,28 @@ export const isAbsoluteUrl = (url: string) => {
...
+    if (hrefValue && !isAbsoluteUrl(hrefValue) &&  sourceUrl) {
+      const resolvedUrl = fromRelativeUrl(sourceUrl, hrefValue);
+      element.setAttribute('href', resolvedUrl);
     }
   });
   return domNode.body.innerHTML;
-}
+};
+
+export const rebaseRelativeResources = (htmlString: string, sourceUrl?: string) => {
```
TomWoodward, line 30: source url should be required

src/gateways/createArchiveLoader.ts
```diff
@@ -58,14 +60,16 @@ export default (url: string) => {
   return {
     book: (bookId: string, bookVersion?: string) => {
       const bookRef = bookVersion ? `${stripIdVersion(bookId)}@${bookVersion}` : stripIdVersion(bookId);
+      const bookAndPageRef = (pageId: string) => `${bookRef}:${pageId}`;
 
       return {
         cached: () => bookCache.get(bookRef),
         load: () => bookLoader(bookRef),
 
         page: (pageId: string) => ({
```
TomWoodward, line 69: give this function a body instead of the return object shorthand and you can define your `bookAndPageRef` at the top and not have the extra function

src/app/content/components/ContentExcerpt.spec.tsx
```diff
@@ -14,6 +14,14 @@ import { findArchiveTreeNodeById } from '../utils/archiveTreeUtils';
 import * as contentManipulation from '../utils/contentManipulation';
 import ContentExcerpt from './ContentExcerpt';
 
+jest.mock('../../../config', () => {
+  const config = jest.requireActual('../../../config');
+  return {
+    ...config,
+    REACT_APP_ARCHIVE_URL: '',
```
TomWoodward, line 21: we should probably deal with this in a better way. i'm realizing that the current logic is appending the query param to routes if `REACT_APP_ARCHIVE_URL` is defined at all, but in the future when adopting rap it'll always be defined....

TomWoodward (changes requested):

src/config.js
```diff
@@ -23,7 +23,7 @@ let config = {
   SEARCH_URL: process.env.SEARCH_URL || 'https://openstax.org',
   HIGHLIGHTS_URL: process.env.HIGHLIGHTS_URL || 'https://openstax.org',
   REACT_APP_ACCOUNTS_URL: '/accounts',
-  REACT_APP_ARCHIVE_URL: '',
+  REACT_APP_ARCHIVE_URL: '/apps/archive/20201029.133542',
```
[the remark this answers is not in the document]
KarinaMendez2: yeah, that's why I put it in a separate commit

src/app/content/utils/contentManipulation.ts
KarinaMendez2: what space?

src/app/content/components/ContentExcerpt.tsx
KarinaMendez2: so remove rebaseRelativeResources? I thought we needed both

src/gateways/createArchiveLoader.ts
KarinaMendez2: I don't understand what you mean about _give it a body_

src/app/content/components/ContentExcerpt.spec.tsx
KarinaMendez2: so like deal with it better in the mocks or everywhere?

src/app/content/components/ContentExcerpt.tsx
TomWoodward: yea no its fine, i don't think it'll break anything

src/app/content/utils/contentManipulation.ts
TomWoodward: the one in the selector

src/gateways/createArchiveLoader.ts
TomWoodward: right now the function is like
```
(pageId: string) => ({
})
```
returning an object literal, you can change it to
```
(pageId: string) => {
  const theThing = 'asfd';
  return {
  };
}
```

TomWoodward (changes requested):

src/app/content/utils/contentManipulation.ts
```diff
@@ -15,13 +15,25 @@ export const isAbsoluteUrl = (url: string) => {
...
-export function fixRelativeURLs(htmlString: string, sourceUrl?: string): string {
+export const rebaseRelativeContentLinks = (htmlString: string, sourceUrl: string) => {
   const domNode = domParser.parseFromString(htmlString, 'text/html');
-  domNode.querySelectorAll('a').forEach((a: HTMLAnchorElement) => {
-    const hrefValue = a.getAttribute('href');
-    if (hrefValue && !isAbsoluteUrl(hrefValue) && sourceUrl) {
-      a.setAttribute('href', fromRelativeUrl(sourceUrl, hrefValue));
+  domNode.querySelectorAll('a').forEach((element: HTMLAnchorElement) => {
+    const hrefValue = element.getAttribute('href');
+    if (hrefValue && !isAbsoluteUrl(hrefValue) &&  sourceUrl) {
```
TomWoodward, line 22: extra whitespace

src/app/content/utils/getCleanContent.ts
```diff
@@ -14,7 +16,9 @@ export default function getCleanContent(
 
   const pageContent = cachedPage ? cachedPage.content : '';
 
-  return transformer(pageContent)
+  const contentUrl = (book && page && archiveLoader.book(book.id, book.version).page(page.id).url()) || '';
```
TomWoodward, line 19: this still isn't handling the null case

src/gateways/createArchiveLoader.ts
```diff
@@ -34,6 +36,11 @@ export default (url: string) => {
   const pageCache = createCache<string, ArchivePage>({maxRecords: 20});
   const pageLoader = contentsLoader<ArchivePage>(pageCache);
 
+  const bookAndPageRef = (bookRef: string, pageId: string) => {
+    const bookAndPageUrl = `${bookRef}:${pageId}`;
```
TomWoodward, line 40: you didn't fix the thing i commented on before

TomWoodward (changes requested):

TomWoodward (approved):

PR #957, opened 2020-12-07
fix build

TomWoodward:
this theoretically shouldn't be used during pre-render but there is some invalid content hitting it

for: openstax/unified#1398

### DIFF
--- a/script/entry.js
+++ b/script/entry.js
@@ -10,7 +10,7 @@ const JSDOM = require('jsdom').JSDOM;
 global.DOMParser = new JSDOM().window.DOMParser;
 
 const URL = require('url');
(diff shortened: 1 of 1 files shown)


script/entry.js
```diff
@@ -10,7 +10,7 @@ const JSDOM = require('jsdom').JSDOM;
 global.DOMParser = new JSDOM().window.DOMParser;
 
 const URL = require('url');
-global.URL = URL;
+global.URL = URL.URL;
```
PiotrKozlowski, line 13: Probably `const { URL } = require('url');` would be a little bit cleaner

PiotrKozlowski (approved):

PR #959, opened 2020-12-08
Issue 1337

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1337


src/app/components/Button.tsx
```diff
@@ -38,12 +38,15 @@ interface ButtonProps<T extends ComponentType | undefined> {
 }
 
 // tslint:disable-next-line:variable-name
-function ButtonHoc<T extends ComponentType | undefined>({variant, size, component, ...props}: ButtonProps<T>) {
+const ButtonHoc = React.forwardRef((
```
TomWoodward, line 41: making this hoc not a generic doesn't seem ok

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -16,16 +16,19 @@ import { getBookPageUrlAndParams } from '../utils/urlUtils';
 
 interface Props {
   content: string;
-  className: string;
+  className?: string;
   source: string | LinkedArchiveTreeSection;
+  forwardedRef?: React.Ref<HTMLElement>;
+  [key: string]: any;
```
TomWoodward, line 22: why this? if you're looking for things you can add to a div then extend the div component props

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -46,17 +49,20 @@ const ContentExcerpt = styled((props: Props) => {
...
+    ref={forwardedRef}
     dangerouslySetInnerHTML={{ __html: fixedContent }}
     className={`content-excerpt ${className}`}
     {...excerptProps}
   />;
-})`
+};
+
+export default styled(React.forwardRef<HTMLElement, Props>(
+  (props, ref) => <ContentExcerpt {...props} forwardedRef={ref} />)
```
TomWoodward, line 60: would this be simpler if you put the forwardRef call around the ContentExcerpt definition?

src/app/content/practiceQuestions/components/Answer/index.tsx
```diff
@@ -63,19 +72,22 @@ const Answer = ({
       id='i18n:practice-questions:popup:answers:choice'
       values={{choiceIndicator: choiceIndicator.toUpperCase()}}
     >{(msg: string) =>
-      <AnswerLabel aria-label={msg}>
-        {choiceIndicator}
+      <React.Fragment>
         <input
+          id={choiceIndicator}
```
TomWoodward, line 77: this seems like a pretty generic thing to make an id, is the id necessary for the label to work?

src/app/content/practiceQuestions/components/Answer/index.tsx
```diff
@@ -63,19 +72,22 @@ const Answer = ({
...
           type='radio'
           name={choiceIndicator}
           checked={isSelected}
           disabled={isSubmitted}
           onChange={onSelect}
         />
-      </AnswerLabel>
+        <AnswerLabel ref={answerRef} aria-label={msg} for={choiceIndicator} tabIndex={-1}>
```
TomWoodward, line 84: when karina did this she used [htmlFor](https://reactjs.org/docs/dom-elements.html#htmlfor)

src/app/content/practiceQuestions/components/Answer/index.tsx
TomWoodward, line 77: this input needs conditional `aria-selected` according to the spec right?

TomWoodward (changes requested):
some warnings in console:

```index.js:1437 Warning: React does not recognize the `mobileOnly` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `mobileonly` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```

```index.js:1437 Warning: React does not recognize the `zIndex` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `zindex` instead. If you accidentally passed it from a parent component, remove it from the DOM element.```

i'm tabbing between the a/b/c/d and the text separately, the tab should highlight the whole row. i think the question text should really be in the label but its not.

the a/b/c/d indicators are being clipped on the sides in their active/selected state

pretty sure its triggering skip when i try to submit the form by pressing the enter key

File: src/app/content/practiceQuestions/components/Answer/index.tsx (hunk 4 above)
Comment (PiotrKozlowski) [reply src/app/content/practiceQuestions/components/Answer/index.tsx:77]: I think it doesn't because screenreader (tested with ChromeVox) will read that you've checked the input radio button and when you TAB to this input you will be informed if it is or it is not selected

File: src/app/content/practiceQuestions/components/Question.tsx [hunk 6]
```diff
@@ -55,6 +56,10 @@ const Question = () => {
     if (container.current) {
       services.promiseCollector.add(typesetMath(container.current, assertWindow()));
     }
+
+    if (questionContent.current) {
+      questionContent.current.focus();
```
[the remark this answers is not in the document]
PiotrKozlowski: There is an requirment for navigation card:
```
If user tabs to and enters Skip, take them to the next question and put the focus at the top of the modal and read the next question.
```
I guess the same should happen for clicking `Next` button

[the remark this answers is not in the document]
TomWoodward: yea but you're not focusing the top of the area you're focusing the excerpt, would it not be the same functionality and simpler code to focus the container instead?

we should also suppress the focused outline unless the user actually tabs to it

src/app/content/practiceQuestions/components/Answer/index.tsx
PiotrKozlowski: if the input is outside of the label then it should be connected via `for` <-> `id` attributes

src/app/content/components/ContentExcerpt.tsx
PiotrKozlowski: I don't think so

src/app/content/practiceQuestions/components/Question.tsx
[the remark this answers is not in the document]
PiotrKozlowski: Focusing the container is not forcing the Chromevox to read it

About suppressing focus after focusing it with js but not when tabbing - I'm not really sure how to do this?

src/app/messages/en/messages.json
```diff
@@ -103,7 +103,9 @@
   "i18n:practice-questions:popup:intro:start": "Start now",
   "i18n:practice-questions:popup:navigation:finish": "Finish",
   "i18n:practice-questions:popup:navigation:next": "Next",
+  "i18n:practice-questions:popup:navigation:next:aria-label:after-correct": "This answer is correct. Next button.",
```
jarosik10, line 106: Is **`after-correct`** in `18n:practice-questions:popup:navigation:next:aria-label:after-correct` necessary? I believe it's about showing some logic like: _this aria-label is added only if the selected answer was correct_. But I'm not sure if logic like that should affect the name.

PiotrKozlowski: I wanted to make it clear that it is used only after certain event
I've change it to: `i18n:practice-questions:popup:navigation:next:after-submit-correct:aria-label` and `i18n:practice-questions:popup:navigation:show-answer:after-submit-incorrect:aria-label` for Show Answer

PiotrKozlowski:
@TomWoodward We can't have answer's content in the label so we need to allow somehow to focus it so the screen reader can read it
Previously I had tabIndex for input and another one for the answer text, now I've changed it to tabIndex on the whole Answer but it still requires selecting `input` separately to select an answer - we could handle Space and Enter key with `onKeyDown` but then screen reader does not read that an answer was selected (even with `aria-selected`)

TomWoodward:
why can't we put the answer content in the label?

src/app/content/practiceQuestions/components/Answer/index.tsx
TomWoodward: yea but can you put it inside?

PiotrKozlowski: In the behavior doc there is a screen showing a desired outline: https://docs.google.com/document/d/1gsfyJi5AdZSMUhbK6TFfzebXq6qU48sh4aD4MOo4EmY/edit#heading=h.hlfm2ekyfgmi

this outline is around the `label` and it should be visible when the user focuses the `input`
If `input` is inside the `label` then there is no way to style it that way

So currently user is focusing invisible `input` that is forcing the screenreader to read if the choice is selected or not and the visual indicator is around the label

PiotrKozlowski:
@TomWoodward We can't put answer content in the label because some of the answers contain math: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1399

btw. sorry, I was sure that I included this information in the previous comment

TomWoodward:
@PiotrKozlowski i'm not talking about the aria-label i'm talking about the `<label>` i think its fine/better to put all answer text including the math inside the label. can't you just move the "choice A" aria-label to the element that is actually displaying the circle with the letter in it? that is what you're visually translating with that text

PiotrKozlowski:
@TomWoodward Please take a look now. Now whole answer is wrapped in a `<lable>`
However Chromevox wasn't reading the answer content when selecting the `<input>` so I had to add `tabIndex={0}` for the `<label>` (`AnswerBlock`)

TomWoodward (approved):

PR #962, opened 2020-12-09
Re import practice

TomWoodward:
for: openstax/unified#1404


concourse/import-exercises/script.deno.ts
```diff
@@ -45,7 +46,7 @@ for await (const row of readCSVObjects(f)) {
     answers: exercise.questions[0].answers,
     group_uuid: exercise.group_uuid,
     stem_html: exercise.questions[0].stem_html,
-    tags: exercise.group_uuid,
+    tags: exercise.tags,
     uid: exercise.uid,
```
jarosik10, line 50: Deconstructing the `exercise` object probably would make the code more readable.

TomWoodward:
```
const { group_uuid, tags, uid, questions } = exercise;

sourceProblems.push({
    answers: questions[0].answers,
    group_uuid: group_uuid,
    stem_html: questions[0].stem_html,
    tags: tags,
    uid: uid,
});
```

does not seem particularly better to me.

in regular js/ts i might do 

```
sourceProblems.push({
    ...pick(['group_uuid', 'tags', 'uid'], exercise),
    answers: exercise.questions[0].answers,
    stem_html: exercise.questions[0].stem_html,
});
```

but i have not found a lodash build i like that works in deno

what formatting would you suggest?

TomWoodward:
merging because kim wanted to do a round of qa with the new content, but i can make another change for feedback

concourse/import-exercises/script.deno.ts
jarosik10: If you always access `questions[0]` you could also make it as: 
```
const { group_uuid, tags, uid, questions: [question] } = exercise;

sourceProblems.push({
       answers: question.answers,
       group_uuid,
       stem_html: question.stem_html,
       tags,
       uid,
});
```
Now we are getting rid of selecting the first element of arary.

PR #966, opened 2020-12-11
Fix math in practice answers

jarosik10:
for: https://github.com/openstax/unified/issues/1399

### DIFF
--- a/src/app/content/practiceQuestions/components/Answer/styled.tsx
+++ b/src/app/content/practiceQuestions/components/Answer/styled.tsx
@@ -9,6 +9,11 @@ export const AnswerExcerpt = styled.span`
   ${textRegularStyle}
   width: 100%;
   padding: 0;
+  overflow: auto;
+
+  * {
+    overflow: initial;
+  }
 `;
 
 // tslint:disable-next-line: variable-name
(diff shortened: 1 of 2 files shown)


src/app/content/practiceQuestions/components/Answer/styled.tsx
```diff
@@ -9,6 +9,11 @@ export const AnswerExcerpt = styled.span`
   ${textRegularStyle}
   width: 100%;
   padding: 0;
+  overflow: auto;
+
+  * {
+    overflow: initial;
+  }
```
PiotrKozlowski, line 16: Do we need this for something?

PiotrKozlowski (changes requested):
I'll wait with approving this pr for Kim to answer: https://github.com/openstax/unified/issues/1399#issuecomment-743091093

src/app/content/practiceQuestions/components/Answer/styled.tsx
jarosik10: It makes the math visible. Similar styles are added to `<ContentExcerpt>` to make the math visible in question content as well.

PiotrKozlowski (approved):

PR #969, opened 2020-12-14
Fixing errors and warnings on unit tests

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1394


PiotrKozlowski (changes requested):
About the issue with `PageWithMockedHighlighter.spec.tsx`:
If you add `Highlighter.mock.instances[1].getOrderedHighlights.mockReturnValue(mockHighlights);` in the line `158` then it doesn't throw an error, however, there is still an issue with
```
Warning: An update to %s inside a test was not wrapped in act(...).
...
```

I didn't found why this is happening

src/app/components/ScrollLock.tsx
```diff
@@ -39,14 +39,19 @@ const fadeIn = keyframes`
...
   className?: string;
+  onClick?: () => void;
   mobileOnly?: boolean;
   zIndex?: number;
 }
 
 // tslint:disable-next-line:variable-name
 export const Overlay = styled(({ mobileOnly, zIndex, ...props}: OverlayProps) => {
   useDisableContentTabbing();
-  return <div {...props} />;
```
TomWoodward, line 49: you should leave this spread, because the props include everything that you can put on a div

TomWoodward (changes requested):

src/app/components/ScrollLock.tsx
KarinaMendez2: the `onClick` was giving a weird console error, refactoring it the way I did fixed it

KarinaMendez2: or `className`, one of those

KarinaMendez2: nvm something from master fixed the issue, changing it back

TomWoodward (approved):

PiotrKozlowski (approved):

PR #971, opened 2020-12-15
Fix scrollable math

jarosik10:
for: https://github.com/openstax/unified/issues/1399


src/app/content/practiceQuestions/components/Answer/styled.tsx
```diff
@@ -9,18 +9,18 @@ export const AnswerExcerpt = styled.span`
...
 // tslint:disable-next-line: variable-name
 export const AnswerAlignment = styled.div`
   display: flex;
   align-items: center;
   min-height: 4rem;
+  flex: 1;
+
+  * {
+    overflow: initial;
+  }
```
PiotrKozlowski, line 23: I think you need `overflow: initial;` only in `AnswerContent`

PiotrKozlowski (changes requested):

src/app/content/practiceQuestions/components/Answer/styled.tsx
jarosik10: Without
```
* {
    overflow: initial;
  }
```
the math will not be displayed. It seems that MathJax doesn't like `overflow: auto`.

PiotrKozlowski: I've meant leave this part where it was before and add `overflow: initial;` to `AnswerContent`

PiotrKozlowski (approved):

PR #972, opened 2020-12-16
Add SW controller check

jarosik10:
for: https://github.com/openstax/unified/issues/1386


jarosik10:
@TomWoodward  @PiotrKozlowski 
What are your thoughts about this kind of solution?

src/index.tsx
```diff
@@ -105,14 +105,16 @@ startMathJax();
...
-    if (registration && (registration.waiting || registration.installing)) {
+    const serviceWorkerController = window.navigator.serviceWorker.controller;
+    if (registration && (registration.waiting || registration.installing) && serviceWorkerController) {
       app.store.dispatch(updateAvailable());
     } else if (registration) {
       // For Chrome and Edge registration.waiting and registration.installing
       // is still null for some time after .register()
       registration.addEventListener('updatefound', () => {
-        app.store.dispatch(updateAvailable());
+        if (serviceWorkerController) {
```
PiotrKozlowski, line 115: Do we need this `if` here?

src/index.tsx
```diff
@@ -105,14 +105,16 @@ startMathJax();
 serviceWorker.register()
   .then((registration) => {
     app.services.serviceWorker = registration;
-
-    if (registration && (registration.waiting || registration.installing)) {
+    const serviceWorkerController = window.navigator.serviceWorker.controller;
+    if (registration && (registration.waiting || registration.installing) && serviceWorkerController) {
```
PiotrKozlowski, line 109: `&& serviceWorkerController` could be before `(registration.waiting || registration.installing)`

PiotrKozlowski (commented):
It seems like this could work

src/index.tsx
jarosik10: Tbh I do not fully understand the purpose of the first `if` statement.
``` const serviceWorkerController = window.navigator.serviceWorker.controller;
    if (registration && (registration.waiting || registration.installing) && serviceWorkerController) {
      app.store.dispatch(updateAvailable());
    }
```
`updatefound` is also getting fierd for new users, so that `if`checks if the `serviceWorker` `controller` is already present (does nothing for new users). 
```
    else if (registration) {
      registration.addEventListener('updatefound', () => {
        if (serviceWorkerController) {
          app.store.dispatch(updateAvailable());
        }
      });
    }
 ```

Comment (PiotrKozlowski) [reply src/index.tsx:115]: I was also wondering about this. I've left a comment about removing the first if statement here: https://github.com/openstax/unified/issues/1386#issuecomment-741848714 but Tom commented that we need it https://github.com/openstax/unified/issues/1386#issuecomment-745382064
I'm not really sure, but I think there might be the case when the `update` event is fired before we add a listener to the `sw` and this is why we check for `waiting` and `installing`? cc @TomWoodward

**Comment by TomWoodward** [comment]
there are a lot of variables here... browsers check for sw updates on page load on their own, or an uninstalled worker might be hanging out from a previous state.

in the existing logic we check for:
- sw is waiting (installed but not active) or installing (browser is downloading/processing), in either of these cases we immediately dispatch updateAvailable, because the helpers in there will make sure everything is all sync'd and ready and then force the update and reload when the button is clicked
- if neither of those states were present on page load then we wait for the event indicating there is an update, and dispatch updateAvailable then, this would happen if our code evaluated before the browsers check that it does

separately, in pollUpdates, we check release ids and dispatch updateAvailable, in that case we didn't detect a sw update but there would be one because of the release, so we manually trigger the update check in response to the action.

the new problem is that for a first time visitor, they won't have a sw running already, so when the browser installs it for the first time we'll get the event and we process it as an update. in reality this doesn't indicate that there was any application deployment and unconditionally prompting all first time users to reload is not ideal. so even though it means that they will be running without a service worker for their first session, we want to suppress all these update notifications in this case. i think we can accomplish this by doing one if statement at the top of the logic looking for a current controller, and if there isn't one return and skip the rest of the logic. theoretically the browser will prepare the sw in the background and when they close the window and open a new one they'll get it automatically.

File: src/index.tsx [hunk 3]
```diff
@@ -105,7 +105,8 @@ startMathJax();
 serviceWorker.register()
   .then((registration) => {
     app.services.serviceWorker = registration;
-
+    const serviceWorkerController = window.navigator.serviceWorker.controller;
+    if (!serviceWorkerController) { return; }
```
PiotrKozlowski, line 109: I think there is no need to store it in the `const`

TomWoodward (changes requested):
i think this solution will work, but i'm not a fan of those one line if statements, can you add some newlines and maybe a blank line to separate from the existing logic?

src/index.tsx
```diff
@@ -105,7 +105,9 @@ startMathJax();
 serviceWorker.register()
   .then((registration) => {
     app.services.serviceWorker = registration;
+
     if (!window.navigator.serviceWorker.controller) { return; }
```
TomWoodward, line 109:
```suggestion
    if (!window.navigator.serviceWorker.controller) {
      return;
    }
```

TomWoodward (changes requested):

TomWoodward (approved):

PR #973, opened 2020-12-16
Fix display of Read link with flexbox magic

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1407


src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -31,6 +31,7 @@ export const ShowPracticeQuestionsBody = styled(PopupBody)`
 // tslint:disable-next-line: variable-name
 export const SectionTitle = styled.div`
   ${h4Style}
+  flex-shrink: 0;
```
jarosik10, line 34: I think that it might be better to use the `flex` shortcut everywhere. Like `flex: 0 0 auto;` instead `flex-shrink: 0;`.

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
```diff
@@ -78,7 +82,7 @@ export const StyledContentLink = styled.a`
   }
 
   ${theme.breakpoints.mobile(css`
-    margin: 1.2rem;
+    padding: 1.2rem;
```
jarosik10, line 85: Padding makes the clickable area of `StyledContentLink` much bigger. I'm not sure if it's wrong, but I think you should keep that in mind.

PiotrKozlowski: that's okay

src/app/content/practiceQuestions/components/ShowPracticeQuestions.tsx
PiotrKozlowski: `flex: 0 0 auto;` is a equivalent of
```
flex-grow: 0;
flex-shrink: 0;
flex-basis: auto;
```
Is there any advantage of using these properties here?

jarosik10: In this example it doesn't really matter, but since these parameters influence each other, I think it is good practice to use the `flex` shorthand.

jarosik10 (approved):

PR #976, opened 2020-12-18
Add disableTabbing prop to Overlay

jarosik10:
for: https://github.com/openstax/unified/issues/1410


src/app/reactUtils.ts
```diff
@@ -245,4 +245,10 @@ export const disableContentTabbingHandler = () => {
   };
 };
 
-export const useDisableContentTabbing = () => React.useEffect(disableContentTabbingHandler);
+export const useDisableContentTabbing = (disableTabbing: boolean) => {
```
PiotrKozlowski, line 248: It should be called `isActive` or `isEnabled` like in other react utils

src/app/reactUtils.ts
```diff
@@ -245,4 +245,10 @@ export const disableContentTabbingHandler = () => {
...
+  React.useEffect(() => {
+    if (disableTabbing) {
+      disableContentTabbingHandler();
+    }
```
PiotrKozlowski, line 252: `disableContentTabbingHandler()` is returning a function that should be called in the `useEffect` for cleaning up effect
sth like 
```js
const cleanUp = disableContentTabbingHandler();
return cleanUp;
```
but since this is in `if` then I think it will mess up with how hooks are working
I think it would be better to do this the same way as with `useFocusLost`

PiotrKozlowski (changes requested):

PiotrKozlowski (changes requested):
There are tests and lint issues but the code looks good

PiotrKozlowski (approved):

PR #978, opened 2020-12-21
tweak focus on question outline

TomWoodward:
for: openstax/unified#1415


KarinaMendez2 (approved):

TomWoodward:
@PiotrKozlowski see this example of how to suppress focus outline conditionally. the trick is that you have different elements for clicking vs keyboard nav and style them differently. clicking will always focus the innermost focusable element, and `tabindex="-1"` is only clickable, not keyboard tabbable, so the inner element is only for mouse events and the outer element will be hit by tabbing.

i checked with CromeVox and tabbing to the outer element does correctly read out the question content.

in this case i made it so the javascript focuses the inner element, so that the initial state does not have the outline, in other cases, like when you click the "skip to content" link in the header, we do the opposite so that its obvious where the focus ended up.

PiotrKozlowski:
@TomWoodward That's cool, I didn't think about a solution like this, thanks!

PR #980, opened 2020-12-22
Change toolbar tab order

jarosik10:
for: https://github.com/openstax/unified/issues/512

> for 1), i think you can accomplish this by re-arranging the toolbar so there are to divs, one for the left and one for the right, and putting them in a flex container with whatever that property is flex-direction: end-to-start or whatever (i'm sure that is wrong but its something like that), you can accomplish the same thing with floats but i think flex will be better. the point is that the stuff on the right side of the toolbar should be defined above the left side in the markup, that will dictate the tab order.

There is a one problem with this solution. It breaks the `justify-content: space-between;` positioning of `TopBar` elements on smaller devices. So I've came up with solution that just changes the `order` parameter of `SidebarControl` component.


jarosik10:
This is how the opening and closing toc controllers are made in `SidebarControl`:
```
export const SidebarControl: React.SFC<InnerProps> = ({message, hideMobileText, children, ...props}) =>
  <FormattedMessage id={message}>
    {(msg: Element | string) => {
      const txt = assertString(msg, 'Aria label only supports strings');
      return <ToCButton aria-label={txt} {...props}>
        <TocIcon />
        <ToCButtonText hideMobileText={!!hideMobileText}>Table of contents</ToCButtonText>
        {children}
      </ToCButton>;
    }}
  </FormattedMessage>;
```
```
const lockControlState = (isOpen: boolean, Control: React.ComponentType<InnerProps>) =>
  connector((props: MiddleProps) => <Control
    {...props}
    data-testid='toc-button'
    message={isOpen ? openMessage : closedMessage}
    onClick={isOpen ? props.closeToc : props.openToc}
  />);

export const OpenSidebarControl = lockControlState(false, SidebarControl);
export const CloseSidebarControl = lockControlState(true, SidebarControl);
```
I'm not sure how to pass a `ref` of ToC opening button to ToC closing button. Instance of `OpenSidebarControl` is created in `Toolbar`, instance of `CloseSidebarControl` is created as: `SidebarHeaderButton` => `ToC` in `Content`. Maybe it would be easier to find the ToC opening button DOM element in ToC close handler function? cc: @TomWoodward

PiotrKozlowski (approved):
I was thinking that we will change the tab order in a way so the ToC is always before the items in the toolbar, but if we want it to be after them then I think your solution is ok

PiotrKozlowski:
About the second task mentioned in the issue:
> 2. focus toc button after closing the toc

I think you should be able to add`React.useEffect` that listens to `isOpen` changes and if it is set to `false` then focus the button.
We need to establish if this task should also include focusing button after opening a ToC?
Then you can add a third argument to the `lockControlState` that will be responsible to focus only the proper button

TomWoodward:
@PiotrKozlowski 
> I think you should be able to addReact.useEffect that listens to isOpen changes and if it is set to false then focus the button.

this seems reasonable to put in the tocopenbutton itself yea

> We need to establish if this task should also include focusing button after opening a ToC?

after opening the toc i think the focus should be moved to the toc body (after the close button, the container around the nav links), but this isn't specified in the card so we should probably wait and talk with people about it.

> Then you can add a third argument to the lockControlState that will be responsible to focus only the proper button

i'm not sure what this is about

PiotrKozlowski:
@TomWoodward 

>  Then you can add a third argument to the lockControlState that will be responsible to focus only the proper button
i'm not sure what this is about

Now we are creating ToC buttons as:
```
export const OpenSidebarControl = lockControlState(false, SidebarControl);

export const CloseSidebarControl = lockControlState(true, SidebarControl);
```

I think we should do something like
```
export const CloseSidebarControl = lockControlState(true, SidebarControl, (isOpen: boolean | null) => isOpen === false);
```

when third argument is a function that is called in the `useEffect` to check if button should be focused. Third argument could be also optional if we don't want to focus `OpenSidebarControl`

TomWoodward:
@PiotrKozlowski i see now... yea we'll need a way to get our custom behavior in there that is different for the two buttons. maybe it could just be some generic function called like `behaviorHooks` or something that would be called unconditionally with the ref in the body of the component.

```
export const OpenSidebarControl = lockControlState(false, SidebarControl, (ref) => {
  const open = selectOpen();
  useEffect(() => {
    if (!open && ref.current) {
      ref.focus()
     }
  }, [open]);
});

export const CloseSidebarControl = lockControlState(true, SidebarControl);
```

does that make sense?

also kim has approved behavior where the toc body should draw focus when mounted. i believe we already have a drawFocus helper right? she approved my suggestion of putting the focus after the close button so that your first tab will go to the first toc link, whatever element we set the focus on should not have a visible focused state

TomWoodward:
not exactly, you're right that we'll have to add forwardref to the control itself, but i was thinking we'd extract the custom logic outside the lockControlState function, and it would look something like

```
const lockControlState = (isOpen: boolean, Control: React.ComponentType<InnerProps>, hooks: (ref) => void = () => null) =>
  connector((props: MiddleProps) => {
    const ref = React.useRef<HTMLElement>();
    
    hooks(ref);
    
    return <Control
      {...props}
      ref={ref}
      data-testid='toc-button'
      message={isOpen ? openMessage : closedMessage}
      onClick={isOpen ? props.closeToc : props.openToc}
    />
  });
```

TomWoodward:
i'm unsure if the react linter is going to yell at us about this... theoretically it should be fine because its not conditional, although it may want us to name the input variable `useInputHook` or something that starts with `use`

PiotrKozlowski:
Please make sure that you are not focusing the button or the ToC body when `isOpen === null` - it is `null` until a user opens or closes it

We have `useDrawFocus` hook, but we would have to extend it to allow `isEnabled` argument (which may be `true` by default) to make sure we are not focusing elements when `isOpen` is `null`

About hiding the outline from the element after focusing it just wrap the element with additional `div`, with `tabIndex={-1}` and hid `outline` with css for this element

TomWoodward:
i think we can use the drawfocus hook unchanged to focus the toc body when it opens

in order to focus the button in the toolbar when it closes i think we'll need a special effect listening for open changing from false to true, i'm not sure the existing hook is useful at all in this case we might just do it inline

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -324,6 +324,7 @@ export const InnerText = styled.div`
 
 // tslint:disable-next-line:variable-name
 export const SidebarControl = styled(OpenSidebarControl)`
+  order: -1;
```
TomWoodward, line 327: i haven't used the order property before but it does seem perfect for what we're doing, good find

TomWoodward (commented):
i think this is actually working well as-is for closing the toc, i don't think we have to change anything there, which simplifies things

i still think it would be nice to not have to tab through the close button when first going into the toc after opening it, i think if we put drawFocus on the toc body it would fix that

this is working really well already though

PiotrKozlowski:
I think `useDrawFocus` will cause focusing ToC on initial page load on desktop, but maybe this is not the case
if it will I propose to add a `isEnabled` argument to this hook

TomWoodward:
i think youre right that it will cause issues on page load. we may have to do something else that looks for the open flag becoming true

TomWoodward:
@jarosik10 since the additional feature i was requesting seems like its going to have complications, kim and i have agreed that we should just merge what you have already done, since it is a significant improvement to the toolbar tab order and closing the toc

TomWoodward:
i'm pretty sure you've already addressed all the ac in the card, is there something you're waiting for before submitting this for review?

PiotrKozlowski (approved):

jarosik10:
Do you want to merge this PR or do you want to put there the entire solution for: https://github.com/openstax/unified/issues/512?  Should I assign myself again to this issue and start working on the ToC button part?
@TomWoodward @kjdav @PiotrKozlowski

PiotrKozlowski:
I think it is waiting for qa approval and then it will be merged before release

PR #1027, opened 2020-12-28
upgrade typescript version: 3.9.6

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1419

Attempt to upgrade to **4.1.3** will be in another PR, due to node issues (also dealt with in another PR)


PiotrKozlowski (commented):
> - The operand of a 'delete' operator must be optional.

Is this because we are doing something like this: ?
```ts
delete window.document.getSelection;
```

since `getSelection` is not an optional argument we shouldn't remove it
if we really want to remove it we should probably change it to:
```ts
delete (window.document as any).getSelection;
```

it seems like this is a good rule to have

src/helpers/applicationUpdates.spec.ts
```diff
@@ -90,7 +90,7 @@ describe('findAndInstallServiceWorkerUpdate', () => {
     const update = jest.fn();
     const sw = {update: update as ServiceWorkerRegistration['update']} as ServiceWorkerRegistration;
 
-    let finishUpdate: () => void = () => null;
+    let finishUpdate: (value?: any) => void = () => null;
```
PiotrKozlowski, line 93: since it is `any` you shouln't need to use `?`

PiotrKozlowski (changes requested):
Some checks are failing, but other than this it looks fine

src/helpers/applicationUpdates.spec.ts
KarinaMendez2: this means calling `finishUpdate({});` instead of `finishUpdate();`, thought it was cleaner adding `?`

src/helpers/loadFont.ts
```diff
@@ -5,7 +5,7 @@ export default (url: string) => new Promise((resolve) => {
   }
 
   if (document.head.querySelector(`link[href="${url}"]`)) {
-    resolve();
+    resolve({});
```
PiotrKozlowski, line 8: Ok, I've looked one more time.
It seems like we need to declare that this promise is returning `void` and then fix a little bit the overwrite for `onload`
```ts
export default (url: string) => new Promise<void>((resolve) => {
  if (typeof(document) === 'undefined' || !document.head) {
    throw new Error('fonts can only be loaded in the browser');
  }

  if (document.head.querySelector(`link[href="${url}"]`)) {
    resolve();
    return;
  }

  const link = document.createElement('link');
  link.setAttribute('rel', 'stylesheet');
  link.setAttribute('href', url);
  link.onload = () => resolve();
  document.head.appendChild(link);
});
```

src/helpers/applicationUpdates.spec.ts
```diff
@@ -90,12 +90,12 @@ describe('findAndInstallServiceWorkerUpdate', () => {
     const update = jest.fn();
     const sw = {update: update as ServiceWorkerRegistration['update']} as ServiceWorkerRegistration;
 
-    let finishUpdate: () => void = () => null;
+    let finishUpdate: (value: any) => void = () => null;
```
PiotrKozlowski, line 93: I've looked again on this
We can do `let finishUpdate: () => void = () => null;`
and then instead of
```
update.mockReturnValue(new Promise((resolve) => finishUpdate = resolve));
```
we have to do:
```
update.mockReturnValue(new Promise<void>((resolve) => finishUpdate = resolve));
```

src/app/content/practiceQuestions/components/Question.tsx
```diff
@@ -24,12 +24,15 @@ export const QuestionWrapper = styled.form`
 `;
 
 // tslint:disable-next-line: variable-name
-export const QuestionContent = styled(ContentExcerpt)`
+export const QuestionContent = styled(React.forwardRef((props, ref) => <div tabIndex={0}>
+  <ContentExcerpt {...props} tabIndex={-1} ref={ref} />
+</div>))`
```
PiotrKozlowski, line 29: I'm not sure why these changes are in this or?

PiotrKozlowski (changes requested):

src/app/content/practiceQuestions/components/Question.tsx
KarinaMendez2: I did something weird while interactive rebasing lol don't worry about this, I'll remove them

KarinaMendez2: I did something weird while interactive rebasing lol don't worry about this file, I'll remove it

package.json
```diff
@@ -39,7 +39,7 @@
     "styled-components": "^4.3.2",
     "styled-icons": "^8.1.0",
     "typesafe-actions": "^4.4.2",
-    "typescript": "~3.5.2",
+    "typescript": "~3.9.6",
```
PiotrKozlowski, line 42: so we are going back to `3` or is this because of rebasing?

KarinaMendez2: node issues with 4.1 version, will deal with node in a separate PR

PiotrKozlowski: should then the `node` be updated before the final version of this or?

KarinaMendez2: 3.9.6 has no issues while running checks locally, Tom wanted to have at least some upgrade so we're merging 3.9.6 while I update node in another PR (there's a card for it since a while ago), and then upgrade typescript again, trying to upgrade both and match dependencies in this PR wasn't working out.

PiotrKozlowski: cool

PiotrKozlowski (approved):

src/app/content/components/BookBanner.spec.tsx
```diff
@@ -26,7 +26,7 @@ describe('BookBanner', () => {
     resetModules();
 
     window = assertWindow();
-    delete window.location;
+    delete (window as any).location;
```
TomWoodward, line 29: why are all these as any casts here?

TomWoodward (changes requested):

src/app/content/components/BookBanner.spec.tsx
PiotrKozlowski: it was added due to the some rule from typescript 4^ that disallows removing properties that are not marked as optional
https://github.com/openstax/rex-web/pull/1027#pullrequestreview-560422300

TomWoodward: So they aren't necessary for the 3.9 upgrade?

src/app/errors/reducer.spec.ts
```diff
@@ -33,7 +42,10 @@ describe('error reducer', () => {
       search: '',
       state: {},
     };
-    const match = {route: notFound};
+    const match = {
+      params: routeParams,
```
TomWoodward, line 46: These aren't required for the not found route right?

src/app/navigation/utils.ts
```diff
@@ -57,10 +58,11 @@ export const findRouteMatch = (routes: AnyRoute[], location: Location): AnyMatch
...
 export const matchSearch = (action: AnyMatch, search: string | undefined) => {
   const previous = querystring.parse(search || '');
-
-  const route = querystring.parse(isMatchWithParams(action)
+  const matchesParams = isMatchWithParams(action);
+  const route = querystring.parse(
+    matchesParams
     ? action.route.getSearch ? action.route.getSearch(action.params) : ''
-    : action.route.getSearch ? action.route.getSearch() : ''
+    : action.route.getSearch ? action.route.getSearch({} as Params) : ''
```
TomWoodward, line 65: These changes don't seem right

TomWoodward (changes requested):

src/app/content/components/BookBanner.spec.tsx
PiotrKozlowski: I think they are not required for 3.9 but from what I've talked with Karina we will still want to update to ts ^4 after we update the `node`
Do you prefer to remove all unnecessary fixes for the current version?

TomWoodward:
closing in favor of https://github.com/openstax/rex-web/pull/1043

PR #1028, opened 2020-12-28
Issue 1389

jarosik10:
for: https://github.com/openstax/unified/issues/1389


PiotrKozlowski (changes requested):
The issue is a little bit more complex than the proposed solution. What we want to achieve is to remove `openPracticeQuestions` action and handle `Practice` button as a `ContentLink` component that will change the location (add `?modal=PQ` param`) and then we will detect that URL was changed and we will open the modal.

Same for closing - it should trigger a `locationChange` without the modal param. So `closePracticeQuestions` should not be used anymore I believe.

I suggest you revert changes to `highlights` and `studyGuides` and focus on implementing these changes only for Practice Questions. When we will accept the solution then you can implement it for other modals (it is not required for this issue)

The final effect is that opening and closing all modals (PQ, SG, MH) should be handled as changing location rather than showing them. In the future maybe we will change the `modal=PQ` to another directory, like `/practice-questions/:questionNumber?selectedAnswer=b` etc. so Google can index it better and we can use links like this to share between users

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
```diff
@@ -25,24 +45,29 @@ const PracticeQuestionsText = styled.span`
...
 
-  if (!isEnabled || !hasPracticeQs) { return null; }
+  if (!isEnabled || !hasPracticeQs || !book || !page) { return null; }
 
-  const openPracticeQuestionsSummary = () => {
-    dispatch(openPracticeQuestions());
-    trackOpenClose();
+  const search = {
+    modal: 'PQ',
+    query: undefined,
```
PiotrKozlowski, line 57: I think you shouldn't have to specify the `query` and there should be a `constant` with a `PQ` value

src/app/content/components/Toolbar/__snapshots__/PracticeQuestions.spec.tsx.snap
```diff
@@ -2,21 +2,4 @@
...
-    src="practiceQuestionsIcon.svg"
-  />
-  <span
-    className="PracticeQuestionsButton__PracticeQuestionsText-vyu0p6-2 fJxtws"
-  >
-    Practice
-  </span>
-</button>
-`;
+exports[`practice questions button renders if feature flag is enabled and there are practice questions 1`] = `null`;
```
PiotrKozlowski, line 5: this shouldn't be `null`

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
```diff
@@ -22,15 +23,18 @@ const PracticeQuestionsPopup = () => {
...
 
   const closeAndTrack = React.useCallback((method: string) => () => {
     if (currentQuestionIndex !== null) {
       const message = intl.formatMessage({ id: 'i18n:practice-questions:popup:warning-before-close' });
       if (!assertWindow().confirm(message)) { return; }
     }
-    dispatch(closePracticeQuestions());
+    if (match) {
+      dispatch(goBack(match));
+    }
```
PiotrKozlowski, line 35: it is unclear to me why we want this dispatch being called only when `match` is truth?
Also, I think it shouldn't be `goBack` action and instead, it should be `PUSH`.
Let's imagine the case when the user went to page A where he found a link to the page with a modal query in a URL
After clicking this link he will go to the page and modal will be opened. Then the user clicks X icon and he goes back to the page where he started, but what he really wanted was to close the modal.
Instead of going back we should remove `modal=PQ` from the current query and call PUSH with the remaining URL

src/app/content/practiceQuestions/reducer.ts
```diff
@@ -21,23 +21,18 @@ export const initialState: State = {
 const reducer: Reducer<State, AnyAction> = (state = initialState, action): State => {
   switch (action.type) {
     case getType(locationChange):
-      const shouldBeOpen = action.payload.query[modalQueryParameterName] === modalUrlName
-        && action.payload.action === 'PUSH';
+      const shouldBeOpen = action.payload.query[modalQueryParameterName] === modalUrlName;
```
PiotrKozlowski, line 24: I think this will not be that simple, because not every book has Practice questions so if someone enters the URL with `modal=PQ` for a book that does not have PQ then we shouldn't show the modal.

PiotrKozlowski (changes requested):

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
jarosik10: Now when it is changed to `push`, it cannot be called without `match`. So i think that the `if` statement should stay.

PiotrKozlowski: Can `match` not be `truthy` when `closeAndTrack` is called?

src/app/content/practiceQuestions/reducer.ts
PiotrKozlowski: exactly, this is a problem to solve
Maybe we should remove `open` property from the redux state and instead have `isOpen` selector that will look for `modal=PQ` and `bookHasPracticeQuestions` etc?

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
jarosik10: Hmm.. I think not, but I will have to make sure. If it can not be `falsy`, then I'll have to use typescript's non-null exclamation mark, right?

src/app/content/practiceQuestions/reducer.ts
jarosik10: But does it fix the timing issue? `bookHasPracticeQuestions` will be `false` during the `locationChange` action.

PiotrKozlowski: selector should change when it will become true

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
PiotrKozlowski: yes

PiotrKozlowski: please use `assertNonNull` helper function

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
```diff
@@ -22,15 +23,18 @@ const PracticeQuestionsPopup = () => {
...
   const intl = useIntl();
+  const match = useSelector(navigation.match);
 
   const closeAndTrack = React.useCallback((method: string) => () => {
     if (currentQuestionIndex !== null) {
       const message = intl.formatMessage({ id: 'i18n:practice-questions:popup:warning-before-close' });
       if (!assertWindow().confirm(message)) { return; }
     }
-    dispatch(closePracticeQuestions());
+    if (match) {
```
PiotrKozlowski, line 33: Can this be replaced with `AssertNotNull`?

src/app/content/practiceQuestions/reducer.ts
PiotrKozlowski: I've meant remove `open` from this reducer and change `practiceQuestionsOpen` selector to look for `modal: PQ` in `navigation.query` + `hasPracticeQuestions` selector
You should be able to do something like:
```ts
export const practiceQuestionsOpen = createSelector(
  hasPracticeQuestions,
  navigationState
  (hasPQ) => hasPQ && navigationState.query && navigationState.query.modal === 'PQ'
);
```

PiotrKozlowski (changes requested):

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
```diff
@@ -1,18 +1,40 @@
...
+import { toolbarIconColor } from '../constants';
+import { buttonMinWidth } from './styled';
 import { PlainButton, toolbarDefaultButton, toolbarDefaultText } from './styled';
 
 // tslint:disable-next-line:variable-name
 export const PracticeQuestionsWrapper = styled(PlainButton)`
   ${toolbarDefaultButton}
 `;
 
+// TODO: refactor the styling of Toolbar ContentLinks
```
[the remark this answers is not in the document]
jarosik10: I wanted to refactor that styling when all modals will be open/close by modal-url.

src/app/content/components/ContentLink.tsx
```diff
@@ -53,8 +53,16 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     ...anchorProps
   } = props;
   const {url, params} = getBookPageUrlAndParams(book, page);
-  const relativeUrl = toRelativeUrl(currentPath, url);
+  let relativeUrl = toRelativeUrl(currentPath, url);
```
PiotrKozlowski, line 56: I think that maybe `toRelativeUrl` should accept the argument with search options or we should leave this as it was (as `const`) and then create `const URL = ${relativeUrl}?${options.search};`

not a big deal

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.spec.tsx
```diff
@@ -26,21 +30,33 @@ jest.mock('react-dom', () => ({
...
   let store: Store;
   let services: ReturnType<typeof createTestServices>;
   let container: HTMLElement;
-  let dispatch: jest.SpyInstance;
 
   beforeEach(() => {
-    store = createTestStore();
-    dispatch = jest.spyOn(store, 'dispatch');
+    store = createTestStore({
+      content: initialContentState,
```
PiotrKozlowski, line 48: this is not needed

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.spec.tsx
```diff
@@ -105,14 +118,13 @@ describe('PracticeQuestions', () => {
       closeButton.props.onClick();
     });
 
-    expect(dispatch).toHaveBeenCalledWith(closePracticeQuestions());
     expect(track).toHaveBeenCalled();
   });
 
-  it('closes popup on esc and tracks analytics', async() => {
+  it('tracks analytics when clicking esc', async() => {
```
PiotrKozlowski, line 124: would be nice to check if `push` action  without modal parameter was called in these tests

src/app/content/practiceQuestions/hooks/locationChange.spec.ts
```diff
@@ -85,8 +99,46 @@ describe('locationChange', () => {
...
+
+    const section = archiveTreeUtils.findArchiveTreeNodeById(book.tree, shortPage.id) as LinkedArchiveTreeSection;
+
+    expect(getSummary).toHaveBeenCalled();
+    expect(dispatch).toHaveBeenCalledWith(receivePracticeQuestionsSummary(mockSummaryResponse));
+    expect(spySection).toHaveBeenCalled();
+    expect(dispatch).not.toHaveBeenCalledWith(setSelectedSection(section));
+  });
+
+  it('doesnt set selected section if there are no practice questions', async() => {
```
PiotrKozlowski, line 128: please add a check to make sure that seSelectedSection is not called

src/app/content/practiceQuestions/hooks/locationChange.spec.ts
```diff
@@ -85,8 +99,46 @@ describe('locationChange', () => {
...
     await hook();
 
+    const section = archiveTreeUtils.findArchiveTreeNodeById(book.tree, shortPage.id) as LinkedArchiveTreeSection;
+
     expect(getSummary).not.toHaveBeenCalled();
     expect(dispatch).not.toHaveBeenCalledWith(receivePracticeQuestionsSummary(mockSummaryResponse));
+    expect(dispatch).toHaveBeenCalledWith(setSelectedSection(section));
+  });
+
+  it('only fetch practice questions if section is undefined', async() => {
```
PiotrKozlowski, line 109: I'm not sure what this test is about? When we are fetching PQ only when `section` is undefined in `locationChange`?

src/app/content/practiceQuestions/reducer.spec.ts
```diff
@@ -1,32 +1,7 @@
-import { locationChange } from '../../navigation/actions';
 import * as actions from './actions';
 import reducer, { initialState } from './reducer';
 
 describe('practice questions reducer', () => {
-  it('keeps modal open on location change if modal query is present', () => {
```
PiotrKozlowski, line 6: you could add check for `locationChange` to make sure that we are handling it properly

src/app/content/practiceQuestions/selectors.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    currentQuestionIndex: null,
+    isEnabled: false,
+    questionAnswers: {},
+    questions: [],
+    selectedSection: null,
+    summary: null,
+  };
+
+describe('isPracticeQuestionsOpen', () => {
+    it('return true if has practice questions and has modal url', () => {
```
PiotrKozlowski, line 15: please also check if it is returning `false` otherwise

PiotrKozlowski (changes requested):

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.spec.tsx
jarosik10: Are you sure? `store` is used in a `Provider`.

src/app/content/practiceQuestions/reducer.spec.ts
jarosik10: Is it really needed? 
`locationChange` reducer is very simple right now:
```
case getType(locationChange):
   return {
      ...state,
      currentQuestionIndex: null,
      questionAnswers: {},
      questions: [],
      selectedSection: null,
   };
```

PiotrKozlowski: I feel like this might be something that will easly break if we update somehow this reducer in the future and we will not get idea why something might not work
This test will be easy to write and just might become usefull in the future
this is not blocking comment

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -22,11 +21,26 @@ const loadSummary = async(services: MiddlewareAPI & AppServices): Promise<Practi
...
+  const state = services.getState();
+  const { book, page } = bookAndPage(state);
+  const isEnabled = practiceQuestionsEnabled(state);
 
-  if (!practiceQuestionsSummary) { return; }
+  if (!book || !page || !isEnabled ) { return; }
 
-  services.dispatch(receivePracticeQuestionsSummary(practiceQuestionsSummary));
+  let practiceQuestionsSummary;
+  if (!hasPracticeQuestions(services.getState())) {
```
TomWoodward, line 31: why do you have all this conditional setting of `practiceQuestionsSummary` like this? the same condition noops down on line 38, can't you just move that up?

src/app/content/practiceQuestions/selectors.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { AppState } from '../../types';
+import { modalUrlName } from './constants';
+import * as selectors from './selectors';
+
+const initialState = {
```
TomWoodward, line 5: whitespace in this whole file is wrong

TomWoodward (changes requested):

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
```diff
@@ -18,19 +20,20 @@ const PracticeQuestionsPopup = () => {
...
+  const match = useSelector(navigation.match);
+  const isPracticeQuestionsOpen = useSelector(pqSelectors.isPracticeQuestionsOpen);
 
   const closeAndTrack = React.useCallback((method: string) => () => {
     if (currentQuestionIndex !== null) {
       const message = intl.formatMessage({ id: 'i18n:practice-questions:popup:warning-before-close' });
       if (!assertWindow().confirm(message)) { return; }
     }
-    dispatch(closePracticeQuestions());
+    dispatch(push(assertDefined(match, 'match should be always defined at this step')));
```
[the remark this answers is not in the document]
TomWoodward: asking about it in slack https://openstax.slack.com/archives/C8V1XET1R/p1610377788352600

src/app/content/components/ContentLink.tsx
```diff
@@ -55,6 +55,11 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
   const {url, params} = getBookPageUrlAndParams(book, page);
   const relativeUrl = toRelativeUrl(currentPath, url);
   const bookUid = stripIdVersion(book.id);
+  // Add options only if linking to the same book
+  const options = currentBook && currentBook.id === bookUid
+    ? createNavigationOptions(search, scrollTarget)
+    : undefined;
+  const URL = options && options.search ? `${relativeUrl}?${options.search}` : relativeUrl;
```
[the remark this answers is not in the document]
jarosik10: Yeah, it makes sense. But how exactly should that parsing work? I mean the order, etc.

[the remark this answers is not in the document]
TomWoodward: pretty sure at this point it is just

```typescript
const navigationOptionsToString = (options: ReturnType<createNavigationOptions>) => options.hash + (options.search ? `?${options.search}` : '')
```

but it'd be nice to define it right next to `createNavigationOptions` so they can be maintained together

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
```diff
@@ -25,24 +42,24 @@ const PracticeQuestionsText = styled.span`
...
-  };
+  if (!isEnabled || !hasPracticeQs || !book || !page) { return null; }
 
   return <FormattedMessage id='i18n:toolbar:practice-questions:button:text'>
     {(msg: Element | string) =>
-      <PracticeQuestionsWrapper onClick={openPracticeQuestionsSummary} aria-label={msg}>
+      <StyledContentLink
+        book={book}
+        page={page}
+        search={{ [modalQueryParameterName]: modalUrlName, query: undefined }}
```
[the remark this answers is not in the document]
jarosik10: I know, but it just doesn't work if `query: undefined` is not included. In my opinion, there must be a logical error in `ContentLink` then.
```
export const ConnectedContentLink = connect(
  (state: AppState, ownProps: {search?: { query?: string | null }}) => ({
    currentBook: select.book(state),
    currentPath: selectNavigation.pathname(state),
    hasUnsavedHighlight: hasUnsavedHighlightSelector(state),
    search: ({
      query: selectSearch.query(state),
      ...(ownProps.search ? ownProps.search : {}),
    }),
  }),
  (dispatch: Dispatch) => ({
    navigate: flow(push, dispatch),
  })
)(ContentLink);
  ```

File: src/app/content/practiceQuestions/hooks/locationChange.ts (hunk 14 above)
Comment (jarosik10) [reply src/app/content/practiceQuestions/hooks/locationChange.ts:31]: 1st `hasPracticeQuestion` could be actually changed into `practiceQuestionsSummary` selector to check if summary isn't already loaded.
2nd `hasPracticeQuestion` checks if questionSummary actually has questions.

When user enters a page with PQ enabled, `loactionChange` hook loads the PQ summary. The first if statement prevents the summary reload after opening the modal.

File: src/app/content/practiceQuestions/components/PracticeQuestionsPopup.spec.tsx [hunk 19]
```diff
@@ -152,16 +188,17 @@ describe('PracticeQuestions', () => {
     element.dispatchEvent(event); // this checks for bindings using addEventListener
     ReactTestUtils.Simulate.click(element, {preventDefault}); // this checks for react onClick prop
 
-    expect(dispatch).toHaveBeenCalledWith(closePracticeQuestions());
     expect(track).toHaveBeenCalled();
+    expect(dispatch).toHaveBeenCalledWith(replace(mockMatch));
   });
 
-  it('show warning prompt and closes popup after confirm', async() => {
+  it('show warning prompt and tracks analytics after confirm', async() => {
```
PiotrKozlowski, line 195: there should be a check for `replace/goBack` action

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.spec.tsx
```diff
@@ -183,17 +220,17 @@ describe('PracticeQuestions', () => {
 
     expect(spyConfirm)
       .toHaveBeenCalledWith('Are you sure you want to exit this page? Your progress will not be saved.');
-    expect(dispatch).toHaveBeenCalledWith(closePracticeQuestions());
     expect(track).toHaveBeenCalled();
   });
 
-  it('show warning prompt and do not close popup after cancel', async() => {
+  it('show warning prompt and do not tracks analytics after cancel', async() => {
```
PiotrKozlowski, line 226: there should be a check that `replace/goBack` action wasn't called

src/app/content/practiceQuestions/hooks/locationChange.spec.ts
PiotrKozlowski: i still don't get it, the only `section` i can see in `locationChange` is this line:
```
const section = findArchiveTreeNodeById(book.tree, page.id);
```
but we are not fetching anything related to this section

maybe the test name should be `do not set selected section if it is undefined`?

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -22,11 +21,26 @@ const loadSummary = async(services: MiddlewareAPI & AppServices): Promise<Practi
...
+    practiceQuestionsSummary = await loadSummary(services, book);
+  }
+  if (practiceQuestionsSummary) {
+    services.dispatch(receivePracticeQuestionsSummary(practiceQuestionsSummary));
+  }
+
+  if (!hasPracticeQuestions(services.getState())) { return; }
```
PiotrKozlowski, line 38:
```suggestion
  if (!hasPracticeQuestions(state)) {
    const practiceQuestionsSummary = await loadSummary(services, book);
    if (practiceQuestionsSummary) {
      services.dispatch(receivePracticeQuestionsSummary(practiceQuestionsSummary));
    }
    const updatedState = services.getState();
    if (!hasPracticeQuestions(updatedState)) { return; }
  }
```

PiotrKozlowski (changes requested):

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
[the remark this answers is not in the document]
jarosik10: I've found the solution. It's about the `createNavigationError`. My suggestion is to change from:
```
export const createNavigationOptions = (
  search: { query?: string | null },
  scrollTarget?: ScrollTarget
) => ({
  hash: scrollTarget ? scrollTarget.elementId : undefined,
    ...(search.query === null ? {} : search),
    target: scrollTarget ? JSON.stringify(omit('elementId', scrollTarget)) : undefined,
  }),
});
```
to:
```
export const createNavigationOptions = (
  search: { query?: string | null },
  scrollTarget?: ScrollTarget
) => {
  const { query, ...otherSearchParams } = search;

  return {
    hash: scrollTarget ? scrollTarget.elementId : undefined,
      ...(query === null ? otherSearchParams : search),
      target: scrollTarget ? JSON.stringify(omit('elementId', scrollTarget)) : undefined,
    }),
  };
};
```
cc: @PiotrKozlowski @TomWoodward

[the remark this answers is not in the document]
PiotrKozlowski: Ok, I think I get it. When you use `ContentLink` and it is by default `ConnectedContentLink` which is always getting the `query` from the state - which is `null` in this case so you have to override it with `undefined` to pass proper argument to the `queryString.stringify`.

I think we just want to pass all search params but replace `null` values with `undefined`

Maybe we could replace
```
...(search.query === null ? {} : search),
```
with 
```
...omitBy(search, isNull),
```
?

src/app/content/practiceQuestions/reducer.ts
```diff
@@ -29,7 +30,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action): State
     case getType(receiveFeatureFlags):
       return {...state, isEnabled: action.payload.includes(practiceQuestionsFeatureFlag)};
     case getType(actions.receivePracticeQuestionsSummary):
-      return {...state, summary: action.payload};
+      return {...state, loading: true, summary: action.payload};
```
PiotrKozlowski, line 33: shouldn't this be set on `locationChange` since we dispatch `receivePracticeQuestionsSummary` only once per book and the screen depends on `questionsCount` selector which depends on `setQuestions`?

There is also `setSelectedSection` hook.
I think we should set loading before we call
```
await practiceQuestionsLoader.getPracticeQuestions
```

It may not be usefull right now since we are getting these questions from prefetched jsons, but the intention of `practiceQuestionsLoader` is that we shouldn't expect anything from it

src/app/content/components/Toolbar/PracticeQuestionsButton.tsx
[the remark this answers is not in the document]
jarosik10: Yeah, it should be fine. Forgot about loadash. :P

src/app/content/practiceQuestions/reducer.ts
jarosik10: There are 3 different actions scenarios:
1. Open modal directly via modal-ulr
`location change` -> `receivePracticeQuestionsSummary` -> `setSelectedSection` -> `setQuestions`
2. Open modal with button
`location change` -> `setSelectedSection` -> `setQuestions`
3. New section with question (Intro or Final)
`setSelectedSection` -> `setQuestions`

Setting `loading: true;` for `location change` and `setSelectedSection` would cover all scenarios, but it gives a little redundancy.

PiotrKozlowski: There will be 4th scenario when we implement #941 - selectin a section from a chapter filter
I think all the cases have call to `setSelectedSection` so I would go with `loading: true;` for `setSelectedSection` and then `loading: false` for `setQuestions`
Will this cause delay in the first scenario?

jarosik10: > There will be 4th scenario when we implement #941 - selecting a section from a chapter filter

Sounds similar to 3rd scenario in terms of actions.
> Will this cause delay in the first scenario?

It would cause delay in 1st and 2nd scenario.

PiotrKozlowski: right, it's basically the same
What about `loading: true` for both `setSelectedSection` and `locationChange` and `loading: false` for `setQuestions`?

jarosik10: It handles all scenarios.

PiotrKozlowski (changes requested):

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -22,11 +21,27 @@ const loadSummary = async(services: MiddlewareAPI & AppServices): Promise<Practi
 };
 
 const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
-  const practiceQuestionsSummary = await loadSummary(services);
+  const state = services.getState();
```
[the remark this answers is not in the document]
jarosik10: Hmm.. but should I move the first `hasPracticeQuestions` check into the `loadSummary` to remove one `if` check from `hookBody` or almost everything like it was before? But in this scenario - if `book` or `page` are `undefined`, it will throw error in `findArchiveTreeNodeById(book.tree, page.id);`. 

But moving the first `hasPracticeQuestions` into the `loadSummary` will make us use `getState()` again (3rd time).

src/app/content/practiceQuestions/reducer.spec.ts
```diff
@@ -41,7 +23,40 @@ describe('practice questions reducer', () => {
       ...initialState,
       currentQuestionIndex: 0,
     };
+
     const newState = reducer(state, actions.nextQuestion());
+
     expect(newState.currentQuestionIndex).toEqual(1);
   });
+
+  it('reduces locationChange', () => {
```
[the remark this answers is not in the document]
jarosik10: I'm checking only the properties which should be changed.
```
    expect(newState.currentQuestionIndex).toEqual(initialState.currentQuestionIndex);
    expect(newState.questionAnswers).toEqual(initialState.questionAnswers);
    expect(newState.questions).toEqual(initialState.questions);
    expect(newState.selectedSection).toEqual(initialState.selectedSection);
    expect(newState.loading).toBe(true);
```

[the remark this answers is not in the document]
PiotrKozlowski: I see only 
```
    expect(newState.currentQuestionIndex).toEqual(initialState.currentQuestionIndex);
    expect(newState.questionAnswers).toEqual(initialState.questionAnswers);
    expect(newState.questions).toEqual(initialState.questions);
    expect(newState.selectedSection).toEqual(initialState.selectedSection);
    ```
    
and this check will pass because it is not checking for other properties so reducer for `locaitonChange` could update 10 other properties and this check would still pass if properties specified here stayed the same

I think it would be better to check:
```
expect(newState).toEqual({
...initialState,
loading: true,
})
```

File: src/app/content/practiceQuestions/hooks/locationChange.ts (hunk 23 above)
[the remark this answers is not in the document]
Comment (PiotrKozlowski) [reply src/app/content/practiceQuestions/hooks/locationChange.ts:24]: In the hookbody you should be able to have something like this:
```ts
  const practiceQuestionsSummary = await loadSummary(services);
  if (!practiceQuestionsSummary) { return; }
  
  services.dispatch(receivePracticeQuestionsSummary(practiceQuestionsSummary));
  
  const section = findArchiveTreeNodeById(book.tree, page.id);
  if (section && isLinkedArchiveTreeSection(section)) {
    services.dispatch(setSelectedSection(section));
  }
```

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -1,32 +1,50 @@
...
   try {
     return await services.practiceQuestionsLoader.getPracticeQuestionsBookSummary(book.id);
   } catch (error) {
     Sentry.captureException(error);
   }
 };
 
+const setSection = (services: MiddlewareAPI & AppServices, book: Book, page: Page) => {
+  // loading hasPracticeQuestions beacuse the state could be changed by receivePracticeQuestionsSummary
+  if (!hasPracticeQuestions(services.getState())) { return; }
```
PiotrKozlowski, line 25: noop also when there is no query.modal=PQ

PiotrKozlowski (changes requested):

PiotrKozlowski (approved):

src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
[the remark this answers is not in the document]
TomWoodward: yea this is good

src/app/navigation/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { flatten, unflatten } from 'flat';
 import { Action, Location } from 'history';
+import { isNull, omitBy } from 'lodash';
```
TomWoodward, line 3: when using lodash functions please import them one by one like this:
```
import isNull from 'lodash/isNull';
import omitBy from 'lodash/omitBy';
```

this reduces the bundle size.

TomWoodward, line 3: also i prefer `lodash/fp`, which auto-curries the functions and allows you to write things like
```
import isNull from 'lodash/fp/isNull';
import omitBy from 'lodash/fp/omitBy';

const omitByNull = omitBy(isNull);

const filteredSearch = omitByNull(search);
```

not that it really matters a lot in this case but consistency is nice

TomWoodward (approved):

src/app/navigation/utils.ts
jarosik10: I see:) Do you want me to change it?

TomWoodward: if you want to search the project and make another pr to fix any places that are doing this, it looks like at least one more in `Dropdown.tsx`

jarosik10: So should I create a special card for it?

TomWoodward: sure, you can make a card in openstax/unified called like `fix imports to reduce bundle size` or something

PR #1031, opened 2021-01-04
Resolve new rap links

KarinaMendez2:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1341


src/app/content/types.ts
```diff
@@ -42,7 +42,10 @@ export interface State {
 }
 
 export interface PageReferenceMap extends PageReference {
-  match: string;
+  match: {
+    rapMatch: string | undefined,
```
TomWoodward, line 46: why are you pulling around two different types of matches? can't you make them both fit into one match format?

TomWoodward (changes requested):

src/app/content/types.ts
KarinaMendez2: because when` contentLinkHandler.ts` searches for hrefs to resolve, rap links come like `href=archive or whatever/something` and the `normal` links in the content come like `href=contents/blah`

TomWoodward: so what is the problem?

KarinaMendez2: That I need to match both href types to replace them with the solved url, I can make it prettier I guess, but I still need to make 2 matches somewhere

TomWoodward: why do you need two? can't you match either link format and put the data in a consistent match format?

KarinaMendez2: Yeah, will work on it.

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -219,8 +224,8 @@ const loadContentReference = async(
       page: getUrlParamForPageId(targetBook, reference.pageUid),
     },
     state: {
-      bookUid: targetBook.id,
-      bookVersion: targetBook.version,
+      bookUid: reference.bookId || targetBook.id,
```
TomWoodward, line 227: Instead of conditions here I think what you want to do is use your reference input to conditionally bypass the target book logic above so that it doesn't make that extra query

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -206,7 +206,12 @@ const loadContentReference = async(
   services: AppServices & MiddlewareAPI,
   book: Book,
   page: ArchivePage,
-  reference: ReturnType<typeof getContentPageReferences>[0]
+  reference: {
```
TomWoodward, line 209: Can you leave this the way it was? Shouldn't it just pick up the new return type?

src/app/content/utils.ts
```diff
@@ -19,8 +19,8 @@ export { getBookPageUrlAndParams, getPageIdFromUrlParam, getUrlParamForPageId, t
 export { stripIdVersion } from './utils/idUtils';
 export { scrollSidebarSectionIntoView } from './utils/domUtils';
 
-export const getContentPageReferences = (content: string) =>
-  (content.match(/"\/contents\/([a-z0-9-]+(@[\d.]+)?)/g) || [])
+export const getContentPageReferences = (content: string) => {
+  const matches = (content.match(/"\/contents\/([a-z0-9-]+(@[\d.]+)?)/g) || [])
```
TomWoodward, line 23: Rename this one legacy matches

src/app/content/utils.ts
```diff
@@ -30,6 +30,24 @@ export const getContentPageReferences = (content: string) =>
       };
     });
 
+  // rap content
+  const rapMatches = (content.match(/.\/([a-z0-9-]+(@[\d.]+)):([a-z0-9-]+.xhtml)/g) || [])
```
TomWoodward, line 34: And this one just matches, the nomenclature rap is already expiring

src/app/content/utils.ts
```diff
@@ -30,6 +30,24 @@ export const getContentPageReferences = (content: string) =>
...
+    .map((match) => {
+      const [bookMatch, pageMatch] = match.split(':');
+      const pageId = pageMatch.substr(0, 36);
+      const [bookId, bookVersion] = bookMatch.split('@');
+
+      return {
+        bookId: bookId.substr(2),
+        bookVersion,
+        match,
```
TomWoodward, line 43: The other one takes a substring here, are you sure this is ok?

TomWoodward (changes requested):

src/app/content/utils.ts
KarinaMendez2: yeah the legacy ones, for some reason, the match comes like this: `"/contents/d0a86917-6447-4b49-bed1-efb5c352cc77` so the `substr` is to remove the `"` but for the rap ones that come like: `./13ac107a-f15f-49d2-97e8-60ab2e3b519c@29.7:99d38770-49c7-49d3-b567-88f393ffb4fe.xhtml` I don't thinks it's needed

src/app/content/hooks/locationChange/resolveContent.ts
KarinaMendez2: it doesn’t catch `reference.bookId` and `reference.bookVersion`

KarinaMendez2: it doesn’t catch `reference.bookId` and `reference.bookVersion`

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -150,27 +150,39 @@ const resolvePage = async(
 
 const getBookInformation = async(
   services: AppServices & MiddlewareAPI,
-  pageId: string
+  reference: {
+    bookId?: string,
+    bookVersion?: string,
+    pageUid: string,
```
TomWoodward, line 156: can you go back and rename `pageUid` to `pageId` so its consistent with the standard naming? i know it was like that already but while we're in here lets fix it

src/app/content/utils.ts
```diff
@@ -30,6 +30,24 @@ export const getContentPageReferences = (content: string) =>
       };
     });
 
+  // rap content
```
TomWoodward, line 33: remove this comment

TomWoodward (changes requested):

TomWoodward (approved):

PR #1039, opened 2021-01-07
Better way of handling errors in hooks

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1354


src/app/content/highlights/hooks/loadMore.ts
```diff
@@ -56,10 +54,13 @@ export const hookBody: ActionHookBody<
     try {
       highlights = await loadMore(services, summaryPageSize);
     } catch (error) {
-      Sentry.captureException(error);
-      services.dispatch(addToast(toastMessageKeys.higlights.failure.popUp.load, {destination: 'myHighlights'}));
       services.dispatch(toggleSummaryHighlightsLoading(false));
-      return;
+
+      if (error instanceof CustomApplicationError) {
```
TomWoodward, line 59: you're doing this if check a lot, i feel like we could do something to streamline it like:

```ts
// not loving the name, maybe something better
throw CustomApplicationError::check(error, new HighlightPopupLoadError({ destination: 'myHighlights' }));
```

```ts
export class CustomApplicationError extends Error {
  public static function check(error: any, customError: CustomApplicationError) {
    if (error instanceof CustomApplicationError) {
      return error;
    }
    return customError;
  }
}
```

or something

src/app/content/highlights/hooks/locationChange.ts
```diff
@@ -41,12 +40,11 @@ const hookBody = (services: MiddlewareAPI & AppServices) => async(action?: AnyAc
...
-    Sentry.captureException(error);
-
-    if (action && action.type !== getType(receivePageFocus)) {
-      dispatch(addToast(toastMessageKeys.higlights.failure.load, {destination: 'page', shouldAutoDismiss: false}));
+    if (action && action.type !== getType(receivePageFocus) && !(error instanceof CustomApplicationError)) {
+      throw new HighlightLoadError({ destination: 'page', shouldAutoDismiss: false });
     }
-    return;
+
+    throw error;
```
TomWoodward, line 47: because of the extra conditions you don't necessarily know this is a `CustomApplicationError` at this point

src/app/content/highlights/hooks/printHighlights.ts
```diff
@@ -1,22 +1,25 @@
...
-import { toastMessageKeys } from '../../../notifications/components/ToastNotifications/constants';
 import { ActionHookBody, AppServices, MiddlewareAPI, Unpromisify } from '../../../types';
-import { actionHook, assertWindow } from '../../../utils';
+import { actionHook, assertWindow, CustomApplicationError } from '../../../utils';
 import { printSummaryHighlights, receiveSummaryHighlights, toggleSummaryHighlightsLoading } from '../actions';
+import { HighlightPopupPrintError } from '../errors';
 import { myHighlightsOpen } from '../selectors';
 import { loadMore, LoadMoreResponse } from './loadMore';
 
+let waitingForPromiseCollector = false;
```
TomWoodward, line 8: what is the purpose of this flag? global flag like this seems worrying to me

src/app/content/highlights/hooks/updateHighlight.ts
```diff
@@ -15,20 +13,22 @@ export const hookBody: ActionHookBody<typeof updateHighlight> =
...
 
+      dispatch(updateHighlight(meta.preUpdateData, {...meta, revertingAfterFailure: true}));
+
+      if (error instanceof CustomApplicationError) {
+        throw error;
+      }
+
       if (payload.highlight.color && oldColor !== payload.highlight.color) {
-        dispatch(addToast(toastMessageKeys.higlights.failure.update.color, {destination}));
+        throw new HighlightUpdateColorError({ destination });
```
TomWoodward, line 28: i'm just now thinking... shouldn't each error type always know what its destination should be? does that have to be defined on the fly?

src/app/content/studyGuides/hooks/printStudyGuides.ts
```diff
@@ -1,22 +1,25 @@
...
-import { toastMessageKeys } from '../../../notifications/components/ToastNotifications/constants';
 import { ActionHookBody, AppServices, MiddlewareAPI, Unpromisify } from '../../../types';
-import { actionHook, assertWindow } from '../../../utils';
+import { actionHook, assertWindow, CustomApplicationError } from '../../../utils';
+import { StudyGuidesPopupPrintError } from '../../highlights/errors';
 import { printStudyGuides, receiveSummaryStudyGuides, toggleStudyGuidesSummaryLoading } from '../actions';
 import { studyGuidesOpen } from '../selectors';
 import { loadMore, LoadMoreResponse } from './loadMore';
 
+let waitingForPromiseCollector = false;
```
TomWoodward, line 8: still worried about this

src/helpers/applicationMessageError.ts
```diff
@@ -0,0 +1,19 @@
+import { ToastMeta } from '../app/notifications/types';
+
+export class ApplicationMesssageError extends Error {
```
TomWoodward, line 3: this could extend `CustomApplicationError`

TomWoodward (changes requested):

src/helpers/applicationMessageError.ts
jarosik10: But don't we separate `CustomApplicationError`s from ones created with `makeApplicationError`?

src/app/notifications/types.ts
```diff
@@ -27,6 +27,11 @@ export interface ToastNotification {
   shouldAutoDismiss: boolean;
 }
 
+export interface ToastMeta {
+  destination: ToastNotification['destination'];
+  shouldAutoDismiss?: boolean;
+}
+
```
jarosik10, line 34: `ToastNotification` extends `ToastMeta` will be handled in a different PR?

src/app/content/highlights/hooks/createHighlight.ts
```diff
@@ -16,10 +14,13 @@ export const hookBody: ActionHookBody<typeof createHighlight> =
     try {
       await highlightClient.addHighlight({highlight: payload});
     } catch (error) {
-      Sentry.captureException(error);
-
-      dispatch(addToast(toastMessageKeys.higlights.failure.create, {destination}));
       dispatch(receiveDeleteHighlight(payload as unknown as Highlight, {...meta, revertingAfterFailure: true}));
+
+      if (error instanceof CustomApplicationError) {
```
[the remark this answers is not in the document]
PiotrKozlowski: I've marked `UnauthenticatedError` and `BookNotFoundError` as `CustomApplicationError`
If we would add any new errors to the API clients then it would be handled by this if statement

src/app/content/highlights/hooks/updateHighlight.ts
PiotrKozlowski: In some cases `destination` is taken from the selector (we are checking if MH popup is open or not) so we could replace this with:
```
if (destination === 'myHighlights) {
  throw ...
} else {
  throw ...
}
```
I think the current solution is easier to ready

src/app/content/highlights/hooks/printHighlights.ts
PiotrKozlowski: `printHighlights` hook is calling `await services.promiseCollector.calm();` and for some reason that was causing infinite loop of waiting for itself if this hook returned the Promise
and since we want the error to be moved up we have to return the Promise so I've created this flag and replaced `await services.promiseCollector.calm();` with 
```
 if (waitingForPromiseCollector) {
    // wait for content to process/load
    await services.promiseCollector.calm();
    waitingForPromiseCollector = false;
  }
 ``` 
 I wanted to refactor this somehow, but I currently don't have more ideas on how to solve this

File: src/helpers/applicationMessageError.ts (hunk 6 above)
Comment (PiotrKozlowski) [reply src/helpers/applicationMessageError.ts:3]: They are separate on purpose, however, maybe they should be named something like `ToastMessageError`?

File: src/app/content/highlights/hooks/locationChange.ts (hunk 2 above)
Comment (PiotrKozlowski) [reply src/app/content/highlights/hooks/locationChange.ts:47]: We can throw it even if this isn't  `CustomApplicationError`

File: src/app/notifications/types.ts (hunk 7 above)
Comment (PiotrKozlowski) [reply src/app/notifications/types.ts:34]: yes

File: src/app/content/highlights/hooks/locationChange.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (PiotrKozlowski) [reply src/app/content/highlights/hooks/locationChange.ts]: what do you mean?

**Comment by PiotrKozlowski** [comment]
I've simplified logic and changed error names, pleas take a look again

File: src/app/content/highlights/hooks/locationChange.ts (hunk 2 above)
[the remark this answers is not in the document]
Comment (jarosik10) [reply src/app/content/highlights/hooks/locationChange.ts]: I just don't fully understand the purpose of this type check.

File: src/helpers/applicationMessageError.ts [hunk 9]
```diff
@@ -12,8 +15,18 @@ export class ApplicationMesssageError extends Error {
...
+export const makeToastMessageError = (messageKey: string) => class extends ToastMesssageError {
   constructor(meta: ToastMeta) {
     super(messageKey, meta);
   }
 };
+
+/**
+ * Return @param error if it is instance of ApplicationError or @param customError in other cases
+ */
+export const makeApplicationError = (error: Error, customError: Error | (() => Error)) => {
```
jarosik10, line 27: Naming is a little bit confusing. It's implying that this function creates instances of `class ApplicationError`, but it doesn't.
Hmm... maybe naming is as a file name: 'makeApplicationMessageError` is a good idea. It's just a suggestion.

PiotrKozlowski: maybe something like `makeApplicationOrCustomError`?

jarosik10: Sounds good :)

TomWoodward: in the case that the first argument is an applicationError, it doesn't actually make anything. maybe something like `ensureApplicationErrorType` or something. a verb like `ensure` or `default` or something that doesn't imply construction

jarosik10 (approved):

PiotrKozlowski:
CI / Test (build) failed at `yarn install`

src/app/content/highlights/hooks/printHighlights.ts
TomWoodward: all hook promises are added to the collector as a way to be able to tell when all logic has settled. so if you await collector.calm() in the hook it causes a dependency

from the previous version of this logic you could have remove the non-async hook wrapper, returned an async function as normal, you just have to remove the `await` before collector.calm() and move the subsequent logic into a `then` and not return that promise, so that it doesn't create a loop.

if there is any async logic after that you'll have to manually add it back to the collector, but in this case it looks like all synchronous logic so it doesn't matter

TomWoodward (changes requested):

TomWoodward (approved):
the update to the promise handling looks good

PR #1040, opened 2021-01-08
Hide content from printing when there is ScrollLock

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1445

### DIFF
--- a/src/app/components/ScrollLock.tsx
+++ b/src/app/components/ScrollLock.tsx
@@ -11,20 +11,24 @@ const ScrollLockBodyClass = createGlobalStyle`
   body.body {
     ${(props: {mobileOnly?: boolean}) => props.mobileOnly && css`
       ${theme.breakpoints.mobile(css`
+        #root {
+          display: none;
+        }
+
         overflow: hidden;
       `)}
     `}
 
     ${(props: {mobileOnly?: boolean}) => props.mobileOnly === false && css`
+      #root {
+        display: none;
+      }
+
       overflow: hidden;
     `}
   }
 
   @media print {
-    #root {
-      display: none;
-    }
-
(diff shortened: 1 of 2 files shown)


src/app/components/ScrollLock.tsx
```diff
@@ -21,6 +21,10 @@ const ScrollLockBodyClass = createGlobalStyle`
   }
 
   @media print {
+    #root {
+      display: none;
```
TomWoodward, line 25: this will hide content when printing unconditionally because of the toc overlay, you need to put it in the conditionals on lines 14 and 19

TomWoodward (changes requested):

src/app/components/ScrollLock.tsx
```diff
@@ -11,11 +11,19 @@ const ScrollLockBodyClass = createGlobalStyle`
   body.body {
     ${(props: {mobileOnly?: boolean}) => props.mobileOnly && css`
       ${theme.breakpoints.mobile(css`
+        #root {
+          display: none;
```
TomWoodward, line 15: both of these need to be in print media queries

TomWoodward (changes requested):

src/app/components/ScrollLock.tsx
```diff
@@ -11,19 +11,29 @@ const ScrollLockBodyClass = createGlobalStyle`
...
+        }
+      }
+
       overflow: hidden;
     `}
   }
 
-  @media print {
```
TomWoodward, line 23: did you intend to remove this?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1042, opened 2021-01-14
fix tests except CustomApplicationError if path

jarosik10:


src/app/content/highlights/hooks/createHighlight.spec.ts
```diff
@@ -76,18 +76,53 @@ describe('createHighlight', () => {
       .mockRejectedValue(error);
     const mock = createMockHighlight();
 
+    try {
     await hook(createHighlight(mock, meta));
+    } catch (error) {
+      expect(createHighlightClient).toHaveBeenCalledWith({highlight: mock});
```
PiotrKozlowski, line 82: there is no `expect` about `error`

src/app/content/highlights/hooks/locationChange.spec.ts
```diff
@@ -149,40 +149,63 @@ describe('locationChange', () => {
   });
 
   describe('error handling', () => {
-    it('doesn\'t show a toast if hook ran because of page\'s focus changing', async() => {
+    it('throws HighlightLoadError', async() => {
```
PiotrKozlowski, line 152: the old test should not be removed

PiotrKozlowski: I see that there some checks for receivePageFocus, but I have to look if they are correct

src/app/content/highlights/hooks/printHighlights.spec.ts
```diff
@@ -60,21 +59,16 @@ describe('printHighlights', () => {
       getState: store.getState,
     };
 
-    addPromise = jest.spyOn(helpers.promiseCollector, 'add');
     dispatch = jest.spyOn(helpers, 'dispatch');
+    calmSpy = jest.spyOn(helpers.promiseCollector, 'calm')
+    .mockImplementation(() => Promise.resolve());
```
PiotrKozlowski, line 64: indentation

src/app/content/highlights/hooks/receiveDeleteHighlight.spec.ts
```diff
@@ -63,17 +61,41 @@ describe('receiveDeleteHighlight', () => {
     const deleteHighlightClient = jest.spyOn(helpers.highlightClient, 'deleteHighlight')
       .mockRejectedValue(error);
 
-    hook(receiveDeleteHighlight(highlight as unknown as Highlight, meta));
-    await Promise.resolve();
+    try {
+      await hook(receiveDeleteHighlight(highlight as unknown as Highlight, meta));
+    } catch (error) {
+      expect(deleteHighlightClient).toHaveBeenCalled();
```
PiotrKozlowski, line 67: there is no assumption on error type

src/app/content/highlights/hooks/updateHighlight.spec.ts
```diff
@@ -88,7 +86,7 @@ describe('updateHighlight', () => {
     expect(updateHighlightClient).not.toHaveBeenCalled();
   });
 
-  it('doesn\'t show toast / revert update if it would result in the same highlight', async() => {
+  it('reverts update if it would result in the same highlight', async() => {
```
PiotrKozlowski, line 89: we should probably expect an error in this test

src/app/content/studyGuides/hooks/printStudyGuides.spec.ts
```diff
@@ -86,9 +107,25 @@ describe('printStudyGuides', () => {
     expect(calmSpy).toHaveBeenCalled();
     await Promise.resolve();
 
+    expect(dispatch).toBeCalledWith(toggleStudyGuidesSummaryLoading(false));
```
PiotrKozlowski, line 110: was this happening in the previous version?

PiotrKozlowski: probably yes but I'll check

src/app/content/studyGuides/hooks/printStudyGuides.spec.ts
```diff
@@ -86,9 +107,25 @@ describe('printStudyGuides', () => {
...
     expect(print).toHaveBeenCalled();
   });
 
+  it('doesn\'t wait for promiseCollector.calm', async() => {
```
PiotrKozlowski, line 114: is this required for coverage for some if path?

src/app/utils.spec.tsx
```diff
@@ -132,6 +135,30 @@ describe('actionHook', () => {
   });
 });
 
+it('handle error if it is instance of ApplicationMesssageError', async() => {
+  const hookSpy = jest.fn(async() => Promise.reject(
+    new HighlightCreateError({ destination: 'myHighlights', shouldAutoDismiss: true }
```
PiotrKozlowski, line 140: probably better to use `createApplicationError` here

PiotrKozlowski (commented):
since `expect` called in `catch {}` block will throw an error only if there is an error, we should make sure that these functions are called

jarosik10:
> since `expect` called in `catch {}` block will throw an error only if there is an error, we should make sure that these functions are called

`expect.assertions(number)` should be fine, right?

src/app/content/highlights/hooks/locationChange.spec.ts
jarosik10: But it doesn't `dispatch(addToast(...)` anymore. It's only about throwing the proper error.

PiotrKozlowski:
@jarosik10 
> expect.assertions(number) should be fine, right?

yes

src/app/content/highlights/hooks/receiveDeleteHighlight.spec.ts
jarosik10: I wanted to test `dispatch(createHighlight(...)` first and then error throwing.
```
} catch (error) {
   dispatch(createHighlight(
      {
          ...payload as unknown as NewHighlight,
          id: payload.id,
       },
       {...meta, revertingAfterFailure: true}
   ));
   
   if (error instanceof CustomApplicationError) {
      throw error;
   }
   
   throw new HighlightDeleteError({ destination });
}
```

src/app/content/studyGuides/hooks/printStudyGuides.spec.ts
jarosik10: Yup. It's about `  if (waitingForPromiseCollector) {...}` else path.
Maybe the test name is misleading or smthing.

src/app/content/highlights/hooks/createHighlight.spec.ts
PiotrKozlowski: its in the separate test below

src/app/content/highlights/hooks/locationChange.spec.ts
PiotrKozlowski: I think we should test if the hook with `receivePageFocus` will not throw `HighlightLoadError`

src/app/content/highlights/hooks/updateHighlight.spec.ts
PiotrKozlowski: nope

PR #1043, opened 2021-01-15
Ts 3.9.6

TomWoodward:
there are very few additional ts compiling errors to get from here to 4.1, but v4 apparently included breaking changes to the typescript tooling apis and will require us to upgrade our build toolchain, not only does react-scripts need the upgraded version of node that we aren't using yet, we also have some hacky stuff going on with our build, so we'll have to tread carefully there.

in here you'll notice some change to the Route and Match types to make them more stable (remove the conditional typing) in the end i'm not sure this was necessary but when i tried it i realized it actually reduced the amount of code necessary to handle those types, even though i feel like the conditional types had a more correct/nice result i kept the change because it simplified other things

there is also a thing where route params must be a `type` not an `interface` for reasons i don't really care to try to understand right now, i disabled a linter rule that was converting simple type objects to interfaces

the change in index.tsx to the `routes` variable definition was one of the crucial changes, for some reason in the previous shape we were getting an intersection of all the members instead of a union and it was breaking many things.

there remains a typing issue in the route utils that i can't figure out, i have a stack overflow answer open for it and am waiting for someone to save me

i refactored a few things along the way that i thought would simplify things

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1419


jarosik10:
Lint error is related to the typing issue you have mentioned?

TomWoodward:
not sure, lint was passing and then master merged and it stopped, maybe a type error from the merge, i haven't looked into it  yet

TomWoodward:
@jarosik10 types fixed

script/prerender/contentPages.tsx
```diff
@@ -113,11 +113,11 @@ type MakeRenderPage = (services: AppOptions['services']) =>
...
-  const matchState = assertDefined(
-    route.state,
-    'match state wasn\'t defined, it should have been'
-  );
-  const {bookUid, bookVersion, pageUid} = matchState;
+  if (!route.state || !('bookUid' in route.state)) {
+    throw new Error('match state wasn\'t defined, it should have been');
+  }
+
+  const {bookUid, bookVersion, pageUid} = route.state;
```
jarosik10, line 120: Why did you make it this way? is it okay to hardcode the check for `bookUid` in `route.state`?

src/app/content/components/__snapshots__/BookBanner.spec.tsx.snap
```diff
@@ -90,25 +90,12 @@ Array [
...
       className="BookBanner__TopBar-sc-1avy0c0-1 jVswHX"
     >
-      <a
-        className="BookBanner__BookTitleLink-sc-1avy0c0-3 iGPDns"
-        data-testid="details-link-expanded"
-        href="/error/404"
-        onClick={[Function]}
+      <span
+        className="BookBanner__BookTitle-sc-1avy0c0-2 gUeKV"
+        data-testid="book-title-expanded"
```
jarosik10, line 95: These changes are intended, right?

src/app/index.tsx
```diff
@@ -69,18 +68,21 @@ export interface AppOptions {
...
     });
 
-  if (initialEntries && initialEntries.length > 0) {
-    const entry = initialEntries[initialEntries.length - 1];
-    if (hasState(entry)) {
-      history.location.state = entry.state;
+    if (initialEntries && initialEntries[0]) {
+      memoryHistory.location.state = initialEntries[0].state;
     }
-  }
```
jarosik10, line 83: So it is no longer necessary for `history` created with`createBrowserHistory`?

script/prerender/contentPages.tsx
TomWoodward: the state type is now `{} | theStateStuff` instead of `undefined` so you can't just do a truthy or undefined check anymore. in this case in prerendering i'm requiring that the uid version of the props is supplied.

its not pretty but it works, we can try to come up with a cleaner way in the future

src/app/content/components/__snapshots__/BookBanner.spec.tsx.snap
TomWoodward: yea i fussed with the bookbanner because there were two variables being used in conditions that were actually associated so i combined the logic, i think its more reliable now

src/app/index.tsx
TomWoodward: there was some separate `initialEntries` logic here that only actually ever worked when memoryHistory was being used, so i combined it into a helper that creates the memory history. we still use browserHistory if the window is defined

src/app/navigation/reducer.ts
```diff
@@ -18,9 +17,7 @@ const addQuery = (location: Location) => ({
 export default (location: Location): Reducer<State, AnyAction> => (state = addQuery(location), action) => {
   switch (action.type) {
     case getType(actions.locationChange):
-      const match = action.payload.match && isMatchWithParams(action.payload.match)
-        ? action.payload.match
-        : undefined;
+      const { match } = action.payload;
```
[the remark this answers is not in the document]
TomWoodward: i mean lol probably, i'll change it

src/app/navigation/types.ts
```diff
@@ -15,24 +15,12 @@ export type RouteState<R> = R extends Route<any, infer S> ? S : never;
...
-      : MatchWithParams<R>)
-    & (RouteState<R> extends undefined
-      ? BasicMatch<R>
-      : MatchWithState<R>)
+export type Match<R> = R extends Route<infer P, infer S>
+  ? {
+    route: R,
+    params: P,
+    state: S,
+  }
```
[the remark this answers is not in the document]
TomWoodward: its less great because you have these params defined when they're not used for some routes, but the code is much simpler with less switches so i think it mostly just offends my human brain

src/app/navigation/utils.ts
```diff
@@ -70,16 +74,18 @@ export const matchSearch = (action: AnyMatch, search: string | undefined) => {
...
 };
 
-export const matchUrl = (action: AnyMatch) => isMatchWithParams(action)
-  ? action.route.getUrl(action.params)
-  : action.route.getUrl();
+// issue with passing AnyMatch into this https://stackoverflow.com/q/65727184/14809536
+export const matchPathname = <M extends Match<Route<any, any>>>(action: M) => action.route.getUrl(action.params);
+
+// issue with passing AnyMatch into this https://stackoverflow.com/q/65727184/14809536
+export const matchUrl = <M extends Match<Route<any, any>>>(action: M) => {
```
TomWoodward, line 81: these `any`s were the issue i was talking about, i haven't had any traction on stack overflow but i have identified that the (or at least one of the) problem(s) with the types is having both the `route.getUrl(params)` (and friends) functions as well as the `params` functions, because of the way function extension works you can never have these extending a base type like `Match<Route<RouteParams, RouteState>>` because the function doesn't match

the union thing i still have no idea about i'm ready to just give up on that and say its some weird ts thing that i'll never understand

but maybe if we restructure `Match` to be created by some kind of `makeMatch` function that curries the params in and then move the `getUrl` function into the match at which point it can take no arguments and use its access the the params from before, maybe then we can remove the extra properties of the match that are causing the types to lock up. i'm continuing to think about this and try to come up with a way to do it where the matches are still structured and you can't just pass any old match into a thing it has to be a match for a registered route. 

definitely want to leave this the way i have it for now, i'll try to improve the types in a later change

jarosik10 (approved):

PR #1044, opened 2021-01-18
Issue 1416 Visual mods to question distractors in Practice 

PiotrKozlowski:
Allow selecting answers with Space, remove onClick, add css for selected state
for: https://github.com/openstax/unified/issues/1416


jarosik10:
When I select the answer with spacebar, it ends up with a focused label. When I select the answer using mouse, it ends up with focused input. It doesn't make a difference to the tabbing after the selection, but maybe it can cause some other problems.

src/app/content/practiceQuestions/components/Answer/index.tsx
```diff
@@ -68,19 +69,31 @@ const Answer = ({
   // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [showCorrect, isCorrect]);
 
+  const onKeyDown = (e: React.KeyboardEvent<HTMLElement>) => {
+    // Space
+    if (e.which === 32) {
```
TomWoodward, line 74: you really should not be having to manually process these key events

TomWoodward (changes requested):

TomWoodward:
we want the focus outline to be browser default, it looks like a different color to me?

src/app/content/practiceQuestions/components/Answer/index.tsx
PiotrKozlowski: it doesn't work without this

PiotrKozlowski:
@TomWoodward 
> we want the focus outline to be browser default, it looks like a different color to me?

I've set this to `outline: auto` for `:focus-visible` and I don't see the difference - it is black for answer like for the questions and buttons in the modal

TomWoodward:
the color seems different to me


image
image
image

PiotrKozlowski:
@TomWoodward they are the same for me :/ 
Hopefully removing `:focus-visible { outline: auto; }` will fix this for you

src/app/content/practiceQuestions/components/Answer/index.tsx
```diff
@@ -81,48 +81,52 @@ const Answer = ({
...
-        <AnswerResult
-          showCorrect={showCorrect}
-          isSelected={isSelected}
-          isSubmitted={isSubmitted}
-          isCorrect={isCorrect}
-        />
-      </AnswerContent>
-    </AnswerAlignment>
-  </AnswerBlock>;
+    <AnswerBlock
```
jarosik10: You are nesting `label` inside a `label` right now.

src/app/content/practiceQuestions/components/Answer/index.tsx
```diff
@@ -81,48 +81,52 @@ const Answer = ({
     }
   };
 
-  return <AnswerBlock
-    showCorrect={showCorrect}
-    isCorrect={isCorrect}
-    isSubmitted={isSubmitted}
-    isSelected={isSelected}
+  return <label
```
jarosik10, line 84: You are nesting `label` inside a `label` right now.

PiotrKozlowski: fixed

PiotrKozlowski:
@jarosik10 outlines should now be even
about their colors - Maybe browsers are specifying another color for not clickable elements like divs?

TomWoodward:
the outline looks like the same color to me now, whatever you changed seems to have fixed it

jarosik10 (approved):

src/app/content/practiceQuestions/components/Answer/index.tsx
TomWoodward: ok i'll look into it

TomWoodward (changes requested):

TomWoodward:
removed necessity for keydown handler and modified aria-label for checkbox based on slack conversation https://openstax.slack.com/archives/C8V1XET1R/p1611177001103700

the issue in my request to piotr initially was that i didn't understand exactly how labels worked, i knew that clicking on a label checks the checkbox, but i didn't know that labels aren't intended to be keyboard navigable at all. the recommendation i found was to simply do keyboard nav to the actual checkbox, with visual users clicking on the label, and style everything with a sibling selector from the checkbox

@PiotrKozlowski @jarosik10

TomWoodward:
tested in chrome/firefox/safari on osx, used chromevox in chrome and it reads out the right stuff. @PiotrKozlowski let me know if i broke anything i shouldn't have

PiotrKozlowski:
@TomWoodward I've removed tests for `onKeyDown` and added missing `tabIndex` to the `<AnswerBlock>`

TomWoodward:
@PiotrKozlowski there should not be a tabIndex on answerBlock

TomWoodward:
for your selecting the correct answer, focus the radio button itself

PiotrKozlowski:
@TomWoodward This works only if I remove `disabled={isSubmitted}` from the input
Do we want to allow users tab through answers after they submitted the form?

TomWoodward:
maybe not, i see you did tabindex=-1 so it won't affect keyboard behavior, lets do what you have

TomWoodward:
nevermind adding the tabindex breaks the flashing outline thing again, we need to more changes to handle this

TomWoodward (approved):

PR #1046, opened 2021-01-19
Issue 1350

PiotrKozlowski:
~I think this pr should be finished after https://github.com/openstax/rex-web/pull/1039 because then we will be able to add the condition in one place to display `errorId` on all toasts~ DONE

for: https://github.com/openstax/unified/issues/1350


src/app/notifications/types.ts
```diff
@@ -25,6 +25,13 @@ export interface ToastNotification {
   timestamp: number;
   destination: 'studyGuides' | 'myHighlights' | 'page';
   shouldAutoDismiss: boolean;
+  errorId?: string;
+}
+
+export interface ToastMeta {
+  destination: ToastNotification['destination'];
+  shouldAutoDismiss?: boolean;
+  errorId?: string;
```
jarosik10, line 34: What about making `ToastNoftification extends ToastMeta` and then remove redundant fields?

jarosik10:
Entering StudyGuides with offline mode shows connection error popup with sentryID, but entering HighLights doesn't. Not sure if it's a bug or a feature;).

PiotrKozlowski:
@jarosik10 I've pushed a fix for this

jarosik10:
As I understood, some of these changes are placeholders, because after https://github.com/openstax/rex-web/pull/1039 all the logic will be moved to one place. So there is nothing else to do but wait, right?

PiotrKozlowski:
> As I understood, some of these changes are placeholders, because after #1039 all the logic will be moved to one place. So there is nothing else to do but wait, right?

This pr should solve the problem, but I think we should merge it after #1039

For now please review and I will pass it to a with this comment

jarosik10:
You are not passing sentry error id in `printHighlights.ts` and `printStudyGuides.ts`.

PiotrKozlowski:
From what I understand the AC is for loading highlights and other toasts are optional
I think there is no sense of doing all modals now because after merging #1039 this will be supported by default

jarosik10 (approved):

jarosik10 (approved):

PiotrKozlowski:
@TomWoodward there is a merge conflict:
```ts
captureException(error: any, level: Sentry.Severity = Severity.Error) {
    if (this.isEnabled) {
<<<<<<< issue-1350
      return Sentry.captureException(error);
=======
      Sentry.withScope((scope) => {
        scope.setLevel(level);
        Sentry.captureException(error);
      });
>>>>>>> master
    } else if (!this.shouldCollectErrors) {
      console.error(error); // tslint:disable-line:no-console
    }
  },
  ```
  TS is saying that `Sentry.withScope` is not returning anything even if I return `Sentry.captureException(error)` 
I can't find if `Sentry.withScope` calls callback synchronously. If yes then we could do something like this:
```
let errorId;
Sentry.withScope((scope) => {
        scope.setLevel(level);
        errorId = Sentry.captureException(error);
});
return errorId;
```
But I think it may not work, because TS is not recognizing assigment to the `errorId` and it is `undefined` when I return it

**Comment by TomWoodward** [comment]
@PiotrKozlowski why do you need to add the return? i don't think it had that before

**Comment by PiotrKozlowski** [comment]
@TomWoodward I need the sentry error id to display it on the toast

File: src/app/utils.ts [hunk 2]
```diff
@@ -60,7 +61,8 @@ const makeCatchError = ({dispatch, getState}: MiddlewareAPI) => (e: Error) => {
     return;
   } else if (e instanceof ToastMesssageError) {
     Sentry.captureException(e);
-    dispatch(addToast(e.messageKey, e.meta));
+    const errorId = getMessageIdStack(getState())[0];
+    dispatch(addToast(e.messageKey, { ...e.meta, errorId }));
```
TomWoodward, line 65: this kind of decoupling is worrying me... we're assuming that captureException is synchronous and there is risk that we could get the wrong or duplicate error ids out of the state if the order is not guaranteed.

going back to the implementation you had before, could we try doing something like this?
```diff
diff --git a/src/helpers/Sentry.ts b/src/helpers/Sentry.ts
index f16875301..d574a8909 100644
--- a/src/helpers/Sentry.ts
+++ b/src/helpers/Sentry.ts
@@ -50,15 +50,19 @@ export default {
     return typeof(window) !== 'undefined' && config.SENTRY_ENABLED;
   },
 
-  captureException(error: any, level: Sentry.Severity = Severity.Error) {
+  captureException(error: any, level: Sentry.Severity = Severity.Error): string | undefined {
+    let eventId;
+
     if (this.isEnabled) {
       Sentry.withScope((scope) => {
         scope.setLevel(level);
-        Sentry.captureException(error);
+        eventId = Sentry.captureException(error);
       });
     } else if (!this.shouldCollectErrors) {
       console.error(error); // tslint:disable-line:no-console
     }
+
+    return eventId;
   },
 
   captureMessage(message: string, level: Sentry.Severity) {
```

TomWoodward (changes requested):

src/app/utils.ts
PiotrKozlowski: I've written a comment about this https://github.com/openstax/rex-web/pull/1046#issuecomment-769821295
I'll push a change so we can test if this is working and typescript is just not recognizing properly the type

TomWoodward: because its a let you have to tell it to be `string | undefined`

src/helpers/Sentry.ts
```diff
@@ -50,15 +50,19 @@ export default {
     return typeof(window) !== 'undefined' && config.SENTRY_ENABLED;
   },
 
-  captureException(error: any, level: Sentry.Severity = Severity.Error) {
+  captureException(error: any, level: Sentry.Severity = Severity.Error): string | undefined {
+    let eventId;
```
TomWoodward, line 54: i realized after my suggestion that the types are tighter if you move the `: string | undefined` from the return type to the eventId definition, otherwise i'll be `any`

TomWoodward (changes requested):

src/helpers/Sentry.ts
PiotrKozlowski: I think it is not because of this
When I've pasted this outside of anything 
```ts
const b = (show: boolean) => {
    let a
    if (show) {
      a = 'asd'
    } else {
      return a
    }
    return a
  }
  ```
  then typescript easily identified that b will return a string or undefined
  
  it seems like there is something wrong with this line:
  ```ts
  eventId = Sentry.captureException(error);
  ```
  When I hover `eventId` here it still says it will be `string | undefined` (as declared after the change) when it should say `string`
  
This also works correctly:
```ts
let eventId;

    if (this.isEnabled) {
      eventId = Sentry.captureException(error);
    } else if (!this.shouldCollectErrors) {
      console.error(error); // tslint:disable-line:no-console
    }

    return eventId;
    ```
    
 so I think it is caused because of wrap in `Sentry.withScope`
    
 Not sure how much do we care, because it seems to be working on the heroku

Comment (TomWoodward) [reply src/helpers/Sentry.ts]: what issue are you talking about? after calling `eventId = Sentry.captureException(error);` it is correct for `eventId` to be `string | undefined`

Comment (PiotrKozlowski) [reply src/helpers/Sentry.ts]: For this code:
```ts
captureException(error: any, level: Sentry.Severity = Severity.Error) {
    let eventId;

    if (this.isEnabled) {
      Sentry.withScope((scope) => {
        scope.setLevel(level);
        eventId = Sentry.captureException(error);
      });
    } else if (!this.shouldCollectErrors) {
      console.error(error); // tslint:disable-line:no-console
    }

    return eventId;
  },
```
typescript is telling that the return type of `captureException` is `undefined`

and for this code:
```ts
captureException(error: any, level: Sentry.Severity = Severity.Error) {
    let eventId;

    if (this.isEnabled) {
      eventId = Sentry.captureException(error);
    } else if (!this.shouldCollectErrors) {
      console.error(error); // tslint:disable-line:no-console
    }

    return eventId;
  },
```
typescript is saying that the return type is `string or undefined`

TomWoodward: oh yea i think that the addition of the withScope callback breaks that inference, but if you have the definition from the start its ok

TomWoodward (approved):

PR #1047, opened 2021-01-20
Intellectual Property updates

staxbot:


TomWoodward:
scan expected to fail on new book

TomWoodward:
failing because its not on cnx.org yet

PR #1049, opened 2021-01-21
Optimize imports

jarosik10:
for: https://github.com/openstax/unified/issues/1459
We should wait for https://github.com/openstax/rex-web/pull/1028, this PR will bring two imports to optimize.
Importing `omitBy` from `lodash/fp` ~~causes Typescript errors during the building.~~

before:
Screenshot from 2021-01-21 13-54-29
after optimizing lodash imports:
Screenshot from 2021-01-21 17-32-32


jarosik10:
@TomWoodward
What do you think about the `analyze:bundle` script? I think that the previous version doesn't work because `--stats` flag got removed ([Link](https://github.com/facebook/create-react-app/issues/6904))
If we want to use `source-map-explorer`, `webpack-bundle-analyzer` should be removed from dependencies, right?

yarn.lock
```diff
@@ -2753,6 +2753,11 @@ ansi-regex@^4.0.0, ansi-regex@^4.1.0:
   resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-4.1.0.tgz#8b9f8f08cf1acb843756a839ca8c7e3168c51997"
   integrity sha512-1apePfXM1UOSqw0o9IiFAovVz9M5S1Dg+4TrDwfMewQ6p/rmMueb7tWZjQ1rx4Loy1ArBggoqGpfqqdI4rondg==
 
+ansi-regex@^5.0.0:
```
PiotrKozlowski, line 2756: all of these changes are coming from adding `source-map-explorer`?

package.json
```diff
@@ -77,7 +77,7 @@
     "test:prerender:browser": "REACT_APP_ENV=test SERVER_MODE=built jest --testPathPattern=\"(\\.|/)browserspec\\.tsx?\" -i",
     "test:prerender:screenshots": "REACT_APP_ENV=test SERVER_MODE=built jest --testPathPattern=\"(\\.|/)screenshotspec\\.tsx?\"",
     "test:lighthouse-manual": "REACT_APP_ENV=test lighthouse --view --config-path=./src/test/audits/index.js",
-    "analyze:bundle": "react-scripts build --stats && webpack-bundle-analyzer build/bundle-stats.json -m static -r build/bundle-stats.html -O",
+    "analyze:bundle": "react-scripts build && source-map-explorer 'build/static/js/*.js' -m && source-map-explorer 'build/static/js/*.js' -m --html build/bundle-stats.html",
```
PiotrKozlowski, line 80: is calling `source-map-explorer` 2 times necessary?
also is this script now checking only /static/js/... ? is this the same to what was the as previous script doing?

PiotrKozlowski (changes requested):

yarn.lock
jarosik10: I hope so. Can I check it by uninstalling `source-map-explorer` and see if all diff are gone?

package.json
jarosik10: I'm going to change this script. First of all, I wanted to know if we would like to work with `source-map-explorer`. Previous version was creating `json` and `html` result file.

> also is this script now checking only /static/js/... ? is this the same to what was the as previous script doing?

I'll check it.

yarn.lock
PiotrKozlowski: to check this you have to restore `yarn.lock` from master and run `yarn add source-map-explorer`

jarosik10: Looks like installing `source-map-explorer` adds all of it.

script/prerender/sitemap.ts
```diff
@@ -1,4 +1,9 @@
-import { filter, flow, get, identity, map, max } from 'lodash/fp';
+import filter from 'lodash/fp/filter';
+import flow from 'lodash/fp/flow';
+import get from 'lodash/fp/get';
+import identity from 'lodash/fp/identity';
+import map from 'lodash/fp/map';
+import max from 'lodash/fp/max';
```
TomWoodward, line 6: fwiw in `/script` we are not in the web code so the bundle doesn't matter, but i'm ok with doing it consistently everywhere so that its easier to find issues

TomWoodward (changes requested):
i haven't tried it locally but the output of that tool you added looks similar to the old one, we can keep it.

make sure to remove any dependencies for the old script

PiotrKozlowski (approved):

TomWoodward (approved):

PR #1051, opened 2021-01-22
Issue 1461

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1461

testing:
```
node ./script/entry.js update-content-versions-and-check-for-archived-slugs --bookId 9ab4ba6d-1e48-486d-a2de-38ae1617ca84 --versionNumber 7.7
```


concourse/update-content/script.bash
```diff
@@ -36,22 +36,10 @@ for book_id in $book_ids; do
...
   desired_version=$(sort --version-sort -r <<< "$approved_versions" | head -n 1)
 
-  current_version=$(jq -r --arg uuid "$book_id" '.[$uuid].defaultVersion' < src/config.books.json)
-
   # approved book format has a "1." on the front of every version
   desired_version=${desired_version:2}
 
-  if [ "$desired_version" == "$current_version" ]; then
-    echo "$book_id alredy at desired version."
-    continue
```
TomWoodward, line 46: in the case that there is no update, we want to continue to not open a pull request... will github reject opening a pull request that has no changes? or do we need to add logic after your new script that noops if there are no staged changes?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+type ArchivedSlugs = ArchivedSlug[];
+
+const { book, newVersion } = argv as any as {
+  book: string
+  newVersion: string | number;
+};
+
+(global as any).fetch = fetch;
```
TomWoodward, line 26: you can move this to `script/entry.js`, it sets some other browser globals we depend on like DOMParser and URL

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const booksPath = path.resolve(__dirname, '../src/config.books.json');
+const archivedSlugsPath = path.resolve(__dirname, '../src/archived-slugs.json');
```
TomWoodward, line 29: can we rename this something more generic... `old-page-urls.json`, `redirects.json`, `page-redirects.json` or something?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+const archiveLoader = createArchiveLoader(`${ARCHIVE_URL}${REACT_APP_ARCHIVE_URL}`);
+const osWebLoader = createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`);
+
+async function processBooks() {
+  const booksData = Object.entries(books);
+  const bookToUpdateIndex = booksData.findIndex(([id]) => id === book);
+  const bookToUpdate = booksData[bookToUpdateIndex] as [string, { defaultVersion: string }] | undefined;
+
+  if (!bookToUpdate) {
+    console.error(`book ${book} not found`); // tslint:disable-line:no-console
```
TomWoodward, line 40: this script should also add new books, sorry i didn't think to mention that it was already handling that

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+(global as any).fetch = fetch;
+
+const booksPath = path.resolve(__dirname, '../src/config.books.json');
+const archivedSlugsPath = path.resolve(__dirname, '../src/archived-slugs.json');
+
+const archiveLoader = createArchiveLoader(`${ARCHIVE_URL}${REACT_APP_ARCHIVE_URL}`);
+const osWebLoader = createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`);
+
+async function processBooks() {
+  const booksData = Object.entries(books);
```
TomWoodward, line 35: why are you converting this to/from and entries list and doing all this searching for the index and stuff? wouldn't it be easier to just use it as an object keyed by the book id?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+      // tslint:disable-next-line: no-console
+      console.log(`error while loading book ${bookId} with defaultVersion ${defaultVersion}`);
+      throw error;
+    });
+  const { tree: newTree } = await archiveLoader.book(bookId, newVersion.toString()).load()
+    .catch((error) => {
+      console.log(`error while loading book ${bookId} with newVersion ${newVersion}`); // tslint:disable-line:no-console
+      throw error;
+    });
+  const bookSlug = await osWebLoader.getBookSlugFromId(bookId)
```
TomWoodward, line 62: you could use the combined book loader and it'd shorten this logic a bit

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+    .catch((error) => {
+      console.log(`error while loading slug for bookId ${bookId}`); // tslint:disable-line:no-console
+      throw error;
+    });
+
+  const flatCurrentTree = flattenArchiveTree(currentTree);
+  const flatNewTree = flattenArchiveTree(newTree);
+  const missingPages = [...archivedSlugs] as ArchivedSlugs;
```
TomWoodward, line 70: why copy this array and not just use the existing variable? if you're just trying to assign a type you might want to instead create a definition file for the redirects file like the books.json has

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  const findArchivedSlug = (section: LinkedArchiveTreeNode) => (
+    { pageId, bookId: pageBookId, pathname }: ArchivedSlug
+  ) => pageId === section.id && pageBookId === bookId && pathname.split('/').pop() === section.slug;
+
+  const formatSection = (section: LinkedArchiveTreeNode) => ({
+    bookId,
+    pageId: section.id,
+    pathname: `/books/${bookSlug}/pages/${section.slug}`,
```
TomWoodward, line 79: you should use the route formatter we already have. `content.getUrl`

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  let countNewArchivedSlugs = 0;
+  for (const section of flatCurrentTree) {
+    if (
+      !flatNewTree.find((newSection) => newSection.slug === section.slug)
```
TomWoodward, line 85: instead of just finding slug by slug, try finding each page from the old version in the new version by id, then comparing the slugs

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,105 @@
...
+      throw error;
+    });
+
+  const flatCurrentTree = flattenArchiveTree(currentTree);
+  const flatNewTree = flattenArchiveTree(newTree);
+  const missingPages = [...archivedSlugs] as ArchivedSlugs;
+
+  const findArchivedSlug = (section: LinkedArchiveTreeNode) => (
+    { pageId, bookId: pageBookId, pathname }: ArchivedSlug
+  ) => pageId === section.id && pageBookId === bookId && pathname.split('/').pop() === section.slug;
```
TomWoodward, line 74: why do you have to pull this slug back out of the path?

TomWoodward (changes requested):

concourse/update-content/script.bash
PiotrKozlowski: > will github reject opening a pull request that has no changes?

there is no way to do this from UI, but it doesn't matter because we would still create and push branch so we have to detect that nothing has changed

script/update-content-versions-and-check-for-archived-slugs.ts
PiotrKozlowski: `section.slug` is for example: ```preface``` and `pathname` is ```/books/physics/pages/preface```

PiotrKozlowski:
@TomWoodward I've applied the requested changes but now I'm thinking that we should probably change a little bit structure of `redirects.json`
I think currently we will have merge conflicts if we will update multiple books that will add to this file.
Maybe we should have a folder `redirects` with `bookId.json`?

I may be wrong and git will figure out how to merge multiple entries in the array

TomWoodward:
hmmm, it would definitely be nice to not have conflicts in that case, go ahead and make a change in whatever way makes sense to you

PiotrKozlowski:
@TomWoodward done

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -50,10 +44,21 @@ async function updateRedirections(_bookId: string, currentVersion: string, _newV
       throw error;
     });
 
+  const redirectsBookPath = path.resolve(redirectsPath, bookId + '.json');
+
+  try {
+    // Create file with empty array only if it does not exixts
+    fs.writeFileSync(redirectsBookPath, '[]', { encoding: 'utf8', flag: 'wx' });
+  } catch {
```
TomWoodward, line 52: this would be less lines of code if you just did an if statement with `existsSync`

TomWoodward: since you have an unconditional write at the end of this logic, you could also do a ternary assignment defaulting `redirects` to `[]` if the file doesn't exist, then it'll be created at the end

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -50,10 +44,21 @@ async function updateRedirections(_bookId: string, currentVersion: string, _newV
...
+    // This will throw if file exists
+  }
+
+  const redirects: Redirects = require(redirectsBookPath);
+
   const flatCurrentTree = flattenArchiveTree(currentTree);
 
   const findRedirect = (section: LinkedArchiveTreeNode) => (
-    { pageId, bookId: pageBookId, pathname }: Redirect
+    { pageId, bookId: pageBookId, pathname }: Redirects[0]
```
TomWoodward, line 61: karina and i discovered that you should use `Redirects[number]` here, or ts might try to be clever and extract the first element of a tuple

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -50,10 +44,21 @@ async function updateRedirections(_bookId: string, currentVersion: string, _newV
...
   ) => pageId === section.id && pageBookId === _bookId && pathname.split('/').pop() === section.slug;
```
TomWoodward, line 62: this manual path extraction is brittle, it'd be better to run `section` through `formatSection` and compare the result to the one you're searching over

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,112 @@
...
+const booksPath = path.resolve(__dirname, '../src/config.books.json');
+const redirectsPath = path.resolve(__dirname, '../src/redirects/');
+
+const bookLoader = makeUnifiedBookLoader(
+  createArchiveLoader(`${ARCHIVE_URL}${REACT_APP_ARCHIVE_URL}`),
+  createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`)
+);
+
+async function updateRedirections(_bookId: string, currentVersion: string, _newVersion: string) {
+  if (currentVersion === _newVersion) {
```
TomWoodward, line 29: all these variables with the `_` on the front, the `_` is used to suppress the unused variable warning, if you're using them you shouldn't start with the underscore

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  fs.writeFileSync(redirectsBookPath, JSON.stringify(redirects, undefined, 2) + '\n', 'utf8');
+
+  return countNewRedirections;
+}
+
+async function processBook() {
+  const { defaultVersion } = books[bookId] || {};
+
+  if (defaultVersion === newVersion.toString()) {
```
TomWoodward, line 90: you're also doing this check above, do you need to do it twice?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,112 @@
...
+    console.log(`${bookId} alredy at desired version.`); // tslint:disable-line:no-console
+    process.exit(0);
+  }
+
+  const { title, version } = await bookLoader(bookId, newVersion.toString())
```
TomWoodward, line 95: why are you loading this twice?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,112 @@
...
+    .catch((error) => {
+      // tslint:disable-next-line: no-console
+      console.log(`error while loading book ${bookId} with version ${newVersion}`);
+      throw error;
+    });
+
+  books[bookId] = { defaultVersion: version };
```
TomWoodward, line 102: why does this need to be changed in the config?

concourse/update-content/script.bash
```diff
@@ -36,33 +36,23 @@ for book_id in $book_ids; do
...
-  fi
-
-  jq \
-    --arg version "$desired_version" \
-    --arg uuid "$book_id" \
-    '.[$uuid].defaultVersion = $version' < src/config.books.json > src/config.books.json.new
+  node script/entry.js update-content-versions-and-check-for-archived-slugs --bookId "$book_id" --newVersion "$desired_version"
 
-  mv src/config.books.json.new src/config.books.json
+  if [[ $(git status --porcelain) ]]; then
```
TomWoodward, line 44: instead of nesting the remaining logic, i'd rather do a conditional short circuit

TomWoodward (changes requested):

script/update-content-versions-and-check-for-archived-slugs.ts
PiotrKozlowski: I wanted to separate logic for updating redirects and updating the config file
I agree that currently, this doesn't matter, but I like when functions don't assume anything unless they are written inside of another function

script/update-content-versions-and-check-for-archived-slugs.ts
PiotrKozlowski: User may provide a version that is not existing so before adding it to the `config.books.json` we check if it is correct
`bookLoader` will cache results so I think it doesn't matter that it is called twice and `updateRedirections` continues to be a standalone function

script/update-content-versions-and-check-for-archived-slugs.ts
PiotrKozlowski: I think I don't understand  the question
it is adding or updating the `defaultVersion` in `books` variable which is then saved as `config.books.json`

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  const findRedirect = (section: LinkedArchiveTreeNode) =>
+    (redirect: Redirects[number]) => isEqual(formatSection(section), redirect);
+
+  let countNewRedirections = 0;
+  for (const section of flatCurrentTree) {
+    const { slug } = findArchiveTreeNodeById(newTree, section.id) || {};
+    if (
+      (slug && slug !== section.slug)
```
PiotrKozlowski, line 66: @TomWoodward What if we don't find section in the newTree? Shouldn't we also redirect it somewhere?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,104 @@
...
+  fs.writeFileSync(redirectsBookPath, JSON.stringify(redirects, undefined, 2) + '\n', 'utf8');
+
+  return countNewRedirections;
+}
+
+async function processBook() {
+  const { defaultVersion } = books[args.bookId] || {};
+
+  if (defaultVersion === args.newVersion.toString()) {
+    console.log(`${args.bookId} alredy at desired version.`); // tslint:disable-line:no-console
```
jarosik10, line 83: Do we want to keep all `tslint:disable` comments above the affected line?

script/update-content-versions-and-check-for-archived-slugs.ts
TomWoodward: not finding the section in the new tree would mean the page was removed, so 404 would be the correct action i think

script/update-content-versions-and-check-for-archived-slugs.ts
TomWoodward: yea, i see now, that wasn't clear to me.

it would have been nice to have a function with a descriptive name and/or create a new variable rather than modifying the global config.

you might have done something like that i haven't looked at the new changes yet

script/update-content-versions-and-check-for-archived-slugs.ts
TomWoodward: yea but couldn't you just pass the book into updateRedirections since you already have it?

PiotrKozlowski: I was thinking about this but then `updateRedirections` will require `id` of the current version and the `ArchiveBook` as the third argument which makes it less generic
I was then thinking about fetching both versions of book in the main function and passing them both but when we add a new book then we don't need to fetch the current version.
I just see that I shouldn't call `updateRedirections` if there is no `defaultVersion` for the provided book since it will not find anything. I'll push a change for this in a minute

script/update-content-versions-and-check-for-archived-slugs.ts
PiotrKozlowski: fixed

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -91,11 +92,16 @@ async function processBook() {
       throw error;
     });
 
-  books[args.bookId] = { defaultVersion: version };
+  const updatedBooksConfig = { ...books };
+  updatedBooksConfig[args.bookId] = { defaultVersion: version };
```
jarosik10, line 96: Shouldn't the `updatedBooksConfig` be in `fs.writeFileSync`?

PiotrKozlowski: ofc it should, thank :facepalm:

jarosik10 (approved):

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,110 @@
...
+import createOSWebLoader from '../src/gateways/createOSWebLoader';
+import { Redirects } from '../src/redirects/types';
+
+const args = argv as any as {
+  bookId: string
+  newVersion: string | number;
+};
+
+const booksPath = path.resolve(__dirname, '../src/config.books.json');
+const redirectsPath = path.resolve(__dirname, '../src/redirects/');
```
TomWoodward, line 22: does it make any sense to use our `data` folder here so we don't overload the `src` directory?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const { tree: newTree } = await bookLoader(bookId, newVersion)
+    .catch((error) => {
+      // tslint:disable-next-line: no-console
+      console.log(`error while loading book ${bookId} with newVersion ${newVersion}`);
+      throw error;
+    });
+
+  const redirectsBookPath = path.resolve(redirectsPath, bookId + '.json');
+  const redirects: Redirects = fs.existsSync(redirectsBookPath) ? require(redirectsBookPath) : [];
```
TomWoodward, line 49: mix/matching import and require is going to cause problems someday, you may just want to change this to `await import` now

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const flatCurrentTree = flattenArchiveTree(currentTree);
+
+  const formatSection = (section: LinkedArchiveTreeNode) => ({
+    bookId,
+    pageId: section.id,
+    pathname: content.getUrl({ book: { slug: bookSlug }, page: { slug: section.slug } }),
+  });
+
+  const findRedirect = (section: LinkedArchiveTreeNode) =>
```
TomWoodward, line 59: this isn't finding anything, its matching, and your last function can be omitted
```
const matchRedirect = (section: LinkedArchiveTreeNode) => isEqual(formatSection(section))
```

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const updatedBooksConfig = { ...books };
+  updatedBooksConfig[args.bookId] = { defaultVersion: version };
+
+  fs.writeFileSync(booksPath, JSON.stringify(updatedBooksConfig, undefined, 2) + '\n', 'utf8');
+
+  // defaultVersion will be undefined when we add a new book.
+  // In this case we don't need to updateRedirections because there is nothing to update.
+  const newRedirectionsCounter = defaultVersion
+    ? await updateRedirections(args.bookId, defaultVersion, args.newVersion.toString())
```
TomWoodward, line 103: you `toString` this more than once, can you hoist a definition to the top of the function to make it more clear?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1052, opened 2021-01-25
add default buy print response

jarosik10:
for: https://github.com/openstax/unified/issues/1457


src/gateways/createBuyPrintConfigLoader.ts
```diff
@@ -18,6 +19,17 @@ export default (url: string) => {
         } else {
           return rejectResponse(response,  (status, message) => `Error response from BuyPrint ${status}: ${message}`);
         }
+      }).catch((e) => {
+        Sentry.captureException(e);
+
+        return Promise.resolve({
+          buy_urls: [{
+            allows_redirects: false,
```
PiotrKozlowski, line 27: it returns `true` for https://buyprint.openstax.org/physcis.json
I think it should be `true` in this case since we will be redirecting by default

src/gateways/createBuyPrintConfigLoader.ts
```diff
@@ -18,6 +19,17 @@ export default (url: string) => {
...
+            disclosure: null,
+            provider: 'amazon',
```
PiotrKozlowski, line 29: should we update this to something like `openstax fallback`?

PiotrKozlowski (changes requested):
You should probably also update the file `buyPrintConfig.ts`

src/gateways/createBuyPrintConfigLoader.ts
TomWoodward: yea this field is irrelevant but i think `true` would be the correct value

src/gateways/createBuyPrintConfigLoader.ts
jarosik10: What exactly do you mean?

TomWoodward: the string `amazon` is not true here, we're directing folks to the openstax backend which will redirect them... somewhere. putting amazon might be confusing. something like `fallback` or `openstax` would be more clear

jarosik10: Okay, i see:)

TomWoodward: the additional background here is that we send people to either amazon or shopify depending on their geographical location, so the provider is supposed to indicate which store they're going to.

in a normal situation the `buy_urls` would be direct urls to the provider, in this case since we're not getting the buy urls we'll send the user to the openstax backend, and it'll redirect them to whichever provider they're supposed to go to

src/gateways/createBuyPrintConfigLoader.ts
```diff
@@ -1,9 +1,10 @@
 import { rejectResponse } from '../helpers/fetch';
+import Sentry from '../helpers/Sentry';
 
 export interface BuyPrintResponse {
   buy_urls: Array<{
     url: string;
-    provider: 'amazon';
+    provider: 'amazon' | 'openstax fallback';
```
PiotrKozlowski, line 7: I'm not sure if this matters but I think we should add `_` between words to match API response style > `openstax_fallback`

PiotrKozlowski (approved):

src/app/content/hooks/locationChange/buyPrintConfig.ts
```diff
@@ -13,20 +12,9 @@ const loadBuyPrintConfig = (services: MiddlewareAPI & AppServices) => async() =>
...
-
-  if (response === null) {
-    Sentry.captureException(new Error(`book: ${book.slug} had no response from buy-book config service`));
-    return;
-  }
-
+  const response = await buyPrintConfigLoader.load(book);
   const config = response.buy_urls[0];
 
-  if (!config) {
```
TomWoodward, line 26: there is a case where we get a response with no buy_urls, this is the case where we want to hide the buy button

TomWoodward (changes requested):

TomWoodward (approved):

PR #1055, opened 2021-01-25
Upgrade node

TomWoodward:


TomWoodward:
i could probably actually clean up some syntax and remove those defer calls in favor of setImmediate, i'm curious if CI passes though

TomWoodward:
hmm probably it won't because CI runs in the container ` openstax/rex-web:CI ` that is pre-built....

TomWoodward:
Fsevents version is causing error starting app in development at least on mac

PiotrKozlowski (commented):
I think `setImmediate` is okay to use in tests so we should replace all `await new Promise(defer);` with it?

There is some `yarn test:build` error about rangy:
`Rangy is not supported in this environment. Reason: Rangy can only run in a browser`

TomWoodward:
@PiotrKozlowski yes i agree i'm going to go back and get rid of the defer calls

i think the rangy thing might have always been there, if not i might have to add some more mocks for it.

right now i'm working on a build error i didn't notice because it wasn't breaking the unit tests, something about fsevents, it looks like i'll have to upgrade react-scripts to fix it

TomWoodward:
closing for now because of major issue with react-intl. they have removed the ability to use html in translations without supporting code in the implementation. having different html per translation is a crucial use case for us, so this isn't really acceptable. we'll have to migrate internationalization systems before proceeding on this

PR #1056, opened 2021-01-26
Add check for ANALYTICS_OPT_OUT cookie

jarosik10:
for: https://github.com/openstax/unified/issues/1463
It seems that `isReadyForCommands()` determines whether GA event is added to the queue or sent.


jarosik10:
@TomWoodward 
1. If this method of disabling GA is fine, where should the cookieName be declared?
2. Is there a easy way to test it? Do I have to build the project and run the server?

src/gateways/googleAnalyticsClient.ts
```diff
@@ -106,7 +107,7 @@ class GoogleAnalyticsClient {
       return;
     }
 
-    if (this.isReadyForCommands()) {
+    if (this.isReadyForCommands() && !Cookies.get('ANALYTICS_OPT_OUT')) {
```
PiotrKozlowski, line 110: I think this check should be moved directly into `isReadyForCommands`
and about the cookie name - we're used to use `constants.ts` file for such things

you should also add a unit test for this to check if `window.ga` is not called if cookie is set and that it is added to the `pendingCommands` instead

PiotrKozlowski (changes requested):

src/gateways/googleAnalyticsClient.ts
jarosik10: I'm not sure where where `constants.ts` should be created. Like `src/constants.ts`, `src/gateways/constants.ts` or `src/app/constants.ts`? It doesn't seem right to me.

PiotrKozlowski: We tend to put them in the folder where are functions that will use them so in this case I think `src/gateways/constants.ts ` will be okay

jarosik10:
I think that there still one problem in `setTrackingIds`:
```
  public setTrackingIds(ids: string[]) {
    // Ignore tracking ID changes for the moment
    if (this.trackerNames.length > 0) { return; }

    for (const id of ids) {
      // Build a tracker for each ID, use the ID as the basis of the
      // tracker name (must be alphanumeric)
      const trackerName = 't' + id.replace( /[^a-z0-9]+/ig, '' );
      this.trackerNames.push(trackerName);
      this.ga('create', id, 'auto', trackerName);
    }

    this.flushPendingCommands();
  }
```
```
  private flushPendingCommands() {
    for (const pendingCommand of this.pendingCommands) {
      this.commandEachTracker(pendingCommand.command, pendingCommand.queueTime());
    }
    this.pendingCommands = [];
  }
```
`flushPendingCommands` ignores the `gaProxy` and calls `commandEachTracker` directly. So during the flush of the pending commands should be another cookie check or call `gaProxy` instead of `commandEachTracker`?
@TomWoodward @PiotrKozlowski

PiotrKozlowski:
hmm, or we could just do this in `this.ga()` method

jarosik10:
But `ga` isn't only used for the send commands. Not sure if we want block all the commands on `ga` and just store them as pending or just make a check like:
` if (commandName === 'send' && Cookies.getCookie(cookie)) { return; } `

src/gateways/constants.ts
```diff
@@ -0,0 +1 @@
+export const cookieGA = 'ANALYTICS_OPT_OUT';
```
PiotrKozlowski, line 1: maybe something like `disableGACookie`?

PiotrKozlowski (changes requested):

PiotrKozlowski:
> But ga isn't only used for the send commands. Not sure if we want block all the commands on ga and just store them as pending or just make a check like:
if (commandName === 'send' && Cookies.getCookie(cookie)) { return; }

I think we need to wait for @TomWoodward to weigh in on that

TomWoodward:
i think you're on the right track, but i'd add a function for like `trackingIsDisabled` and short circuit `setTrackingIds` if that is set, and also reference it in `isReadyForCommands` where you already made a change

also please make the cookie something more generic because we'll use the same to disable other analytics systems, you could define it somewhere in the helpers/analytics area

jarosik10:
@TomWoodward 
So if `trackingIsDisabled` is true, we will never want to flush the `pendingCommands` list?

TomWoodward:
I think that is comfortable... I would prefer to preserve the option, but there are complexities in this case.

We can't simply queue the tracker creation by using gaproxy because it needs to be created before sending any events

jarosik10:
@TomWoodward @PiotrKozlowski 
I've made few changes.

src/gateways/googleAnalyticsClient.ts
```diff
@@ -102,6 +102,7 @@ export const campaignFromQuery: (query: Query) => SetPayload = flow(
 class GoogleAnalyticsClient {
   private trackerNames: string[] = [];
   private pendingCommands: PendingCommand[] = [];
+  private trackingIsDisabled: boolean = !!Cookies.get(disableAnalyticsCookie);
```
PiotrKozlowski, line 105: This checks for the cookie only on the initialization of the class.
The cookie in theory may be set also in the middle of work so I think it should be a method

PiotrKozlowski (changes requested):

src/gateways/googleAnalyticsClient.ts
```diff
@@ -182,6 +181,10 @@ class GoogleAnalyticsClient {
     this.flushPendingCommands();
   }
 
+  private trackingIsDisabled(): boolean {
```
PiotrKozlowski, line 184: I just realised that Tom might have in mind to make this a normal function in `helpers/analytics.ts`  since he mentioned "other analytics systems"

jarosik10: But the cookie in `helpers/analyticsEvents/constants.ts` should stay, right? and just make `trackingIsDisabled` as a exported function in `helpers/analytics.ts`.

PiotrKozlowski: I think in this case we should move it to the `helpers/contants.ts`

PiotrKozlowski (commented):
Now I think it needs only unit tests

src/gateways/googleAnalyticsClient.spec.ts
```diff
@@ -122,6 +123,20 @@ describe('GoogleAnalyticsClient', () => {
...
+      it('doens\'t send and flush the commands if trackingIsDisabled is true', async() => {
+        jest.spyOn(analytics, 'trackingIsDisabled').mockReturnValueOnce(true);
+
+        client.trackPageView('/some/path');
+        expect(mockGa).not.toHaveBeenCalled();
+        expect(client.getPendingCommands().length).toBe(1);
+
+        client.setTrackingIds(['foo']);
+        expect(mockGa).not.toHaveBeenCalledWith('tfoo.set', 'queueTime', expect.any(Number));
+        expect(mockGa).not.toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/some/path'});
```
PiotrKozlowski, line 136: maybe change this to `expect(mockGa).not.toHaveBeenCalled()` to be sure that it is really not called cuz this test will pass if someone mistypes the name

PiotrKozlowski: same for the other test

PiotrKozlowski (changes requested):

src/helpers/analytics.ts
```diff
@@ -21,6 +22,7 @@ import * as search from './analyticsEvents/search';
 import * as openCloseStudyGuides from './analyticsEvents/studyGuides/openClosePopUp';
 import * as openUTG from './analyticsEvents/studyGuides/openUTG';
 import * as unload from './analyticsEvents/unload';
+import { disableAnalyticsCookie } from './constants';
```
TomWoodward, line 25: you can define this in this same file, `helpers` don't really have a common theme so a shared constants file might be confusing

TomWoodward (changes requested):

PiotrKozlowski (commented):
No more comments from me, there is one from Tom

src/helpers/analytics.ts
jarosik10: But we want to have that `disableAnalyticsCookie` name exported, right?

PiotrKozlowski: I think linter will yell at you if you export a const that is not used in another file so for now, I think it shouldn't be exported

PiotrKozlowski (approved):

TomWoodward (approved):

PR #1059, opened 2021-01-29
Add script for creating redirects.json file

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1466

testing:
Add this to the `createRedirects.ts`
```
createRedirects(
  createArchiveLoader(`${ARCHIVE_URL}${REACT_APP_ARCHIVE_URL}`),
  createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`)
);
```
run:
```
node ./script/entry.js prerender/createRedirects
```

To make it run you'll need some json files in `data/redirects/` folder:
```
data/redirects/9ab4ba6d-1e48-486d-a2de-38ae1617ca84.json

[
  {
    "bookId": "9ab4ba6d-1e48-486d-a2de-38ae1617ca84",
    "pageId": "7b0acdae-a602-4ff3-ad0f-814d36f6fea4",
    "pathname": "some-non-existing-link"
  },
  {
    "bookId": "9ab4ba6d-1e48-486d-a2de-38ae1617ca84",
    "pageId": "beaf2bad-3650-4195-8467-28c50ead4999",
    "pathname": "some-non-existing-link2"
  }
]
```


script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,45 @@
...
+          // tslint:disable-next-line: no-console
+          console.log(`Couldn't find page ${pageId} in book ${bookId}`);
+          return null;
+        }
+
+        return {
+          from: pathname,
+          to: content.getUrl({ book: { slug: bookSlug }, page: { slug: page.slug } }),
+        };
+      }).filter((data) => Boolean(data)) as Array<{ from: string, to: string }>
```
TomWoodward, line 38: probably cleaner to rewrite this as a for loop so your `!page` can noop and push the other records since you already have a for loop over books

the map/filter approach would be cleaner if you mapped `{pageId, pathname}` to `{page, pathname}` then filtered the null records, then mapped to `{from, to}`

TomWoodward (changes requested):

TomWoodward:
changed base so other change would auto merge

script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const createRedirects = async(archiveLoader: AppServices['archiveLoader'], osWebLoader: AppServices['osWebLoader']) => {
+  const bookLoader = makeUnifiedBookLoader(archiveLoader, osWebLoader);
+
+  const books = fs.readdirSync(redirectsPath).filter((name) => name.match('.json'));
+
+  const redirects: Array<{ from: string, to: string }> = [];
+
+  for (const bookName of books) {
+    const bookRedirects: Redirects = require(redirectsPath + '/' + bookName);
```
TomWoodward, line 20: same thing about the mixed require/import, we should be using import consistently or we'll have compatibility issues

script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,43 @@
...
+    const bookId = bookName.replace('.json', '');
```
TomWoodward, line 21: this is confusing, can `bookName` be `fileName`?

script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { tree, slug: bookSlug } = await bookLoader(bookId);
```
TomWoodward, line 22: you want to use the current version here, omitting it will get the _latest_ version which we may not have adopted yet, and the ability to not pass a version will go away when we adopt the new web pipeline archive

TomWoodward (changes requested):

script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (!configForBook) {
+      // tslint:disable-next-line: no-console
+      console.log(`Couldn't find version for book: ${bookId}`);
+      continue;
+    }
+
+    const { tree, slug: bookSlug } = await bookLoader(bookId, configForBook.defaultVersion);
+
+    for (const { pageId, pathname } of bookRedirects) {
```
TomWoodward, line 33: maybe we should use the bookId from here instead of the filename? then the way that we break down the files won't necessarily be coupled to the logic in this script

TomWoodward (changes requested):

script/prerender/createRedirects.ts
PiotrKozlowski: oh, wait, you want the `bookLoader` to use the `bookId` from the `bookRedirects`?

This will require call `bookLoader` for each entry in `bookRedirects` - it shouldn't be a problem since we are caching the results - do I understand correctly?

script/prerender/createRedirects.ts
```diff
@@ -0,0 +1,52 @@
...
+    const bookIdFromFilename = fileName.replace('.json', '');
+    const configForBook: { defaultVersion: string } | undefined = config[bookIdFromFilename];
+
+    if (!configForBook) {
+      // tslint:disable-next-line: no-console
+      console.log(`Couldn't find version for book: ${bookIdFromFilename}`);
+      continue;
+    }
+
+    const { tree, slug: bookSlug } = await bookLoader(bookIdFromFilename, configForBook.defaultVersion);
```
TomWoodward, line 31: yea my intent was to not use the filename for anything. i see now that would mean moving the async call into the page loop.... i guess thats probably fine because of the caching. i like that it removes the assumption that the files are broken down by book and uses the bookId that was put in the json structure

TomWoodward (changes requested):

TomWoodward (approved):

PR #1060, opened 2021-01-29
Issue 1358

PiotrKozlowski:
TODO: 
- [x] Render html passed as variable

for: https://github.com/openstax/unified/issues/1358


src/app/content/components/__snapshots__/Attribution.spec.tsx.snap
```diff
@@ -0,0 +1,460 @@
...
+        Authors: {bookAuthors}
+      </li>
+      <li>
+        Publisher/website: OpenStax
+      </li>
+      <li>
+        Book title: {bookTitle}
+      </li>
+      <li>
+        Publication date: {bookPublishDate, date, medium}
```
jarosik10, line 319: Is it expected behavior?

PiotrKozlowski: nice catch, fixed

jarosik10 (approved):
Looks fine to me:)

src/app/messages/en/attributionText.ts
```diff
@@ -62,15 +62,17 @@ export const defaultText = `
 </p>
 `;
 
-export const teaText = `
+export const specialText = `
 <p>
   Want to cite, share, or modify this book? This book is
   <a target="_blank" rel="noopener" href="https://creativecommons.org/licenses/by/4.0/">
```
TomWoodward, line 68: why is there a link here? it might be wrong if the book isn't actually creative commons

src/app/messages/en/attributionText.ts
```diff
@@ -62,15 +62,17 @@ export const defaultText = `
...
    {bookLicenseName} {bookLicenseVersion}
   </a>
-  and you must attribute “Texas Education Agency (TEA)." The original material is available at: <a target="_blank"
-  rel="noopener" href="https://www.texasgateway.org/book/{teaBookName}">https://www.texasgateway.org/book/{teaBookName}
-  </a>. Changes were made to the original material, including updates to art, structure, and other content updates.
+  and you must attribute {copyrightHolder}.{displayOriginalMaterialInformation, select, yes {
```
TomWoodward, line 71: i feel like you can just switch on the presence of the `originalMaterialLink` link here.

src/app/messages/en/attributionText.ts
```diff
@@ -62,15 +62,17 @@ export const defaultText = `
...
+    The original material is available at: <a target="_blank" rel="noopener" href="{originalMaterialLink}">
+    {originalMaterialLink}
+  </a>.} no {}} Changes were made to the original material, including updates to art, structure, and
```
TomWoodward, line 74: maybe add another variable, or use that `displayOriginalMaterialInformation` one, for this text. i'd call it like `isDerivativeWork` or something

src/app/messages/en/attributionText.ts
```diff
@@ -124,7 +126,7 @@ export const teaText = `
 </ul>
 
 <p>
-  © {bookLatestRevision, date, medium} Texas Education Agency (TEA). <strong>The OpenStax name, OpenStax logo,
+  © {bookLatestRevision, date, medium} {copyrightHolder}. <strong>The OpenStax name, OpenStax logo,
```
TomWoodward, line 129: use the `isDerivativeWork` again here to omit the part that is omitted

TomWoodward (changes requested):
i think if you address the comments the two attribution messages can be combined

src/app/messages/en/attributionText.ts
PiotrKozlowski: I've doubled checked [the documentation](https://formatjs.io/docs/getting-started/installation) and I didn't found anything like this

src/app/messages/en/attributionText.ts
PiotrKozlowski: this link was hardcoded for tea books. I've removed it, but maybe it would be better to add more variables like `showLicenceLink` and `licenseLink`?

TomWoodward (changes requested):

src/app/messages/en/attributionText.ts
```diff
@@ -55,77 +60,10 @@ export const defaultText = `
...
-  {bookLicenseName} {bookLicenseVersion} license. <strong>The OpenStax name, OpenStax logo, OpenStax book
-  covers, OpenStax CNX name, and OpenStax CNX logo are not subject to the Creative Commons license and may
-  not be reproduced without the prior and express written consent of Rice University.</strong>
-</p>
-`;
-
-export const teaText = `
-<p>
-  Want to cite, share, or modify this book? This book is
-  <a target="_blank" rel="noopener" href="https://creativecommons.org/licenses/by/4.0/">
```
TomWoodward, line 68: the TEA physics book requires that this link be here. we can add the link unconditionally to the unified attribution message but we have to use the license link out of the archive data `book.license.url`

TomWoodward (changes requested):

src/app/content/components/Attribution.tsx
```diff
@@ -97,9 +97,30 @@ interface Props {
 
 class Attribution extends Component<Props> {
   public container = React.createRef<HTMLDetailsElement>();
-  private bookIdsWithTEAAttributionText: {[key: string]: string} = {
-    '394a1101-fd8f-4875-84fa-55f15b06ba66': 'tea-statistics',
-    'cce64fde-f448-43b8-ae88-27705cceb0da': 'tea-physics',
+  private bookIdsWithSpecialAttributionText: {
+    [key: string]: {
+      copyrightHolder?: string,
+      displayOriginalMaterialInformation?: 'yes' | 'no',
```
TomWoodward, line 103: why are you using yes/no instead of true/false?

src/app/messages/en/attributionText.ts
```diff
@@ -1,76 +1,14 @@
...
   Want to cite, share, or modify this book? This book is
-  <a target="_blank" rel="noopener" href="https://creativecommons.org/licenses/by/4.0/">
-   {bookLicenseName} {bookLicenseVersion}
-  </a>
-  and you must attribute “Texas Education Agency (TEA)." The original material is available at: <a target="_blank"
-  rel="noopener" href="https://www.texasgateway.org/book/{teaBookName}">https://www.texasgateway.org/book/{teaBookName}
-  </a>. Changes were made to the original material, including updates to art, structure, and other content updates.
+  <a target="_blank" rel="noopener" href="{bookLicenseUrl}">
+    {bookLicenseName} {bookLicenseVersion}
+  </a> and you must attribute {copyrightHolder}.{displayOriginalMaterialInformation, select, yes {
```
TomWoodward, line 6: do you really need displayOriginalMaterialInformation here? can you just check if `originalMaterialLink` is defined?

TomWoodward (changes requested):
this is looking *_so much_* better

src/app/messages/en/attributionText.ts
PiotrKozlowski: how? I didn't found anything about displaying text conditionally for defined / not defined vars
I've found out that if you use a variable in the text and you don't provide this variable to the component then it will fail

I think I could do something like `{originalMaterialLink, select, false {} other {link provided}}` and then set `originalMaterialLink: 'false'` as a default prop. Is this what you had in mind?

TomWoodward: yea some kind of empty value not necessarily undefined/missing

src/app/content/components/Attribution.tsx
```diff
@@ -97,9 +97,30 @@ interface Props {
 
 class Attribution extends Component<Props> {
   public container = React.createRef<HTMLDetailsElement>();
-  private bookIdsWithTEAAttributionText: {[key: string]: string} = {
-    '394a1101-fd8f-4875-84fa-55f15b06ba66': 'tea-statistics',
-    'cce64fde-f448-43b8-ae88-27705cceb0da': 'tea-physics',
+  private bookIdsWithSpecialAttributionText: {
+    [key: string]: {
+      copyrightHolder?: string,
+      displayOriginalMaterialInformation?: boolean,
```
TomWoodward, line 103: this is unused now right?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1098, opened 2021-02-02
Event capture

TomWoodward:
for: openstax/unified#1479

- had to separate some combined open/close events into their own events
- added flag for default color being used as part of wrangling required data for create highlight event
- updated some types for highlight actions
- moved highlight create tracking into redux hook because previous location in the ui didn't have access to the rest of the highlight data, had to add some metadata to the action
- moved around/broke down some of the analytics helper files

i want to verify the reporting works but after that the environment.json should be reverted


bethshook (commented):
This looks good to me but I'll defer to @PiotrKozlowski for approval

src/helpers/analytics/events/highlighting/createNote.ts
```diff
@@ -0,0 +1,33 @@
...
+  highlight: NewHighlightPayload,
+  {isDefaultColor}: {isDefaultColor?: boolean}
+): AnalyticsEvent | void => {
+  return {
+    getEventCapturePayload: () => createdHighlight({
+      ...highlight,
+      annotation: highlight.annotation || '',
+      highlightId: highlight.id,
+      locationStrategies: JSON.stringify(highlight.locationStrategies),
+      sourceMetadata: '', // idk what this is for
```
PiotrKozlowski, line 25: do we want to do anything about this?

bethshook (approved):

PR #1100, opened 2021-02-03
Filter out approved-books.json by tutor_only and server fields

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1472

### DIFF
--- a/concourse/update-content/script.bash
+++ b/concourse/update-content/script.bash
@@ -17,7 +17,7 @@ approved_books_default_branch=$(curl -s https://api.github.com/repos/openstax/co
 rex_default_branch=$(curl -s https://api.github.com/repos/openstax/rex-web | jq -r .default_branch)
 
 approved_books=$(curl -sL "https://github.com/openstax/content-manager-approved-books/raw/$approved_books_default_branch/approved-books.json")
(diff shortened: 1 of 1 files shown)


concourse/update-content/script.bash
```diff
@@ -17,7 +17,7 @@ approved_books_default_branch=$(curl -s https://api.github.com/repos/openstax/co
 rex_default_branch=$(curl -s https://api.github.com/repos/openstax/rex-web | jq -r .default_branch)
 
 approved_books=$(curl -sL "https://github.com/openstax/content-manager-approved-books/raw/$approved_books_default_branch/approved-books.json")
-book_ids=$(jq -r '.[].uuid' <<< "$approved_books")
+book_ids=$(jq -r 'map(select(.tutor_only == false)) | map(select(.server == "cnx.org")) | .[].uuid' <<< "$approved_books")
```
TomWoodward, line 20: this filtering looks right but it needs to be done on `approved_books` not the ids, otherwise we might still use the versions pulled out down on line 33

this line should unique the ids though, otherwise we might process some of them more than once

TomWoodward (changes requested):

TomWoodward (approved):

PR #1105, opened 2021-02-08
Add validate-modified-books script

jarosik10:
for: https://github.com/openstax/unified/issues/1477


.github/workflows/ci.yml
```diff
@@ -70,26 +64,9 @@ jobs:
         path: node_modules
         key: node_modules3
     - run: yarn install
-    - run: |
-        node script/entry.js domVisitor errorsExist \
-          --bookId="$BOOK_ID" \
-          --bookVersion="$BOOK_VERSION" \
-          --rootUrl "$BASE_URL" \
-          --queryString="validateLinks"
+    - run: bash ./script/validate-modified-books.bash "$BASE_URL"
```
TomWoodward, line 67: pretty sure you don't need `bash` on the front of this. and i think it would be fine to use `BASE_URL` from the environment instead of a numbered param inside the script

script/validate-modified-books.bash
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+modified_books=$(echo $(./script/get-modified-books.bash))
```
TomWoodward, line 2: why the extra echo?

script/validate-modified-books.bash
```diff
@@ -0,0 +1,19 @@
...
+    --bookVersion="$book_version" \
+    --rootUrl $1 \
+    --queryString="validateLinks"
+
+  node script/entry.js urlChecker \
+    --rootUrl="https://openstax.org" \
+    --bookId="$book_id" \
+    --bookVersion="$book_version" \
+    --useUnversionedUrls
+done
```
TomWoodward, line 19: this script won't fail if the commands here return errors. you're gonna have to do some more work to capture the exit codes and then exit non-zero at the end if any of the books failed

TomWoodward (changes requested):

jarosik10:
I see that Lint doesn't like this type of exit code checking `if [ $? != 0 ]; then`

TomWoodward:
yea i think it would be better to have like a `code=0` at the top and then put `|| code=1` after the commands and `exit "$code"` at the end, that way if any of the books fail it still runs the rest of the books before exiting, so we can see all the potential errors instead of jus the first one

TomWoodward (changes requested):
the comment i just left

TomWoodward:
can you try pushing some book versions to this branch and see if this script's CI fails correctly?

it think these versions were broken:
Principles of Accounting, Volume 1: Financial Accounting (9ab4ba6d-1e48-486d-a2de-38ae1617ca84) 7.7
Principles of Accounting, Volume 2: Managerial Accounting (920d1c8a-606c-4888-bfd4-d1ee27ce1795) 17.5

jarosik10:
@TomWoodward i've pushed the broken versions.

script/validate-modified-books.bash
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+modified_books=$(./script/get-modified-books.bash)
+code=0
+
+for row in $(jq -c '.[]' <<< "$modified_books"); do
+  book_id=$(jq -r '.book_id' <<< "$row")
+  book_version=$(jq -r '.book_version' <<< "$row")
+
+  node script/domVisitor errorsExist \
```
PiotrKozlowski, line 9: `node script/entry.js domVisitor errorsExist`

PiotrKozlowski (dismissed):

script/validate-modified-books.bash
jarosik10: Thanks

TomWoodward:
if it was running the scans i would expect to see the output coming through, i think it might be hanging up on something before that. consider adding `set -x` to the top of the script so you can see what it is doing

TomWoodward:
not really i suspect its hanging

TomWoodward:
yea you can see in the output the book id and version are not being set

image

jarosik10:
@TomWoodward I think it might be a problem with the whitespaces inside json `"title":"Principles of Accounting, Volume 1: Financial Accounting"`.

TomWoodward:
the internet is telling me that we're supposed to use a `while read -r line` loop instead of a `for` if we really want the lines (which we do)

https://stackoverflow.com/a/10752407/14809536

jarosik10:
@TomWoodward it seems to work fine. Do you want me to undo the changes in `config.books.json`?

TomWoodward:
yea it looks like the scan correctly failed, lets put the book config back

TomWoodward (changes requested):
the book config

TomWoodward (approved):

PR #1107, opened 2021-02-11
Allow focusing highlights with keyboard

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1491

requires: https://github.com/openstax/highlighter/pull/42 

TODO:
- [x] Fix tests
- [x] Update version of @openstax/highlighter when required PR is merged


PiotrKozlowski:
@jarosik10 this should be a Draft pr, work is still in progress

PiotrKozlowski:
`lint` and `build` CIs are failing because of old version of `highlighter`

PiotrKozlowski:
@jarosik10 , @TomWoodward this is ready to review

src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -81,11 +82,12 @@ const handleUpdate = (services: Services) => (
   selectResult(services, previous, current, options);
 };
 
-const searchHighlightManager = (container: HTMLElement) => {
+const searchHighlightManager = (container: HTMLElement, intl: IntlShape) => {
   const services = {
     container,
     highlighter: new Highlighter(container, {
       className: 'search-highlight',
+      formatMessage: ({ id }) => intl.formatMessage({ id }, { style: 'search' }),
```
jarosik10, line 90: Could you give some explanation about `{ style: 'search' }`?

src/app/content/components/Page/highlightManager.ts
```diff
@@ -46,7 +49,7 @@ export type HighlightProp = ReturnType<typeof mapStateToHighlightProp>
   & ReturnType<typeof mapDispatchToHighlightProp>;
 
 // deferred so any cards that are going to blur themselves will have done so before this is processed
-const onClickHighlight = (services: HighlightManagerServices, highlight: Highlight | undefined) => defer(async() => {
+const onFocusHighlight = (services: HighlightManagerServices, highlight: Highlight | undefined) => defer(async() => {
```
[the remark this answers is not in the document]
PiotrKozlowski: my thought process was like this: both, focus and click events are focusing a button so let's call it `onFocusHighlight` :D

src/app/content/components/Page/searchHighlightManager.ts
PiotrKozlowski: `Highlighter` will call `formatMessage` with `{id: string, style: IHighlightData['style'] | undefined}`
This mean that `style` is basically set by us, and normal highlights have styles like `yellow`, `green`, etc. but search highlight doesn't have `style` property and we need it to create text https://github.com/openstax/rex-web/pull/1107/files#diff-51e75076ae4cd9a4653ce8a1403c42dab5790b384f89645d40abb917e68dde80R36

The `id` is created in `Highlighter`: https://github.com/openstax/highlighter/blob/master/src/injectHighlightWrappers.ts#L68

jarosik10: Ok, I see. Thanks for the explanation.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -76,10 +79,29 @@ const onSelectHighlight = (
...
 });
 
-const createHighlighter = (services: Omit<HighlightManagerServices, 'highlighter'>) => {
+const createHighlighter = (services: Omit<HighlightManagerServices, 'highlighter'>, intl: IntlShape) => {
 
   const highlighter: Highlighter = new Highlighter(services.container, {
-    onClick: (highlight) => onClickHighlight({ ...services, highlighter }, highlight),
+    formatMessage: intl.formatMessage,
+    onClick: (highlight) => onFocusHighlight({ ...services, highlighter }, highlight),
+    onFocusIn: (highlight) => onFocusHighlight({ ...services, highlighter }, highlight),
```
jarosik10, line 87: Btw. doesn't click automatically fire `focusin` event?

TomWoodward: theoretically in normal circumstances yes, but the way its implemented in highlighter the elements you click on and the element you tab focus aren't the same, so in this case you need both.

@PiotrKozlowski it might be a good idea for the future to simplify the interface to get rid of `onClick` and just call `onFocusIn` from the click handler instead

src/app/content/components/Page/highlightManager.ts
```diff
@@ -76,10 +79,29 @@ const onSelectHighlight = (
...
+      if (
+        activeElement
+        && findFirstAncestorOrSelf(
+          activeElement,
+          (node) => node.hasAttribute('data-highlight-card') || node.hasAttribute('data-for-screenreaders')
+        )
+      ) {
+        return;
+      }
+      services.getProp().clearFocus();
```
TomWoodward, line 103: the part about not clearing if the focus was moved to another highlight is supposed to have been done already because of this https://github.com/openstax/rex-web/pull/1107/files#diff-69f29777fa0e95134d5f3990ffbc23328b6078e8f0fbaf3d54faf3219e56941bR51

can you also move this whole `onFocusOut` handler up next to `onFocusHighlight`?

TomWoodward (changes requested):
this looks good, i have some small concern about the `onFocusOut` logic but it doesn't seem like a big deal

src/app/content/components/Page/highlightManager.ts
PiotrKozlowski: We could remove `|| node.hasAttribute('data-for-screenreaders')` from this function and it will work however in this case it will first dispatch `clear` and then it will `focus` new highlight
with this check it will only `focus` new highlight, without clearing the old one before
this shouldn't change anything for the user but it is stopping unnecessary `dispatch`

I've moved the handler to separate function

jarosik10 (approved):

PiotrKozlowski:
@TomWoodward can you restart the CI for `test:build` - it timed out

TomWoodward:
i'm worried about the codecov checks that dont seem to be reporting on this, @jarosik10 do you know why that might be happening?

TomWoodward (approved):

PR #1108, opened 2021-02-19
Issue 1493

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1493


PiotrKozlowski:
I'm not sure why the build failed. It is passing locally

TomWoodward (changes requested):
i get the same error as CI when running `yarn test:build` locally, seems to be breaking on pre-render

PiotrKozlowski:
@TomWoodward 65eb842 this commit fixes issue with `yarn test:build`, but I don't understand why - this code works on `master`. Do you want me to research that?

TomWoodward:
you could try bisecting down the exact change on this branch that caused it to start failing, could be some related thing

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -32,20 +32,31 @@ export type ContentLinkProp =
...
-    const search = content.getSearch && content.getSearch(reference.params);
-    const query = search ? `?${search}` : '';
-    const a = assertNotNull(
-      document.querySelector(`[href^='${reference.match}']`),
-      'references are created from hrefs');
-    const href = assertNotNull(a.getAttribute('href'), 'it was found by href value')
-      .replace(reference.match, toRelativeUrl(currentPath, path) + query);
-    a.setAttribute('href', href);
+    // references may contain PageReferenceMapError only if UNLIMITED_CONTENT is set to true
+    if (isPageReferenceMapError(reference)) {
```
TomWoodward, line 36: this function would be less intimidating if you split the two sections out into separate little helpers

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -32,20 +32,31 @@ export type ContentLinkProp =
...
+      const href = assertNotNull(a.getAttribute('href'), 'it was found by href value')
+        .replace(reference.match, toRelativeUrl(currentPath, path) + query);
+      a.setAttribute('href', href);
+    }
   }
 };
 
-const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageReferenceMap) =>
-  content.getUrl(ref.params) === pathname
+const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageReferenceMap | PageReferenceMapError) =>
```
TomWoodward, line 56: can we rename `PageReferenceMapError` to `PageReferenceError`?

src/app/content/types.ts
```diff
@@ -37,14 +38,25 @@ export interface State {
...
+export interface PageReferenceMapError {
+  reference: ReturnType<typeof getContentPageReferences>[number];
+  type: 'error';
+}
+
+/**
+ * Error created when we can't load a reference. For example when target book doesn't exist
+ * or the rex config doesn't have defaultVersion for this book.
+ */
+export class ReferenceLoadingError extends Error {}
```
TomWoodward, line 58: this isn't a type

src/app/content/guards.ts
```diff
@@ -18,3 +22,16 @@ export const isLinkedArchiveTreeSection =
 
 export const hasOSWebData = (book: Book | undefined): book is BookWithOSWebData =>
   book ? 'slug' in book : false;
+
+export const isReferenceLoadingError = (something: any): something is ReferenceLoadingError =>
```
TomWoodward, line 26: is there value in this guard? you can just do an inline instanceof check

src/app/content/guards.ts
```diff
@@ -18,3 +22,16 @@ export const isLinkedArchiveTreeSection =
...
+  something instanceof ReferenceLoadingError;
+
+export const isContentPageRefencesType = (something: any): something is ContentPageRefencesType =>
+  isObject(something)
+  && typeof (something as ContentPageRefencesType).match === 'string'
+  && typeof (something as ContentPageRefencesType).pageId === 'string';
+
+export const isPageReferenceMapError = (something: any): something is PageReferenceMapError =>
```
TomWoodward, line 34: you can start with the interface from the state here instead of `any` and it'll probably be a lot cleaner 

i think maybe just `{match: string}` would be the place to start

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -178,14 +178,17 @@ export const getBookInformation = async(
 
   } else if (UNLIMITED_CONTENT) {
     for (const {id, bookVersion} of allReferences) {
+      // Omit references that don't have bookVersion because new pipeline will not support requests without a version
+      if (!bookVersion) { continue; }
```
TomWoodward, line 182: for now we support both, if we didn't support the old one i'm not sure this check would belong here though, what is it here for?

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -178,14 +178,17 @@ export const getBookInformation = async(
...
+
       const osWebBook =  await services.osWebLoader.getBookFromId(id).catch(() => undefined);
       const archiveBook = await services.archiveLoader.book(id, bookVersion).load();
       if (archiveBook && archiveTreeSectionIsBook(archiveBook.tree)) {
         return {osWebBook, archiveBook};
       }
     }
   }
-  return undefined;
+  return new ReferenceLoadingError();
```
TomWoodward, line 191: should this throw instead of return?

actually i feel like it would have been easier to just leave this as `undefined`

src/app/content/components/Page.spec.tsx
```diff
@@ -61,6 +62,32 @@ const makeEvent = (doc: Document) => {
...
+      },
+    },
+    state: {
+      bookUid: 'book',
+      bookVersion: 'version',
+      pageUid: 'page',
+    },
+  },
+  {
+    reference: {
```
TomWoodward, line 83: can you change the error format to be more consistent with the regular format? otherwise you get stilly stuff like `reference.reference`, i feel like the error should just have keys `match` and `type` right?

TomWoodward (changes requested):

src/app/content/components/Page.spec.tsx
```diff
@@ -549,6 +560,35 @@ describe('Page', () => {
     }));
   });
 
+  it('interceptes clicking links that failed due to ReferenceLoadingError', async() => {
+    const {root} = renderDomWithReferences();
+
+    const spyAlert = jest.spyOn(globalThis as any, 'alert');
+
+    // page lifecycle hooks
+    await Promise.resolve();
```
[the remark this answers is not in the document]
PiotrKozlowski: I've copied this from another test, but it look like this is working even without this line
This line is used 44 times in this file.
Removing all of them breaks 10 tests - should I remove it and fix the tests with `setImmediate` in this PR?

src/app/content/hooks/locationChange/resolveContent.ts
PiotrKozlowski: If I understand correctly, the `bookVersion` may be `undefined` if a book is not in the rex config and the AC for this issue is to change links to pop alerts in this case

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -196,6 +199,12 @@ export const resolveExternalBookReference = async(
 ) => {
   const bookInformation = await getBookInformation(services, reference);
 
+  // Don't throw an error if reference couldn't be loaded when UNLIMITED_CONTENT is truthy
+  // It will be processed in contentLinkHandler.ts
+  if (UNLIMITED_CONTENT && !bookInformation) {
+    return bookInformation;
+  }
```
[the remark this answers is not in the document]
PiotrKozlowski: yes, if `UNLIMITED_CONTENT` is set to `true` we will process `undefined` data from `getBookInformation`
if `UNLIMITED_CONTENT` is not set then we will throw an error

src/app/content/components/Page.spec.tsx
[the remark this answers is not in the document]
TomWoodward: you don't have to change stuff that you aren't touching

src/app/content/hooks/locationChange/resolveContent.ts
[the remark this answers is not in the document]
TomWoodward: yea the code would be cleaner if `resolveExternalBookReference` returned a reference instead of a book, then you'd need less conditionals in `loadContentReference`, but the way this is set up is fine, its because of the way the logic was being done already

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
PiotrKozlowski: `services.archiveLoader.getBookIdsForPage` is returning `Promise<BookReference[]>` and `BookReference` is typed as `{id: string; bookVersion: string | undefined; }`

is this possible that we will not have `bookVersion` at this stage?

TomWoodward: `getBookIdsForPage` is not used for web pipeline content. at this point the `bookVersion` key is optional, but it should be ok because `services.archiveLoader.book` accepts `bookVersion` as optional

TomWoodward (approved):

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -168,23 +168,27 @@ export const getBookInformation = async(
...
-
-    return {osWebBook, archiveBook};
-
-  } else if (UNLIMITED_CONTENT) {
-    for (const {id, bookVersion} of allReferences) {
-      const osWebBook =  await services.osWebLoader.getBookFromId(id).catch(() => undefined);
-      const archiveBook = await services.archiveLoader.book(id, bookVersion).load();
-      if (archiveBook && archiveTreeSectionIsBook(archiveBook.tree)) {
-        return {osWebBook, archiveBook};
+  try {
```
TomWoodward, line 171: this is a very aggressive try/catch, i think it would also work to change the one line (182 in this diff)

```
const archiveBook = await services.archiveLoader.book(id, bookVersion).load();
```

to

```
const archiveBook = await services.archiveLoader.book(id, bookVersion).load().catch(() => undefined);
```

this way the change will only affect the one query we actually expect to fail within the `UNLIMITED_CONTENT` portion of the logic

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
PiotrKozlowski: For the page that you've posted in the slack: `/books/university-physics-volume-2@86cddf1/pages/1-6-mechanisms-of-heat-transfer?archive=https%3A%2F%2Fopenstax.org%2Fapps%2Farchive-preview%2F20210208.162118` it is failing for lines 174-176 - so it is in `configuredReference`

Will it be okay to remove `try catch` and replace the whole block with: ?
```ts
if (configuredReference) {
    const osWebBook =  await services.osWebLoader.getBookFromId(configuredReference.id);
    const archiveBook = await services.archiveLoader.book(
      configuredReference.id, BOOKS[configuredReference.id].defaultVersion
    ).load().catch(() => undefined);

    if (archiveBook && archiveTreeSectionIsBook(archiveBook.tree)) {
      return {osWebBook, archiveBook};
    }
  } else if (UNLIMITED_CONTENT) {
    for (const {id, bookVersion} of allReferences) {
      const osWebBook =  await services.osWebLoader.getBookFromId(id).catch(() => undefined);
      const archiveBook = await services.archiveLoader.book(id, bookVersion).load().catch(() => undefined);
      if (archiveBook && archiveTreeSectionIsBook(archiveBook.tree)) {
        return {osWebBook, archiveBook};
      }
    }
  }
```

So we add `.catch(() => undefined)` in both branches of if statement and in the first one we also return only if `archiveBook && archiveTreeSectionIsBook(archiveBook.tree)`

TomWoodward: uhhh let me think this through...

- the `getBookInformation` is only used for other-book links
- web-pipeline content doesn't have a version in the link, the version comes from the book config in `getInputReferenceInfo`
- so the `else if (UNLIMITED_CONTENT) {` block would never be hit for web-pipeline content, it would only be used for cnx.org content that was loading arbitrary book ids through `archiveLoader.getBookIdsForPage`

so lets:
- do the change with the catch on `archiveLoader` within `UNLIMITED_CONTENT` as we discussed already
- also add a comment at the top of that block `//this section only used for cnx.org content, delete when web-pipeline (RAP) is adopted

and:
- add the catch to `services.archiveLoader` in the `configuredReference` block as you described, but instead of just returning undefined, rethrow the error if `UNLIMITED_CONTENT` is unset. (because this logic is also used in production)

TomWoodward (changes requested):

TomWoodward (approved):

PR #1110, opened 2021-02-22
Issue 1495 - allow moving focus between highlights in document and their edit popup 

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1495

~This branch is based on branch issue-1491 since it requires highlight manager settings for screenreaders~

Update 6th April: I've changed the base to `master` so pr for issue 1491 can be merged

### Testing

~You have to link @openstax/highlighter from branch `highlights-tabbing`~

Use the shortcut from file `src/app/content/highlights/constants.ts` (currently <kbd>H</kbd> key) to move the focus between highlight and card.


PiotrKozlowski:
@TomWoodward @jarosik10 it is still a draft because I didn't add tests, but I would like to have you review this early

TomWoodward (changes requested):
overall i think this looks reasonable. i am a little confused because there was some focus management stuff in there already for highlights and now we're adding some new stuff and i'm not sure how it should interact with the stuff that was already there, what the difference in functionality is, maybe some of the stuff can be combined somehow?

src/app/content/highlights/components/Card.tsx
```diff
@@ -43,9 +44,12 @@ export interface CardProps {
   data?: HighlightData;
   className: string;
   zIndex: number;
+  // If this prop is true then focus will be moved to this card.
+  shouldFocusCard: boolean;
   topOffset?: number;
   highlightOffsets?: { top: number, bottom: number };
   onHeightChange: (ref: React.RefObject<HTMLElement>) => void;
+  moveFocusToTheHighlight: () => void;
```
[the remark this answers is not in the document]
PiotrKozlowski: it is moving focus to the screenreader node before highlight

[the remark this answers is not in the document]
PiotrKozlowski: I could modify `Highlight.focus()` method to do the same, but we still have to reset `shouldFocusCard` state because if we don't then focusing another highlight would automatically move focus to the card
It is behaving like this because it is implemented as a hook that triggers focus if this props is set.
I was thinking on giving unique data-attr for each card and focus it using normal js, but then we can have problems related to showing cards without notes (we can't focus them until they show up)

src/app/content/components/Page/highlightManager.ts
```diff
@@ -81,9 +82,24 @@ const createHighlighter = (services: Omit<HighlightManagerServices, 'highlighter
...
+    onFocusOut: () => {
+      defer(() => {
+        // Do not clear focus from highlight if it was moved to the related card or to another highlight
+        // This function is defered because moving focus is done in a hook inside Card component
+        // which is triggered by a hook in CardWrapper component.
+        // We still want to clear focused highlight if user TAB outside of it, for example to figure link.
+        const activeElement = assertDocument().activeElement;
+        if (
+          activeElement
+          && (activeElement.hasAttribute('data-highlight-card') || activeElement.hasAttribute('data-highlighted'))
```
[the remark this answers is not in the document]
PiotrKozlowski: If they click on another card then focus will be moved to this card and it will trigger focusing proper highlight
Clicking "H" will move focus to the highlight related to the focused card

PiotrKozlowski:
Before this PR logic was as follows:
click on highlight
-> dispatch `focusHighlight`
-> update `focused` in redux state
-> `highlightManager` is updated on `focused` change
-> call `highlight.focus()`

This PR is making sure that these criteria are met:
- Highlight can become focused when users click on it or when user TABs to it
- When a highlight is focused clicking "H" will do one of two things:
  1. If focus is outside of the CardWrapper component it will move focus to the Card component for the focused highlight
  2. If focus is within CardWrapper then the focus will be moved to the highlight in the content for the focused highlight
- Clicking or focusing element outside of the CardWrapper which is not a highlight will clear the focused highlight

This PR is not touching previous logic. Instead, it is adding a new behavior:
1. focusing highlight with a keyboard will trigger `Highlighter.option.onFocusIn` which will trigger the same behavior as clicking on a highlight
2. leaving focus from the focused highlight will trigger `onFocusOut` that will clear the `focused` highlight (dispatch `clearFocusedHighlight`)
This will noop if a newly focused element is another highlight (`<span data-highlighted>` ) or a Card component

How it is implemented:
In CardWrapper component we listen to the keyboard shortcut for moving focus to the highlight
-> if it is clicked we check if focus in in CardWrapper and if there is `focusedHighlight`
-> if focus is outside of the CardWrapper and there is focused highlight we set `shouldFocusCard` to `true` in CardWrapper and pass it as a prop to the proper Card
-> if focus is inside of the CardWrapper and there is focused highlight we move focus to the highlight in the content - we also set `shouldFocusCard` to `false` because if we don't the it trigger hook for moving focus to the card on every rerender

src/app/content/components/Page/highlightManager.ts
```diff
@@ -81,9 +82,24 @@ const createHighlighter = (services: Omit<HighlightManagerServices, 'highlighter
 
   const highlighter: Highlighter = new Highlighter(services.container, {
     formatMessage: (id: string, style?: string) => intl.formatMessage({ id }, { style }),
-    onClick: (highlight) => onClickHighlight({ ...services, highlighter }, highlight),
-    onFocusIn: (highlight) => services.getProp().focus(highlight.id),
-    onFocusOut: () => services.getProp().clearFocus(),
+    onClick: (highlight) => onFocusHighlight({ ...services, highlighter }, highlight),
+    onFocusIn: (highlight) => onFocusHighlight({ ...services, highlighter }, highlight),
+    onFocusOut: () => {
+      defer(() => {
```
TomWoodward, line 88: i don't understand why this would have to be deferred, if you pressed the key combination, sure the movement of the focus is deferred by the hook handling, but focusOut won't be fired until after that is processed right? so by the time we get in here the focus has already been moved?

src/app/content/highlights/components/Card.tsx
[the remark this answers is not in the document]
TomWoodward: `shouldFocusCard` makes some sense to me. i think `moveFocusToTheHighlight` would make more sense if we just called the regular focus function and had that focus the SR element.

for `shouldFocusCard`, you might want to consider putting it in redux state and resetting it to false on any highlight focus in/out actions and then only set it to true when the key combo is pressed. you might find that easier to manage than all the react-hooks stuff

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -29,12 +31,47 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+    }
+
+    if (element.current.contains(activeElement)) {
+      let highlightElement: HTMLElement | undefined | null;
+      focusedHighlight.elements.some((el) => {
+        highlightElement = (el as HTMLElement).querySelector('[data-for-screenreaders]') as HTMLElement | null;
+        return highlightElement;
+      });
+      if (highlightElement) {
+        highlightElement.focus();
```
TomWoodward, line 61: yea this logic is gnarly, i'd rather use highlighter.focus() and make that focus the proper element

src/app/content/highlights/components/Card.tsx
```diff
@@ -121,13 +135,14 @@ const Card = (props: CardProps) => {
     className: props.className,
     highlight: props.highlight,
     isFocused: props.isFocused,
+    moveFocusToTheHighlight: props.moveFocusToTheHighlight,
```
TomWoodward, line 138: you don't have to pass this down, you could handle it with `setEditing` on line 160

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -71,3 +71,12 @@ export const generateUpdatePayload = (
 
   return {updatePayload, preUpdateData};
 };
+
+export const noopKeyCombinationHandler = (activeElement: Element | null): boolean => {
+  if (!activeElement) { return false; }
+  if (activeElement.nodeName === 'TEXTAREA') { return true; }
+  if (activeElement.nodeName === 'INPUT') {
```
TomWoodward, line 78: i feel as if there should be a way to check if an element accepts keyboard input without whitelisting them... but i can't find anything online.

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -71,3 +71,12 @@ export const generateUpdatePayload = (
...
+    return activeElement.getAttribute('type') === 'text';
```
TomWoodward, line 79: what about other text like inputs such as 'password'

src/app/reactUtils.ts
```diff
@@ -18,25 +18,42 @@ export const useDrawFocus = <E extends HTMLElement = HTMLElement>() => {
...
 
-    if (!isElement(relatedTarget) || !ref.current!.contains(relatedTarget)) {
+    if (
+      type === 'focusout'
+      && (!isElement(relatedTarget) || !ref.current!.contains(relatedTarget))
+    ) {
+      cb();
+    } else if (
+      type === 'focusin'
+      && (isElement(relatedTarget) && ref.current!.contains(assertDocument().activeElement))
```
TomWoodward, line 40: on focusin i think e.target is the element receiving focus

src/app/reactUtils.ts
```diff
@@ -249,3 +266,41 @@ export const disableContentTabbingHandler = (isEnabled: boolean) => () => {
...
+ * Value of @param options.key is converted to lowercase before comparing.
+ * @param {KeyCombinationOptions} options
+ * @param {Function} callback
+ */
+export const useKeyCombination = (
+  options: KeyCombinationOptions,
+  callback: () => void,
+  noopHandler?: (activeElement: Element | null) => boolean
+) => {
+  const document = assertDocument();
```
TomWoodward, line 283: you're doing a lot of document stuff, i feel like you don't need that.

on focusin you can use event.target instead of document.activeElement

you could bind TWO handlers, one on the content and one of the cardwrapper, that each do ONE thing instead of switching based on complex logic

src/app/reactUtils.ts
```diff
@@ -249,3 +266,41 @@ export const disableContentTabbingHandler = (isEnabled: boolean) => () => {
...
+
+  const handler = React.useCallback((event: KeyboardEvent) => {
+    if (noopHandler && noopHandler(document.activeElement)) {
+      return;
+    }
+    for (const option in options) {
+      if (
+        (option === 'key' && event.key.toLowerCase() !== options.key!.toLowerCase())
+        || (event[option as keyof KeyCombinationOptions] !== options[option as keyof KeyCombinationOptions])
```
TomWoodward, line 292: i'm not seeing a great way to iterate like this and preserve types. you might try some formatting more like

```
for (const [option, value] of Object.entries(options)) {
```

```suggestion
        || (!(option in event) || event[option] !== value)
```

and maybe some additional thing that checks the types at runtime and does toLowerCase on both if they're strings. 

also consider extracting some of this into a `keybaordEventMatchesCombination` util

TomWoodward (changes requested):

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: My intent was to allow passing `noopHandler` as an argument to `useKeyCombination` so this function is passed only for a hook that is used specifically in the `CardWrapper` so we don't have to care about other cases
In Card components we have textareas and checkboxes

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: I think that we shouldn't disable `useKeyCombination` on every editable element because we could use it with another keyboard shortcut that uses for example `ctrlKey` so it can be used while editing
`useKeyCombination` is expecting _optional_ `noopHandler` function so it can be customized like in this case

src/app/content/components/Page/highlightManager.ts
PiotrKozlowski: I've dived deeper in this.
We are returning `null` instead of component for cards without annotation that are not focused
then we have a hook that focuses this card if `shouldFocusCard && element.current`
This hook is doing it's job fine but moving focus to this component is causing all cards to be rerendered for some reason 
This is causing `document.activeElement` to be `body`, not the card

This is not happening for other components that are in the dom all the time

---

I've managed to fix it by hiding cards without annotations with css

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: Do you mean `highlighter.focus()` or `highlight.focus()`?
We have `highlight.focus()` but it is used in `highlightManager.update()` and it is called multiple times when interacting with highlights' cards due to rerenders, so we won't be able to move focus in this method unless we want to work on `highlightManager` to debug this

On the other hand, I could add `focus` method on `highlighter` that does what you say - I'm just confused with naming, because we already have `highlighter.clearFocus` and `highligh.focus` methods that are removing/adding `FOCUS_CSS` classes to the elements.

src/app/reactUtils.ts
PiotrKozlowski: > you could bind TWO handlers, one on the content and one of the cardwrapper, that each do ONE thing instead of switching based on complex logic

So one `useKeyCombination` on `<PageContent>` and the second on `<CardWrapper>`?
I'm not convinced to use it - It would split the focusing logic into one more place and require using redux to trigger focus in the card when we can have everything in one place
Splitting code into smaller parts is a good thing, but it will require us to remember of all places which can be affected

If we add a method like `highlighter.focus()` and pass `event` to `moveFocuse` then the "complex logic" from `CardWrapper` could become:
```ts
const moveFocus = React.useCallback((event: KeyboardEvent) => {
    if (!focusedHighlight || !event.target || !element.current) {
      return;
    }

    if (element.current.contains(event.target)) {
        highlighter.focus(focusedHighlight);
        setShouldFocusCard(false); // I think it can be also removed since we have `onFocusLost` in this component that does the same
    } else {
      setShouldFocusCard(true);
    }
  }, [element, focusedHighlight]);

File: src/app/content/highlights/components/CardWrapper.tsx (hunk 4 above)
Comment (PiotrKozlowski) [reply src/app/content/highlights/components/CardWrapper.tsx:61]: or we could add a method on a highlight that will return it's screen reader nodes, sth like
`highlight.getScreenReaderNodes(): [startNode, endNode]` and we could use it like `highlight.getScreenReaderNodes()[0].focus`

Comment (PiotrKozlowski) [reply src/app/content/highlights/components/CardWrapper.tsx:61]: One more idea would be to change current name from `focus` to `applyFocusedStyles` and then add `focus` method that actually focuses the highlight

**Comment by jarosik10** [comment]
To reproduce:
1. Not logged in
2. Select some text (it shows a popup "Log in to hightlight..."
3. Press `H`
Zrzut ekranu z 2021-03-31 17-14-01

**Comment by jarosik10** [comment]
`Uncaught TypeError: this.elements[0] is undefined` can be also triggered when user is logged in:
1. Select a text which is not a highlight
2. Press `H`
3. Press `H` again

**Comment by jarosik10** [comment]
After pressing `H` it is really hard to tell what is focused: highlighted text or highlight edit-card.

**Comment by PiotrKozlowski** [comment]
@jarosik10 Good catch.
It will do the same if you are logged in and double press <kbd>h</kbd>
First <kbd>h</kbd> will move focus to the new card and the second will try to move it to the highlight but since there isn't one yet it will throw an error

Basically, there is a way to create an instance of `Highlight` that doesn't have any elements and we should at least noop in `focus` method in this case

Maybe we should also create a highlight (with default yellow color) after pressing <kbd>h</kbd> on the newly selected fragment?
Or maybe when user is creating a new highlight we should expect another shortcut like <kbd>h</kbd> and <kbd>+</kbd> and noop when only <kbd>h</kbd> is pressed? cc @TomWoodward

**Comment by TomWoodward** [comment]
@PiotrKozlowski when using a mouse the highlight is created as soon as you interact with the floating card, so i think the analogous keyboard behavior _would_ be to create the highlight as soon as `h` is pressed.

**Comment by PiotrKozlowski** [comment]
> After pressing H it is really hard to tell what is focused: highlighted text or highlight edit-card (not sure if it should be considered as a issue).

> @PiotrKozlowski when using a mouse the highlight is created as soon as you interact with the floating card, so i think the analogous keyboard behavior would be to create the highlight as soon as h is pressed.

I've updated the code to focus the textarea when we focus EditCard by default

File: src/app/content/highlights/components/Card.tsx [hunk 11]
```diff
@@ -56,6 +59,22 @@ const Card = (props: CardProps) => {
   const locationFilters = useSelector(selectHighlights.highlightLocationFilters);
   const hasUnsavedHighlight = useSelector(selectHighlights.hasUnsavedHighlight);
 
+  const { isFocused, highlight: { id }, focus, shouldFocusCard } = props;
```
TomWoodward, line 62: so i think `isFocused` is really like.... "highlight or card are focused so the card should be displayed in its 'active' state where it is indented a little or is shown at all on smaller screens"

i think this is rightfully a different thing than your `shouldFocusCard` which indicates moving the actual browser focus into the card, but the nomenclature is confusing to me. maybe we should rename `isFocused` to `isActive` ? or `isSelected` ?

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -97,9 +122,14 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
     if (!position) { return; }
     const topOffset = getOffsetsForHighlight(focusedHighlight).top;
 
+    let transform = '';
```
TomWoodward, line 125: part of me just hates `let` and would prefer this be done in a ternary assignment, but i think thats just personal preference on my part

src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -88,7 +88,7 @@ const DisplayNote = React.forwardRef<HTMLElement, DisplayNoteProps>((
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [element, confirmationRef, confirmingDelete, textToggle, width, isTocOpen, searchQuery]);
 
-  return <div className={className} ref={mergeRefs(ref, element)} data-highlight-card>
+  return <div className={className} ref={mergeRefs(ref, element)} tabIndex={0} data-highlight-card>
```
TomWoodward, line 91: hmmm i'm not sure about allowing users to tab through the cards with the keybard... i'm not sure if that is desirable or if we'd rather force them to tab through the highlights in the content. we might need bey or victoria to decide on this

https://openstax.slack.com/archives/C8V1XET1R/p1617824288341300

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -75,6 +76,19 @@ const EditCard = React.forwardRef<HTMLElement, EditCardProps>((props, ref) => {
 
   useOnEsc(element, props.isFocused, cancelEditing);
 
+  // Overwrite .focus method on element.current to focus textarea when focusing this element.
+  // It is done this way instead of using React.useImperativeHandle because there are other
+  // functions that are being used and not all of them require overrides.
+  React.useEffect(() => {
+    if (element.current) {
+      (element.current as any).focus = () => {
```
TomWoodward, line 84: i don't think we should be monkeypatching the element focus function, this is terrifying to me.

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -97,6 +97,9 @@ const fadeInAnimation = css`
 `;
 
 export const mainCardStyles = css`
+  ${(props: CardProps) => !props.isFocused && (!props.data || !props.data.annotation)
```
TomWoodward, line 100: this is hiding unfocused cards that have no note? why did you have to add this wasn't that the behavior before?

src/app/content/highlights/components/cardUtils.ts
TomWoodward: but no text inputs right? so if its supposed to be super specific for that use case then why are you checking for text inputs?

TomWoodward (changes requested):

src/app/content/highlights/components/CardWrapper.tsx
PiotrKozlowski: np :D

src/app/content/highlights/components/cardStyles.ts
PiotrKozlowski: previously this was done at the component level (`return null`) for this statement
At some point, you asked to remove `defer` from `onFocusOutHighlight` and I found out that this approach was why `defer` was needed in the first place https://github.com/openstax/rex-web/pull/1110#discussion_r586500175
So hiding cards with css and removing `defer` worked for situations when clicking `DisplayNote` components but it didn't help when clicking `EditNote` components.
I've left `defer` in `onFocusOutHighlight` and this css part so we don't rerender cards if we don't have to, but I can roll this back if you want

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: right, that can be confusing, we should either check for all text like inputs or check only for specific ones

jarosik10 (commented):
Everything looks fine to me. I would mark it as approved, but I'm scared that It will auto-merge :)

TomWoodward:
@jarosik10 it only auto merges if the review checkboxes in the issue are all checked, they're not on this one.

either of our pending reviews, or my change request, will block it too

jarosik10 (approved):

src/app/content/highlights/components/CardWrapper.spec.tsx
```diff
@@ -6,11 +7,37 @@ import createMockHighlight from '../../../../test/mocks/highlight';
...
 import CardWrapper from './CardWrapper';
 
+const dispatchKeyDownEvent = (
+  window: Window,
+  element: Document | HTMLElement,
+  key: string,
+  target?: HTMLElement
+) => {
+  const keyboardEvent = window.document.createEvent('KeyboardEvent');
+  keyboardEvent.initKeyboardEvent('keydown', true, true, window, key, 0, '', false, '');
```
TomWoodward, line 23: mdn says this function is deprecated, is there another approach we should use?

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/initKeyboardEvent

edit: nvm this is a spec, i don't care

TomWoodward (approved):
this seems fine, lets get it into acceptance

PR #1112, opened 2021-02-23
make some sw errors warnings in sentry

TomWoodward:
for: openstax/unified#1509

### DIFF
--- a/src/helpers/Sentry.spec.ts
+++ b/src/helpers/Sentry.spec.ts
@@ -3,7 +3,7 @@ import { recordSentryMessage } from '../app/errors/actions';
 import { Store } from '../app/types';
 import config from '../config';
 import createTestStore from '../test/createTestStore';
-import Sentry, { onBeforeSend } from './Sentry';
+import Sentry, { onBeforeSend, Severity } from './Sentry';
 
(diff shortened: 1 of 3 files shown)


PiotrKozlowski (approved):

TomWoodward:
You can actually throw anything in javascript so it's probably better to
leave it as any even if we always use the error type

PiotrKozlowski:
I know, I was thinking of making sure that we throw only what we really want to throw, but I don't think it should be implemented in this PR or implemented at all

TomWoodward:
casting doesn't do that, because if you did like `throw 'asdf'` then your catch is going to get a string, casting it to an error is just going to mask errors. we'd have to make an error constructor that takes `any` and returns errors or converts them if they're not errors similar to how we did for the application errors, but i believe sentry already has that logic in their client so i don't think it makes sense to do it on our end as well

PiotrKozlowski:
I was thinking of something like this:
```ts
try {
  await someAsyncFunction().catch(convertErrorToOurErrorTypeAndThrowItAgain) 
} catch (e) {
  const error = <OurErrorThrownByOurLongNamedFunction>e;
  ...
}
```

PR #1117, opened 2021-03-08
Get codecov check working again

jarosik10:
for: https://github.com/openstax/unified/issues/1511


jarosik10:
It seems to load the coverage file and send it to the codecov repo. I'm testing with:
```
yarn codecov --token="<token>"
```
 `token` is taken from https://app.codecov.io/gh/openstax/rex-web/settings.

@TomWoodward is codecov used in CI or is it just always sent manually?

TomWoodward:
@jarosik10 i would like for it to show up as a PR check automatically

jarosik10:
So it's about adding it as a github action?

jarosik10:
@TomWoodward There is [covecov github action](https://github.com/codecov/codecov-action) avaible. We could make the codedcov job `needs: Test (unit)`.
Do we have the codecov token in repo secrets so I can use it in gh action (like: `${{ secrets.CODECOV_TOKEN }}`)?

TomWoodward:
i don't think its possible to specify a `needs` relationship on a specific instance of a matrix job, i think you can only specify the order. also, `needs` only establishes the order, and will skip the second job if the first one fails, it doesn't actually make the jobs share a filesystem, so you'd have to pass the coverage files as an artifact

you could approach it that way if you want, but i think it might be easier to extend our `test:unit` command, perhaps extract it to a script, make it check for the `CI` environment variable, and if it is set just run `yarn codecov`. according to https://docs.codecov.io/docs/about-the-codecov-bash-uploader we do not have to worry about the token since we're in a public repo

jarosik10:
@TomWoodward 
Test (unit) CI check:
```
> if [[ ${CI} = true ]]; then npm run-script codecov ; fi

sh: 1: [[: not found
```
do you know if i can make it run properly or it just needs to be extracted to a separate script?

TomWoodward:
I think maybe the syntax is not supported by sh, probably cleaner to extract a script anyway

package.json
```diff
@@ -58,12 +58,12 @@
...
     "prerender": "REACT_APP_ENV=${REACT_APP_ENV:-production} node ./script/entry prerender/index",
     "server": "REACT_APP_ENV=${REACT_APP_ENV:-development} node ./script/entry server/cli",
-    "codecov": "codecov",
+    "codecov": "codecov -f ./coverage/lcov.info",
     "coverage-report": "open coverage/index.html",
     "pretest:sims": "./script/build-simulations.bash",
     "pretest": "npm run-script build:css",
     "test": "npm run-script test:unit && npm run-script test:browser && npm run-script test:build",
-    "test:unit": "REACT_APP_ENV=test jest --coverage --testPathPattern=\"(\\.|/)spec\\.tsx?\"",
+    "test:unit": "REACT_APP_ENV=test jest --coverage --testPathPattern=\"(\\.|/)spec\\.tsx?\" && CI=${CI} ./script/push-codecov.bash",
```
TomWoodward, line 66: you don't need the `CI=${CI}` here, the variable is in the environment

script/push-codecov.bash
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -x
+
+if [[ ${CI} = true ]]; then
```
TomWoodward, line 4:
```suggestion
if [[ "${CI}" = "true" ]]; then 
```

i think, i'm not positive `=` is the right operator but if it works its probably fine

TomWoodward (changes requested):

script/push-codecov.bash
jarosik10: I think it should be `[ "$string1" = "string2" ]` or ` [[ "$string1" == "string2" ]]`

jarosik10:
@TomWoodward there is something wrong with commit version. Not sure why.

jarosik10:
@TomWoodward hmm. it [shows](https://codecov.io/gh/openstax/rex-web/commit/9cb05a42981dbdb4c7f48366b14f8058e6500205) the code diff caused by [#d8bb831](https://github.com/openstax/rex-web/pull/1117/commits/d8bb83193cc16f8a6d25185efa825db39cd92f73) properly. But I don't know why the results are not shown as one of the checks (like before).

TomWoodward:
i think the default behavior is the commit statuses (https://docs.codecov.io/docs/commit-status), i think this is the system we used before and we frequently had this problem where the checks would not show up.

maybe we should look into using their integration with the actual checks api here? https://docs.codecov.io/docs/github-checks-beta

jarosik10:
@TomWoodward It seems that adding `codecov.yml` fixed it. Codecov status appeared in the last 4 commits. But do we want to integrate the https://docs.codecov.io/docs/github-checks-beta anyway?

TomWoodward:
no if it works i say leave it

TomWoodward (approved):

PR #1118, opened 2021-03-08
Issue 1498

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1498

### Testing:
Go to `/books/physics/pages/1-introduction301` - you should be redirected to `/books/physics/pages/1-introduction`

Only this URL is set up


TomWoodward (approved):
this is fine the way it is, you can ignore my comment about the browserspec

jarosik10 (approved):
Looks good.

PiotrKozlowski:
`yarn test:unit` and `yarn test:browser` are failing with `Error: ENOSPC: System limit for number of file watchers reached, watch '/__w/rex-web/rex-web/public'` but they are passing locally, hmm

script/prepareRedirects.ts
```diff
@@ -0,0 +1,53 @@
...
+import { RedirectsData } from '../data/redirects/types';
+import { content } from '../src/app/content/routes';
+import { makeUnifiedBookLoader } from '../src/app/content/utils';
+import { findArchiveTreeNodeById } from '../src/app/content/utils/archiveTreeUtils';
+import { AppServices } from '../src/app/types';
+import config from '../src/config.books';
+
+const redirectsPath = path.resolve(__dirname, '../data/redirects/');
+
+const prepareRedirects = async(
```
TomWoodward, line 12: scripts should be executable, this could be moved into /script/util

src/setupProxy.js
```diff
@@ -18,8 +17,37 @@ const {
   REACT_APP_SEARCH_URL,
   REACT_APP_HIGHLIGHTS_URL,
   REACT_APP_OS_WEB_API_URL,
+  REACT_APP_ARCHIVE_URL,
 } = require('./config');
 
+// lots of stuff relies on this
+const JSDOM = require('jsdom').JSDOM;
```
TomWoodward, line 24: all this seems duplicated from entry.js, this is the third babel config in the project. we should find a way to share this with the entry.js config

TomWoodward (changes requested):

PiotrKozlowski:
@TomWoodward I've applied the changes that you've requested but we have a problem with file watchers.
I think it is caused by requiring babel config twice. I've tried increasing the file watchers limit, but it doesn't work: https://github.com/openstax/rex-web/pull/1118/commits/3067341e4c07d8528f22c1ed1e0a110f22903cd1 https://github.com/openstax/rex-web/runs/2096701566

TomWoodward:
did you look at the logs? your change to the watcher limit didn't work because `sudo` couldn't be found, i'm not sure if its necessary. in this context you might also need to wrap that whole command in `bash -c ""`

you could also make the change in the rex dockerfile, i can upload a new version for you

TomWoodward:
i don't think it would be from the babel thing, i don't think that adds any watchers, but idk maybe, thats one of those errors that is hard to track down

src/setupProxy.js
```diff
@@ -166,6 +181,22 @@ function stubEnvironment(app) {
   });
 }
 
+function stubRedirects(app) {
+  app.use(async(req, res, next) => {
+    const {pathname} = url.parse(req.url);
+    const redirects = await prepareRedirects(archiveLoader, osWebLoader);
```
TomWoodward, line 187: you could move this outside the middleware, up in the `stubRedirects` function

src/app/content/hooks/receivePageNotFoundId.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      ...createTestServices(),
+      dispatch: store.dispatch,
+      getState: store.getState,
+    };
+
+    helpers.history.location = {
+      pathname: '/books/physics/pages/1-introduction301',
+    } as any;
+
+    historyReplaceSpy = jest.spyOn(helpers.history, 'replace');
```
TomWoodward, line 33: try throwing a `mockImplementation` on this, i wonder if the history implementation is causing issues for you

TomWoodward (changes requested):

TomWoodward:
i'm seeing the fs watchers stuff on other PRs too, it could be something we merged or an environmental thing that changed with github. other PRs are getting the watcher errors but still passing the check, so you may be able to ignore it

src/app/content/hooks/receivePageNotFoundId.spec.ts
PiotrKozlowski: it didn't help, i'll keep trying

PiotrKozlowski:
Importing `prepareRedirects` and/or `createArchiveLoader` is causing the error
I will continue testing

PiotrKozlowski:
So this error is about something that is inside of `preapreRedirects` and `createArchiveLoader` and it is not about `fs.readdirSync`.

PiotrKozlowski:
Hmm, removing usage of `makeUnifiedBookLoader` seems to help. `makeUnifiedBookLoader` is using `createArchiveLoader` and since importing `createArchiveLoader` also breaks tests I guess that I should start searching in `createArchiveLoader` for something that can cause the error.

PiotrKozlowski:
I've reset all changes related to debugging.
I'm leaving notes for myself for tomorrow.
The error starts to show up only after importing `preapreRedirects` or `createArchiveLoader` in `setupProxy.js`. If I don't import them then tests are passing.
I've found out that in `preapreRedirects` if we remove `makeUnifiedBookLoader` call (and import statement - I didn't check them separately) then importing `preapreRedirects` in `setupProxy.js` is okay.

I think the error is related to something in `ArchiveLoader`, but I don't know what it can be. I also don't know if there is a better way of testing this than pushing every single change to the pr (after changing `ci.yml` file to run only `setup` and `tests` jobs)

PiotrKozlowski:
I don't have more ideas on how to fix CI so I'm leaving it until I come up with something.

TomWoodward:
have you tried using the sysctl option in container.options ?

https://github.community/t/increase-watchers-in-github-workflow-when-running-from-a-docker-image/16691
https://docs.docker.com/engine/reference/commandline/create/#options
https://www.cloudbees.com/blog/adjusting-linux-kernel-parameters-with-docker-compose/

PiotrKozlowski:
It seems like it is not allowed: 
9bc0cf3 https://github.com/openstax/rex-web/runs/2166342368#step:2:55 
From what I'm reading [here](https://github.com/emqx/emqx/issues/1727#issuecomment-432539500) and [here](https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime) we are not allowed to change settings that would change host settings.
As mentioned [here](https://github.com/expo/expo-cli/issues/277#issuecomment-477057752) there is a pretty low limit on GH for the file watchers option.

From my research I've understood that we can't increased this limit so I've spent time debugging why it was showing in the first place. I've left few comments during this time - [summary](https://github.com/openstax/rex-web/pull/1118#issuecomment-799556137)
TLDR. Importing `prepareRedirects` or/and `createArchiveLoader` in `setupProxy.js` (even if we don't use them) is causing this error. If we don't import these files then everything works fine.
cc @TomWoodward

TomWoodward:
according to some docs watch mode should be disabled anyway when the CI environment variable is set

https://github.com/facebook/create-react-app/issues/2336#issuecomment-303482868
https://github.com/facebook/create-react-app/issues/784

can you follow up on that?

PiotrKozlowski:
I think this is about `--watch` option for `jest` and we have a problem with Linux system that is not allowing loading more than 8192 files at one time https://unix.stackexchange.com/questions/444998/how-to-set-and-understand-fs-notify-max-user-watches
I'll can take another look at it when I finish 1518

TomWoodward:
@PiotrKozlowski it seems that CRA enables watch by default for jest unless `CI` is set, we do set it, so the watching should be disabled. that seems like a better avenue to troubleshoot than raising the watcher limit. there is no reason for the watching to be enabled in this case anyway

PiotrKozlowski:
@TomWoodward 
I did some more research.
I think that something is adding files to [Inotify](https://en.wikipedia.org/wiki/Inotify) even if `jest` is run with `CI: true` - it is the same as running it with `--watchAll=false`. My guess is that it is something from `jest` package or from `webpack`

TomWoodward:
have you tried adding `--watchAll=false` to the script execution? maybe the environment variable is getting messed up somehow

PiotrKozlowski:
It didn't help: https://github.com/openstax/rex-web/runs/2175652528
I think it may be just related to how jest is loading all files at the beginning:
```
Determining test suites to run...
 RUNS  src/app/content/components/Page.spec.tsx
 RUNS  src/app/content/highlights/components/Highlights.spec.tsx
 RUNS  src/app/content/components/BookBanner.spec.tsx
 RUNS  src/app/content/hooks/locationChange.spec.ts
 RUNS  src/app/content/studyGuides/components/Filters.spec.tsx
 RUNS  src/app/content/components/Toolbar.spec.tsx
 RUNS  src/app/content/components/NudgeStudyTools/index.spec.tsx
```

It would really help if I could debug this locally somehow.

PiotrKozlowski:
@TomWoodward One more thing I can think of is that the server that we run for tests (custom `server.command` in `jest-puppeteer` preset) has a hot reloading feature enabled even if it is not required for tests to run

TomWoodward:
had you seen this? https://github.com/facebook/create-react-app/issues/8689#issuecomment-602167930

also yes, the `yarn start` command in jest-puppeteer.config.js probably has watching and it would be nice to disable that

PiotrKozlowski:
@TomWoodward Changing from `yarn start` to `yarn server` in `jest-puppeteer.config.js` helped with the file watchers issue but now few browser tests are failing. I'm debugging it.

PiotrKozlowski:
Some notes:
I think that using `yarn server` for browsers tests isn't the best solution so I've made some research to find out how to disabled the file watching in react-scripts.
It is impossible to do this without `eject` in a standard way because react-scripts have hardcoded configs for webpack.
I've found [craco](https://github.com/gsoft-inc/craco) and I'm trying to find the proper option.

TomWoodward:
@PiotrKozlowski you can't just change it from `yarn start` to `yarn server` those are testing two different things, it already does `yarn server` for testing prerendered files as part of `yarn test:build`

you need to find a way to make `yarn start` not do the watching

PiotrKozlowski:
Craco is not working in this case... https://github.com/gsoft-inc/craco/issues/143#issuecomment-806767979
I'll try creating a fork of react-scripts as advised in the documentation and modify it to check for CI variable and disable hot reloading if it is set

TomWoodward:
@PiotrKozlowski this is getting out of control, if necessary we could disable test:browser which i think is the only one using the dev server.

it looks like you must have done something to fix `test:unit`, because that is passing now, that should be good enough

TomWoodward:
I see you added server mode to test:unit and that seems to have helped, but that confuses me because the unit tests shouldn't be using the puppeteer environment. If the unit tests are running the puppeteer environment that is a problem

PiotrKozlowski:
Idk why this is happening, but I think I've just found a solution. I'll push a commit in a sec

PiotrKozlowski:
Oh no :( I really thought this was gonna work: https://github.com/openstax/rex-web/pull/1118/commits/7d8fa6f1c0ab359a58c2011336596153567ea3a2#diff-9531c01f38fb3a33c0a25db522822e353b143e69b9d43246d8e32bcb66662bfdR1 

One more thing I can think of is this comment: https://github.com/webpack/webpack-dev-server/issues/1744#issuecomment-477341127

I didn't experience this locally but maybe this is what is happening in CI? I think we are running unit, browser and build tests simultaneously?

TomWoodward:
maybe try turning the actual hot param off in your config override? https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpackDevServer.config.js#L73

PiotrKozlowski:
@TomWoodward This is not working: https://github.com/gsoft-inc/craco/issues/143#issuecomment-806767979

I've tested this locally on all different ways and even by modifying node_modules package - I didn't found a way to disable hot reloading for webpack-dev-server

TomWoodward (approved):

PR #1120, opened 2021-03-10
Implement the glow around focused colors in highlight popup

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1494


src/app/content/highlights/components/ColorIndicator.tsx
```diff
@@ -20,21 +20,44 @@ const CheckIcon = styled(Check)`
   display: none;
 `;
 
+interface FocusedStyleProps extends Pick<Props, 'style' | 'size' | 'shape'> {}
```
jarosik10, line 23: Making
```
interface FocusedStyleProps {
  style: typeof highlightStyles[number];
  size?: 'small'
  shape?: 'square' | 'circle';
}
```
and then `Props` to extend it feels more clear, but I think it's debatable and it doesn't really matter. :P

jarosik10: Btw. should `size` and `shape` be optional?

PiotrKozlowski: I've did what you've suggested but I've changed the name of the interface to `StyleProps` so it makes more sense he they are separated from the `Props`

PiotrKozlowski: > Btw. should size and shape be optional?

There is support for `undefined` values and this PR is not really doing anything about these values so I think yes, they should stay optional

jarosik10 (approved):

PR #1132, opened 2021-03-16
Console log the PQ modal window.confirm result on exit

jarosik10:
for: https://github.com/openstax/unified/issues/1412

### DIFF
--- a/src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
+++ b/src/app/content/practiceQuestions/components/PracticeQuestionsPopup.tsx
@@ -28,7 +28,10 @@ const PracticeQuestionsPopup = () => {
   const closeAndTrack = React.useCallback((method: string) => () => {
     if (currentQuestionIndex !== null) {
       const message = intl.formatMessage({ id: 'i18n:practice-questions:popup:warning-before-close' });
-      if (!assertWindow().confirm(message)) { return; }
+      const confirmExit = assertWindow().confirm(message);
+      // tslint:disable-next-line: no-console
+      console.log('Exit confirmed: ', confirmExit);
+      if (!confirmExit) { return; }
     }
 
     dispatch(push(assertDefined(match, 'match should be always defined at this step')));


jarosik10:
I've asked Lena to check it and:
- submitting `Cancel` with space-bar works
- submitting `Cancel` with enter doesn't work

@TomWoodward @PiotrKozlowski

PiotrKozlowski:
What do you mean by "it doesn't work"?

jarosik10:
**works** => submitting `Cancel` button in `window.confirm` doesn't close PQ modal
**doesnt' work** =>submitting `Cancel` button in `window.confirm` closes PQ modal

PiotrKozlowski:
It seems like it may be related to how it is implemented in Safari. We are using standard web API (window.confirm) and as far I know we can't control events in this popup and even if we could I don't think we should.

A possible solution to standardized it would be to create a custom popup (similar to what we are using for warning users before closing unsaved note for highlights)

@TomWoodward what do you think?

jarosik10:
Okay. I've done some testing with Lena and it seems to be a default Mac dialog navigation [behavior](https://superuser.com/questions/24159/osx-keyboard-shortcuts-in-dialogs).

`Enter` -> selects default button `OK` (has different color - depends on browser)
`Esc` -> selects `Cancel`
`Spacebar` -> selects chosen button (selected with `tab`)

Safari:
Screenshot from 2021-03-17 14-51-58

Chrome:
Screenshot from 2021-03-17 14-53-59

Firefox:
Screenshot from 2021-03-17 14-54-30

TomWoodward:
is this just a matter of different browsers selecting different options by default? does it work consistently if you tab over the the consistent option? 

@jarosik10 @lenka222

lenka222:
All browsers behave the same on Mac. Tabbing over the options doesn't change behavior.

TomWoodward:
if you select with spacebar it works correctly, pressing enter apparently always does the same thing regardless of selection

Bkblodget:
Closing this as not a bug - this is a functionality of using a Mac/not a REX related fix.

PR #1133, opened 2021-03-17
Issue 1518 - add support for new web pipeline with archive code versions

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1518


src/app/content/utils.ts
```diff
@@ -20,37 +21,35 @@ export { stripIdVersion } from './utils/idUtils';
 export { scrollSidebarSectionIntoView } from './utils/domUtils';
 
 export interface ContentPageRefencesType {
-  bookId?: string;
-  bookVersion?: string;
+  bookId: string;
+  bookVersion: string;
```
TomWoodward, line 25: book version is still optional in this structure because it is not in the cross-book links

src/app/content/utils.ts
```diff
@@ -20,37 +21,35 @@ export { stripIdVersion } from './utils/idUtils';
...
     .map((match) => {
       const [bookMatch, pageMatch] = match.split(':');
-      const pageId = pageMatch.substr(0, 36);
-      const [bookId, bookVersion] = bookMatch.split('@');
+      const pageId = pageMatch.split('.xhtml')[0];
+      let [bookId, bookVersion] = bookMatch.split('@') as [string, string | undefined];
+      bookId = bookId.substr(2);
+
+      if (!bookVersion && BOOKS[bookId]) {
+        bookVersion = BOOKS[bookId].defaultVersion;
```
TomWoodward, line 39: can we leave this logic in `resolveContent.ts`? i liked it better there

src/test/mocks/archiveLoader.ts
```diff
@@ -1,30 +1,31 @@
...
 import { ArchiveBook, ArchivePage } from '../../app/content/types';
 
 export const book = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, '../fixtures/contents/testbook1-shortid.json'), 'utf8')
+  fs.readFileSync(path.resolve(__dirname, '../fixtures/apps/archive/20210224.204120/contents/testbook1-shortid.json'), 'utf8')
 ) as ArchiveBook;
 
 export const page = JSON.parse(
-  fs.readFileSync(path.resolve(__dirname, '../fixtures/contents/testbook1-shortid:testpage1-shortid.json'), 'utf8')
+  fs.readFileSync(path.resolve(__dirname, '../fixtures/apps/archive/20210224.204120/contents/testbook1-shortid:testpage1-shortid.json'), 'utf8')
```
TomWoodward, line 12: you might want to use a more placeholdery name for the test archive url, like `apps/archive/codeversion/contents` or something, otherwise this version is going to be saved for the rest of time in the tests or we'll have to update it when we update the prod one even if we don't really have to

TomWoodward (changes requested):
also we need the other change about detecting page renames on archive backend change before we can merge this because there are a few that we know are in the first rap version we're adopting

PiotrKozlowski:
I've pushed requested changes. Do you want to review now or after #1519 is finished?

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -31,6 +31,29 @@ export default async(
   return {book, page};
 };
 
+export function getContentPageReferences(htmlContent: string) {
```
TomWoodward, line 34: i didn't mean to move this whole function, just the `defaultVersion` part that you added to it

TomWoodward (changes requested):
this logic is fine, there is that one comment we're still going back and forth on, but after the redirect script is done this change will trigger some redirect data changes that will be added as part of this PR so i'll wait to do final approval until then

jarosik10 (approved):
Couldn't find anything that should be change. Looks good to me.

concourse/update-content/script.bash
```diff
@@ -49,7 +49,7 @@ for book_id in $book_ids; do
   git commit -m "update content" || true
   git push --set-upstream origin "$branch"
 
-  book_title=$(node script/entry.js book-info "$book_id" --field title)
+  book_title=$(node script/entry.js book-info "$book_id" "$desired_version" --field title)
```
TomWoodward, line 52: cant `book-info` just use the books config?

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -96,16 +96,12 @@ export const resolveBookReference = async(
     throw new BookNotFoundError(`Could not resolve uuid for slug: ${bookSlug}`);
   }
 
-  const bookVersion = 'version' in match.params.book
-    ? match.params.book.version === 'latest'
```
TomWoodward, line 100: removing this breaks specifying a book@version in the url that is not in the config right? we need that

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -148,50 +144,31 @@ const resolvePage = async(
...
-  const configuredReference = allReferences.filter((item) => BOOKS[item.id])[0];
+  const { bookId, bookVersion } = getInputReferenceInfo(reference.bookId, reference.bookVersion);
 
-  if (configuredReference) {
-    const osWebBook =  await services.osWebLoader.getBookFromId(configuredReference.id);
-    const archiveBook = await services.archiveLoader.book(
-      configuredReference.id, BOOKS[configuredReference.id].defaultVersion
-    ).load().catch((error) => {
+  const osWebBook =  await services.osWebLoader.getBookFromId(bookId);
+  const archiveBook = await services.archiveLoader.book(bookId, bookVersion || 'undefined')
```
TomWoodward, line 160: passing "undefined" here will always fail, you could probably short circuit earlier

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent-in-development.spec.ts
```diff
@@ -151,57 +153,12 @@ describe('locationChange', () => {
       expect(getBookSlugFromId).toHaveBeenCalledTimes(1);
     });
 
-    it('uses param version if there is one', async() => {
```
TomWoodward, line 154: we need this for that use case i asked you to put back in right?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1134, opened 2021-03-26
Issue 1519 - add update-archive-version script to handle page url changes after archive code version updates

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1519

Testing:
- change value of `REACT_APP_ARCHIVE_URL` in `config.archive-url.json` to an empty string `""`
- remove all `.json` files from `/data/redirects/*`
- run `node ./script/entry.js update-archive-version --newArchiveUrl /apps/archive/20210224.204120`
- expected output should be similar to this:
```
Preparing books...
Updating redirects [===================                                         ] 19/60 (49.4s ETA) | Finished processing Principles of Accounting, Volume 1: Financial Accounting | 9ab4ba6d-1e48-486d-a2de-38ae16
Updating redirects [=========================                                   ] 25/60 (46.5s ETA) | Finished processing Principles of Accounting, Volume 2: Managerial Accounting | 920d1c8a-606c-4888-bfd4-d1ee2
Updating redirects [============================================================] 60/60 (0.0s ETA) | Finished processing Principles of Economics | 69619d2b-68f0-44b0-b074-a9b2bf90b2c6


Added 14 redirects for book Business Law I Essentials | 464a3fba-68c1-426a-99f9-597e739dc911
Added 11 redirects for book Business Ethics | 914ac66e-e1ec-486d-8a9c-97b0f7a99774
Added 1 redirects for book Introduction to Sociology 2e | 02040312-72c8-441e-a685-20e9333f3e1d
Added 1 redirects for book Chemistry: Atoms First | 4539ae23-1ccc-421e-9b25-843acbb6c4b0
Added 1 redirects for book Principles of Microeconomics | ea2f225e-6063-41ca-bcd8-36482e15ef65
Added 1 redirects for book Principles of Macroeconomics | 4061c832-098e-4b3c-a1d9-7eb593a2cb31
Updating config.archive-url.json with /apps/archive/20210224.204120
```
- run `git status`
- expected: no changes to commit


script/utils/update-redirects-data.ts
```diff
@@ -0,0 +1,52 @@
...
+  });
+
+  const matchRedirect = (section: LinkedArchiveTreeNode) => isEqual(formatSection(section));
+
+  let countNewRedirections = 0;
+  for (const section of flatCurrentTree) {
+    const { slug } = findArchiveTreeNodeById(newBook.tree, section.id) || {};
+    // findArchiveTreeNodeById will cache book structure and we don't want this
+    // because we are comparing the same book with different version which may have different content
+    CACHED_FLATTENED_TREES.clear();
```
PiotrKozlowski, line 36: This part wasn't originally in the script. I've added it today.
I think that if we ever used `update-content-versions-and-check-for-archived-slugs.ts` before then we weren't searching for missing pages properly

script/prerender/prerender.tsx
```diff
@@ -73,7 +74,10 @@ async function render() {
 
   await renderSitemapIndex();
   await renderManifest();
-  await createRedirects(archiveLoader, osWebLoader);
+
+  if (APP_ENV === 'production') {
+    await createRedirects(archiveLoader, osWebLoader);
+  }
```
PiotrKozlowski, line 80: Since I've added bunch of files to `/data/redirects/` this script is trying now create redirects for them and it fails because we use fixtures files for archive and os web loaders

We can either do the `if` like I did or we can change `books` variable in `createRedirects` function to an empty array in test env. Another option would be to create mock file in `data/redirects/` and change this `books` variable to `['testbook1-uuid.json']`

I don't think it really matters

PiotrKozlowski: Actually, it may be good to make sure that `createRedirects` is run properly in CI. I'll make sure it is

PiotrKozlowski: resolved with c25b347

script/utils/update-redirects-data.ts
TomWoodward: really it would be more correct to add the version to the cache key

TomWoodward (changes requested):

script/utils/update-redirects-data.ts
PiotrKozlowski: Function which does the caching is that one:
```ts
function flattenArchiveTree(tree: LinkedArchiveTree): Array<LinkedArchiveTree | LinkedArchiveTreeSection> 
```
There are two things that we would have to add:
1. book version - we don't have access to this in `tree: LinkedArchiveTree`
2. archive URL - we have access only to `REACT_APP_ARCHIVE_URL` which is not enough when updating the archive URL because we need both at the same time

There are multiple functions that depend on `flattenArchiveTree` so all of them would have to accept additional `bookVersion` and `archiveUrl`. I've tried that and it is a mess, even with default parameters :D

Another solution would be to wrap all archive utils in a class with a constructor that accept these parameters, but this will affect tens of files since we will have to create archive utils instance for all of them.

Do you prefer any of these solutions?

PiotrKozlowski: `flattenArchiveTree` is called recursively for every chapter - `tree` is a little bit confusing name because we have `book.tree` but in this case, it corresponds to "book tree" or "chapter"
`version` is only in the `id` itself but since it is called for chapters then it is a version of a section, not book

One more thing which we could do is change how we decide to cache tree - currently we are doing:
```
 if (process.env.NODE_ENV !== 'test') {
    CACHED_FLATTENED_TREES.set(tree.id, flattened);
 }
 ```
 
 I think I'll just change it to check if `APP_ENV` from `config` is set to `development` or `production`

**Comment by TomWoodward** [comment]
why is it generating empty files? can we make it not?

**Review by TomWoodward — CHANGES_REQUESTED** [review]

File: src/config.development.js [hunk 3]
```diff
@@ -21,6 +21,7 @@ const UNLIMITED_CONTENT = UNLIMITED_CONTENT_QUERY === undefined
 ;
 
 module.exports = {
+  APP_ENV: 'development',
```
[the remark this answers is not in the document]
PiotrKozlowski: we are loading `config.development` in `else` statement, before that we are checking for `production` and `test`
```js
if (process.env.REACT_APP_ENV === 'production') {
  Object.assign(config, require('./config.production'));
} else if (process.env.REACT_APP_ENV === 'test') {
  Object.assign(config, require('./config.test'));
} else {
  Object.assign(config, require('./config.development'));
}
```

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -40,7 +41,7 @@ export function flattenArchiveTree(tree: LinkedArchiveTree): Array<LinkedArchive
   }));
   // Cache is disabled for testing
   /* istanbul ignore next */
-  if (process.env.NODE_ENV !== 'test') {
+  if (APP_ENV === 'production' || APP_ENV === 'development') {
```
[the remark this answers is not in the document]
PiotrKozlowski: previously it was "cache always if it isn't `test` env" which was also true for `undefined` (for example when running scripts) and now it is specified to cache for `development` and `production` only

src/config.development.js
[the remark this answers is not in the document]
PiotrKozlowski: ugh, since we can start scripts without specifying env then it will default to `development` which is wrong because it will then cache trees..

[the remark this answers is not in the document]
PiotrKozlowski: I've removed `APP_ENV: 'development'` from the config and set `REACT_APP_ENV=development` in `yarn start` script
so now `APP_ENV` will be `undefined` when running scripts (as it was before this change) and it will be set to `development` when running with `yarn start`

script/utils/update-redirects-data.ts
TomWoodward: so it IS caching based on the version right? because of the version in the id parameter? so the only issue with the caching is if you're running multiple archive connections right?

PiotrKozlowski: If version of a section/chapter is always updated when there is a slug change then I think the version shouldn't be a problem

Initially caching was a problem for unit tests. If I remember correctly it was because we were modifying book tree in order to test something

Caching is generally a problem when we can get two sections with the same id and the same version that are expected to have different properties (like when fetching from different archive version)

TomWoodward: when a book version is updated all of the page versions should be updated too. i think in testing the tree is modified for different tests and that creates issues, and yes when you have an archive upgrade the version isn't changed but the tree might be.

the changes that you're making to the environmental handling are confusing to me, i think having complicated environmentally aware logic is not ideal, and introducing an `undefined` environment for scripting and relying on that for functionality seems unclear to me.

i think it might be more clear if we change the caching line to be 

```diff
-export const CACHED_FLATTENED_TREES = new Map<string, Array<LinkedArchiveTree | LinkedArchiveTreeSection>>();  
+const CACHED_FLATTENED_TREES = new Map<string, Array<LinkedArchiveTree | LinkedArchiveTreeSection>>();  
+let cacheArchiveTrees=true
+export const disableArchiveTreeCaching = () => cacheArchiveTrees = false;
```

then
```diff
-if (process.env.NODE_ENV !== 'test') {
-  CACHED_FLATTENED_TREES.set(tree.id, flattened);
-}
+if (cacheArchiveTrees) {
+  CACHED_FLATTENED_TREES.set(tree.id, flattened);
+}
```

and call that from the beginning of your script and `/test/setup.ts`

PiotrKozlowski: this makes a lot of sense, thanks

PiotrKozlowski: I've pushed a change. `CACHED_FLATTENED_TREES` is still exported so we can clear it after updating titles in the tree in `content/utils.ts`

data/redirects/02040312-72c8-441e-a685-20e9333f3e1d.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "bookId": "02040312-72c8-441e-a685-20e9333f3e1d",
+    "pageId": "325e4afd-80b6-44dd-87b6-35aff4f40eac",
+    "pathname": "/books/introduction-sociology-2e/pages/preface"
```
TomWoodward, line 5: this is a mistake on the content side, we're going to have to also do a book version update to get the new content so this rename doesn't happen

script/entry.js
```diff
@@ -8,6 +8,8 @@ const requireBabelConfig = require('../src/babel-config');
 
 requireBabelConfig(extensions);
 
+const { disableArchiveTreeCaching } = require('../src/app/content/utils/archiveTreeUtils');
```
TomWoodward, line 11: this will make pre-rendering very slow, i think we should only call it on the script you have that is dealing with the two archive versions

script/utils/prepareRedirects.ts
```diff
@@ -5,25 +5,29 @@ import { content } from '../../src/app/content/routes';
...
-import config from '../../src/config.books';
+import { APP_ENV, BOOKS } from '../../src/config';
 
-const redirectsPath = path.resolve(__dirname, '../../data/redirects/');
+const redirectsDataFolderPath = path.resolve(__dirname, '../../data/redirects/');
+
+const redirectsDataFiles = APP_ENV === 'test'
+  ? [path.resolve(__dirname, '../../src/mock-redirects.json')]
+  : fs.readdirSync(redirectsDataFolderPath)
+    .filter((name) => name.match('.json'))
```
TomWoodward, line 15: can you indent this and the map one more time because of the ternary?

src/test/setup.ts
```diff
@@ -81,6 +82,17 @@ let scrollBy: jest.SpyInstance;
 let mockGa: any;
 
 resetModules();
+
+beforeAll(async() => {
+  // Add DOMParser if it doesn't exists since it is require by archiveTreeUtils
+  if (!(global as any).DOMParser) {
+    (global as any).DOMParser = new JSDOM().window.DOMParser;
```
TomWoodward, line 89: doesn't jest add this already?

src/test/setup.ts
```diff
@@ -81,6 +82,17 @@ let scrollBy: jest.SpyInstance;
...
+  }
+  // import has to be here because we are using resetModules()
+  const { disableArchiveTreeCaching } = await import('../app/content/utils/archiveTreeUtils');
+  disableArchiveTreeCaching();
```
TomWoodward, line 93: could this be done in the `resetModules` function instead?

TomWoodward (changes requested):

data/redirects/02040312-72c8-441e-a685-20e9333f3e1d.json
PiotrKozlowski: I should just re-run the script after https://github.com/openstax/rex-web/pull/1143 is merged, right?

src/test/setup.ts
PiotrKozlowski: it started to fail for `test:browser` with `ReferenceError: DOMParser is not defined` after adding `disableArchiveTreeCaching`. I guess this is because of importing the `archiveTreeUtils` after resetting modules
I'll try to find another workaround

TomWoodward:
Does the puppeteer preset cause jest to not load jsdom with it's default
behavior?

PiotrKozlowski:
> Does the puppeteer preset cause jest to not load jsdom with it's default behavior?

So it was failing at test:browser and since on CI it is no longer used it was failing at `test:prerender:specs` which is part of `test:build`
If I remove `--preset jest-puppeteer` then it will fail with: 
```
wsEndpoint not found

      at PuppeteerEnvironment.setup (node_modules/jest-environment-puppeteer/lib/PuppeteerEnvironment.js:45:13)
 ```
 
 It works normally after removing these lines from `resetModules()`
 ```js
const { disableArchiveTreeCaching } = await import('../app/content/utils/archiveTreeUtils');
disableArchiveTreeCaching();
  ```

**Comment by TomWoodward** [comment]
well yea it needs the puppeteer preset, i'm just wondering if we were not depending on jsdom at all for any of the puppeteer tests before

after some research, i think the problem is the environment. https://jestjs.io/docs/configuration#testenvironment-string jsdom is the default test environment, and we are overriding that with the puppeteer environment that does the browser setup for each test

so you can either try to find a new way to use both environments, or to follow up on your previous approach you may need a `beforeEach` instead of `beforeAll` because i think that is more in line with how the jest test environment would have worked

**Comment by PiotrKozlowski** [comment]
What do you think about this approach: https://github.com/openstax/rex-web/blob/4edf2568ab6425e6c489fd9b034960847076b2be/src/app/content/utils/archiveTreeUtils.ts#L17-L24 ?

`initializeDOMParser` is used inside specific functions that need it

I think it is working because without `jsdom` test environment it wasn't loading but since the DOMParser is now created only when calling the function then it is called inside puppeteer which just has normal DOMParser

**Comment by TomWoodward** [comment]
that fix is working for you because you're never trying to actually use the utils on that test run, you're just including the file and calling the function to not cache. all of the actual execution during puppeteer tests is in chrome, which wouldn't be affected by any setup you do outside chrome

i think the call to disable caching is irrelevant for the puppeteer tests, because the setup.ts file is the same you're running into the issue where it tries to run it anyway even though it doesn't have to

File: data/redirects/02040312-72c8-441e-a685-20e9333f3e1d.json (hunk 5 above)
Comment (PiotrKozlowski) [reply data/redirects/02040312-72c8-441e-a685-20e9333f3e1d.json:5]: this pr is not merged yet, but I've checked that running a script again after this book is updated to `15.7` will result in not creating this file

Comment (PiotrKozlowski) [reply data/redirects/02040312-72c8-441e-a685-20e9333f3e1d.json]: pr is merged, I've re-runed the script and file wasn't created

File: script/update-archive-version.ts [hunk 10]
```diff
@@ -0,0 +1,102 @@
...
+        updatingRedirectsBar.tick({
+          msg: `Finished processing ${book.title} | ${book.id}`,
+        });
+        if (redirects > 0) {
+          newRedirects.push([book, redirects]);
+        }
+      })
+      .catch((error) => {
+        updatingRedirectsBar.interrupt(`Error while loading: ${bookId} | ${error}`);
+      });
```
PiotrKozlowski, line 77: Should I catch this error or just throw it so the CI fails in these cases?

src/test/setup-for-puppeteer.ts
```diff
@@ -0,0 +1,9 @@
+if (process.env.CI) {
+  // set default timeout to something quite large in CI
+  jest.setTimeout(90 * 1000);
+} else {
+  jest.setTimeout(120 * 1000);
+}
+
+// export to prevent isolatedModules error
+export {};
```
jarosik10, line 9: What is the purpose of this file?

script/update-archive-version.ts
TomWoodward: i think we want CI to fail if any books fail to load

src/test/setup-for-puppeteer.ts
TomWoodward: i think the idea is to separate the puppeteer setup from `src/test/setup.ts` because some of the stuff in there is unnecessary for puppeteer and potentially causing issues or at least confusing

TomWoodward (changes requested):
looks good, i just commented on re-throwing that error

TomWoodward (approved):

jarosik10 (approved):

PR #1139, opened 2021-04-01
Issue 1525 - Handle new ABL schema

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1525


script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,160 @@
...
+      // books from collections have versions starting with 1.
+      return isCollection ? versionData.content_version.slice(2) : versionData.content_version;
+    })
+    // sort and revert so results are descending
+    .sort()
+    .reverse();
+
+  // in general we want the most recent one, in the future we'll have to filter this by which ones
+  // are available for the recipe code version REX supports
+  return versions[0];
```
PiotrKozlowski, line 111: This is something that we could do after emerging #1133 and #1134 , right?

script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+const { data } = argv as { data?: string };
```
jarosik10, line 121: I think the script arguments should be read inside the main `transformData` function or maybe make the `transformData` function to receive the `data` as argument. 

Doesn't change anything but seems clearer to me ;)

jarosik10 (commented):
The typing is very precise, but it seems to be a little overkill for me.
It's just script for a data manipulation, maybe we should simplify the typing a little bit.
**I'm not sure about it - it's just my thought**

script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,168 @@
...
+      if (!configuredBooks[book.uuid] || configuredBooks[book.uuid].defaultVersion !== desiredVersion) {
+        results[book.uuid] = desiredVersion;
+      }
+    }
+  }
+
+  // this script is used in a update-content/script.bash
+  // so we write results to the terminal to assign them to a variable
+  // tslint:disable-next-line: no-console
+  console.log(JSON.stringify(results));
```
TomWoodward, line 164: `console.log` includes formatting, you're better off calling `process.stdout.write` directly

script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,168 @@
...
+    // books from collections have versions starting with 1.
+    return isCollection ? versionData.content_version.slice(2) : versionData.content_version;
+  };
+
+  const versions = approvedVersions
+    // only versions for current repo / collection
+    .filter(filter)
+    .map(transformVersion)
+    // sort and revert so results are descending
+    .sort()
```
TomWoodward, line 113: naive sort doesn't work on semver

script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,168 @@
...
+};
+
+const getDesiredVersion = (
+  approvedVersions: Array<ApprovedVersionCollection | ApprovedVersionRepo>,
+  colOrRepo: ApprovedRepo | ApprovedCollection
+): string | undefined => {
+  const isCollection = isApprovedCollection(colOrRepo);
+
+  const filter = isCollection
+    ? matchCollectionVersion(colOrRepo as ApprovedCollection)
```
TomWoodward, line 100: you have a guard on line 97, you shouldn't have to cast (i mean you do here, but you should be able to restructure and not)

TomWoodward (changes requested):
you should also add the thing about only reading versions that are available in the archive code version that we're using

script/transform-approved-books-data.ts
PiotrKozlowski: thats true, however at this step they are already transformed - we are removing first 2 characters so it should be `X.X` for collections and `X` for repos and this should work with `.sort()`

PiotrKozlowski:
> The typing is very precise, but it seems to be a little overkill for me.
It's just script for a data manipulation, maybe we should simplify the typing a little bit.
I'm not sure about it - it's just my thought

You have a point - we probably could create a single interface and assume that someone who writes the code will know what should be passed to the script, however, writing those interfaces and guards took me like ~15 minutes and we don't have to assume anything anymore :D

script/transform-approved-books-data.ts
PiotrKozlowski: @TomWoodward I've added the archive code version to the filter but I'm not sure if there is still a possibility that we will get more than one version matching the archive code version?

PiotrKozlowski:
@TomWoodward 
> you should also add the thing about only reading versions that are available in the archive code version that we're using

done, but please take a look at the comment: https://github.com/openstax/rex-web/pull/1139#discussion_r608608270

jarosik10 (approved):

script/transform-approved-books-data.ts
TomWoodward: there could be more than one version, we want the biggest version that matches our criteria

script/transform-approved-books-data.ts
TomWoodward: yea but like `4.56` is bigger than `4.9` and string or number sort will both get that wrong. you need a semver specific sort

PiotrKozlowski: right... that's obvious but I didn't know that :grimacing: 
fixed

concourse/update-content/script.bash
```diff
@@ -16,30 +16,25 @@ yarn
 approved_books_default_branch=$(curl -s https://api.github.com/repos/openstax/content-manager-approved-books | jq -r .default_branch)
 rex_default_branch=$(curl -s https://api.github.com/repos/openstax/rex-web | jq -r .default_branch)
 
-approved_books=$(curl -sL "https://github.com/openstax/content-manager-approved-books/raw/$approved_books_default_branch/approved-books.json" | jq 'map(select(.tutor_only == false)) | map(select(.server == "cnx.org"))')
-book_ids=$(jq -r 'map(.uuid) | unique | .[]' <<< "$approved_books")
+data=$(curl -sL "https://github.com/openstax/content-manager-approved-books/raw/$approved_books_default_branch/approved-book-list.json")
+
+# script will return a JSON object with book ids and new versions only for books that doesn't mach current config
+book_ids_and_versions=$(node script/entry.js transform-approved-books-data --data "$data")
+book_ids=$(echo "$book_ids_and_versions" | jq -r 'keys | .[]')
```
TomWoodward, line 23: if you use a -c flag you could use `entries` or something here and iterate over each json object, rather than just the key and having to look it up in the big response each time

script/transform-approved-books-data.ts
```diff
@@ -0,0 +1,178 @@
...
+  return isApprovedVersionCollection(versionData) && versionData.collection_id === collectionData.collection_id;
+};
+
+const getDesiredVersion = (
+  approvedVersions: Array<ApprovedVersionCollection | ApprovedVersionRepo>,
+  colOrRepo: ApprovedRepo | ApprovedCollection
+): string | undefined => {
+  // Example archive code version: 20210224.204120
+  const archiveVersionMatch = REACT_APP_ARCHIVE_URL.match(/[0-9]+\.[0-9]+/);
+  const archiveVersion = archiveVersionMatch ? archiveVersionMatch[0] : undefined;
```
TomWoodward, line 103: we should probably stop supporting this being undefined now. also, if it makes life easier we can have the archive version in a separate variable, but that would also require modifying the archive update script so if you want to just add an `assertDefined` here thats fine too

TomWoodward (changes requested):

TomWoodward (approved):

PR #1141, opened 2021-04-02
Pass Sentry.Severity to Sentry.captureException

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1527

### DIFF
--- a/src/app/content/components/utils/attachHighlight.ts
+++ b/src/app/content/components/utils/attachHighlight.ts
@@ -6,18 +6,18 @@ const attachHighlight = <T extends Highlight | SerializedHighlight>(
   highlighter: Highlighter,
   errorMsg?: (highlight: T) => string
 ) => {
-  highlighter.highlight(highlight);
-  const result = highlighter.getHighlight(highlight.id);
-  if (!result || !result.isAttached()) {
-    Sentry.captureException(
-      new Error(errorMsg
+  try {
+    highlighter.highlight(highlight);
+    const result = highlighter.getHighlight(highlight.id);
+    if (!result || !result.isAttached()) {
+      throw new Error(errorMsg
         ?  errorMsg(highlight)
-        : `Highlight with id: ${highlight.id} has not been attached.`),
-      Severity.Warning
-    );
-    return;
+        : `Highlight with id: ${highlight.id} has not been attached.`);
+    }
+    return result;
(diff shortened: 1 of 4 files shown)


TomWoodward:
@PiotrKozlowski can you also change this logic https://github.com/openstax/rex-web/blob/master/src/app/content/components/utils/attachHighlight.ts#L13 to just hit sentry instead of throwing the error and immediately catching it and passing to sentry? (remove the try/catch)?

PiotrKozlowski:
@TomWoodward I think it is in a `try catch` because `highlighter.highlight()` may throw as well: https://github.com/openstax/rex-web/blob/master/src/app/content/components/utils/attachHighlight.ts#L10

we have some tests for this: https://github.com/openstax/rex-web/blame/master/src/app/content/search/utils.spec.ts#L232
added in https://github.com/openstax/rex-web/pull/465

Is that a thing? I can research that next week

TomWoodward:
> @TomWoodward I think it is in a `try catch` because `highlighter.highlight()` may throw as well: https://github.com/openstax/rex-web/blob/master/src/app/content/components/utils/attachHighlight.ts#L10
> 
> we have some tests for this: https://github.com/openstax/rex-web/blame/master/src/app/content/search/utils.spec.ts#L232
> added in #465
> 
> Is that a thing? I can research that next week

i think you're right, that the try/catch is there to catch issues from the highlighter, then we added the throw for the 'cant be mounted' error and it would have been silly to do that outside the try/catch and have another sentry call, but then i decided the 'can't be mounted' error should be a warning and i updated the sentry call in the catch to be a warning but that is wrong because now any error from the highlighter will be a warning 😬 big yikes

i think the try/catch needs to stay but the level needs to be removed from the sentry call in the catch and the throw for the 'cant be mounted' needs to be done elsewhere (without the throw, just passing an error into sentry with the warning level)

i think this is all messed up because of my previous change, but if you could fix it as part of this PR i'd appreciate it @PiotrKozlowski

PiotrKozlowski:
@TomWoodward I've extracted the `if` check outside of the `try catch` so now it is more obvious that it is there for highlighter errors

jarosik10 (approved):
Everything looks fine ;) . But it'd be good to restart the checks - `Selenium` failed.

PR #1144, opened 2021-04-07
 issue-1499

jarosik10:
for: https://github.com/openstax/unified/issues/1499

**Testing:**
Go to `config.books.json` and delete two books: `13ac107a-f15f-49d2-97e8-60ab2e3b519c`, `9d8df601-4f12-4ac1-8224-b450bf739e5f`.

`13ac107a-f15f-49d2-97e8-60ab2e3b519c` has a backup in `release.development.json`.

`13ac107a-f15f-49d2-97e8-60ab2e3b519c` stands for Algebra and Trigonometry `../books/algebra-and-trigonometry/pages/1-introduction-to-prerequisites`
`9d8df601-4f12-4ac1-8224-b450bf739e5f` stands for American Government 2e `../books/american-government-2e/pages/1-introduction`


jarosik10:
@TomWoodward do we want to add the book config gateway as a `AppService`?

TomWoodward:
yea you'll need to configure the base url and pass it through like the other gateways

jarosik10:
From where can I retrieve the base url for `.../rex/release.json`?

TomWoodward:
you can load it relatively in the browser, but for example it can be found here https://openstax.org/rex/release.json

jarosik10:
@TomWoodward 
How can it be tested? because `rex/release.json` can't be accessed on `localhost` or `herokuapp`.

TomWoodward:
you should set it up so that if it fails to load release.json it fails silently and doesn't cause an error to be displayed

to test that it actually works you can add a test file similar to the way that environment.development.json works

jarosik10:
@TomWoodward could you generally review the logic? And do we want to completely remove importing BOOKS config directly and replace it with the gateway?

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -163,15 +168,20 @@ export const getBookInformation = async(
...
-  const configuredReference = allReferences.filter((item) => BOOKS[item.id])[0];
+  const configuredReference = allReferences.filter(async(item) =>
+    await services.bookConfigLoader.getBookVersionFromUUID(item.id))[0];
 
   if (configuredReference) {
     const osWebBook =  await services.osWebLoader.getBookFromId(configuredReference.id);
     const archiveBook = await services.archiveLoader.book(
-      configuredReference.id, BOOKS[configuredReference.id].defaultVersion
+      configuredReference.id,
+      assertDefined(
```
TomWoodward, line 181: i think the reference has a version on it, you might not have to do this here

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,37 @@
...
+  books: BookConfig;
+  code: string;
+  id: string;
+}
+
+export default (baseUrl: string) => {
+  const loadBookRemoteBookConfig = (url: string): Promise<BookConfig | undefined> => {
+    return fetch(url)
+      .then((response) => {
+        if (response.status === 200) {
```
TomWoodward, line 21: there is an `acceptStatus` helper that you could use here

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,37 @@
...
+          return response.json() as Promise<ReleaseJsonStructure>;
+        }
+      })
+      .then((response) => response && response.books)
+      .catch(() => {
+        return Promise.resolve(undefined);
```
TomWoodward, line 27: it would be nice to have this put a warning in sentry, but definitely not pop an error to the user

TomWoodward (changes requested):
you may want to actually base your changes on piotr's rap refactor because it simplifies the types in resolveContent i think

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -152,8 +153,12 @@ export const getBookInformation = async(
   services: AppServices & MiddlewareAPI,
   reference: ReturnType<typeof getContentPageReferences>[number]
 ) => {
-  const getInputReferenceInfo = (id: string, inputVersion?: string) => {
-    const defaultVersion = BOOKS[id] ? BOOKS[id].defaultVersion : undefined;
+  const getInputReferenceInfo = async(id: string, inputVersion?: string) => {
+    const defaultVersion = assertDefined(
+      await services.bookConfigLoader.getBookVersionFromUUID(id),
+      `BUG: ${id} is not in BOOKS configuration`
```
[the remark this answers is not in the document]
jarosik10: oh yah. It should accept undefined here.

src/test/mocks/bookConfigLoader.ts
```diff
@@ -0,0 +1,9 @@
+import { BookVersion } from '../../gateways/createBookConfigLoader';
+
+export const mockBookVersionFromConfig: BookVersion = {
+  defaultVersion: 'testbook1-version',
+};
+
+export default () => ({
+  getBookVersionFromUUID: jest.fn(() => Promise.resolve(mockBookVersionFromConfig)),
```
PiotrKozlowski, line 8: Probably it should be mocked in the similar way to `mocks/archiveLoader.ts`

src/index.tsx
```diff
@@ -51,6 +52,7 @@ const app = createApp({
   initialState: window.__PRELOADED_STATE__,
   services: {
     archiveLoader: createArchiveLoader(archiveUrl),
+    bookConfigLoader: createBookConfigLoader(window.location.origin),
```
PiotrKozlowski, line 55: this should be configured like other urls

src/app/content/hooks/locationChange/resolveContent.ts
PiotrKozlowski: this is added in https://github.com/openstax/rex-web/pull/1133

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,37 @@
...
+      .then((response) => response && response.books)
+      .catch((e) => {
+        Sentry.captureException(e);
+        return Promise.resolve(undefined);
+      });
+  };
+
+  return {
+    getBookVersionFromUUID: (uuid: string): BookVersion | Promise<BookVersion | undefined> => {
+      const url = `${baseUrl}/rex/release.json`;
```
PiotrKozlowski, line 33: does it need to be here? it seems to me that it can be moved directly to the `loadRemoteBookConfig`

PiotrKozlowski (changes requested):
I didn't look at `resolveContent.ts ` because it will change after merging https://github.com/openstax/rex-web/pull/1133

src/app/content/hooks/locationChange/resolveContent.ts
jarosik10: Should I wait for #1133 to be merged? @PiotrKozlowski @TomWoodward

src/index.tsx
jarosik10: What exactly do you mean? via config?

PiotrKozlowski: you can do it similar to `archiveUrl`, `buyPrintConfigUrl` etc. from this file

src/app/content/hooks/locationChange/resolveContent.ts
PiotrKozlowski: You could either merge my branch into yours and change the base for this PR or you could just wait and resolve conflicts when my PR is merged - idk when this will happen

src/index.tsx
jarosik10: @PiotrKozlowski 
I wonder if the path to `release.json` file should be placed in config or just move all the logic for creating the `fetchURL` to the `bookConfigLoader` gateway?

PiotrKozlowski: I would stick to the same format as for other gateways, so you pass a `baseUrl` as you currently do, and `/rex/release.json` is added in `createBookConfigLoader` as it is now
the only change is to replace `window.location.origin` with a const `bookConfigUrl` that will be created with `assertDefined` in the same way like for example `buyPrintConfigUrl`

jarosik10: Okay, but there is question if we would like to make the `BOOK_CONFIG_URL` from `config.js` to receive the url as:
- `BOOK_CONFIG_URL: window.location.origin`
- or `BOOK_CONFIG_URL: typeof(window) === 'undefined' ? 'https://openstax.org' : window.location.origin`

And the naming seems weird. If this property only contains the origin path, maybe it should be named the way showing it. To make it feel reusable.

However it still feels awkward form me to store it in config, but maybe I'm missing something :D

jarosik10: @TomWoodward what do you think?

src/gateways/createBookConfigLoader.ts
PiotrKozlowski: this is still not resolved

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,37 @@
...
+      return BOOKS[uuid] ? BOOKS[uuid] : loadRemoteBookConfig(url).then((books) => books && books[uuid]);
```
PiotrKozlowski, line 34: I think it would be good to create `const cachedBooks = {...BOOKS}` and use this variable to store there versions loaded from the remote book configs so we don't request 2 times the same book from the remote config

src/config.js
```diff
@@ -24,6 +24,7 @@ let config = {
   OS_WEB_URL: process.env.OS_WEB_URL || 'https://openstax.org',
   SEARCH_URL: process.env.SEARCH_URL || 'https://openstax.org',
   HIGHLIGHTS_URL: process.env.HIGHLIGHTS_URL || 'https://openstax.org',
+  ORIGIN_URL: typeof(window) === 'undefined' ? 'https://openstax.org' : window.location.origin,
```
PiotrKozlowski, line 27: this would be probably `process.env.ORIGIN_URL || 'https://openstax.org'`
and in the dev config it could be `typeof(window) === 'undefined' ? 'https://openstax.org' : window.location.origin,` or 
`typeof(window) === 'undefined' ? 'https://openstax.org' : '',`

PiotrKozlowski (changes requested):
Should this new gateway be used in places like `getUrlParamsForBook`, `getCanonicalUrlParams` functios?

jarosik10:
I've find out that when there is a missing book in local book config (look testing in pr's description) the `href`s of links in ToC change from:
`https://localhost:3000/books/algebra-and-trigonometry/pages/1-1-real-numbers-algebra-essentials`
to:
`https://localhost:3000/books/13ac107a-f15f-49d2-97e8-60ab2e3b519c@32.2/pages/1-1-real-numbers-algebra-essentials`

TomWoodward:
@jarosik10 the reverse routing checks to see if the book is at the configured version and uses that to decide if it renders the regular link or the versioned link. did you not update that to use your new config gateway?

jarosik10:
@TomWoodward 
I'm working on it. But it will require a lot of code changing. Changin _getUrlParamsForBook_ and _getBookPageUrlAndParams_ in `urlUtils.ts` will affect a lot of components and _domVisitor_, _urlChecker_ scripts.

TomWoodward:
any scripts that aren't running the app are probably fine to continue accessing the config json directly. anything in the app that would also be running in the browser needs to use the gateway. the logic in the url param helpers that decides if it needs the versions in the url, that should use the gateway. unfortunately that will mean making a lot of url logic async and i'm unsure how gnarly that is going to be

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,46 @@
...
+      .then((response) => response.json() as Promise<ReleaseJsonStructure>)
+      .then((response) => response && response.books)
+      .catch((e) => {
+        Sentry.captureException(e);
+        return Promise.resolve(undefined);
+      });
+  };
+
+  return {
+    getBookVersionFromUUID: (uuid: string): BookVersion | Promise<BookVersion | undefined> => {
```
PiotrKozlowski, line 35: my understanding was that we should have 2 methods - one synchronous and the second one asynchronous 
both of them should access `cachedBooks` variable
the async one will be used in `resolveContent` and it will update `cachedBooks` if it wasn't found previously
the sync one will just return version from `cachedBooks`

PiotrKozlowski (changes requested):

src/gateways/createBookConfigLoader.ts
jarosik10: To be consistent I will have to replace all `BOOKS[id]` calls in the project with the sync function from the `bookConfigLoader`, right? But in the other hand it will make some components need to receive or load `AppServices['BookConfigLoader']`, which will require some refactoring (like in this [PR](https://github.com/openstax/rex-web/pull/1153), but without making some functions async).

PiotrKozlowski: I think this will be okay. The other option would be to just create a module that exports these functions instead of creating it as a gateway and do not pass it to the services

PiotrKozlowski: @TomWoodward Do you have an opinion on this?

TomWoodward: it might be ok for the synchronous function to be a regular export that references a `let books` at the top of the module

src/gateways/createBookConfigLoader.ts
```diff
@@ -44,3 +44,5 @@ export default (baseUrl: string) => {
     },
   };
 };
+
+export const getBookVersionFromUUID = (uuid: string): BookVersion | undefined => cachedBooks[uuid];
```
PiotrKozlowski, line 48: I would call this function differently and change the type of the first one to always return a promise

src/app/content/utils/canonicalUrl.ts
```diff
@@ -16,14 +16,15 @@ export async function getCanonicalUrlParams(
...
     ...(CANONICAL_MAP[book.id] || []),
-    ...(BOOKS[book.id] && bookVersion === BOOKS[book.id].defaultVersion ? bookDefaultMap : []),
+    ...(bookVersionFromConfig && bookVersion === bookVersionFromConfig.defaultVersion ? bookDefaultMap : []),
     // use the current book as a last resort if it has the same version as in books config
-  ]).filter(([id]) => !!BOOKS[id]);
+  ]).filter(([id]) => !!getBookVersionFromUUID(id));
 
   for (const [id, CANONICAL_PAGES_MAP] of canonicals) {
-    const version = BOOKS[id].defaultVersion;
+    const version = getBookVersionFromUUID(id)!.defaultVersion;
```
PiotrKozlowski, line 27: `assertDefined` or map through `cannonicals` and add `version` to them

src/app/content/hooks/locationChange.spec.ts
```diff
@@ -203,11 +203,13 @@ describe('locationChange', () => {
       publish_date: '2012-06-21',
     };
 
+    mockConfig.BOOKS[mockOtherBook.id] = {defaultVersion: mockOtherBook.version};
+    jest.doMock('../../../config', () => mockConfig);
+
```
PiotrKozlowski, line 208: I think you could just `mockImplementation` on `getBookVersionFromUUID`

PiotrKozlowski (changes requested):

src/app/content/hooks/locationChange.spec.ts
jarosik10: Is it really needed? It seems easier to mock the config there.

src/app/content/utils/canonicalUrl.ts
```diff
@@ -16,15 +16,18 @@ export async function getCanonicalUrlParams(
...
     ...(bookVersionFromConfig && bookVersion === bookVersionFromConfig.defaultVersion ? bookDefaultMap : []),
     // use the current book as a last resort if it has the same version as in books config
-  ]).filter(([id]) => !!getBookVersionFromUUID(id));
+  ]).filter(([id]) => !!getBookVersionFromUUIDSync(id));
 
   for (const [id, CANONICAL_PAGES_MAP] of canonicals) {
-    const version = getBookVersionFromUUID(id)!.defaultVersion;
+    const version = assertDefined(
+      getBookVersionFromUUIDSync(id),
+      `BUG: Book ${id} is not in BOOKS configuration`
```
PiotrKozlowski, line 29:
```suggestion
      `We've already filtered out books that are not in the `BOOK` configuration`
```

PiotrKozlowski (approved):

src/app/content/hooks/locationChange.spec.ts
TomWoodward: i agree that it would be more straight forward to use a spy on `getBookVersionFromUUID`, but its fine like this

src/config.d.ts
```diff
@@ -10,6 +10,7 @@ interface Config {
   REACT_APP_OS_WEB_API_URL: string;
   REACT_APP_SEARCH_URL: string;
   REACT_APP_HIGHLIGHTS_URL: string;
+  ORIGIN_URL: string;
```
TomWoodward, line 13: origin url for what? is this used for loading the rex configs? oh i see below you're handling undefined window, is this only for loading rex configs when pre-rendering? because when not pre-rendering you could just not put a host and use a relative url like `/rex/release.json`

when pre-rendering our local config will definitely be newer than the remote, we should not go out of our way to make fetching the remote work. in the gateway we could skip the remote check if window is undefined

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default (baseUrl: string) => {
+  const loadRemoteBookConfig = (): Promise<BookConfig | undefined> => {
+    const url = `${baseUrl}/rex/release.json`;
+    return fetch(url)
+      .then(acceptStatus(200, (status, message) => `Error response from "${url}" ${status}: ${message}`))
+      .then((response) => response.json() as Promise<ReleaseJsonStructure>)
+      .then((response) => response && response.books)
+      .catch((e) => {
+        Sentry.captureException(e);
```
TomWoodward, line 29: i can't imagine this being anything but a network failure, which will just be noise in sentry, worth reporting it but maybe make it a warning

TomWoodward (changes requested):

src/config.d.ts
TomWoodward: if you're not using the gateway in the scripts, then you don't need the skip check in the gateway, but i don't understand why you need the origin url config, you can just reference with a relative url like `/rex/release.json`

PiotrKozlowski: I think I've suggested adding it to the config since I saw a bunch of this type of variables there, but indeed relative url seems better

TomWoodward:
@jarosik10 it doesn't look like you've updated this since my review

jarosik10:
@TomWoodward Sorry. Changes are pushed.

src/config.js
```diff
@@ -24,6 +24,7 @@ let config = {
   OS_WEB_URL: process.env.OS_WEB_URL || 'https://openstax.org',
   SEARCH_URL: process.env.SEARCH_URL || 'https://openstax.org',
   HIGHLIGHTS_URL: process.env.HIGHLIGHTS_URL || 'https://openstax.org',
+  ORIGIN_URL: process.env.ORIGIN_URL || 'https://openstax.org',
```
TomWoodward, line 27: this isn't necessary right?

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,48 @@
...
+  [key: string]: BookVersion;
+}
+
+interface ReleaseJsonStructure {
+  books: BookConfig;
+  code: string;
+  id: string;
+}
+
+const cachedBooks = { ...BOOKS } as BookConfig;
```
TomWoodward, line 19: this is the type that the books object already has, why do you need to redeclare `BookConfig` and `BookVersion` here?

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  return {
+    getBookVersionFromUUID: (uuid: string): Promise<BookVersion | undefined> => {
+      return cachedBooks[uuid] ? Promise.resolve(cachedBooks[uuid]) : loadRemoteBookConfig().then((books) => {
+        let bookVersion;
+        if (books && uuid in books) {
+          bookVersion = books[uuid];
+          cachedBooks[uuid] = bookVersion;
+        }
+        return bookVersion;
```
TomWoodward, line 42: why do you have all this logic in here? why don't you just assign books over cachedBooks and return `getBookVersionFromUUIDSync`?

TomWoodward (changes requested):

src/gateways/createBookConfigLoader.ts
jarosik10: yea, actually it is not needed to point it here `const cachedBooks = { ...BOOKS } as BookConfig;`, but these interfaces are showing what data types the function returns.

src/gateways/createBookConfigLoader.ts
jarosik10: this is what you mean?
```js
getBookVersionFromUUID: (uuid: string): Promise<BookVersion | undefined> => {
      return cachedBooks[uuid] ? Promise.resolve(cachedBooks[uuid]) : loadRemoteBookConfig().then((books) => {
        if (books) {
          cachedBooks = books;
        }
        return getBookVersionFromUUIDSync(uuid);
      });
    },
});
```

TomWoodward: i think so, that works right?

src/gateways/createBookConfigLoader.ts
TomWoodward: in the case of the function you can just use `typeof BOOKS` right?

TomWoodward (changes requested):
there are still some outstanding comments

src/gateways/createBookConfigLoader.ts
```diff
@@ -0,0 +1,40 @@
+import { BOOKS } from '../config';
+import { acceptStatus } from '../helpers/fetch';
+import Sentry, { Severity } from '../helpers/Sentry';
+
+export type BookVersion = typeof BOOKS[0];
```
TomWoodward, line 5: i don't think this needs to be exported

src/setupProxy.js
```diff
@@ -202,6 +202,19 @@ function stubRedirects(app) {
   });
 }
 
+function stubRelease(app) {
+  app.use((req, res, next) => {
+    const {pathname} = url.parse(req.url);
+
+    if (pathname === '/rex/release.json') {
+      const releaseFile = path.join(__dirname, 'release.development.json');
```
TomWoodward, line 210: this json file will become old and weird very quickly, can you construct a json response here based on the actual books config? 

or maybe since the json config will never really be used in development, since it is only ever used when the book check is missed, maybe leave the release.development.json but leave the books section empty, that way its there if someone wants to test something and add a book but it doesn't have content that will need to be updated to continue making sense

TomWoodward (changes requested):

src/gateways/createBookConfigLoader.ts
jarosik10: It is imported in `src/test/mocks/bookConfigLoader.ts`.

TomWoodward: you'd be better off setting the mock `getBookVersionFromUUID` to `ReturnType<typeof createBookConfigLoader>['getBookVersionFromUUID']` so that its forced to match the actual type instead of deriving the same type again

TomWoodward (changes requested):

TomWoodward (approved):

PR #1147, opened 2021-04-09
Issue 969 - updated meta descriptions

bethshook:
for: https://github.com/openstax/unified/issues/969


src/app/content/utils/seoUtils.ts
```diff
@@ -25,6 +27,54 @@ const getParentPrefix = (node: LinkedArchiveTreeNode | undefined): string => {
 
 };
 
+const hideMath = (node: any) => {
```
bethshook, line 30: @TomWoodward what is the correct type for a DOM node/HTML element? TS complained about everything I tried.

src/app/content/utils/seoUtils.ts
```diff
@@ -25,6 +27,54 @@ const getParentPrefix = (node: LinkedArchiveTreeNode | undefined): string => {
...
+  const node = assertDefined(
+    findArchiveTreeNodeById(book.tree, page.id),
+    `couldn't find node for a page id: ${page.id}`
+  );
+  const sectionTitle = getArchiveTreeSectionTitle(node);
+  const nodeTitleDoc = domParser.parseFromString(node.title, 'text/html');
+  const titleNode = nodeTitleDoc.body.children[0];
+  const chapterFromTitle = titleNode.innerText.split('.')[0] || '';
+  const isAnswerKey = contentNode.classList.contains('os-solution-container')
+    || contentNode.classList.contains('os-solutions-container');
```
bethshook, line 63: @TomWoodward These are the two classes I've seen applied to answer keys.

TomWoodward: lol the pluralization. ok.

src/app/content/hooks/receiveContent.ts
```diff
@@ -42,10 +35,9 @@ const hookBody: ActionHookBody<typeof receivePage> = ({
   }
 
   const title = createTitle(page, book);
-
-  // the abstract could be '<div/>'.
-  const abstract = stripHtmlAndTrim(page.abstract ? page.abstract : '');
-  const description = abstract || stripHtmlAndTrim(getCleanContent(book, page, archiveLoader));
+  const cleanContent = getCleanContent(book, page, archiveLoader);
```
TomWoodward, line 38: i think it'd be slightly cleaner to put the `getCleanContent` inside `createDescription` unless there is some other use for it out here

src/app/content/utils/seoUtils.ts
TomWoodward: import HTMLElement from that openstax place

src/app/content/utils/seoUtils.ts
```diff
@@ -25,6 +27,57 @@ const getParentPrefix = (node: LinkedArchiveTreeNode | undefined): string => {
...
+  );
+  const sectionTitle = getArchiveTreeSectionTitle(node);
+  const nodeTitleDoc = domParser.parseFromString(node.title, 'text/html');
+  const titleNode = nodeTitleDoc.body.children[0];
+  const titleNodeNum = parseInt(titleNode.innerText.split('.')[0]);
+  const prefix = getParentPrefix(node.parent);
+  const prefixNum = parseInt(prefix.trim().split(' ')[1]);
+
+  // Check for numbers indicating chapter in title node and then prefix.
+  const chapterNum = !isNaN(titleNodeNum) ? titleNodeNum : (!isNaN(prefixNum) ? prefixNum : '');
```
TomWoodward, line 66: lets make like a `getTemplateValues` function or something like that

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,149 @@ import {
...
+  } else {
+    let descriptionPhrase = '';
+    if (pageType === 'answer-key') {
+      descriptionPhrase = `the Answer Key for ${sectionTitle} of`;
+    } else if (!parentIsChapter && !parentIsBook) {
+      descriptionPhrase = `${parentTitle}: ${sectionTitle} for ${parentPrefix} of`;
+    } else if (sectionTitle !== parentPrefix) {
+      descriptionPhrase = `the ${sectionTitle} for ${parentPrefix} of`;
+    } else {
+      descriptionPhrase = `the ${sectionTitle} for`;
```
TomWoodward, line 154: we're gonna have to integrate this into the internationalization system, so for now what i'd like to see is helpers like this:

```ts
type PageTypes = 'answer-key' | 'section' | 'introduction'; // or whatever, i'm making these up
interface DescriptionTemplateValues { // or whatever, i'm making these up
  parentType: 'chapter' | 'book';
  pageTitle: string;
  bookTitle: string;
  chapterTitle?: string;
}
const generateDescriptionFromTemplate = (pageType: PageTypes, values: DescriptionTemplateValues) {
  switch (pageType) {
    case 'answer-key':
      return `the whole template string`;
    case 'section':
      return `the whole template string`;
    default:
       throw new Error('unknown page type');
  }
}
const getPageType = (treeNode: ArchiveTreeSection, contentNode: HTMLElement): PageTypes => {
  if (whatever) {
    return something;
  }

  return stuff
}
const getPageDescriptionValues = (treeNode: ArchiveTreeSection, contentNode: HTMLElement): DescriptionTemplateValues => {
  return ...;
}
const getPageDescriptionFromContent = (pageType: PageTypes, contentNode: HTMLElement): string | null => {
  // try to find paragraph, if found sanitize and return snippet. maybe use pagetype to filter certain types of content per page, maybe not necessary idk

 // if no paragraph found or content is too short return null
}
const getPageDescription = (... whatever arguments) => {
  const pageType = getPageType(stuff)
  if (['section', 'introduction'].contains(pageType) {

  }
  const contentDescription: string | null = ['section', 'introduction'].contains(pageType)
    ? getPageDescriptionFromContent(stuff)
    : null;

  return contentDescription || generateDescriptionFromTemplate(stuff)
}
```

take all that with a grain of salt i'm trying to outline a sort of logical structure but don't follow through on anything i did if it feels weird or bad when you get into it. do you understand kind of where i'm going with this? eventually we'll replace a chunk of `generateDescriptionFromTemplate` with 
```
intl.formatMessage(`intl:meta-description:${pageType}`, descriptionValues)
```
but i don't want to think about how to get the intl object in there quite yet

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
bethshook: yep, sounds good

bethshook: still need to do some testing but I thing it's much more legible now. thanks for the outline.

src/test/mocks/archiveLoader.ts
```diff
@@ -85,13 +85,25 @@ export default () => {
       localBooks[`${newBook.id}@${newBook.version}`] = newBook;
       localBookPages[`${newBook.id}@${newBook.version}`] = {};
     },
-    mockPage: (parentBook: ArchiveBook, newPage: ArchivePage, pageSlug: string) => {
-      localBookPages[`${parentBook.id}@${parentBook.version}`][newPage.id] = newPage;
-      localBooks[`${parentBook.id}@${parentBook.version}`].tree.contents.push({
+    // tslint:disable-next-line: max-line-length
+    mockPage: (parentBook: ArchiveBook, newPage: ArchivePage, pageSlug: string, parentChapter?: ArchiveTree, parentSection?: ArchiveTree) => {
```
bethshook, line 89: @TomWoodward curious if there is a better way to do this or if I should write a new function. `mockPage()` was built to add a page at the book level and I wanted options for adding book subtrees.

src/app/content/utils/seoUtils.constants.ts
```diff
@@ -0,0 +1,107 @@
+// tslint:disable: object-literal-sort-keys max-line-length
```
TomWoodward, line 1: should name this file something to indicate that it is used in specs maybe `setUtils.spec.data.ts` or something like that

src/app/content/utils/seoUtils.spec.ts
```diff
@@ -1,7 +1,83 @@
...
+  contentPage,
+  contentPageShort,
+  contentPageWithObjectives,
+  eobPage,
+  eocPage,
+  eocSubPage,
+  mockBook,
+} from './seoUtils.constants';
+
+// tslint:disable: object-literal-sort-keys max-line-length
```
TomWoodward, line 17: you don't need to ignore the sort key anymore right?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+
+export const getParentPrefix = (node: LinkedArchiveTreeNode | undefined, includeTitle: boolean = false): string => {
   if (!node) {
     return '';
   }
 
   if (archiveTreeSectionIsChapter(node)) {
-    const number = getArchiveTreeSectionNumber(node);
-    return `Ch. ${number} `;
+    const number = getArchiveTreeSectionNumber(node).replace('Chapter', '').trim();
```
TomWoodward, line 33: where did you see something with the word `chapter` in the number?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  return node.textContent;
+};
+
+export const getTextContent = (str: string) => {
+  const parsed = domParser.parseFromString(str, 'text/html');
+  const text = parsed.body.textContent;
+  return text;
+};
+
+const removeIntroContent = (node: HTMLElement) => {
```
TomWoodward, line 61: can you name this something else? it sounds like its only used on introduction pages, but some of this stuff is relevant to all content pages

incidentally if you wanted to make different cleanup logic per page type that might be a good idea... i could go either way on that

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+    return null;
+  }
+
+  const introContentList = node.querySelectorAll(
+    '[data-type="abstract"], .learning-objectives, .chapter-objectives, .be-prepared, .os-teacher'
+    ) || [];
+
+  for (let i = 0; i <= introContentList.length; i++) {
+    if (introContentList[i]) {
+      const parent = introContentList[i].parentNode;
```
TomWoodward, line 72: is there not a remove self or something? https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove

is there a reason you're not using that?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+      if (parent === null) {
+        break;
+      }
+      parent.removeChild(introContentList[i]);
+    }
+  }
+};
+
+export const generateExcerpt = (str: string) => {
+  return str.replace(/\n/g, ' ').replace(/\s\s/g, ' ').trim().substring(0, 152) + '...';
```
TomWoodward, line 82: should `.replace(/\s\s/g, ' ')` be `.replace(/\s+/g, ' ')` ?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+    nodeClasses.contains('os-solution-container')
+      || nodeClasses.contains('os-solutions-container')
+      ) {
+    return 'answer-key';
+  } else if (parentType !== 'chapter' && parentType !== 'book') {
+    return 'eoc-sub-page';
+  } else if (pageTitle !== chapterTitle) {
+    return 'eoc-page';
+  } else {
+    return 'eob-page';
```
TomWoodward, line 105: probably don't make this a fall through case. i think it might be better to throw if you can't figure out what kind of page it is

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+      parent.removeChild(introContentList[i]);
+    }
+  }
+};
+
+export const generateExcerpt = (str: string) => {
+  return str.replace(/\n/g, ' ').replace(/\s\s/g, ' ').trim().substring(0, 152) + '...';
+};
+
+// tslint:disable-next-line: max-line-length
```
TomWoodward, line 85: i don't think you need this

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  const nodeType = node.getAttribute('data-type');
+
+  if (nodeType === 'page') {
+    return 'page';
+  } else if (nodeClasses.contains('appendix')) {
+    return 'appendix';
+  } else if (
+    nodeClasses.contains('os-solution-container')
+      || nodeClasses.contains('os-solutions-container')
+      ) {
```
TomWoodward, line 98: pull the `) {` back to the first indentation level

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+    return 'answer-key';
+  } else if (parentType !== 'chapter' && parentType !== 'book') {
+    return 'eoc-sub-page';
```
TomWoodward, line 101: this will return true for answer key pages, whose parent is the answer key section. shouldn't those be like... `eob-subpage` or something?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  } else if (pageTitle !== chapterTitle) {
```
TomWoodward, line 102: when does the page title match the chapter title?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  }
+};
+
+const getPageDescriptionFromContent = (node: HTMLElement): string | null => {
+  if (!node) {
+    return null;
+  }
+  removeIntroContent(node);
+  const firstSection = node.querySelector('section');
+  let para: Element | null = firstSection
```
TomWoodward, line 115: does it work to do `node.querySelector(':scope>section>p,:scope>p')` so that you get either at the same time?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+    const mathless = hideMath(para);
+    return mathless ? generateExcerpt(mathless) : null;
+  } else {
+    const text = para.textContent;
+    const next = para.nextElementSibling;
+    while (text && text.length < 90 && next) {
+      if (next.matches('p')) {
+        para = next;
+      }
+    }
```
TomWoodward, line 134: i think you could replace most of this logic by initially doing `querySelectorAll` up on line 115, and then doing `find` with your length criteria

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  switch (pageType) {
+    case 'page':
+      return `On this page you will discover the ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'answer-key':
+      return `On this page you will discover the Answer Key for ${pageTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eoc-sub-page':
+      return `On this page you will discover the ${parentTitle}: ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eoc-page':
+      return `On this page you will discover the ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eob-page' || 'appendix':
```
TomWoodward, line 151: is there a reason to not just generally treat appendix as `eob-page`? as in return `eob-page` from its page type logic?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+      ? 'book'
+      : 'other');
+
+  const values: DescriptionTemplateValues = {
+    bookTitle: book.title,
+    chapterTitle,
+    pageTitle,
+    parentTitle,
+    parentType,
+  };
```
TomWoodward, line 182: please extract this to another function for getting the template variables

src/test/mocks/archiveLoader.ts
```diff
@@ -49,7 +49,7 @@ export default () => {
     return bookData
       ? Promise.resolve(bookData)
       : Promise.reject(new Error(`failed to load book data ${bookId}@${bookVersion}`))
-    ;
+      ;
```
TomWoodward, line 52: i liked most of this formatting the way it was before

src/test/mocks/archiveLoader.ts
TomWoodward: it looks like perhaps you don't even need to call `mockPage` because you're setting up your tree structure manually right?

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,182 @@ import {
...
+  const {parentTitle, pageTitle, bookTitle, chapterTitle} = values;
+  switch (pageType) {
+    case 'page':
+      return `On this page you will discover the ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'answer-key':
+      return `On this page you will discover the Answer Key for ${pageTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eoc-sub-page':
+      return `On this page you will discover ${parentTitle}: ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eoc-page':
+      return `On this page you will discover the ${pageTitle} for ${chapterTitle} of OpenStax's ${bookTitle} free textbook.`;
```
TomWoodward, line 150: do you have feelings about the presence of this `the` vs it not being in `eoc-sub-page`?

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
bethshook: It only happened when I ran the metadata dump script we wrote that pulls books from config.books.json

src/app/content/utils/seoUtils.ts
bethshook: This is only used for page type "page" (see line 186)

src/app/content/utils/seoUtils.ts
bethshook: you're right - I may have been worried about IE compatibility but I see it's used elsewhere

src/app/content/utils/seoUtils.ts
bethshook: I think there are more instances of weirdness when "the" is included

src/app/content/utils/seoUtils.ts
bethshook: eob pages. maybe the problem is the name "chapterTitle." I'll restore it to "parentPrefix" even though I don't really understand that name.

src/app/content/utils/seoUtils.ts
bethshook: renamed to subpage since these can have a grandparent of chapter or book

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,167 @@ import {
...
+    throw new Error('no page type detected');
+  }
+};
+
+const getPageDescriptionFromContent = (node: HTMLElement): string | null => {
+  if (!node) {
+    return null;
+  }
+  removeExcludedContent(node);
+  const paragraphs = node.querySelectorAll(':scope>section>p,:scope>p');
```
bethshook, line 108: @TomWoodward belated question about this: didn't we agree to prioritize paragraphs within sections over other paragraphs? This won't do that.

(It also throws an error because `:scope` yields an ID selector beginning with a digit, which [isn't allowed in CSS3](https://stackoverflow.com/a/37271406).)

TomWoodward: what element has a numeric id?

src/app/content/utils/seoUtils.ts
TomWoodward: do you know what book it was in?

src/app/content/utils/seoUtils.ts
bethshook: the content div: https://share.getcloudapp.com/z8u6rmlY

bethshook: this seems ok unless we're concerned about paragraphs that are children of some other element: `const paragraphs = node.querySelectorAll('section>p') || node.querySelectorAll('p');`

src/app/content/utils/seoUtils.ts
bethshook: All of them. When I log `getArchiveTreeSectionNumber(node)`, which pulls the text content of the from the first `.os-number` element, on my local env it contains only a number (eg "2"). When I log it via the script it logs "Chapter 2"

src/test/mocks/archiveLoader.ts
bethshook: I'm trying to implement it this way but without `mockPage` the loader just has empty pages via `mockBook`: 
```
mockBook: (newBook: ArchiveBook) => {
      localBooks[`${newBook.id}@${newBook.version}`] = newBook;
      localBookPages[`${newBook.id}@${newBook.version}`] = {};
    },
```

This causes `getCleanContent()` to return empty because each page cache returns undefined: 
```
const cachedPage = book && page &&
    archiveLoader.book(book.id, book.version).page(page.id).cached()
  ;
```

src/app/content/utils/seoUtils.ts
TomWoodward: ah, i was wondering how you could not be getting the logic that cleans up the tree, but apparently its not set up to run until you combine the archive book with the cms data. if you look in ` src/app/content/utils.ts` there is some logic there that cleans the tree titles. i think all the logic in the app is going to expect that cleaning logic to have been run.

i think that if your logic were to run on the actual page, the place where the data comes from would have been cleaned already, but in your script you're loading it from the archive loader directly and its not going through the same logic

src/app/content/utils/seoUtils.ts
TomWoodward: you're right that we need to prioritize paragraphs in sections, in order to simplify your filtering logic you might want to initialize as `[...node.querySelectorAll('section>p'), ...node.querySelectorAll('p')]` so that you can then just filter by text length

if you see anything with an invalid id that is a bug we should bring up with the content team. in your screenshot i think its valid because it starts with a `d`, but it looks like they're just using a uuid and _sometimes_ that will start with a number.

src/test/mocks/archiveLoader.ts
TomWoodward: oh yea, `mockPage` does some stuff to the tree that you've already handled and is probably bad for it to try to do again, but you need the part where it updates `localBookPages`.

you may need to update `mockPage` to conditionally not update the tree

TomWoodward (changes requested):

src/app/content/utils/seoUtils.ts
bethshook: do you think we need that cleanup logic here or will it suffice to just tidy the `parentPrefix` string within the script file? I'm struggling to find which util function I would need to apply

bethshook: ah but then I have the same issue when running tests

src/test/mocks/archiveLoader.ts
```diff
@@ -85,13 +85,18 @@ export default () => {
...
+      const currentBook = localBooks[`${parentBook.id}@${parentBook.version}`];
+      if (tree) {
+        currentBook.tree.contents.push(tree);
+      } else {
+        currentBook.tree.contents.push({
+          id: `${newPage.id}@${newPage.version}`,
+          slug: pageSlug,
+          title: newPage.title,
+        });
+      }
```
bethshook, line 99: @TomWoodward this way a page is only added directly to a book if no tree is passed in

src/app/content/utils/seoUtils.ts
TomWoodward: honestly you might want to change your meta util to accept the combined book and then run everything including your test data through `formatBookData` and that'll handle it for you. you'll see other tests using `formatBookData` as well

src/test/mocks/archiveLoader.ts
TomWoodward: alternatively you could use something like `findArchiveTreeNode` to see if the node already exists before adding it.

i'm not sure how i feel about passing the tree in, since there is nothing that says it would have to be the same tree as the passed book, and if it was different the test page wouldn't work.

bethshook: got it, makes sense

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,168 @@ import {
...
+    : getParentPrefix(node.parent, includeTitle);
+
+};
+
+const hideMath = (node: Element) => {
+  if (!node) {
+    return '';
+  }
+
+  node.querySelectorAll('math, .os-math-in-para').forEach((el: Element) => {
```
bethshook, line 49: @TomWoodward the json files that are used for testing the `receiveContent` hook don't apply the `os-math-in-para` class to math. Instead of adding another selector here like `<math>` I'm wondering if the test page markup should be updated to use the class.

bethshook: (this is all because there is a meta description test for that hook and I don't want to "expect" math gibberish)

src/app/content/utils.ts
```diff
@@ -71,19 +72,19 @@ const pickArchvieFields = (archiveBook: ArchiveBook) => ({
...
+};
 
 export const formatBookData = <O extends OSWebBook | undefined>(
   archiveBook: ArchiveBook,
   osWebBook: O
 ): O extends OSWebBook ? BookWithOSWebData : ArchiveBook => {
   if (osWebBook === undefined) {
     // as any necessary https://github.com/Microsoft/TypeScript/issues/13995
-    return pickArchvieFields(archiveBook) as ArchiveBook as any;
+    return pickArchiveFields(archiveBook) as ArchiveBook as any;
```
TomWoodward, line 84: lol spelling is hard

src/app/content/utils/seoUtils.ts
TomWoodward: you could probably just update that test to mock your description util so that you don't have to worry about testing it from more than one place

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+
+const removeExcludedContent = (node: HTMLElement) => {
+  if (!node) {
+    return null;
+  }
+  const excludedContent = node.querySelectorAll(
+    '[data-type="abstract"], .learning-objectives, .chapter-objectives, .be-prepared, .os-teacher'
+    ) || [];
+
+  for (let i = 0; i <= excludedContent.length; i++) {
```
TomWoodward, line 69: you can use a for `in` or `of` or whatever the right word is here

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+      excludedContent[i].remove();
+    }
+  }
+};
+
+export const generateExcerpt = (str: string) => {
+  return str.replace(/\n+/g, ' ').replace(/\s+/g, ' ').trim().substring(0, 152) + '...';
+};
+
+const getPageType = (node: HTMLElement, values: DescriptionTemplateValues): PageTypes => {
```
TomWoodward, line 80: we're overloading `DescriptionTemplateValues` by using it to decide the page type and render the template

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+    return 'subpage';
+  } else if (pageTitle !== parentPrefix) {
+    return 'eoc-page';
+  } else {
+    throw new Error('no page type detected');
+  }
+};
+
+const getPageDescriptionFromContent = (node: HTMLElement): string | null => {
+  if (!node) {
```
TomWoodward, line 104: how could it be falsy?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+    return null;
+  }
+  removeExcludedContent(node);
+  const paragraphs = [...Array.from(node.querySelectorAll('section>p')), ...Array.from(node.querySelectorAll('p'))];
```
TomWoodward, line 108: i'm not loving that we're gonna get all paragraphs, not just the ones that are direct children of the page. i think there were some in like notes or whatever that we wanted to ignore ghat this would pick up.

maybe you should pass in the next node up so that you could include the page in the selector (since `:scope` is broken by the id)

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+
+  if (!paragraphs) {
+    return null;
+  }
+
+  const foundByLength = Array.from(paragraphs).find((p) => p.textContent ? p.textContent.length >= 90 : null);
+  const mathless = foundByLength ? hideMath(foundByLength) : null;
```
TomWoodward, line 115: maybe do hidemath before the length trimming, i think that'll give better results

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
 
   return archiveTreeSectionIsBook(node.parent)
     ? getArchiveTreeSectionTitle(node) + ' '
-    : getParentPrefix(node.parent);
+    : getParentPrefix(node.parent, includeTitle);
+
+};
+
+const hideMath = (node: Element) => {
+  if (!node) {
```
TomWoodward, line 45: how could this be falsy? the type isn't nullable

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,166 @@ import {
...
+  const {parentTitle, pageTitle, bookTitle, parentPrefix} = values;
+  switch (pageType) {
+    case 'page':
+      return `On this page you will discover the ${pageTitle} for ${parentPrefix} of OpenStax's ${bookTitle} free textbook.`;
+    case 'answer-key':
+      return `On this page you will discover the Answer Key for ${pageTitle} of OpenStax's ${bookTitle} free textbook.`;
+    case 'subpage':
+      return `On this page you will discover ${parentTitle}: ${pageTitle} for ${parentPrefix} of OpenStax's ${bookTitle} free textbook.`;
+    case 'eoc-page':
+      return `On this page you will discover the ${pageTitle} for ${parentPrefix} of OpenStax's ${bookTitle} free textbook.`;
```
TomWoodward, line 130: are `page` and `eoc-page` the same?

src/test/mocks/archiveLoader.ts
```diff
@@ -83,11 +84,17 @@ export default () => {
...
       localBookPages[`${parentBook.id}@${parentBook.version}`][newPage.id] = newPage;
-      localBooks[`${parentBook.id}@${parentBook.version}`].tree.contents.push({
-        id: `${newPage.id}@${newPage.version}`,
-        slug: pageSlug,
-        title: newPage.title,
-      });
+      const currentBook = localBooks[`${parentBook.id}@${parentBook.version}`];
+      const treeNode = findArchiveTreeNodeById(parentBook.tree, newPage.id);
+      if (treeNode) {
+        currentBook.tree.contents.push(treeNode);
```
TomWoodward, line 90: if you've found the treenode in the tree already, you don' t have to push it on.

i guess we're assuming that the `parentBook` is already in `localBooks`, and not some other object with the same id and version, but idk what to do about that.

TomWoodward (changes requested):

src/app/content/utils.ts
```diff
@@ -61,7 +61,8 @@ export const parseContents = (book: Book, contents: Array<ArchiveTree | ArchiveT
   return contents;
 };
 
-const pickArchvieFields = (archiveBook: ArchiveBook) => ({
+const pickArchiveFields = (archiveBook: ArchiveBook) => {
+  return {
```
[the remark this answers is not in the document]
bethshook: I added a log and forgot to restore : (

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,154 @@ import {
...
+  const nodeType = node.getAttribute('data-type');
+
+  if (nodeType === 'page' && !nodeClasses.contains('appendix')) {
+    return 'page';
+  } else if (pageTitle === parentPrefix) {
+    return 'eob-page';
+  } else if (
+    nodeClasses.contains('os-solution-container')
+    || nodeClasses.contains('os-solutions-container')
+      || nodeClasses.contains('os-end-of-book-solutions-container')
```
bethshook, line 94: @TomWoodward the gift that keeps on giving

src/app/MessageProvider.tsx
```diff
@@ -17,8 +17,17 @@ async function polyfill(locale: 'en') {
...
+export const intl = createIntl({
+  locale: 'en',
+  messages: enMessages,
+}, cache);
+
 // tslint:disable-next-line:variable-name
-const MessageProvider: React.SFC<{onError?: () => void}> = (props) =>
-  <IntlProvider onError={props.onError} locale='en' messages={enMessages}>{props.children}</IntlProvider>;
+const MessageProvider: React.FC<{}> = (props) =>
+  <RawIntlProvider value={intl}>
```
bethshook, line 29: @TomWoodward Importing the useServices() hook here instead created hook usage errors in some tests =/

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,155 @@ import {
...
+  const foundByLength = Array.from(paragraphs).find((p) => {
+    const mathlessP = hideMath(p);
+    return mathlessP.textContent && mathlessP.textContent.length >= 90 ? mathlessP : null;
+  });
+  return foundByLength && foundByLength.textContent ? generateExcerpt(foundByLength.textContent) : null;
+};
+
+const getTemplateVars = (book: Book, node: LinkedArchiveTreeNode) => {
+  const parentTitle = node.parent ? getTextContent(node.parent.title) : null;
+  const parentPrefix = getParentPrefix(node, true).replace('Ch.', 'Chapter').trim();
```
PiotrKozlowski, line 125: The `Ch.` and `Chapter` should also come from `intl` service

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,155 @@ import {
...
+export const getPageDescription = (services: Services, book: Book, page: Page) => {
+  const intl = services.intl;
+  const cleanContent = getCleanContent(book, page, services.loader);
+  const doc = domParser.parseFromString(cleanContent, 'text/html');
+  const node = doc.body.children[0];
+  const treeNode = findArchiveTreeNodeById(book.tree, page.id);
+  const values = treeNode ? getTemplateVars(book, treeNode) : undefined;
+  if (!values || !treeNode) {
+    return '';
+  }
```
PiotrKozlowski, line 153: You could return after `treeNode` and get rid of ternary

bethshook:
Splitting into separate PRs

PR #1148, opened 2021-04-13
Issue 1227 - when user focuses a highlight then move only these cards that needs to be moved 

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1227

Currently, we are calculating card positions only when their height changes, and when the user focuses the highlight we are moving the whole sidebar with all cards.
We should update card positions when focused highlight is changing and move only those cards that are needed.

New logic:
1. calculate all positions like we currently do

do next steps only if there is a focused highlight:

2. split highlights into 2 arrays: before and after focused highlight (the first array includes focused highlight)
3. take the first array, reverse it and then start adjusting highlights positions - break the loop if the next highlight doesn't have to be updated (because it has enough space)

adjusting highlights positions at step 3 may create a space for highlights after this one so we have to recalculate their positions one more time

4. take the second array and calculate all positions with the current algorithm but adjust initial variables to take into consideration that we are starting not from the first highlight

## Video with how it works
https://drive.google.com/file/d/10OJe9gkyvgsVZL4sUs7PnhJ2ltDbGIAM/view?usp=sharing


src/app/content/highlights/components/cardUtils.ts
```diff
@@ -71,3 +73,121 @@ export const generateUpdatePayload = (
 
   return {updatePayload, preUpdateData};
 };
+
+/**
+ * Calculate positions for the cards according to the location of associated highlights in the document.
+ * Positions will be adjusted to the presence of the other cards in cases like multiple highlights in the same line.
+ */
+const updateStackedCardsPostions = (
```
bethshook, line 81: Typo in "Positions"

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -71,3 +73,121 @@ export const generateUpdatePayload = (
...
+  highlightsElements: Highlight[],
+  heights: Map<string, number>,
+  getHighlightPosition: (highlight: Highlight) => { top: number, bottom: number },
+  initialPositions?: Map<string, number>,
+  addMarginBottomToTheFirstCard = false,
```
bethshook, line 86: What determines this? Just want to make sure I understand.

TomWoodward (approved):
transition is lit 🔥 🔥 🚒

src/app/content/highlights/components/cardUtils.ts
PiotrKozlowski: I've changed it to `addAditionalMarginForTheFirstCard`

When we calculate `stackedTopOffset` on line `97` we want to make sure that it includes minimum margin-bottom from the previous card. 
I'm adding this margin to all cards except the first one because there is nothing above it.
This logic has changed since now I'm using this function two times - the second call is only for the cards after the `focusedHighlight` so I actually want the first card to be adjusted by this margin-bottom value

PR #1149, opened 2021-04-14
Do not reset PQ when ff are fetched

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1536


src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -38,19 +38,28 @@ const setSection = (services: MiddlewareAPI & AppServices, book: Book, page: Pag
...
+    services.dispatch(receivePracticeQuestionsSummary(practiceQuestionsSummary));
   }
+};
+
+const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
+  const state = services.getState();
+  const { book, page } = bookAndPage(state);
+  const isEnabled = practiceQuestionsEnabled(state);
+
+  if (!book || !page || !isEnabled) { return; }
```
TomWoodward, line 60: you don't need this check here right? it'll get caught by the one in the helper?

src/app/content/practiceQuestions/hooks/locationChange.ts
```diff
@@ -38,19 +38,28 @@ const setSection = (services: MiddlewareAPI & AppServices, book: Book, page: Pag
...
+
+  await loadPracticeQuestionsSummaryHookBody(services)();
 
   setSection(services, book, page);
```
TomWoodward, line 64: won't this make it so that if location change fires before feature flags are received and then the dialog is opened the section won't be set?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1151, opened 2021-04-15
Dont use buttons aria-label as analytics label

jarosik10:
for: https://github.com/openstax/unified/issues/1535


src/app/content/components/popUp/FiltersList.tsx
```diff
@@ -65,6 +65,7 @@ export const FiltersListColor = (props: FiltersListColorProps) => (
     <StyledPlainButton
       aria-label={useIntl().formatMessage({id: props.ariaLabelKey(props.color)}, {filterValue: props.color})}
       onClick={props.onRemove}
+      data-analytics-label={`Remove breadcrumb for color ${props.color}`}
     >
```
jarosik10, line 69: It doesn't work properly for Study guides. Example:
```
aria-label="Remove breadcrumb for label Connections & Relationships"
data-analytics-label="Remove breadcrumb for color green"
```

TomWoodward: what does the intl message look like? does it have a big condition in there with every label type by color?

PiotrKozlowski (changes requested):

src/app/content/components/popUp/ChapterFilter.tsx
```diff
@@ -130,6 +130,7 @@ const ChapterFilter = (props: ChapterFilterProps) => {
               title={section.title}
               onChange={() => handleChange(section)}
               ariaLabel={getAriaLabel(section)}
+              dataAnalyticsLabel={`Filter PQ by ${splitTitleParts(section.title).join(' ')}`}
```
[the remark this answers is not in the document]
PiotrKozlowski: or just create function like `prepareSectionTitle` and use it here and in `getAriaLabel`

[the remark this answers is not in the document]
jarosik10: > connect these functions somehow so we are sure that if one changes then we change another one

What exactly do you mean? As I understand there is no other option than to _hardcode_ the `data-analytics-label`.

src/app/content/components/popUp/FiltersList.tsx
```diff
@@ -91,6 +92,7 @@ export const FiltersListChapter = (props: FiltersListChapterProps) => (
         id: 'i18n:highlighting:filters:remove:chapter'},
```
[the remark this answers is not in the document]
jarosik10: Do we want to add a separate key in `messages.json` `i18n:studyguides:filters:remove:chapter` for SG? Even if they will be the same as `i18n:highlighting:filters:remove:chapter`?

src/app/content/components/popUp/ChapterFilter.tsx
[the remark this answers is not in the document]
PiotrKozlowski: I mean that you are reusing `splitTitleParts(section.title).join(' ')` from `getAriaLabel` function
Instead of doing this you could have `getAriaLabel(section)` and `getAnalyticsLabel(section)` and both of them would use `prepareSectionTitle(section) => splitTitleParts(section.title).join(' ')`
However I have second thoughts on this because we don't want to change analytics labels when we change aria labels so it might be better to leave it as it is now

src/app/content/components/popUp/FiltersList.tsx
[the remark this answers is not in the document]
PiotrKozlowski: Yes, I think this is okay

src/app/content/components/popUp/FiltersList.tsx
```diff
@@ -88,9 +89,9 @@ interface FiltersListChapterProps {
...
   <FilterListItem>
     <StyledPlainButton
-      aria-label={useIntl().formatMessage({
-        id: 'i18n:highlighting:filters:remove:chapter'},
-        {filterValue: splitTitleParts(props.title).join(' ')}
+      aria-label={useIntl().formatMessage(
+        { id: props.ariaLabelKey },
+        { filterValue: splitTitleParts(props.title).join(' ') }
       )}
       data-analytics-label={`Remove breadcrumb for chapter ${splitTitleParts(props.title).join(' ')}`}
```
PiotrKozlowski, line 96: I think this analytics label and the second one for `FiltersListColor` should be generated in 
`src/app/content/highlights/components/SummaryPopup/Filters.tsx` and
`src/app/content/studyGuides/components/Filters.tsx `

src/app/content/highlights/components/SummaryPopup/Filters.tsx
```diff
@@ -97,7 +97,8 @@ export default () =>
       <ConnectedPrintButton />
     </FiltersTopBar>
     <ConnectedFilterList
-      colorAriaLabelKey={() => 'i18n:highlighting:filters:remove:color'}
+      colorAriaLabelKey='i18n:highlighting:filters:remove:color'
```
PiotrKozlowski, line 100: this was a function to unify type with `colorLabelKey`
I think you should add here also `analyticsLabel` props that should also accept a function  with `color` parameter and return something based on this

PiotrKozlowski (changes requested):

src/app/content/components/popUp/FiltersList.tsx
```diff
@@ -90,10 +92,10 @@ export const FiltersListChapter = (props: FiltersListChapterProps) => (
   <FilterListItem>
     <StyledPlainButton
       aria-label={useIntl().formatMessage(
-        { id: props.ariaLabelKey },
+        { id: props.ariaLabelKey(props.title) },
         { filterValue: splitTitleParts(props.title).join(' ') }
       )}
-      data-analytics-label={`Remove breadcrumb for chapter ${splitTitleParts(props.title).join(' ')}`}
+      data-analytics-label={props.dataAnalyticsLabel(props.title)}
```
PiotrKozlowski, line 98: We could probably pass here `splitTitleParts(props.title).join(' ')` to keep this logic in one place

src/app/content/studyGuides/components/Filters.tsx
```diff
@@ -14,13 +14,34 @@ import PrintButton from '../../components/popUp/PrintButton';
...
 import { LinkedArchiveTreeNode } from '../../types';
+import { splitTitleParts } from '../../utils/archiveTreeUtils';
 import { printStudyGuides, updateSummaryFilters } from '../actions';
 import { highlightStyles } from '../constants';
 import * as selectors from '../selectors';
 import { cookieUTG } from './UsingThisGuide/constants';
 import UsingThisGuideBanner from './UsingThisGuide/UsingThisGuideBanner';
 import UsingThisGuideButton from './UsingThisGuide/UsingThisGuideButton';
 
+const createColorDataAnalyticsLabel = (color: HighlightColorEnum): string => {
```
PiotrKozlowski, line 25: It would be good to add a comment here explaining that this is created based on current values from `message.json`

PiotrKozlowski (changes requested):
Looks pretty good. I've left two more, really small commentes

PiotrKozlowski (approved):

TomWoodward (approved):

PR #1153, opened 2021-04-20
Make urlUtils to use bookConfigLoader and adjust the rest of the files

jarosik10:
for: https://github.com/openstax/unified/issues/1499

### DIFF
--- a/src/app/content/components/ContentExcerpt.tsx
+++ b/src/app/content/components/ContentExcerpt.tsx
@@ -1,8 +1,9 @@
 import flow from 'lodash/fp/flow';
-import React, { HTMLAttributes } from 'react';
+import React, { HTMLAttributes, useEffect } from 'react';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components/macro';
 import { bodyCopyRegularStyle } from '../../components/Typography';
+import { useServices } from '../../context/Services';
 import { assertDefined } from '../../utils';
 import { book } from '../selectors';
 import { LinkedArchiveTreeSection } from '../types';
@@ -30,22 +31,30 @@ const ContentExcerpt = (props: Props) => {
     forwardedRef,
     ...excerptProps
   } = props;
-
+  const services = useServices();
+  const [excerptSourceUrl, setExcerptSourceUrl] = React.useState('');
   const currentBook = assertDefined(useSelector(book), 'book not loaded');
   const sourcePage = typeof source === 'string'
     ? assertDefined(findArchiveTreeNodeById(currentBook.tree, source), 'page not found in book')
     : source;
 
-  const excerptSource = getBookPageUrlAndParams(
-    currentBook,
-    sourcePage
-  );
+  useEffect(() => {
+    const setBookPageUrlAndParams = async() => {
+      const { url } = await getBookPageUrlAndParams(
+        currentBook,
+        sourcePage,
+        services.bookConfigLoader
+      );
+      setExcerptSourceUrl(url);
+    };
+    setBookPageUrlAndParams();
+  });
 
   const fixedContent = React.useMemo(() => flow(
     addTargetBlankToLinks,
-    (newContent) => rebaseRelativeContentLinks(newContent, excerptSource.url),
-    (newContent) => resolveRelativeResources(newContent, excerptSource.url)
-  )(props.content), [props.content, excerptSource.url]);
+    (newContent) => rebaseRelativeContentLinks(newContent, excerptSourceUrl),
+    (newContent) => resolveRelativeResources(newContent, excerptSourceUrl)
+  )(props.content), [props.content, excerptSourceUrl]);
 
   return <div
     ref={forwardedRef}
--- a/src/app/content/components/ContentLink.tsx
+++ b/src/app/content/components/ContentLink.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import styled from 'styled-components/macro';
 import { linkStyle } from '../../components/Typography';
+import { useServices } from '../../context/Services';
 import { push } from '../../navigation/actions';
 import * as selectNavigation from '../../navigation/selectors';
 import { ScrollTarget } from '../../navigation/types';
@@ -15,7 +16,7 @@ import {
 import { content } from '../routes';
 import * as selectSearch from '../search/selectors';
 import * as select from '../selectors';
-import { Book } from '../types';
+import { Book, Params } from '../types';
 import { getBookPageUrlAndParams, stripIdVersion, toRelativeUrl } from '../utils';
 import { isClickWithModifierKeys } from '../utils/domUtils';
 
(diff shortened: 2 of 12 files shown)


src/app/content/components/ContentExcerpt.tsx
```diff
@@ -30,22 +31,30 @@ const ContentExcerpt = (props: Props) => {
...
   const sourcePage = typeof source === 'string'
     ? assertDefined(findArchiveTreeNodeById(currentBook.tree, source), 'page not found in book')
     : source;
 
-  const excerptSource = getBookPageUrlAndParams(
-    currentBook,
-    sourcePage
-  );
+  useEffect(() => {
+    const setBookPageUrlAndParams = async() => {
```
TomWoodward, line 42: is it not ok for the effect itself to be async?

src/app/content/components/ContentExcerpt.tsx
```diff
@@ -30,22 +31,30 @@ const ContentExcerpt = (props: Props) => {
...
   const fixedContent = React.useMemo(() => flow(
     addTargetBlankToLinks,
-    (newContent) => rebaseRelativeContentLinks(newContent, excerptSource.url),
-    (newContent) => resolveRelativeResources(newContent, excerptSource.url)
-  )(props.content), [props.content, excerptSource.url]);
+    (newContent) => rebaseRelativeContentLinks(newContent, excerptSourceUrl),
+    (newContent) => resolveRelativeResources(newContent, excerptSourceUrl)
+  )(props.content), [props.content, excerptSourceUrl]);
 
   return <div
```
TomWoodward, line 59: should probably not proceed with render until the url loads or the content could be invalid

src/app/content/components/ContentLink.tsx
```diff
@@ -53,7 +54,23 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     hasUnsavedHighlight,
     ...anchorProps
   } = props;
-  const {url, params} = getBookPageUrlAndParams(book, page);
+  const services = useServices();
+  const [url, setUrl] = React.useState('');
+  const [params, setParams] = React.useState<Params>({} as any);
+
+  React.useEffect(() => {
+    const setBookPageUrlAndParams = async() => {
```
TomWoodward, line 62: all the same comments as the other place

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -126,12 +129,14 @@ export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => Co
       }, {hash, search: searchString}));
     } else {
       // defer to allow other handlers to execute before nav happens
+      const bookParams = (await getBookPageUrlAndParams(book, page, bookConfigLoader)).params;
+      const bookState = (await getBookPageUrlAndParams(book, page, bookConfigLoader)).params;
```
TomWoodward, line 133: why are you running the same query twice?

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -126,12 +129,14 @@ export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => Co
...
       defer(() => navigate({
-        params: getBookPageUrlAndParams(book, page).params,
+        params: bookParams,
         route: content,
         state: {
           ...locationState,
-          ...getBookPageUrlAndParams(book, page).state,
+          ...bookState,
```
TomWoodward, line 139: i feel like you're passing `params` instead of state here

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -77,7 +78,16 @@ const HighlightListElement = ({ highlight, locationFilterId, pageId }: Highlight
...
+  const services = useServices();
+  const [linkToHighlight, setLinkToHighlight] = useState('');
+
+  React.useEffect(() => {
+    const getLinkToHighlight = async() => {
+      const highLightLink = await createHighlightLink(services.bookConfigLoader, highlight, book);
+      setLinkToHighlight(highLightLink);
+    };
+    getLinkToHighlight();
+  }, [book, highlight, services.bookConfigLoader]);
```
TomWoodward, line 90: i feel like a `useContentLink(book, page)` hook might be in order, although i see that is not exactly what you're doing here, maybe you could compose it and still have less logic

src/app/content/search/hooks/index.ts
```diff
@@ -74,7 +74,7 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
   const page = currentPage || targetPage;
 
   const navigation = {
-    params: getBookPageUrlAndParams(book, targetPage).params,
+    params: (await getBookPageUrlAndParams(book, targetPage, services.bookConfigLoader)).params,
```
TomWoodward, line 77: this is a situation where it would never be valid for it to be async, because the content has to be loaded before we get this far, so the book must be in the config already

something to keep in mind, maybe we can simplify some logic if we build on that assumption

TomWoodward (changes requested):

src/app/content/components/ContentExcerpt.tsx
jarosik10: Making the effect itself to be async triggers eslint warning: `Effect callbacks are synchronous to prevent race conditions. Put the async function inside.` 
There is an [article](https://www.robinwieruch.de/react-hooks-fetch-data) about it.

src/app/content/components/Page/contentLinkHandler.ts
jarosik10: oops mistake.

src/app/content/search/hooks/index.ts
jarosik10: @TomWoodward 
I'm thinking about creating a separate functions inside `urlUtils` like: `getBookPageUrlAndParams` and async version `cacheSafeGetBookPageUrlAndParams which uses bookConfigLoader`. But it will cause a lot of code reduplication. 
Generally, it's all about making making the book config loading optional.

jarosik10: It would remove changes in `domVisitor` and `urlChecker`

src/app/content/components/ContentLink.tsx
PiotrKozlowski: We could create `useGetBookPageUrlAndParams` hook that will return `{isLoading: boolean, url: string, params: ...}` and reuse it

src/app/content/utils/urlUtils.ts
```diff
@@ -28,11 +30,16 @@ export const getBookPageUrlAndParams = (
...
 
-export const getUrlParamsForBook = (
-  book: Pick<Book, 'id' | 'tree' | 'title' | 'version'> & Partial<{slug: string}>
-): Params['book'] => {
-  if ('slug' in book && book.slug && BOOKS[book.id]) {
-    return book.version === BOOKS[book.id].defaultVersion
+export const getUrlParamsForBook = async(
+  book: Pick<Book, 'id' | 'tree' | 'title' | 'version'> & Partial<{slug: string}>,
+  bookConfigLoader?: AppServices['bookConfigLoader']
+): Promise<Params['book']> => {
```
PiotrKozlowski, line 36: I think there should be a way to declare this function and `getBookPageUrlAndParams` to be synchronous if `bookConfigLoader` is not passed and `async` if it is
This way we could use the same function for both cases

jarosik10: @PiotrKozlowski 
I was thinking about that, but i'm not sure how to make the function to be optional async. Maybe by returning the result in `Promise.resolve` if `bookConfigLoader` is passed? I was experimenting with this approach and it makes the function unintelligible, but maybe it can be improved.

PiotrKozlowski: You can try something like this:
```ts
function a(b: 1): 1;
function a(b: 1, v: boolean): Promise<1>;
function a(b: 1, v?: boolean) {
  if (v) {
    return Promise.resolve(b)
  }
  return b
}

const z = a(1);
const zz = a(1, true);
```

PiotrKozlowski: it is called Function Overloading - theoretically number of arguments should be always the same, but you can experiment with it

jarosik10: Yeah, i know what function overloading is. I thought that there is no function overloading in js, but as I understand _here comes the typescript!_

PR #1156, opened 2021-04-28
Upgrade typescript to ^4.2.0 and fix lint issues

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1469

For adding support for SW I've followed instructions from this comment: https://github.com/facebook/create-react-app/issues/10032#issuecomment-723433217


src/service-worker.ts
```diff
@@ -0,0 +1,122 @@
+/// <reference lib="webworker" />
```
PiotrKozlowski, line 1: Previous service worker content can be found here: http://rex-web.herokuapp.com/books/service-worker.js

script/doctor-workbox.bash
```diff
@@ -3,82 +3,13 @@
...
-echo "remove this when offical solution is merged https://github.com/facebook/create-react-app/pull/5369"
-
 function abs_path {
   (cd "$(dirname '$1')" &>/dev/null && printf "%s/%s" "$PWD" "${1##*/}")
 }
 
 build=$(abs_path "${BASH_SOURCE%/*}/../build")
 worker="$build"/service-worker.js
 
-# add cache behaviors for 100% offline load
```
PiotrKozlowski, line 16: I've moved it to `src/service-worker.ts`

src/app/content/highlights/components/Note.tsx
```diff
@@ -36,7 +36,7 @@ const TextArea = styled.textarea`
 
 // tslint:disable-next-line:variable-name
 const Note = ({onChange, onFocus, note, textareaRef}: Props) => {
-  const setTextAreaHeight = () => {
+  const setTextAreaHeight = React.useCallback(() => {
```
jarosik10, line 39: Why did you make this change?

src/app/index.spec.tsx
```diff
@@ -119,7 +119,7 @@ describe('create app', () => {
         pathname: notFound.getUrl({url: 'url'}),
         replace: jest.fn(),
       };
-      delete window.location;
+      delete (window as any).location;
```
jarosik10, line 122: It's a fix for `The operand of a 'delete' operator must be optional.`. Is it something we want to intentionally ignore?

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "isolatedModules": true,
-    "jsx": "preserve",
+    "jsx": "react",
```
jarosik10, line 9: Why did you change it?

src/service-worker.ts
```diff
@@ -0,0 +1,122 @@
...
+// Precache all of the assets generated by your build process.
+// Their URLs are injected into the manifest variable below.
+// This variable must be present somewhere in your service worker file,
+// even if you decide not to use precaching. See https://cra.link/PWA
+precacheAndRoute(self.__WB_MANIFEST);
+
+// Set up App Shell-style routing, so that all navigation requests
+// are fulfilled with your index.html shell. Learn more at
+// https://developers.google.com/web/fundamentals/architecture/app-shell
+const fileExtensionRegexp = new RegExp('/[^/?]+\\.[^/]+$');
```
jarosik10, line 30: from: http://rex-web.herokuapp.com/books/service-worker.js: `blacklist: [/^\/_/,/\/[^/?]+\.[^/]+$/]`
```suggestion
const fileExtensionRegexp = new RegExp('/[^/?]+\.[^/]+$');
```
it's a difference i've found.

jarosik10 (changes requested):
Did you test if SW works properly?

src/app/utils/assertions.ts
```diff
@@ -27,3 +27,13 @@ export const assertString = <X>(x: X, message: string): string => {
 
   return x;
 };
+
+export const assertNonNullableArray = <X>(arr: X[], message: string) => {
```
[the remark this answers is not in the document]
PiotrKozlowski: `NonNullable` is a utility type:
```ts
/**
 * Exclude null and undefined from T
 */
type NonNullable<T> = T extends null | undefined ? never : T;
```

src/app/content/highlights/components/Note.tsx
PiotrKozlowski: `react-hooks/exhaustive-deps` error in useEffect below

src/app/index.spec.tsx
PiotrKozlowski: I think so

tsconfig.json
PiotrKozlowski: that was automatically changed after running updated react-scripts

src/service-worker.ts
PiotrKozlowski: I think both of these regexp were automatically generated do I'm not sure if we should change that cc @TomWoodward

jarosik10 (approved):

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -88,7 +88,7 @@ export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => Co
       return;
     }
 
-    const base = new URL(assertWindow().location);
+    const base = new URL(assertWindow().location.href);
```
TomWoodward, line 91: this seems bad

src/app/content/highlights/components/utils/onClickOutside.ts
```diff
@@ -46,6 +46,7 @@ export const useOnClickOutside = (
   cb: (e: MouseEvent) => void,
   eventOptions?: EventListenerOptions
 ) => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
TomWoodward, line 49: its necessary to put this just because `onClickOutside` is a separate function? or is it because of eventOptions?

src/app/content/search/components/SearchResultsSidebar/index.spec.tsx
```diff
@@ -195,7 +195,7 @@ describe('SearchResultsSidebar', () => {
 
     jest.useFakeTimers();
     if (sidebar.searchSidebar.current) {
-      sidebar.searchSidebar.current.dispatchEvent(animationEvent());
+      sidebar.searchSidebar.current.dispatchEvent(animationEvent() as any);
```
TomWoodward, line 198: why is the any needed for this?

src/app/content/utils/urlUtils.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('getUrlParamForPageId', () => {
   });
 
   it('returns uuid param if tree is missing a slug', () => {
-    delete book.tree.contents[0].slug;
+    delete (book as any).tree.contents[0].slug;
```
TomWoodward, line 73: why is this necessary?

src/app/developer/components/Home.spec.tsx
```diff
@@ -16,7 +16,7 @@ describe('Home', () => {
 
   it('matches snapshot', async() => {
     jest.spyOn(Date.prototype, 'getFullYear').mockReturnValue(2021);
-    const store = createTestStore({navigation: new URL('https://localhost')});
+    const store = createTestStore({navigation: new URL('https://localhost') as any});
```
TomWoodward, line 19: why is this necessary?

src/app/index.tsx
```diff
@@ -32,15 +33,15 @@ export const actions = {
...
+export const routes = assertNonNullableArray(Object.values({
   ...(
     process.env.REACT_APP_ENV !== 'production'
       ? developer.routes
       : /* istanbul ignore next */ {}
   ),
   ...content.routes,
   ...errors.routes,
-});
+}), 'some of the required routes are not exported');
```
TomWoodward, line 44: how is it possible that these values would be null?

src/app/reactUtils.ts
```diff
@@ -51,10 +51,12 @@ export const onFocusInOrOutHandler = (
 };
 
 export const useFocusLost = (ref: React.RefObject<HTMLElement>, isEnabled: boolean, cb: () => void) => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   React.useEffect(onFocusInOrOutHandler(ref, isEnabled, cb, 'focusout'), [ref, isEnabled]);
```
TomWoodward, line 55: is this because of the callback?

src/app/reactUtils.ts
```diff
@@ -166,7 +170,7 @@ export const useMatchMobileQuery = () => {
 
   React.useEffect(() => {
     matchMedia.addListener(listener);
-    return () => { matchMedia.removeListener(listener); };
+    return () => { matchMedia.removeListener(listener as any); };
```
TomWoodward, line 173: why is the any necessary?

src/helpers/loadFont.ts
```diff
@@ -5,7 +5,7 @@ export default (url: string) => new Promise((resolve) => {
   }
 
   if (document.head.querySelector(`link[href="${url}"]`)) {
-    resolve();
+    resolve({});
```
TomWoodward, line 8: why?

src/service-worker.ts
TomWoodward: the regexes look the same, this one is just a string so it needs the slash escaped

src/service-worker.ts
```diff
@@ -0,0 +1,122 @@
...
+    plugins: [
+      new ExpirationPlugin({
+        maxEntries: 20,
+      }),
+    ],
+  })
+);
+
+registerRoute(
+  /\/contents|resources|extras\//,
```
TomWoodward, line 101: incidentally, i forgot to check if this still works on the rap urls, because they have a prefix on them. does it?

src/serviceWorker.ts
```diff
@@ -44,7 +44,7 @@ export function register(): Promise<ServiceWorkerRegistration> {
 
   // The URL constructor is available in all browsers that support SW.
   const publicUrl = new URL(
-    process.env.PUBLIC_URL,
+    process.env.PUBLIC_URL!,
```
TomWoodward, line 47: can you handle this null by nooping or throwing or something before this

craco.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+    style: {
+        css: {
+            loaderOptions: {
+                // https://github.com/openstax/unified/issues/1469
+                url: false,
```
TomWoodward, line 6: 👍

package.json
```diff
@@ -36,10 +37,22 @@
...
+    "workbox-cacheable-response": "^5.1.3",
+    "workbox-core": "^5.1.3",
+    "workbox-expiration": "^5.1.3",
+    "workbox-google-analytics": "^5.1.3",
+    "workbox-navigation-preload": "^5.1.3",
+    "workbox-precaching": "^5.1.3",
+    "workbox-range-requests": "^5.1.3",
+    "workbox-routing": "^5.1.3",
+    "workbox-strategies": "^5.1.3",
+    "workbox-streams": "^5.1.3"
```
TomWoodward, line 55: i feel like some of these workbox plugins aren't used, are they?

script/doctor-workbox.bash
```diff
@@ -3,82 +3,15 @@
 # sed expression needs $s
```
TomWoodward, line 3: why is this script needed at all? i thought new react-scripts handles it correctly now?

TomWoodward (changes requested):

src/service-worker.ts
PiotrKozlowski: yes, it does - I've checked on Firefox and Chrome

src/app/content/components/Page/contentLinkHandler.ts
PiotrKozlowski: the `URL` accepts `string` so without it there is an error `Argument of type 'Location' is not assignable to parameter of type 'string'.ts(2345)`
`location.href` contains the whole URL so I think it is okay, because we are reseting `href` and `search` below anyway

src/app/content/highlights/components/utils/onClickOutside.ts
PiotrKozlowski: Yes, it is because of `onClickOutside`
This error is showing up for all useEffects that are using this structure:
```
React Hook useEffect received a function whose dependencies are unknown. Pass an inline function instead.
```

src/app/content/search/components/SearchResultsSidebar/index.spec.tsx
PiotrKozlowski: It is because `animationEvent` is returning `CustomEvent` instead of `Event`. I'll update the function instead using `any`
```
Argument of type 'CustomEvent<unknown>' is not assignable to parameter of type 'Event'.
  Types of property 'srcElement' are incompatible.
    Type 'EventTarget | null' is not assignable to type 'Element | null'.
          Type 'EventTarget' is missing the following properties from type 'Element': assignedSlot, attributes, classList, className, and 119 more.ts(2345)
    ```

File: src/app/content/utils/urlUtils.spec.ts (hunk 11 above)
Comment (PiotrKozlowski) [reply src/app/content/utils/urlUtils.spec.ts:73]: `slug` isn't declared as optional so you can't `delete` it

File: src/app/developer/components/Home.spec.tsx (hunk 12 above)
Comment (PiotrKozlowski) [reply src/app/developer/components/Home.spec.tsx:19]:
```
Type 'URL' is not assignable to type 'State'.
  Property 'state' is missing in type 'URL' but required in type 'Location<any>'.ts(2322)
  ```
  I guess this didn't change recently but now it is an error so it should be okay to just use `any`

src/app/index.tsx
PiotrKozlowski: it more about ts not being able to recognize that we are exporting `routes` from each of these routes so when we want to use it it throws:
```ts
const routes: (navigation.types.Route<{}, {}> | navigation.types.Route<content.types.Params, State> | navigation.types.Route<Params, {}> | undefined)[]
```
```
Argument of type '(Route<{}, {}> | Route<Params, State> | Route<Params, {}> | undefined)[]' is not assignable to parameter of type 'AnyRoute[]'.
  Type 'Route<{}, {}> | Route<Params, State> | Route<Params, {}> | undefined' is not assignable to type 'AnyRoute'.
    Type 'undefined' is not assignable to type 'AnyRoute'.ts(2345)
```
I've used `assertNonNullableArray` name because I've also used `NonNullable` for `export type AnyRoute = NonNullable<typeof routes[number]>;`

src/app/reactUtils.ts
PiotrKozlowski: This is because `onFocusInOrOutHandler` is a separate function

src/app/reactUtils.ts
PiotrKozlowski: `addListener` and `removeListener` have different signatures, however, it shouldn't matter because both of these methods are deprecated and we should change them when we update `@openstax/types`

```
Argument of type '(e: MediaQueryListEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject | null'.
  Type '(e: MediaQueryListEvent) => void' is not assignable to type 'EventListener'.
    Types of parameters 'e' and 'evt' are incompatible.
      Type 'Event' is missing the following properties from type 'MediaQueryListEvent': matches, mediats(2345)
```

src/helpers/loadFont.ts
PiotrKozlowski: `resolve` expects a `value: unknown` passed unless we declare `Promise<void>` but if we do so then TS will complain about  `link.onLoad = resolve`:
```
Type '(value: void | PromiseLike<void>) => void' is not assignable to type '(this: GlobalEventHandlers, ev: Event) => any'.
  Types of parameters 'value' and 'ev' are incompatible.
    Type 'Event' is not assignable to type 'void | PromiseLike<void>'.
      Property 'then' is missing in type 'Event' but required in type 'PromiseLike<void>'.ts(2322)
 ```
 
 Passing anything to `resolve(...)` helps, but I see we can also change mentioned line to `link.onload = () => resolve();`

File: package.json (hunk 20 above)
Comment (PiotrKozlowski) [reply package.json]: Indeed it seems like this. I've copied all of them from the template created with `yarn create react-app my-app --template cra-template-pwa-typescript` and they are not used there either 
I've removed them for now and we can check if they work

File: script/doctor-workbox.bash (hunk 21 above)
Comment (PiotrKozlowski) [reply script/doctor-workbox.bash:3]: I'm using it to move `service-worker.js` to the `books` folder in the `build`
I could try to find another way of doing it, maybe using `craco` for this

File: package.json (hunk 20 above)
Comment (PiotrKozlowski) [reply package.json]: After removing them service worker is not being registered. I'll revert this change

Comment (PiotrKozlowski) [reply package.json:55]: Oh, it is actually not working after reverting. I think it may be related to the change about PUBLIC_URL:
```
if (!process.env.PUBLIC_URL) {
    return Promise.reject(
      new Error('service worker won\'t work if PUBLIC_URL is not defined.')
    );
  }
```

Comment (PiotrKozlowski) [reply package.json]: I've ended up changing the `if` statement to `if (process.env.PUBLIC_URL === undefined) {` and removed these workbox plugins

Btw. `process.env.PUBLIC_URL` is set to an empty string on `rex-web.herokuapp.com`

File: script/doctor-workbox.bash (hunk 21 above)
Comment (PiotrKozlowski) [reply script/doctor-workbox.bash]: I've changed this with craco

File: src/app/index.tsx (hunk 13 above)
Comment (TomWoodward) [reply src/app/index.tsx:44]: what is actually causing that undefined to be in the type? is it one of the modules in particular that has it?

File: src/app/reactUtils.ts (hunk 14 above)
Comment (TomWoodward) [reply src/app/reactUtils.ts:55]: it should be correct to add that to deps, if it is re-defined it might have new scope that is necessary, you can wrap it in React.useCallback to reduce reloading

Comment (TomWoodward) [reply src/app/reactUtils.ts:55]: or if you really want to ignore its scope changes you could wrap it in a ref

Comment (PiotrKozlowski) [reply src/app/reactUtils.ts:55]: I can add `cb` to dependencies but the issue will still exist. It is the same as for `onClickOutside`:
https://github.com/openstax/rex-web/pull/1156#discussion_r625741872

File: src/app/reactUtils.ts (hunk 15 above)
Comment (PiotrKozlowski) [reply src/app/reactUtils.ts:173]: @TomWoodward Yes, please review, merge and deploy https://github.com/openstax/typescript-types/pull/4, then I can cherry pick changes from this PR https://github.com/openstax/rex-web/pull/1167 and fix it

Comment (TomWoodward) [reply src/app/reactUtils.ts:173]: i approved and merged the one in the types repo

File: src/app/index.tsx (hunk 13 above)
Comment (TomWoodward) [reply src/app/index.tsx:44]: is it because of the ternary? that is very odd

it would be safe to cast the ternary result to `typeof developer.routes` if that helps

File: src/app/developer/components/Home.spec.tsx (hunk 12 above)
Comment (TomWoodward) [reply src/app/developer/components/Home.spec.tsx]: if url legitimately doesn't have a field that state requires, it would probably be better to add that field by doing
```
{...new URL('https://localhost'), state: {}}
```
or whatever

File: src/app/reactUtils.ts (hunk 15 above)
Comment (PiotrKozlowski) [reply src/app/reactUtils.ts:173]: You still need to deploy it right? https://www.npmjs.com/package/@openstax/types

File: src/app/head/hooks/locationChange.spec.ts [hunk 22]
```diff
@@ -6,7 +6,7 @@ import { assertNotNull } from '../../utils';
 describe('setHead hook', () => {
   let hookBody: ActionHookBody<typeof locationChange>;
   const helpers = {} as MiddlewareAPI & AppServices;
-  const action = locationChange({location: new URL('http://localhost/'), action: 'PUSH'});
+  const action = locationChange({location: new URL('http://localhost/') as any, action: 'PUSH'});
```
TomWoodward, line 9: same deal with the location/url situation

src/app/navigation/types.ts
```diff
@@ -59,7 +59,7 @@ export interface LocationChange<M = AnyMatch> {
   action: Action;
 }
 
-export type AnyRoute = typeof routes[number];
+export type AnyRoute = NonNullable<typeof routes[number]>;
```
TomWoodward, line 62: this isn't necessary anymore right?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1163, opened 2021-04-30
refactor MessageProvider to use intl provided from appServices

bethshook:
for: openstax/unified#1554


src/app/index.tsx
```diff
@@ -107,15 +108,15 @@ export default (options: AppOptions) => {
     reducer,
   });
 
-  const container = () => (
+  const container: React.FC<{}> = () => (
```
TomWoodward, line 111: it shouldn't really need the type at all on the declaration, but in any case i'm pretty sure the empty props is the default
```suggestion
  const container: React.FC = () => (
```

TomWoodward (changes requested):
i haven't looked at the individual tests yet but the helpers look good

TomWoodward:
unit test CI is failign

PiotrKozlowski (commented):
The individual unit test changes look good. I'm wondering why the test for `Attribution.spec.tsx` is not passing

TomWoodward (changes requested):
the attribution spec issue is because of the `scroll-to-element` mock. something about the changes is probably causing domUtils to be included in the first beforeEach, it is then cached and the mock set up in the second beforeEach is not used. all the logic in the test is actually happening fine

src/app/content/components/Attribution.spec.tsx
```diff
@@ -2,29 +2,23 @@ import cloneDeep from 'lodash/fp/cloneDeep';
...
   beforeEach(() => {
     jest.resetAllMocks();
     resetModules();
-    React = require('react');
-    ReactDOM = require('react-dom');
-    Provider = require('react-redux').Provider;
-    renderer = require('react-test-renderer');
-    renderToDom = require('../../../test/reactutils').renderToDom;
-    MessageProvider = require('../../MessageProvider').default;
+    ({React, ReactDOM, renderer, renderToDom, TestContainer} = reactAndFriends());
```
bethshook, line 21: @TomWoodward based on some reading I thought something like this in the second `beforeEach` would work but it's having no effect:
```
    const mockScrollTo = scrollTo = jest.fn();
      jest.isolateModules(() => {
        jest.doMock('scroll-to-element', () => mockScrollTo);
        scrollToElem = require('scroll-to-element').default;
      });
```

bethshook: https://stackoverflow.com/questions/45571816/reset-single-module-with-jest

TomWoodward: the issue has to do with the files you're testing, it might help to put `isolateModules` around where you are requiring `Attribution` but putting it around the mock will not impact the stuff required by the code you're testing.

when i was looking at this i just moved the `scroll-to-element` mock up the logic until it started working, which would be an acceptable fix even though it would put it in scope of more tests, whatever.

TomWoodward (approved):

PR #1164, opened 2021-04-30
Improve meta descriptions

bethshook:
for: openstax/unified#969


src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +12,82 @@ import {
...
+  return str.replace(/\n+/g, ' ').replace(/\s+/g, ' ').trim().substring(0, 152) + '...';
+};
+
+const getPageDescriptionFromContent = (page: HTMLElement): string | null => {
+  if (!page) {
+    return null;
+  }
+  removeExcludedContent(page);
+  // tslint:disable-next-line: max-line-length
+  const paragraphs = [...Array.from(page.querySelectorAll(':scope>section>p')), ...Array.from(page.querySelectorAll(':scope>p'))];
```
bethshook, line 72: @TomWoodward I had to revert to using `:scope` because the `div:first-child>p` selector was finding direct-descendent paragraphs of _all_ first-child divs on the page. But based on what you said about page content IDs I think this is ok.

bethshook: I lied, my metadata dump script complains about invalid IDs when I run it : (

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +14,101 @@ import {
...
+  const cleanContent = getCleanContent(book, page, loader);
+  const doc = domParser.parseFromString(cleanContent, 'text/html');
+  const treeNode = findArchiveTreeNodeById(book.tree, page.id);
+  if (!treeNode) {
+    return '';
+  }
+  const contentDescription: string | null = getPageDescriptionFromContent(doc.body.firstElementChild);
 
+  return contentDescription || 'OpenStax is a non-profit organization committed to improving student access to quality learning materials. Our free textbooks are developed and peer-reviewed by educators to ensure they are readable and accurate.';
+  // return contentDescription || useIntl().formatMessage({id: 'i18n:metadata:description'});
```
PiotrKozlowski, line 111: `useIntl` is a hook so this will not work
This probably waits for the other or to use `intl` from `services`?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +14,101 @@ import {
...
   if (!node) {
     return '';
   }
 
   if (archiveTreeSectionIsChapter(node)) {
-    const number = getArchiveTreeSectionNumber(node);
-    return `Ch. ${number} `;
+    const number = getArchiveTreeSectionNumber(node).trim();
+    const name = getArchiveTreeSectionTitle(node);
+    return includeTitle ? `Ch. ${number}. ${name}` : `Ch. ${number} `;
```
PiotrKozlowski, line 27: This fragment should be also internationalized

src/app/content/utils/seoUtils.ts
```diff
@@ -9,20 +14,101 @@ import {
...
+
+const removeExcludedContent = (node: Element) => {
+  if (!node) {
+    return null;
+  }
+  const excludedContent = node.querySelectorAll(
+    '[data-type="abstract"], .learning-objectives, .chapter-objectives, .be-prepared, .os-teacher'
+  ) || [];
+
+  for (const item of Array.from(excludedContent)) {
```
PiotrKozlowski, line 57: `Array.from` is not needed

src/test/mocks/archiveLoader.ts
```diff
@@ -83,11 +84,15 @@ export default () => {
     },
     mockPage: (parentBook: ArchiveBook, newPage: ArchivePage, pageSlug: string) => {
       localBookPages[`${parentBook.id}@${parentBook.version}`][newPage.id] = newPage;
-      localBooks[`${parentBook.id}@${parentBook.version}`].tree.contents.push({
-        id: `${newPage.id}@${newPage.version}`,
-        slug: pageSlug,
-        title: newPage.title,
-      });
+      const currentBook = localBooks[`${parentBook.id}@${parentBook.version}`];
+      const treeNode = findArchiveTreeNodeById(parentBook.tree, newPage.id);
```
PiotrKozlowski, line 88: shouldn't we check if `newPage.id` is in `currentBook.tree` instead?

PiotrKozlowski (changes requested):

src/app/content/utils/seoUtils.ts
TomWoodward: yea i got ahead of myself thinking that the intl changes would not be relevant here, we do need the other PR merged so we can access intl via services for this

bethshook: Right, I'm just waiting for 1554

src/app/content/utils/seoUtils.ts
bethshook: Without it I get a TS error: `Type 'NodeListOf<Element>' must have a '[Symbol.iterator]()' method that returns an iterator.`

TomWoodward: thats weird.... mdn is pretty sure `for...of` loops should work for nodelists

PiotrKozlowski: I assume it may be related to ts version

src/app/content/highlights/components/Confirmation.spec.tsx
```diff
@@ -14,28 +20,54 @@ jest.mock('react', () => {
...
+  let services: ReturnType<typeof createTestServices>;
+
+  const intl = createIntl({
+    locale: 'en',
+    messages: {
+      ...en,
+      'i18n:confirm': 'confirm',
+      'i18n:message': 'message',
+    },
+  }, cache);
```
bethshook, line 34: @TomWoodward this was the offending file re "missing message" errors. Is this the fix you had in mind?

TomWoodward: sort of, although https://formatjs.io/docs/intl#definemessagesdefinemessage is also a thing that i was imagining using instead of replacing/duplicating the initialization

TomWoodward (changes requested):

src/app/content/highlights/components/Confirmation.spec.tsx
bethshook: I don't understand how that function helps—wouldn't I still need to pass the new messages to a provider so `<FormattedMessage>` can locate the ID?

src/app/content/utils/seoUtils.ts
bethshook: Most solutions I've found recommend setting es6 as the "target" or as a "lib" in tsconfig, but so far that's not working for me. Fwiw it seems like `Array.from` is used across the project where we iterate over or destructure nodelists.

src/app/messages/en/messages.json
```diff
@@ -64,6 +64,9 @@
   "i18n:highlighting:filters:remove:color": "Remove breadcrumb for color {filterValue}",
   "i18n:highlighting:login:link": "Log in",
   "i18n:highlighting:login:prompt": "Log in to highlight and take notes. It’s 100% free.",
+  "i18n:metadata:description": "OpenStax is a non-profit organization committed to improving student access to quality learning materials. Our free textbooks are developed and peer-reviewed by educators to ensure they are readable and accurate.",
+  "i18n:metadata:title:with-name": "Ch. {number}. {name}",
+  "i18n:metadata:title:without-name": "Ch. {number}",
```
bethshook, line 69: @PiotrKozlowski does this look ok?

script/dump-book-metadata.ts
```diff
@@ -0,0 +1,49 @@
...
+const getBookMetadata = async(id: string, version: string) => {
+  const loader = archiveLoader.book(id, version);
+  const singleBook = await loader.load();
+  const osWebBook = singleBook.tree.slug ? await osWebLoader.getBookFromSlug(singleBook.tree.slug) : undefined;
+  const book = formatBookData(singleBook, osWebBook);
+
+  const bookPages = findTreePages(book.tree);
+  for (const page of bookPages) {
+    getPageMetadata(page, book, loader);
+  }
```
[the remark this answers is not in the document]
bethshook: but then the rows won't be logged until all promises are resolved, no?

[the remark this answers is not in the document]
PiotrKozlowski: logs should be still called in the same way but `getBookMetadata` would now return a promise that would resolve only after all pages are processed so you would process one book at a time and don't mix the logs

I think that scripts run from the terminal will wait for all promises anyway so if you don't care about mixing longs it shouldn't be a problem in this case

In other places calling `await getBookMetadata(...)` would just start bunch of promises for all pages and return without waiting for processing them

src/app/content/utils/seoUtils.ts
PiotrKozlowski: Yea, I think it is not a big deal if it is not working by default yet

src/app/messages/en/messages.json
PiotrKozlowski: yes, thanks

src/app/content/highlights/components/Confirmation.spec.tsx
bethshook: @TomWoodward don't think this is a blocker for the PR but I'm still not clear on how to implement this in jest

TomWoodward (changes requested):

script/dump-book-metadata.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const getBookMetadata = async(id: string, version: string) => {
+  const loader = archiveLoader.book(id, version);
+  const singleBook = await loader.load();
+  const osWebBook = singleBook.tree.slug ? await osWebLoader.getBookFromSlug(singleBook.tree.slug) : undefined;
+  const book = formatBookData(singleBook, osWebBook);
+  const bookPages = findTreePages(book.tree);
+
+  const bookRows = await Promise.all(bookPages.map(async(page) => {
+    getPageMetadata(page, book, loader);
```
PiotrKozlowski, line 44: you need a `return` here

src/app/content/utils/seoUtils.ts
```diff
@@ -9,23 +14,125 @@ import {
...
     return '';
   }
 
   if (archiveTreeSectionIsChapter(node)) {
-    const number = getArchiveTreeSectionNumber(node);
-    return `Ch. ${number} `;
+    const number = getArchiveTreeSectionNumber(node).trim();
+    const name = getArchiveTreeSectionTitle(node);
+    const prefixVariant = includeTitle ? 'with-name' : 'without-name';
+    return intl.formatMessage({id: `i18n:metadata:title:${prefixVariant}`}, {number, name});
```
PiotrKozlowski, line 33: The difference is that we will be able to write a script that will search through the keys and find those that are not used in application
```suggestion
    const i18nTitle = includeTitle ? 'i18n:metadata:title:with-name' : 'i18n:metadata:title:without-name';
    return intl.formatMessage({id: i18nTitle}, {number, name});
```

src/app/content/utils/seoUtils.spec.ts
```diff
@@ -1,7 +1,72 @@
...
+  eocPage,
+  mockBook,
+  mockOsWebBook,
+} from './seoUtils.spec.data';
+
+// tslint:disable: max-line-length
+describe('getDescription', () => {
+  const loader = makeArchiveLoader();
+  const book = formatBookData(mockBook, mockOsWebBook);
+  const services = {
```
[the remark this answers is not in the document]
bethshook: see my comment on receiveContent.ts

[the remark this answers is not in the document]
TomWoodward: you mean the one you're getting from `createTestServices`?

src/app/content/utils/seoUtils.ts
```diff
@@ -9,23 +13,125 @@ import {
...
+  const {intl, archiveLoader} = services;
+  const cleanContent = getCleanContent(book, page, archiveLoader);
+  const doc = domParser.parseFromString(cleanContent, 'text/html');
+  const treeNode = findArchiveTreeNodeById(book.tree, page.id);
+  if (!treeNode) {
+    return '';
+  }
+  const pageNode = doc.body.firstElementChild;
+  const isAppendix = pageNode.classList.contains('appendix');
+  const contentDescription: string | null = isAppendix ? null : getPageDescriptionFromContent(pageNode);
```
TomWoodward, line 129: all the other filtering on content type is happening in `getParagraphs`, i feel like this should be there too. maybe change one of your selectors to add a `:not('.appendix')` on the page node?

src/app/content/highlights/components/Confirmation.spec.tsx
TomWoodward: i thought defineMessage could be called on `intl` to add more messages, it doesn't work like that i was wrong. but couldn't you overwrite the intl in the test services with the one you're creating and still use `TestContainer` by passing services in?

TomWoodward (changes requested):

src/app/content/utils/seoUtils.spec.ts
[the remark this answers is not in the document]
bethshook: right, if I just pass in `createTestServices()` to `getPageDescription` then the loader doesn't have the cached pages. or am I missing something?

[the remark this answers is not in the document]
TomWoodward: can't you do 
```
const services = createTestServices();
services.archiveLoader.mockBook(book)
...
it('asdf', () => {
  services.archiveLoader.mockPage(...);
});
```
?

TomWoodward (changes requested):

src/app/content/utils/seoUtils.spec.ts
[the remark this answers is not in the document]
bethshook: if my brain were working, yes, that would be a way

TomWoodward (approved):

PiotrKozlowski (approved):

TomWoodward:
@bethshook lint and coverage are failing

PR #1168, opened 2021-05-05
Throw at the end of update-content script

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1525


concourse/update-content/script.bash
```diff
@@ -55,3 +67,13 @@ for book_and_version in $book_entries; do
     }
 JSON
```
TomWoodward, line 68: i'm not used to using `trap`, i think that this is working because you can only have one trap per signal, so you're overwriting it on each loop, but it will persist after the loop, maybe you should call `trap -` after the loop to deregister the last trap? maybe better to actually not call `set -e`

TomWoodward (changes requested):

concourse/update-content/script.bash
PiotrKozlowski: I've pushed the change about `trap`
> maybe better to actually not call set -e

Do you mean not to call it at the top of the file or after the loop?

TomWoodward: i mean after the loop

PiotrKozlowski: `set -e` is supposed to return to the initial state from line 12: `set -ex` after being changed in line 30: `set +e`
What I meant is: Disable exiting when an error occurs in a loop because I'll handle it with a `trap` and then reenable initial settings
Does it make sense? My bash knowledge is very little

TomWoodward: oh i see in your initial set you're _disabling_ errors. yea i'd like to see `set -e` at the top of the file, actually i usually put
```
 set -e; if [ -n "$DEBUG" ]; then set -x; fi
```
at the top of my scripts

then you could even do everything on the same lines so that its more clear:
```
set +e; trap "report_error; continue" ERR

...

set -e; trap - ERR
```

right?

TomWoodward (changes requested):

concourse/update-content/script.bash
PiotrKozlowski: looks better

TomWoodward (approved):

PR #1170, opened 2021-05-06
Handle redirects in book scan

jarosik10:
for: https://github.com/openstax/unified/issues/1556


script/utils/bookUtils.ts
```diff
@@ -1,21 +1,22 @@
...
   bookVersion,
+  osWebLoader,
+  rootUrl,
 }: {
-  rootUrl: string,
-  archiveUrl?: string,
+  archiveLoader: AppServices['archiveLoader'],
   bookId?: string,
   bookVersion?: string,
+  osWebLoader: AppServices['osWebLoader']
```
PiotrKozlowski, line 18: loaders could be next to each other and rootUrl may be still first parameter
they don't have to be in alphabetic order and it will create less diff

script/domVisitor.ts
```diff
@@ -73,7 +80,8 @@ async function visitPages(
       const appendQueryString =
         queryString ? (archiveUrl ? `?archive=${archiveUrl}&${queryString}` : `?${queryString}`)
                     : archiveUrl ? `?archive=${archiveUrl}` : '';
-      await page.goto(`${rootUrl}${pageUrl}${appendQueryString}`);
+      const pageRedirection = checkIfPageHasRedirect(redirects, pageUrl);
+      await page.goto(`${rootUrl}${pageRedirection ? pageRedirection : pageUrl}${appendQueryString}`);
```
PiotrKozlowski, line 84: You could change the `checkIfPageHasRedirect` a little bit and do something like this:
```suggestion
      const validatedUrl = getPageUrlOrRedirection(redirects, pageUrl);
      await page.goto(`${rootUrl}${validatedUrl}${appendQueryString}`);
```

jarosik10:
@TomWoodward @PiotrKozlowski 
To be honest I'm not sure how test properly.

script/urlChecker.ts
```diff
@@ -38,15 +44,23 @@ async function checkPages(bookSlug: string, pages: string[]) {
...
+      try {
+        const response = await fetch(url);
+        return response.status === 404;
+      } catch {
+        anyFailures = true;
+        bar.interrupt(`- (error loading) ${page}`);
+      }
+    };
+
+    if (checkIfPageDoesntExist(`${rootUrl}${page}`)) {
```
PiotrKozlowski, line 57: I think there is no need to first go to the page and then go to the redirected page in case of 404
You should do the same as in `domVisitior` where you first check for the redirected url

PiotrKozlowski (changes requested):

script/utils/getPageUrlOrRedirection.ts
```diff
@@ -0,0 +1,7 @@
+import { Redirects } from '../../data/redirects/types';
+
+const getPageUrlOrRedirection = (redirects: Redirects, page: string) => {
+    return redirects.find(({ from }) => from === page) || page;
```
PiotrKozlowski, line 4: it is returning `Redirect | string`, but should always return `string`
Also, I think you should compare `page` to `redirect.to` and return `redirect.from` in order to check if it is working because it is checking on `openstax.org` that doesn't have a new version yet

PiotrKozlowski (changes requested):
You can test it by merging https://github.com/openstax/rex-web/pull/1165/ to your branch and running `validate-modified-books.bash` script.

You can see that it is failing on one page: https://github.com/openstax/rex-web/pull/1165/checks?check_run_id=2481010263

You'll need to update `getPageUrlOrRedirection.ts`

script/urlChecker.ts
```diff
@@ -38,15 +44,17 @@ async function checkPages(bookSlug: string, pages: string[]) {
   const notFound: string[] = [];
 
   const visitPage = async(page: string) => {
+    const validatedUrl = getPageOrRedirectedUrl(redirects, page);
```
TomWoodward, line 47: this is gonna return _one of_ the redirects targeting the current page in no particular order, i think we need to check all of them

script/urlChecker.ts
```diff
@@ -38,15 +44,17 @@ async function checkPages(bookSlug: string, pages: string[]) {
...
     try {
-      const response = await fetch(`${rootUrl}${page}`);
+      const response = await fetch(`${rootUrl}${validatedUrl}`);
```
TomWoodward, line 49: the criteria should be
1) if the page url returns a 200 on prod, 👍 
2) if the page url is a redirect target, one of the redirects returns a 200 on prod, 👍

script/urlChecker.ts
```diff
@@ -77,19 +85,23 @@ const getUrl = (book: Book) => useUnversionedUrls
...
   const url = assertDefined(rootUrl, 'please define a rootUrl parameter, format: http://host:port');
   const books = await findBooks({
-    archiveUrl,
+    archiveLoader,
     bookId,
     bookVersion,
+    osWebLoader,
     rootUrl: url,
   });
+  const redirects = await prepareRedirects(archiveLoader, osWebLoader);
```
TomWoodward, line 98: idk if there is any point in running this

TomWoodward (changes requested):

script/urlChecker.ts
jarosik10: @TomWoodward 
What do you mean? Do I have access to `rex/redirects.json` during CI?

TomWoodward: `prepareRedirects` doesn't add any information you don't already have, it just renders the book id and page id into a url, you're comparing the pages to urls and comparing the urls, you could just compare the ids

jarosik10:
There must be a issue with `status === 200`?

script/urlChecker.ts
```diff
@@ -66,6 +91,23 @@ async function checkPages(bookSlug: string, pages: string[]) {
...
+  const fileName = path.resolve(__dirname, `../data/redirects/${bookID}.json`);
+  if (!fs.existsSync(fileName)) { return; }
+
+  const bookRedirects: RedirectsData = await import(fileName);
+  const redirectedPages: Array<{pageId: string, redirectedURL: string}> = [];
+
+  for (const { pageId, pathname } of bookRedirects) {
+    redirectedPages.push({
+      pageId,
+      redirectedURL: pathname,
```
TomWoodward, line 104: what is the point of reformatting this array?

TomWoodward (changes requested):

script/urlChecker.ts
```diff
@@ -74,6 +91,23 @@ async function checkPages(bookSlug: string, pages: string[], redirects: Redirect
   return anyFailures || notFound.length > 0;
 }
 
+const loadRedirectedPagesForBookId = async(bookID: string) => {
+  const fileName = path.resolve(__dirname, `../data/redirects/${bookID}.json`);
+  if (!fs.existsSync(fileName)) { return; }
```
PiotrKozlowski, line 96: you can return an empty array here and you will not need and `if` statement in `visitPage`

PiotrKozlowski (changes requested):

script/urlChecker.ts
```diff
@@ -37,16 +47,30 @@ async function checkPages(bookSlug: string, pages: string[]) {
 
   const notFound: string[] = [];
 
-  const visitPage = async(page: string) => {
+  const validatePage = async(page: LinkedArchiveTreeSection) => {
+    const pageURL = getUrl(book)(page);
+    const redirectedURLs = redirectedPages.filter(({ pageId }) => pageId === page.id)
+      .map(({ pathname }) => pathname);
```
PiotrKozlowski, line 53: You don't have to use `map`, you can destructure below

jarosik10:
Reminder: undo commit https://github.com/openstax/rex-web/pull/1170/commits/fc2f223b61ed1d9e6cbadabff026e6a229a5ada1 before merging.

script/urlChecker.ts
```diff
@@ -37,16 +47,30 @@ async function checkPages(bookSlug: string, pages: string[]) {
 
   const notFound: string[] = [];
 
-  const visitPage = async(page: string) => {
+  const validatePage = async(page: LinkedArchiveTreeSection) => {
+    const pageURL = getUrl(book)(page);
+    const redirectedURLs = redirectedPages.filter(({ pageId }) => pageId === page.id)
```
PiotrKozlowski, line 52: you could change the name of this variable and unshift it with `pageURL` so you can do everything in one for loop

PiotrKozlowski (changes requested):
It looks good. I've left 2 small comments.

We have to remember about removing updates to `/data/redirects/..` and `config.books.json`

script/urlChecker.ts
jarosik10: I'd have to add `pageURL` as `urls.unshift({pathname: pageURL} as any)`. Not sure if it is okay to have such a weird array. Leaving `map` and then adding `unshift` will create array of strings, which seems to be more clear.

PiotrKozlowski: right, in this case, it makes sense to use `map`
I actually wrote these comments in another order :D

script/urlChecker.ts
```diff
@@ -37,16 +47,29 @@ async function checkPages(bookSlug: string, pages: string[]) {
...
+  const validatePage = async(page: LinkedArchiveTreeSection) => {
+    const pageURL = getUrl(book)(page);
+    const urls = [pageURL, ...redirectedPages.filter(({ pageId }) => pageId === page.id)
+      .map(({ pathname }) => pathname)];
+
+    for (const url of urls) {
+      if ((await fetch(`${rootUrl}${url}`)).status === 200) {
+        return true;
+      }
+    }
```
PiotrKozlowski, line 59: you probably want to `return false` after the loop to keep types consistent

PiotrKozlowski (changes requested):
+ https://github.com/openstax/rex-web/pull/1170#issuecomment-839740016

TomWoodward (commented):
looks good, need to remove that debug commit before approval

PiotrKozlowski (approved):

TomWoodward (approved):

PR #1176, opened 2021-05-07
add allowedUrls list for sentry

bethshook:
for: openstax/unified#1553


src/helpers/Sentry.ts
```diff
@@ -38,6 +38,14 @@ export default {
...
         release: normalize(`rex@${config.RELEASE_ID}`),
         tracesSampleRate: 0.1,
+        whitelistUrls: [
+          /https?:\/\/localhost:3000/,
+          /https?:\/\/(www\.)?openstax\.org/,
+          /https?:\/\/dev\.openstax\.org/,
+          /https?:\/\/(.*\.)?sandbox\.openstax\.org/,
+          /https?:\/\/staging\.openstax\.org/,
+          /https?:\/\/rex-web\.herokuapp\.com/,
+        ],
```
bethshook, line 48: @TomWoodward I'm probably missing things...

src/helpers/Sentry.ts
```diff
@@ -38,6 +38,14 @@ export default {
         ],
         release: normalize(`rex@${config.RELEASE_ID}`),
         tracesSampleRate: 0.1,
+        whitelistUrls: [
+          /https?:\/\/localhost:3000/,
+          /https?:\/\/(www\.)?openstax\.org/,
+          /https?:\/\/dev\.openstax\.org/,
+          /https?:\/\/(.*\.)?sandbox\.openstax\.org/,
+          /https?:\/\/staging\.openstax\.org/,
```
TomWoodward, line 46: i'd combine these into _anything_ dot openstax dot org

src/helpers/Sentry.ts
```diff
@@ -38,6 +38,14 @@ export default {
...
+          /https?:\/\/rex-web\.herokuapp\.com/,
```
TomWoodward, line 47: missing the PR environments on heroku

TomWoodward (changes requested):

src/helpers/Sentry.ts
```diff
@@ -27,6 +27,11 @@ export default {
   initializeWithMiddleware(): Middleware {
     return (store) => {
       Sentry.init({
+        allowUrls: [
+          /localhost/,
+          /openstax.org/,
+          /https?:\/\/rex-web(.*\.)?\.herokuapp\.com/,
+        ],
```
bethshook, line 34: @TomWoodward I'm still confused about what this list is doing exactly. Some errors are definitely still reported for URLs not on the list (ie when I remove `localhost`), but it seems like errors that are manually reported via `Sentry.captureException` are filtered out.

Would it be worth first trying some custom event filters [here](https://sentry.io/settings/openstax/projects/rex/fi) before we rely on a whitelist or does that seem too limiting/unpredictable?

src/helpers/Sentry.ts
```diff
@@ -27,6 +27,11 @@ export default {
   initializeWithMiddleware(): Middleware {
     return (store) => {
       Sentry.init({
+        allowUrls: [
+          /localhost/,
+          /openstax.org/,
+          /https?:\/\/rex-web(.*\.)?\.herokuapp\.com/,
```
PiotrKozlowski, line 33:
```suggestion
          /https?:\/\/rex-web(.*)?\.herokuapp\.com/,
```

src/helpers/Sentry.ts
PiotrKozlowski: [It seems like `allowURL` works only for captured events](whitelistUrls)
I think that maybe we want to use [`whitelistUrls`](https://docs.sentry.io/clients/javascript/tips/#decluttering-sentry)?

PiotrKozlowski (changes requested):

src/helpers/Sentry.ts
jarosik10: [docs](https://docs.sentry.io/platforms/javascript/guides/angular/configuration/filtering/#decluttering-sentry):
_Prior to version 5.17.0, allowUrls and denyUrls were called whitelistUrls and blacklistUrls respectively. These options are still supported for backwards compatibility reasons, but they will be removed in version 6.0._ 

What is the reason behind switching to `allowUrls`?

jarosik10: Okay, i see:)

jarosik10 (approved):

TomWoodward (approved):

PiotrKozlowski (approved):

PR #1181, opened 2021-05-14
Search button A/B test

bethshook:
for: openstax/unified#1565


src/app/content/experiments/selectors.ts
```diff
@@ -0,0 +1,13 @@
...
+export const localState = createSelector(
+  parentSelectors.localState,
+  (parentState) => parentState.experiments
+);
+
+// how to create selectors for particular variants
+export const experimentsEnabled = createSelector(
+  localState,
+  (state) => state
+);
```
bethshook, line 13: @TomWoodward on Monday I'd like to talk about the best way to employ selectors to get the feature variant info we need

src/app/content/experiments/reducer.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const reducer: Reducer<State, AnyAction> = (state = initialState, action): any => {
+    switch (action.type) {
+        case getType(receiveFeatureFlags):
+            const [id, variant] = action.payload;
+            if (experimentIds[id]) {
+                const experiment = experimentIds[id];
+                const idx = parseInt(variant, 10);
+                return {...state, [experiment]: idx};
+            }
```
bethshook, line 18: @TomWoodward imagine you will have thoughts on how to better organize this. [here is how the store looks.](https://share.getcloudapp.com/4gu2glWG)

TomWoodward: if you're gonna have it doing processing of the experiment id/mapping stuff, i'd put that behind an action named more specifically just for the experiments, because feature flags could also indicate the release flags and there is no reason for them to go through the same logic

i also think it would be neat for the value to be mapped before going into the state, so for debugging it will have a more human readable value

src/app/content/experiments/selectors.ts
TomWoodward: yea so by design selectors can't take arguments, what you'd want is a selector that just returns all the flags and their values, and then maybe you could have a helper like `getVariantForFeatureFlag` or something that takes the flag name and internally uses the selector

src/app/content/reducer.ts
```diff
@@ -20,6 +21,7 @@ import { State } from './types';
 
 export const initialState = {
   buyPrint: null,
+  experiments: initialExperimentsState,
```
TomWoodward, line 24: i think i want this outside of `content` and maybe we should name it `featureFlags` instead of `experiments`

src/index.tsx
```diff
@@ -101,6 +102,10 @@ window.__APP_ANALYTICS = registerGlobalAnalytics(window, app.store);
 pollUpdates(app.store);
 startMathJax();
 
+// load optimize
+loadOptimize(config.DEPLOYED_ENV === 'server' ? null
+  : (window.location.hostname === 'openstax.org' ? 'OPT-NFHSM4B' : 'OPT-W65B3CP'));
```
TomWoodward, line 107: so in this file we know we're in a browser, pre-rendering doesn't use this file 

i'd do like... `getOptimizeContainerByEnv` or something within your `src/helpers/loadOptimize.ts` file and make this here just a one like function call with no conditions

src/helpers/loadOptimize.ts
```diff
@@ -0,0 +1,15 @@
+export default (containerId: string | null) => new Promise((resolve) => {
+    if (!containerId) {
+        return;
+    }
+
+    if (typeof(document) === 'undefined' || !document.head) {
+      throw new Error('Google Optimize can only be loaded in the browser');
+    }
```
TomWoodward, line 8: there is an `assertDocument` helper that does this for you

TomWoodward (changes requested):

src/app/content/experiments/reducer.ts
bethshook: done. makes sense.

src/index.tsx
bethshook: whoops, yeah, forgot to make the switch when I changed this from `APP_ENV` to `DEPLOYED_ENV`

src/app/content/constants.ts
```diff
@@ -18,3 +19,17 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
...
+  // tslint:disable-next-line: max-line-length
+  Object.keys(primaryColors).forEach((color) => {primaryBases[color] = primaryColors[color].base; });
+  return primaryBases;
+};
+
+export const searchButtonStyles: {[key: string]: string} = {
+  gray: '#5D6062',
+  transparent: 'transparent',
+  ...getPrimaryBaseColors(theme.color.primary) as {},
+};
```
bethshook, line 35: @TomWoodward this is along the lines of how the `<BookBanner />` text color is assigned except I am merging theme colors with the new button colors. I tried using the `SearchButtonColor` type below for the keys of the exported object but I ran into some ts errors I had trouble resolving

src/app/content/constants.ts
```diff
@@ -18,3 +19,15 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
...
+  const bases: { [key: string]: string } = {};
+
+  Object.keys(colors).forEach((color) => {bases[color] = colors[color].base; });
+  return bases;
+};
+
+export const searchButtonStyles: {[key: string]: string} = {
+  gray: '#5D6062',
+  ...getBaseColors(theme.color.primary) as {},
+};
```
bethshook, line 33: I'm not sure whether this new gray merits updating the theme config

src/app/content/constants.ts
```diff
@@ -18,3 +19,15 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
...
 ];
+
+const getBaseColors = (colors: {[key: string]: {[key: string]: string}}) => {
+  const bases: { [key: string]: string } = {};
+
+  Object.keys(colors).forEach((color) => {bases[color] = colors[color].base; });
+  return bases;
+};
+
+export const searchButtonStyles: {[key: string]: string} = {
```
bethshook, line 30: @TomWoodward what is the correct way to apply a custom type (in this case `SearchButtonColor` defined below) for object keys? I tried but ran into ts errors that I struggled to resolve

src/app/content/constants.ts
```diff
@@ -18,3 +19,11 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
   {label: HighlightColorEnum.Purple, passive: '#cbcfff', focused: '#545ec8'},
   {label: HighlightColorEnum.Pink, passive: '#ffc5e1', focused: '#de017e'},
 ];
+
+export const getBaseColors = () => {
```
bethshook, line 23: @TomWoodward is there a theme utils file where this should live?

PiotrKozlowski: it seems to me like this could be implemented as a `getter` somewhere in the `color` object - `theme.color.primary.baseColors`

src/app/actions.ts
```diff
@@ -1,4 +1,5 @@
 import { createStandardAction } from 'typesafe-actions';
 
 export const receiveFeatureFlags = createStandardAction('app/receiveFeatureFlags')<string[]>();
+export const receiveExperiments = createStandardAction('app/receiveExperiments')<string[]>();
```
TomWoodward, line 4: move these two actions into featureFlags/actions

src/app/content/components/Toolbar/index.tsx
```diff
@@ -3,6 +3,8 @@ import flow from 'lodash/fp/flow';
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { connect } from 'react-redux';
+import { featureFlagsEnabled as featureFlagsEnabledSelector } from '../../../featureFlags/selectors';
```
TomWoodward, line 6: i like to treat selectors like verbs, often i'll do like:
```
import * as selectFeatureFlags from '../../../featureFlags/selectors';

...

selectFeatureFlags.enabled(state)
```

or in this case if you want to only pull in the one function maybe `selectEnabledFeatureFlags`

src/app/content/constants.ts
```diff
@@ -18,3 +19,11 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
...
+  const bases: { [key: string]: string } = {};
+  const primaryColors: {[key: string]: {[key: string]: string}} = theme.color.primary;
+
+  Object.keys(primaryColors).forEach((color) => {bases[color] = primaryColors[color].base; });
```
TomWoodward: is all this getting the base property of the primary colors? maybe:
```
Object.fromEntries(Object.entries(theme.color.primary).map(([key, {base}]) => [key, base]))
```

or you could use the `mapValues` from lodash that works on objects

src/app/content/search/selectors.ts
```diff
@@ -58,3 +60,8 @@ export const currentPageResults = createSelector(
   parentSelectors.page,
   (rawResults, page) => rawResults && page ? getSearchResultsForPage(page, rawResults) : []
 );
+
+export const bookTheme = createSelector(
```
TomWoodward, line 64: this shouldn't be a search selector, it should be in content/selectors

src/app/content/types.ts
```diff
@@ -132,3 +132,6 @@ export interface ArchivePage {
 }
 
 export type ArchiveContent = ArchivePage | ArchiveBook;
+
+// tslint:disable-next-line: max-line-length
+export type SearchButtonColor = 'blue' | 'green' | 'gray' | 'yellow' | 'deep-green' | 'light-blue' | 'orange' | 'red' | 'gray';
```
TomWoodward, line 137: is this the same as the theme colors? is it the theme colors plus gray? you have gray in there twice. can you derive it from the theme colors?

also this could go under content/search

src/app/featureFlags/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { FeatureFlagVariantValue } from './types';
+
+export const experiments: { [key: string]: FeatureFlagVariantValue[] } = {
```
TomWoodward, line 3: i think typescript's [record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) type is supposed to be used for situations like this

src/app/featureFlags/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+  searchButton: [
+    'originalButton',
+    'bannerColorButton',
+    'grayButton',
+  ],
+};
+
+export const experimentIds: { [key: string]: string } = {
```
TomWoodward, line 11: you could make the values here `keyof experiments` and then typescript will complain if there are any mismatches

src/app/featureFlags/reducer.ts
```diff
@@ -0,0 +1,26 @@
...
+import { receiveExperiments } from '../actions';
+import { AnyAction } from '../types';
+import { experimentIds, experiments } from './constants';
+import { State } from './types';
+
+export const initialState: State = {};
+
+const reducer: Reducer<State, AnyAction> = (state = initialState, action): any => {
+    switch (action.type) {
+        case getType(receiveExperiments):
```
TomWoodward, line 12: can you also reduce receiveFeatureFlags? even if we don't use them for anything i'd be nice to have them and migrate towards using them from here, and having them here for debugging wouldn't suck

src/app/featureFlags/reducer.ts
```diff
@@ -0,0 +1,26 @@
...
+            const [id, variant] = action.payload;
+            if (experimentIds[id]) {
```
TomWoodward, line 14: would be better to also check if the variant exists

TomWoodward (changes requested):

src/app/content/constants.ts
bethshook: I'm running into issues with both of those options. As I mentioned, `Object.fromEntries` is from a newer ES version than the one we're targeting, and `mapValues` returns an empty object always, including when I use an example from the lodash docs:
```
const users = {
  fred: { user: 'fred', age: 40 },
  pebbles: { user: 'pebbles', age: 1 },
};

console.log(mapValues(users, (o: any) => o.age));
// => {}

console.log(mapValues(users, 'age'));
  // => {0: false, 1: false, 2: false}
```

src/app/featureFlags/reducer.ts
```diff
@@ -0,0 +1,30 @@
...
+                const experimentName = experimentIds[id];
+                const variantIndex = parseInt(variant, 10);
+                const variantName = experiments[experimentName][variantIndex];
+                return {...state, [experimentName]: variantName};
+            }
+            return state;
+        case getType(receiveFeatureFlags):
+            const flags: { [key: string]: boolean } = {};
+            action.payload.forEach((flag: string) => flags[flag] = true);
+            return {...state, ...flags};
```
bethshook, line 24: @TomWoodward is this ok for now?
```
featureFlags: {
   'searchButton': 'grayButton',
   'studyGuidesEnabled': true,
}
```

src/helpers/analytics.ts
```diff
@@ -75,6 +76,15 @@ const analytics = {
 export const registerGlobalAnalytics = (window: Window, store: Store) => {
   const document = window.document;
 
+  // google optimize - will dispatch a feature flag on variant receieved
+  window.gtag('event', 'optimize.callback', {
+    callback: (idx: string, id: string) => {
+      store.dispatch(receiveExperiments([id, idx]));
+    },
+  });
```
bethshook, line 84: @TomWoodward the test for this function fails on the exception `TypeError: window.gtag is not a function` because `gtag()` is defined in the head html. Is there a way I can mock it?

TomWoodward: you could throw it in `src/test/setup.ts` but to please typescript in either the tests or the regular app you'll need to add its type to `src/typings/global.d.ts`

src/app/featureFlags/reducer.ts
TomWoodward: yea that state looks good. i'd write the logic as
```ts
return action.payload.reduce((result, flag) => ({...result, [flag]: true}), state)
```

src/app/content/constants.ts
TomWoodward: the way you have it done here you'd just do 
```

File: src/app/featureFlags/reducer.ts [hunk 21]
```diff
@@ -0,0 +1,28 @@
...
+            const [id, variant] = action.payload;
+            if (variant && experimentIds[id]) {
+                const experimentName = experimentIds[id];
+                const variantIndex = parseInt(variant, 10);
+                const variantName = experiments[experimentName][variantIndex];
+                return {...state, [experimentName]: variantName};
+            }
+            return state;
+        case getType(receiveFeatureFlags):
+            return action.payload.reduce((result, flag) => ({...result, [flag]: true}), state);
```
TomWoodward, line 22: i misled you here, the code i suggested is modifying the existing `state` which is bad. you can do either:
```suggestion
            return action.payload.reduce((result, flag) => ({...result, [flag]: true}), {...state});
```
or:
```suggestion
            return {...state, ...action.payload.reduce<State>((result, flag) => ({...result, [flag]: true}), {})};
```

src/helpers/analytics.ts
```diff
@@ -75,6 +76,16 @@ const analytics = {
...
 
+  // google optimize - will dispatch a feature flag on variant receieved
+  if (window.gtag && window.dataLayer) {
+    window.gtag('event', 'optimize.callback', {
+      callback: (idx: string, id: string) => {
+        store.dispatch(receiveExperiments([id, idx]));
+      },
+    });
+    window.dataLayer.push({event: 'optimize.activate'});
+  }
```
TomWoodward, line 87: would this logic maybe fit better in the `loadOptimize` function?

src/typings/global.d.ts
```diff
@@ -54,6 +54,8 @@ declare global {
     };
     MathJax: any;
     ga: UniversalAnalytics.ga;
+    dataLayer:  object[];
+    gtag: any;
```
TomWoodward, line 58: are there types in `UniversalAnalytics` to help with this? if not then you could at least type gtag as a function accepting `string, string, object` which i think is what we understand to be the right arguments

TomWoodward (changes requested):

src/app/content/constants.ts
bethshook: No, first I got the error `An index signature parameter type cannot be a union type. Consider using a mapped object type instead.` and then others as I tried to resolve it. But this code isn't needed anymore.

TomWoodward: did you try the mapped type as the error suggests? by doing: 
```
export const searchButtonStyles: {[key in SearchButtonColor]: string} = {
```

src/helpers/loadOptimize.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const script = assertDocument().createElement('script');
+    script.setAttribute('type', 'text/javascript');
+    script.setAttribute('src', `https://www.googleoptimize.com/optimize.js?id=${containerId}`);
+    script.onload = resolve;
+    assertDocument().head.appendChild(script);
+
+    window.gtag('event', 'optimize.callback', {
+      callback: (idx: string, id: string) => {
+        store.dispatch(receiveExperiments([id, idx]));
```
bethshook, line 19: @TomWoodward I'm not sure how to mock the callback so that this line is covered by the test

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  beforeEach(() => {
+    store = createTestStore();
+    window = assertWindow();
+    window.gtag = jest.fn();
+    window.dataLayer = [];
+
+    if (typeof document === 'undefined') {
+      throw new Error('JSDom not loaded');
+    }
```
PiotrKozlowski, line 18: You could use `assertDocument`, similar to `assertWindow` - or just use `window.document`

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    await loadOptimize(window, store);
+
+    if (document && document.head) {
+      const style = document.head.querySelector('style');
+      if (style) {
+        style.remove();
+      }
+      // tslint:disable: max-line-length
+      expect(document.head.innerHTML).toMatchInlineSnapshot(
+        `"<script type=\\"text/javascript\\" src=\\"https://www.googleoptimize.com/optimize.js?id=OPT-NFHSM4B\\"></script>"`
```
PiotrKozlowski, line 58: you could expect something is returned from `document.querySelector` for this specific attribute - you wouldn't have to remove `style`

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    const newLocation = {
+      ...window.location,
+      hostname: 'openstax.org',
+    };
+    delete (window as any).location;
+    window.location = newLocation;
+
+    await loadOptimize(window, store);
+
+    if (document && document.head) {
```
PiotrKozlowski, line 51: if you would use `assertDocument` at the top then you don't have to check for this

src/helpers/loadOptimize.ts
PiotrKozlowski: As I understand calling `window.gtag(...)` should push all arguments to the `window.dataLayer` so you could create a test in which you call callback from the first item, like `window.dataLayer[0][2].callback(idx, id)` and expect that receiveExperiments was dispatched

bethshook: That worked once I defined the function in test/setup.ts 👍

public/index.html
```diff
@@ -30,6 +30,12 @@
     </script>
     <!-- End Google Analytics -->
 
+    <!-- Google Optmize (declaring gtag helper function) -->
+    <script>
+      function gtag() {if(dataLayer)dataLayer.push(arguments);}
```
PiotrKozlowski, line 35: I've found:
```
window.dataLayer = window.dataLayer || [];
function gtag(){dataLayer.push(arguments);}
```
https://developers.google.com/gtagjs/devguide/snippet#google-analytics

Idk which one should be used

src/test/setup.ts
```diff
@@ -93,6 +93,13 @@ beforeEach(() => {
 
   scrollTo = window.scrollTo = jest.fn();
   scrollBy = window.scrollBy = jest.fn();
+  window.dataLayer = [];
+
+  function gtag() {
+    if (window) {
+      window.dataLayer.push(arguments);
+    }
+  }
```
PiotrKozlowski, line 102: Should we add some comment here that it is also added in the `index.html`?

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    dispatch = jest.spyOn(store, 'dispatch');
+
+    const createEvent = document.createEvent.bind(document);
+    const originalCreateElement = document.createElement.bind(document);
+
+    document.createElement = jest.fn((type: string) => {
+      const element = originalCreateElement(type);
+      const event = createEvent('Event');
+      event.initEvent('load');
+      setTimeout(() => element.onload && element.onload(event), 1);
```
PiotrKozlowski, line 26: `0` should be enough

src/app/content/components/Toolbar/index.tsx
```diff
@@ -94,13 +100,19 @@ class Toolbar extends React.Component<Props, State> {
     const hideFromFocus = this.props.tocOpen === true
       || (this.props.tocOpen === null && !this.props.searchSidebarOpen);
 
+    const searchButtonColor = this.props.searchButtonStyle === 'grayButton' ? ('gray' as BookWithOSWebData['theme']) : (
+      this.props.searchButtonStyle === 'bannerColorButton' ? this.props.bookTheme : null
```
TomWoodward, line 104: i feel a little weird about allowing the third variant to fall through the null case, but its probably fine

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -131,6 +133,12 @@ export const SearchButton = styled(({ desktop, mobile, ariaLabelId, ...props })
     <SearchIcon/>
   </PlainButton>;
 })`
+  height: 3.2rem;
+  background:
```
TomWoodward, line 137: does this look horrible when the flag loads and it suddenly changes to orange? maybe we could add a transition on the background color?

src/app/content/components/utils/applySearchIconColor.ts
```diff
@@ -0,0 +1,28 @@
+import { css } from 'styled-components/macro';
+import theme from '../../../theme';
+import { BookWithOSWebData } from '../../types';
+import { toolbarIconColor } from '../constants';
+
+export const applySearchIconColor = (props: {colorSchema: BookWithOSWebData['theme']}) =>
+  props.colorSchema ? css`
```
TomWoodward, line 7: unless they recently fixed this the arguments to the functions you call in a styled-component block don't have strong typing, so i prefer not to hide them in helpers. for instance here you might make the function accept just the `schema` and do the prop extraction in the styled block

src/app/featureFlags/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+export const experiments: Record<string, FeatureFlagVariantValue[]> = {
+  searchButton: [
+    'originalButton',
+    'bannerColorButton',
+    'grayButton',
+  ],
+};
+
+export const experimentIds: { [key: string]: keyof typeof experiments } = {
+  '473Jeu7ORta1TTuJ5UjL6w': 'searchButton', // prod
```
TomWoodward, line 12: this test experiement had already been started, here is a new one i just made that is in draft state and we'll start it after the release goes out `zHtr33wcS1mVgFn7b2nU3A`

src/app/featureFlags/types.ts
```diff
@@ -0,0 +1,7 @@
+export type FeatureFlagVariantValue = string | number | boolean;
+
+export type FeatureFlag = string;
+
+export interface State {
+    [key: string]: FeatureFlagVariantValue;
+   }
```
TomWoodward, line 7: closing bracket indentation

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    dispatch = jest.spyOn(store, 'dispatch');
+
+    const createEvent = document.createEvent.bind(document);
+    const originalCreateElement = document.createElement.bind(document);
+
+    document.createElement = jest.fn((type: string) => {
+      const element = originalCreateElement(type);
+      const event = createEvent('Event');
+      event.initEvent('load');
+      setTimeout(() => element.onload && element.onload(event), 0);
```
TomWoodward, line 26: does it work if you do `setImmediate`?

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  afterEach(() => {
+    document.head.innerHTML = '';
+    window.dataLayer = [];
+  });
+
+  it('injects correct <script> into head if in production', async() => {
+    const newLocation = {
+      ...window.location,
+      hostname: 'openstax.org',
```
TomWoodward, line 40: i think this is actually registering sessions in the results

src/helpers/loadOptimize.ts
```diff
@@ -0,0 +1,23 @@
+import { receiveExperiments } from '../app/featureFlags/actions';
+import { Store } from '../app/types';
+import { assertDocument } from '../app/utils';
+
+export const getOptimizeContainerByEnv = (window: Window) =>
+  window.location.hostname === 'openstax.org' ? 'OPT-NFHSM4B' : 'OPT-W65B3CP';
```
TomWoodward, line 6: any reason not to use the deployed_env configuration that we already have?

TomWoodward (changes requested):

src/app/content/components/Toolbar/styled.tsx
bethshook: It loads the same as the BookBanner color so I think we should either add a transition to both or neither.

src/helpers/loadOptimize.ts
bethshook: won't `window` always be defined when this is called? I made this change to avoid having to do a null check for `containerId` below

src/app/content/components/Toolbar/index.tsx
```diff
@@ -94,13 +103,30 @@ class Toolbar extends React.Component<Props, State> {
     const hideFromFocus = this.props.tocOpen === true
       || (this.props.tocOpen === null && !this.props.searchSidebarOpen);
 
+    let searchButtonColor = null;
+
+    if (book) {
+      switch (searchButtonStyle) {
+        case 'grayButton':
```
PiotrKozlowski, line 110: This could be written with a ternary, but I guess an even better option would be to have this written like:
```
const searchButtonColor = book ? getSearchButtonColor(searchButtonStyle, bookTheme) : null;
```
or maybe even better: create a selector called something like `searchButtonColorExperiment` that checks for `book`, `bookTheme` and `searchButtonStyle`

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -131,6 +133,12 @@ export const SearchButton = styled(({ desktop, mobile, ariaLabelId, ...props })
...
+    ${(props: {colorSchema: BookWithOSWebData['theme'] }) => props.colorSchema
```
PiotrKozlowski, line 138: If `colorSchema` may be null then I think we should type it like this

src/app/content/components/utils/applySearchIconColor.ts
```diff
@@ -0,0 +1,29 @@
+import { css } from 'styled-components/macro';
+import theme from '../../../theme';
+import { toolbarIconColor } from '../constants';
+
+const primaryColors = theme.color.primary;
+
+export const applySearchIconColor = (colorSchema: keyof typeof primaryColors) =>
+  colorSchema ? css`
```
PiotrKozlowski, line 8:
```suggestion
export const applySearchIconColor = (colorSchema: keyof typeof primaryColors | null) =>
  colorSchema && primaryColors[colorSchema] ? css`
```

PiotrKozlowski (changes requested):
https://github.com/openstax/rex-web/pull/1181/files#r639890475

src/helpers/loadOptimize.ts
TomWoodward: if you just do:
```suggestion
  DEPLOYED_ENV === 'openstax.org' ? 'OPT-NFHSM4B' : 'OPT-W65B3CP';
```
that doesn't change your null check requirements right?

i know there isn't _a lot_ of behind that `DEPLOYED_ENV` variable, but there is a little, and idk maybe would be a logical place to add more logic in the future, since its just a case of using one variable instead of another here i think its worth using the derived value

TomWoodward (changes requested):

src/app/content/components/Toolbar/styled.tsx
TomWoodward: but book banner will have its color before the feature flags are loaded

src/app/featureFlags/selectors.ts
```diff
@@ -0,0 +1,24 @@
...
+  parentSelectors.localState,
+  (parentState) => parentState.featureFlags
+);
+
+export const searchButtonStyle = createSelector(
+  enabled,
+  (featureFlags) => featureFlags.searchButton || null
+);
+
+export const searchButtonColor = createSelector(
```
bethshook, line 16: @PiotrKozlowski is this what you had in mind?

src/helpers/loadOptimize.ts
bethshook: makes sense

src/helpers/loadOptimize.spec.ts
bethshook: yes, did not know this existed

src/app/content/components/Toolbar/styled.tsx
bethshook: ah that's right. so far it looks fine but I'll add a .2s background transition.

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  afterEach(() => {
+    document.head.innerHTML = '';
+    window.dataLayer = [];
+  });
+
+  it('injects correct <script> into head if in production', async() => {
+    config.DEPLOYED_ENV = 'openstax.org';
+    window.dataLayer.push = jest.fn();
```
bethshook, line 39: @TomWoodward added this to prevent registering sessions

I'm also wondering why overwriting `config.DEPLOYED_ENV` works above but `jest.mock('../config', () => ({DEPLOYED_ENV: 'openstax.org'}));` does not

src/helpers/loadOptimize.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      'https://www.googleoptimize.com/optimize.js?id=OPT-W65B3CP'
+    );
+  });
+
+  it('registers optimize callback and correctly dispatches action', async() => {
+    config.DEPLOYED_ENV = 'foo';
+    await loadOptimize(window, store);
+
+    ((window.dataLayer[0] as any)[2] as any).callback('1', 'OCCkMMCZSwW87szzpniCow');
+    expect(dispatch).toHaveBeenCalledWith(receiveExperiments(['OCCkMMCZSwW87szzpniCow', '1']));
```
bethshook, line 66: @TomWoodward is it ok to register sessions to dev here, or can you recommend a better way to test the optimize callback?

src/app/featureFlags/selectors.ts
PiotrKozlowski: yes!

src/helpers/loadOptimize.spec.ts
PiotrKozlowski: Is it loading the actual script when you call `await loadOptimize`? Maybe we could just prevent this?

TomWoodward: to make it a unit test i think you'd mock the appendElement and asset the given script tag against a snapshot, you'd have to also mock the gtag and dataLayer calls, invoke the callback in the test with some known input, and then assert that dispatch is correctly called.

that is probably the right way to do it, although i'm a little sad about losing the confirmation that we're doing the calls correctly for the library because they are a little arcane, it is really better to not have the unit tests depending on loading external scripts

TomWoodward (changes requested):

src/helpers/loadOptimize.spec.ts
bethshook: updated

PiotrKozlowski (approved):

src/helpers/loadOptimize.spec.ts
TomWoodward: probably depends on the order of the importing and the mocking

TomWoodward (approved):

PR #1183, opened 2021-05-17
handle search that has no results

jarosik10:
for: https://github.com/openstax/unified/issues/1380


PiotrKozlowski (changes requested):

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +52,60 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-    query,
-    target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
-  });
-  const hash = selectedResult.result.source.elementId;
+  if (selectedResult) {
+    services.dispatch(selectSearchResult(selectedResult));
+  }
 
-  services.dispatch(action(navigation, { hash, search }));
+  let navigation = selectNavigation.match(services.getState());
```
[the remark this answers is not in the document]
jarosik10: I don't think that I can use `createNavigationOptions` here. I can't make 
`search.target:` to be  `JSON.stringify({ type: 'search', index: selectedResult.highlight }),` using `createNavigationOptions`.

> You can use state variable and also make it a `const currentNavigation`

So you want the navigation variable to not be `let`?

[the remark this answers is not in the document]
PiotrKozlowski: oh, I didn't know there is `createNavigationOptions` function. I've meant to create one and get rid of the `let`

src/app/content/search/hooks.spec.ts
```diff
@@ -310,7 +312,7 @@ describe('hooks', () => {
       );
     });
 
-    it('dispatches REPLACE with search query when page is undefined', () => {
+    it('dispatches PUSH with search query when page is undefined', () => {
```
jarosik10, line 315: @TomWoodward this how it should be, right?

src/app/content/search/hooks/index.ts
```diff
@@ -65,28 +65,29 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-      params: getBookPageUrlAndParams(book, targetPage).params,
-      route: content,
-      state : {
-        bookUid: book.id,
-        bookVersion: book.version,
-        pageUid: stripIdVersion(targetPage.id),
-      },
-    };
-  }
+  const createUpdatedNavigation = () => {
```
PiotrKozlowski, line 70: This function will be defined every time when you call `receiveSearchHook`
If you extract this function to another file then it will be defined only once + logic in `receiveSearchHook` will become more straightforward

PiotrKozlowski (changes requested):

src/app/content/search/hooks/index.ts
jarosik10: You are right. I'm wondering if I should create a separate file in `search/hooks/utils.ts` or just use the existing `search/utils.ts` file, because I'm not sure if this util-function will used anywhere else.

PiotrKozlowski: I guess `search/utils.ts` is fine

src/app/content/search/utils.ts
```diff
@@ -175,3 +182,21 @@ export const findSearchResultHit = (
 ): SearchResultHit | undefined => {
   return results.find((result) => result.source.elementId === target.elementId);
 };
+
+export const createUpdatedNavigation = (targetPageId: string | null, book: Book) => {
```
PiotrKozlowski, line 186: This function is not updating `navigation` so I think it should be maybe called something like `createRouteMatchOptions`
It should probably also take `book: Book, pageId: string` and return null if `page` was not found in a book.
This way it would be more generic. You could also declare the returned type so it is more clear what we expect

PiotrKozlowski (changes requested):

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,41 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-  const search = queryString.stringify({
-    query,
-    target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
-  });
-  const hash = selectedResult.result.source.elementId;
+  if (selectedResult) {
+    services.dispatch(selectSearchResult(selectedResult));
+  }
+
+  const targetPageId = selectedResult ? selectedResult.result.source.pageId : currentPage ? currentPage.id : null;
```
PiotrKozlowski, line 66: You can use optional chaining now

src/app/content/search/utils.ts
```diff
@@ -175,3 +183,20 @@ export const findSearchResultHit = (
 ): SearchResultHit | undefined => {
   return results.find((result) => result.source.elementId === target.elementId);
 };
+
+export const createRouteMatchOptions = (pageId: string, book: Book): AnyMatch => {
+  const page = assertDefined(
+    findArchiveTreeNodeById(book.tree, pageId),
+    'search result pointed to page that wasn\'t in book'
+  );
```
PiotrKozlowski, line 191: No need for assert, just return `null` or `undefined if it wasn't found

PiotrKozlowski (changes requested):

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,41 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-    target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
-  });
-  const hash = selectedResult.result.source.elementId;
+  if (selectedResult) {
+    services.dispatch(selectSearchResult(selectedResult));
+  }
+
+  const targetPageId = selectedResult?.result.source.pageId || currentPage?.id;
+
+  const navigation = targetPageId ? createRouteMatchOptions(targetPageId, book) : selectNavigation.match(state);
```
PiotrKozlowski, line 68: This const could be directly above the `if (navigation)`

src/app/content/search/utils.ts
```diff
@@ -175,3 +183,19 @@ export const findSearchResultHit = (
 ): SearchResultHit | undefined => {
   return results.find((result) => result.source.elementId === target.elementId);
 };
+
+export const createRouteMatchOptions = (pageId: string, book: Book): AnyMatch | undefined => {
+  const page = findArchiveTreeNodeById(book.tree, pageId);
+
+  return page ?
```
PiotrKozlowski, line 190: You could do `if (!page) { return }` or put the `?` and the same level as `:`

PiotrKozlowski (approved):

PiotrKozlowski (approved):

src/app/content/search/hooks.spec.ts
TomWoodward: whether the page is defined or not isn't the criteria, the way we have it set up it should be a `push` if the url location is changing, or a `replace` if it is only modifying the query.

so if you're on a 404 page and you do a search that has no results, that would be a replace still, but if it has results and it goes somewhere it'd `push`

TomWoodward (changes requested):

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,40 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
     : getFirstResult(book, payload);
 
   if (
-    // selectedResult may equal to null if api did not return any results
-    !selectedResult
+    selectedResult &&
```
TomWoodward, line 53: operators on the beginning of the lines

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,40 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
     // We are clearing selected result when requesting a new search so in the theory this should never happen
-    || isEqual(select.selectedResult(state), selectedResult)
+    (isEqual(select.selectedResult(state), selectedResult)
```
TomWoodward, line 55: indent the parenthesis contents to make them more obviously in parenthesis

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,40 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
+    services.dispatch(selectSearchResult(selectedResult));
+  }
 
-  services.dispatch(action(navigation, { hash, search }));
+  const targetPageId = selectedResult?.result.source.pageId || currentPage?.id;
+
+  const action = (targetPageId && stripIdVersion(targetPageId)) === (currentPage && stripIdVersion(currentPage.id))
+    ? replace : push;
+
+  const options = selectedResult ?
```
TomWoodward, line 71: i format ternaries as:
```
const thing = condition
  ? first case
  : second case
;
```
with the operators on the front

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +50,40 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
+      search: queryString.stringify({
+        query,
+        target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
+      }),
+    } :
+    {
+      search: queryString.stringify({ query }),
+    };
+
+  const navigation = targetPageId ? createRouteMatchOptions(targetPageId, book) : selectNavigation.match(state);
```
TomWoodward, line 83: this method name is `createRouteMatchOptions` but it is being assigned to something called `navigation`, and another thing called `options` is defined above, something is not right here

src/app/content/search/utils.ts
```diff
@@ -175,3 +183,18 @@ export const findSearchResultHit = (
 ): SearchResultHit | undefined => {
   return results.find((result) => result.source.elementId === target.elementId);
 };
+
+export const createRouteMatchOptions = (pageId: string, book: Book): AnyMatch | undefined => {
```
TomWoodward, line 187: this logic isn't actually specific to search, and overlaps with stuff in ContentLink i think, maybe some other places, you could move the util up and use it in more places

TomWoodward (changes requested):
the logic is looking good, i left some comments on formatting and i'd love to see that util used in more places

src/app/content/components/ContentLink.tsx
```diff
@@ -54,6 +55,9 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     ...anchorProps
   } = props;
   const {url, params} = getBookPageUrlAndParams(book, page);
+  const navigationMatch = assertDefined(
+    createNavigationMatch(page.id, book, params),
+    `Page ${page.id} does not exist in book ${book.id}`);
```
TomWoodward, line 60: why might this not return anything? would it help to pass the whole page instead of just the id? we weren't previously checking if the given page was in the given book.

src/app/content/search/hooks/index.ts
```diff
@@ -52,45 +51,38 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
+      hash: selectedResult.result.source.elementId,
+      search: queryString.stringify({
+        query,
+        target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
+      }),
+    }
+    : { search: queryString.stringify({ query }) };
+
+  const navigationMatch = targetPageId ? createNavigationMatch(targetPageId, book) : selectNavigation.match(state);
+  if (navigationMatch) {
```
TomWoodward, line 83: the only reason this would be nullish would be if your `createNavigationMatch` helper decided the args were invalid right? in that case it'd be an error 

maybe that function should throw instead of returning undefined so you don't have to do null checks

TomWoodward (changes requested):

src/app/content/search/hooks/index.ts
jarosik10: `selectNavigation.match(state)` returns `AnyMatch | undefined`

TomWoodward: hmmm... i think ideally `selectNavigation.match` would not be able to return undefined because it should always at least be the 404 route... but that isn't something we should look into on this pr.

for now lets make `createNavigationMatch` always return a match but leave the conditional afterwards

TomWoodward (changes requested):

src/app/content/search/hooks/index.ts
jarosik10: This is how it is handled right now - it's ready for your review.

src/app/content/components/ContentLink.tsx
```diff
@@ -54,6 +54,7 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     ...anchorProps
   } = props;
   const {url, params} = getBookPageUrlAndParams(book, page);
+  const navigationMatch = createNavigationMatch(page as Page, book, params);
```
TomWoodward, line 57: you can't just cast this. either change the `createNavigationMatch` to only pick what it needs and hopefully that is only id and title or update Props here to add what `createNavigationMatch` needs

TomWoodward (changes requested):

TomWoodward (approved):

PR #1191, opened 2021-05-24
Do not run urlChecker on books that are newly added to the config

PiotrKozlowski:
https://openstax.slack.com/archives/C8V1XET1R/p1621864771075700?thread_ts=1621864107.075500&cid=C8V1XET1R

for: openstax/unified#1590

### DIFF
--- a/script/validate-modified-books.bash
+++ b/script/validate-modified-books.bash
@@ -7,6 +7,7 @@ code=0
 while read -r row; do
   book_id=$(jq -r '.book_id' <<< "$row")
   book_version=$(jq -r '.book_version' <<< "$row")
+  is_new=$(jq -r '.is_new' <<< "$row")
 
   node script/entry.js domVisitor errorsExist \
     --bookId="$book_id" \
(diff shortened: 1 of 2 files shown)


script/validate-modified-books.bash
```diff
@@ -15,12 +16,14 @@ while read -r row; do
     --queryString="validateLinks" \
     || code=1
 
-  node script/entry.js urlChecker \
-    --rootUrl="https://openstax.org" \
-    --bookId="$book_id" \
-    --bookVersion="$book_version" \
-    --useUnversionedUrls \
-    || code=1
+  if [ "$is_new" == false ]; then
```
TomWoodward, line 19: how does this actually look when it comes out? can a string equal a boolean? does bash have good boolean support?

TomWoodward (changes requested):

script/validate-modified-books.bash
PiotrKozlowski: I've added the sociology book to the config, run the script and this is the output:
```
...
+ book_version=3.7
++ jq -r .is_new
+ is_new=true
+ '[' true == false ']'
+ read -r row
+ exit 0
```

I've changed the check to `.. == true`, and it worked:
```
+ is_new=true
+ '[' true == true ']'
+ node script/entry.js urlChecker --rootUrl=https://openstax.org --bookId=746f171e-0d6a-4ef2-b69d-367880872f4a --bookVersion=3.7 --useUnversionedUrls
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade caniuse-lite browserslist`
checking introduction-to-sociology-3e [===                 ] 39/243 (3.8s ETA)^C
```

PiotrKozlowski: > can a string equal a boolean

The `""` around the variable are required by linter so I think it is not evaluated to the string

> does bash have good boolean support

I'm trying to find something about this. I see that using variables in if statements my be problematic since you can pass a command and then it's result will be evaluated, but this is not our case

https://stackoverflow.com/questions/2953646/how-can-i-declare-and-use-boolean-variables-in-a-shell-script#2953673

TomWoodward (approved):

PR #1196, opened 2021-05-31
Change how scrolling to the highlights work

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1147

To discuss: https://github.com/openstax/unified/issues/1147#issuecomment-851468302


jarosik10 (commented):
The code look fine, but as you said it requires some [discussion](https://github.com/openstax/unified/issues/1147#issuecomment-851468302). I'm not sure if hardcoding `100px` is a good idea.

jarosik10 (approved):

src/app/domUtils.ts
```diff
@@ -97,18 +97,25 @@ export const scrollTo = (elem: HTMLElement | Element | string) => {
...
 
-export const scrollIntoView = (elem: HTMLElement) => {
+/**
+ * This function will scroll into @param firstElement view but it will take into consideration
+ * the @param lastElement position to check if scrolling is required.
+ */
+export const scrollIntoView = (firstElem: HTMLElement, lastElem?: HTMLElement) => {
   const window = assertWindow();
-  if (!window.document.body.contains(elem)) { return; }
+  if (!window.document.body.contains(firstElem)) { return; }
```
jarosik10, line 106: Maybe we should make the same check if `lastElem` is passed.

jarosik10 (changes requested):

src/app/domUtils.ts
PiotrKozlowski: I think we would like to scroll to the `firstElement` if the `lastElement` is not defined, since it is supposed to be used only for calculation purposes

jarosik10 (approved):

PiotrKozlowski:
@jarosik10 You've made me think that if a user provides an element that is not in the DOM then its rects will be set to 0 which would interfere with calculations, so I've added a check for this: https://github.com/openstax/rex-web/pull/1196/commits/d93d5a1c699f665cbd9f8eaf1ba19573f2b2eefa

PR #1197, opened 2021-05-31
Issue 1596 - handle content-style query param

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1596


bethshook (approved):

src/app/components/DynamicContentStyles.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import styled from 'styled-components';
+import { query } from '../navigation/selectors';
+
+// tslint:disable-next-line: variable-name
+export const WithStyles = styled.div`
+  ${(props: { styles: string }) => props.styles}
```
TomWoodward, line 8: can you rearrange this so that it doesn't add yet another arbitrary div to the dom structure?

src/app/content/components/ContentLink.tsx
```diff
@@ -30,7 +30,7 @@ interface Props {
   navigate: typeof push;
   currentPath: string;
   hasUnsavedHighlight: boolean;
-  search: { query: string | null };
+  search: Record<string, string | null | undefined>;
```
TomWoodward, line 33: probably `null` is a mistake here, i don't see how it could be null right?`

src/app/content/search/hooks/index.ts
```diff
@@ -84,9 +85,11 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
   };
 
   const action = stripIdVersion(page.id) === stripIdVersion(targetPage.id) ? replace : push;
+  const contentStyle = navigationQuery['content-style'] as string | undefined;
   const search = queryString.stringify({
+    'content-style': contentStyle,
```
TomWoodward, line 90: i feel like you want a centralized config somewhere for like... 'persistentQueryParameters` that includes the search query and content style, and merge that under the given nav options in the router dispatch, rather than having to update all these different places

TomWoodward (changes requested):

jarosik10 (changes requested):
There is `test:build` to fix. 
```
-           <div class="sc-bwzfXH bGgvvP">
+           <div class="sc-bdVaJa ibTDPh">
```

src/app/content/components/ContentLink.tsx
PiotrKozlowski: It is because we are doing:
```
search: ({
      ...
      'query': selectSearch.query(state),
      ...
    }),
```
and `selectSearch.query(state)` may return `null`

TomWoodward: maybe it should be changed to `undefined`?

src/app/components/DynamicContentStyles.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+interface DynamicContentStylesProps extends React.HTMLAttributes<HTMLDivElement> {}
+
+// tslint:disable-next-line: variable-name
+const DynamicContentStyles = ({ children, ...otherProps }: React.PropsWithChildren<DynamicContentStylesProps>) => {
+  const queryParams = useSelector(query);
+  const [styles, setStyles] = React.useState('');
+
+  React.useEffect(() => {
+    const cssfileUrl = queryParams['content-style'] as string | undefined;
```
TomWoodward, line 19: does this need the cast?

src/app/components/DynamicContentStyles.tsx
```diff
@@ -0,0 +1,32 @@
...
+    if (cssfileUrl) {
+      fetch(cssfileUrl)
+        .then((res) => res.text())
+        .then((data) => setStyles(data));
+    }
+  }, [queryParams]);
+
+  return <WithStyles styles={styles} data-dynamic-style={!!styles} {...otherProps}>
+    {children}
```
TomWoodward, line 28: it doesn't matter so much for testing, but if we were to use this component with the baked styles in production we'd want to find a way to prevent the flash of unstyled content while this loads

src/app/content/highlights/components/SummaryPopup/utils.ts
```diff
@@ -1,12 +1,25 @@
...
 import { getBookPageUrlAndParams } from '../../../utils/urlUtils';
 
-export const createHighlightLink = ({ sourceId, id, anchor: elementId }: Highlight, book?: Book) => {
-  const page = book ? findArchiveTreeNodeById(book.tree, sourceId) : undefined;
-  const data = { target: JSON.stringify({ id, type: 'highlight' }) };
-  const target = queryString.stringify(data) + `#${elementId}`;
-  return `${page && book ? getBookPageUrlAndParams(book, page).url : ''}?${target}`;
+export const useCreateHighlightLink = ({ sourceId, id, anchor: elementId }: Highlight, book?: Book) => {
+  const [link, setLink] = React.useState('');
+  const persistentQueryParams = useSelector(persistentQueryParameters);
```
TomWoodward, line 12: im making things up a little bit here but i think when opening in a new tab you probably don't want search to be pulled with you. but you do want the content styles and archive though. unless i guess you're clicking a search result, then you'd want the search.

maybe separate your selectors into `systemQueryParameters` and rely on the router to add the `persistentQueryParameters` as it processes the navigation, that way opening a link in a new tab vs navigating through js will work right

src/app/content/search/hooks/index.ts
```diff
@@ -87,6 +87,7 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
   const search = queryString.stringify({
     query,
     target: JSON.stringify({ type: 'search', index: selectedResult.highlight }),
+    ...selectNavigation.persistentQueryParameters(state),
```
TomWoodward, line 90: i was thinking you'd do this in the router

TomWoodward (changes requested):

src/app/navigation/middleware.ts
```diff
@@ -28,6 +30,11 @@ export default (routes: AnyRoute[], history: History): Middleware => ({dispatch}
       return;
     }
 
+    action.payload.search = queryString.stringify({
```
TomWoodward, line 33: you're stringifying this just so `matchSearch` can immediately parse it again. the second argument to `matchSearch` is only used from here, so you could change the way it works to be less silly

TomWoodward (changes requested):

src/app/components/DynamicContentStyles.tsx
TomWoodward: you reacted +1 but the cast is still here, does that mean you do need it? i thought it was handled by the state types

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
```diff
@@ -141,11 +142,13 @@ const HighlightListElement = ({ highlight, locationFilterId, pageId }: Highlight
       onColorChange={updateColor}
     />}
     <HighlightContentWrapper color={highlight.color}>
-      <ContentExcerpt
-        data-highlight-id={highlight.id}
-        content={highlight.highlightedContent}
-        source={highlight.sourceId}
-      />
+      <DynamicContentStyles>
```
TomWoodward, line 145: this is still rendering an extra div here. you may want to change to something more extensible like a higher order component or function-as-child pattern that would allow you to just compose the classname onto other elements

TomWoodward (changes requested):

src/app/content/highlights/components/SummaryPopup/HighlightListElement.tsx
TomWoodward: i haven't looked in contentExcerpt but maybe you could just update that too

src/app/components/DynamicContentStyles.tsx
PiotrKozlowski: I probably didn't stage this file before committing, thanks

src/app/components/DynamicContentStyles.spec.tsx
```diff
@@ -29,9 +29,26 @@ describe('DynamicContentStyles', () => {
     // tslint:disable-next-line: no-empty
     await renderer.act(async() => {});
 
+    expect(spyFetch).toHaveBeenCalledTimes(1);
+
     const withStyles = componenet.root.findByType(WithStyles);
     expect(withStyles.props.styles).toEqual('.cool { color: red; }');
     expect(spyFetch).toHaveBeenCalledWith('file.css');
+
+    // tslint:disable-next-line: no-empty
```
jarosik10, line 38: not needed

src/app/components/DynamicContentStyles.spec.tsx
```diff
@@ -29,9 +29,26 @@ describe('DynamicContentStyles', () => {
...
+    await renderer.act(async() => {
+      store.dispatch(locationChange({ location: { search: 'content-style=file2.css' } } as any));
+    });
+
+    expect(spyFetch).toHaveBeenCalledTimes(2);
+
+    // tslint:disable-next-line: no-empty
```
jarosik10, line 45: not needed

jarosik10 (changes requested):

TomWoodward (approved):

jarosik10 (approved):

bethshook (approved):

src/app/navigation/selectors.ts
```diff
@@ -38,4 +40,16 @@ export const match = createSelector(
   (state) => state.match
 );
 
+export const systemQueryParameters = createSelector(
+  query,
+  (navQuery) => pickBy(isDefined, {
+    'content-style': navQuery['content-style'] as string | undefined,
```
TomWoodward, line 46: archive is also a system param

TomWoodward: also why does this need a cast

src/app/navigation/middleware.ts
```diff
@@ -31,7 +33,12 @@ export default (routes: AnyRoute[], history: History): Middleware => ({dispatch}
     history[action.payload.method]({
       hash: action.payload.hash,
       pathname: matchPathname(action.payload),
-      search: matchSearch(action.payload, action.payload.search),
+      search: matchSearch(
+        action.payload,
+        {
+          ...persistenQueryParameters(getState()),
+          ...queryString.parse(action.payload.search || ''),
```
TomWoodward, line 40: i mean this is less of a big deal but it'd be nice if we werent doing all this parsing and hashing of query arguments. all throughout the historyActions its a string but i'm not sure it has to be, i think it might be better as an object everywhere until the return result of `matchSearch`

src/app/navigation/middleware.ts
```diff
@@ -31,7 +33,12 @@ export default (routes: AnyRoute[], history: History): Middleware => ({dispatch}
     history[action.payload.method]({
       hash: action.payload.hash,
       pathname: matchPathname(action.payload),
-      search: matchSearch(action.payload, action.payload.search),
+      search: matchSearch(
+        action.payload,
+        {
```
TomWoodward, line 38: so here is the problem malar found, all params are "persistent" so if you try to remove a query param by dispatching a location change without it, it'll be re-added by the middleware.

src/app/navigation/selectors.ts
```diff
@@ -38,4 +40,16 @@ export const match = createSelector(
...
+  })
+);
+
+export const persistenQueryParameters = createSelector(
+  query,
+  (navQuery) => pickBy(isDefined, navQuery)
```
TomWoodward, line 52: i'm not sure i totally agree with all parameters being "persistent", i forget if i had listed a few before, but search definitely. the `modal` params i don't think really need to be here, because they aren't the kind of thing you have to reach across the app to persist on navigation without knowing what its for. if you're doing navigation in a modal and want to make sure it stays open, you'll be doing it in the modal

src/app/navigation/middleware.ts
TomWoodward: maybe if we're going to follow through with this pattern, we need to put a filter on undefined in here and then the only way to remove a "persistent" query param is to explicitly pass undefined so it does get re-added.

i'm not sure i really like that, but it would work

TomWoodward (changes requested):

src/app/navigation/selectors.ts
```diff
@@ -38,4 +40,19 @@ export const match = createSelector(
   (state) => state.match
 );
 
+export const systemQueryParameters = createSelector(
+  query,
+  (navQuery) => pickBy(isDefined, {
+    'content-style': navQuery['content-style'],
```
bethshook, line 46: when I included `navQuery.archive` here, it ended up duplicated in the query string on navigation

src/app/navigation/middleware.ts
bethshook: I'm struggling with finding a solution outside of inserting the system params here. Specifically I haven't found a way to persist the `content-style` query param once `locationChange` is dispatched after a search.

TomWoodward: you'd have to put it back in the location change dispatch in the search hook right?

src/app/content/components/ContentLink.tsx
```diff
@@ -59,7 +61,7 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
   const bookUid = stripIdVersion(book.id);
   // Add options only if linking to the same book
   const options = currentBook && currentBook.id === bookUid
-    ? createNavigationOptions(search, scrollTarget)
+    ? createNavigationOptions({...search, ...systemQueryParams}, scrollTarget)
```
TomWoodward, line 64: i don't think `contentlink` should be handling "search" specifically, maybe we can't select persistent params from within here, because maybe we'd be trying to remove them, but we should accept any query params as input to the component, not typed just to search

i would like to have a type somewhere that encapsulates all the possible query params that we support on the content route

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -38,11 +39,12 @@ export default class PageComponent extends Component<PagePropTypes> {
...
 
   public getTransformedContent = () => {
-    const {book, page, services} = this.props;
+    const {book, page, services, systemQueryParams} = this.props;
 
     return getCleanContent(book, page, services.archiveLoader, (content) => {
       const parsedContent = parser.parseFromString(content, 'text/html');
-      contentLinks.reduceReferences(parsedContent, this.props.contentLinks);
+      const systemQueryString = queryString.stringify(systemQueryParams);
+      contentLinks.reduceReferences(parsedContent, this.props.contentLinks, systemQueryString);
```
TomWoodward, line 47: `systemQueryParams` could be part of `this.props.contentLinks `

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -38,25 +38,25 @@ const reducePageReferenceError = (reference: PageReferenceError, document: Docum
...
   for (const reference of references) {
     // references may contain PageReferenceError only if UNLIMITED_CONTENT is set to true
     if (isPageReferenceError(reference)) {
       reducePageReferenceError(reference, document);
     } else {
-      reduceReference(reference, currentPath, document);
+      reduceReference(reference, currentPath, document, systemQueryString);
     }
   }
 };
```
TomWoodward, line 62: does the navigation from `contentLinkHandler` need to be updated to include persistent params such as search? (edit: this is working correctly, i'm not sure why, but it is.)

i think its better/correct to remove the search query from the actual in content link (as you have done already), so that if  you open in new tab it doesn't bring the search with you, that seems good

src/app/content/search/hooks/index.ts
```diff
@@ -77,6 +78,7 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
       search: queryString.stringify({
```
TomWoodward, line 78: do we still have the concept of persistent params? right this second i can't think of any other than the search ones... but i feel like having them as a construct wouldn't be bad. you could apply them at the beginning of this object and they'd get overridden as necessary

TomWoodward (approved):
i would like to address my comments separately but none of them are blocking merge. i'm going to merge this as is.

src/app/content/components/Page/contentLinkHandler.ts
bethshook: In-content links should still include the scroll `target` right? Just not `query`

bethshook: nvm, I get it

PR #1198, opened 2021-06-01
Update highlights cards position on container size change

jarosik10:
for: https://github.com/openstax/unified/issues/748

It requires https://github.com/openstax/typescript-types/pull/5 to be merged.


src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -37,6 +37,8 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
     () => highlights.find((highlight) => highlight.id === focusedId),
     [focusedId, highlights]);
   const prevFocusedHighlightId = React.useRef(focusedId);
+  // tslint:disable-next-line: no-empty
+  const setNewCardsPositionsRef = React.useRef(() => {});
```
PiotrKozlowski, line 41: will it require tslint disable if you return `null` or `undefined`?

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -68,23 +70,24 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+  }, [container]);
 
-  React.useEffect(() => {
+  const setNewCardsPositions = React.useCallback(() => {
     const positions = updateCardsPositions(focusedHighlight, highlights, cardsHeights, getOffsetsForHighlight);
     setCardsPositions(positions);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [highlights, cardsHeights, focusedHighlight]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [cardsHeights, focusedHighlight, highlights]);
```
PiotrKozlowski, line 85: I think it is okay to add `getOffsetsForHighlight` to the deps array since now it will change only once per container

PiotrKozlowski (changes requested):
I can see that there is one more bug related to Show/Hide Solutions.
When you expand the solution container and create a highlight inside of it and then collapse the solution, the highlight will float at the position at which it was created. The fix for this should be probably as easy as "if offset for highlight wasn't found then use -1000px" - it should hide the card and the proper value will be recalculated when the user expands the solution

src/app/content/highlights/components/CardWrapper.tsx
jarosik10: it causes the ts error: `Type '() => void' is not assignable to type '() => undefined'.`

PiotrKozlowski: You can type the ref to `() => void | undefined` or something

jarosik10:
> The fix for this should be probably as easy as "if offset for highlight wasn't found then use -1000px" - it should hide the card and the proper value will be recalculated when the user expands the solution

It's not like the offset for highlight cannot be found. It just doesn't change its position.

PiotrKozlowski:
@jarosik10 Ugh, it seems like solutions are hidden with `overflow: hidden; height: 0;` so the rects for the highlight may still be found.  
You'd have to find a way to tell if the highlight is in solution which is not expanded. Maybe by modifying `getHighlightOffset` to accept an argument with selectors in which highlights are treated as hidden

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -9,13 +9,32 @@ import { assertDefined } from '../../../utils/assertions';
...
+) => {
   if (!container || !highlight.range || !highlight.range.getBoundingClientRect) {
     return;
   }
 
   const {top, bottom } = highlight.range.getBoundingClientRect();
 
+  if (expandableAncestorsSelector) {
+    const highlightElement = highlight.elements[0] as HTMLElement;
+    const expandableAncestor = highlightElement && highlightElement.closest(expandableAncestorsSelector);
```
jarosik10, line 25: @TomWoodward  @PiotrKozlowski 
Do we want to handle expandable containers nested in other expandable container?

PiotrKozlowski: I think we have any so I thnk we don't care about them

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -9,13 +9,32 @@ import { assertDefined } from '../../../utils/assertions';
 import { cardMarginBottom } from '../constants';
 import { HighlightData } from '../types';
 
-export const getHighlightOffset = (container: HTMLElement | undefined, highlight: Highlight) => {
+export const getHighlightOffset = (
+  container: HTMLElement | undefined,
+  highlight: Highlight,
+  expandableAncestorsSelector?: string
```
PiotrKozlowski, line 15: I think this doesn't have to be a parameter and can be hardcoded in this function
also, you could use the `[data-type="solution"][aria-expanded="false"]` to narrow it down only to sections that are collapsed and then you don't have to calculate their getBoundingClientRect

PiotrKozlowski (changes requested):

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -11,28 +11,26 @@ import { HighlightData } from '../types';
...
-        top: -9999,
-      };
-    }
+  const highlightElement = highlight.elements[0] as HTMLElement;
+  const expandableAncestor = highlightElement
+    && highlightElement.closest('[data-type="solution"][aria-expanded="false"]');
+
+  if (expandableAncestor
+    && expandableAncestor.getBoundingClientRect().bottom
+    < highlightElement.getBoundingClientRect().top
```
PiotrKozlowski, line 28: do we have to check for this? the selector that you've used indicates that it is in a collapsed solution

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -11,28 +11,26 @@ import { HighlightData } from '../types';
...
+  ) {
+    return {
+      bottom: -9999,
```
PiotrKozlowski, line 31: this value could be a constants since it is used in `updateStackedCardsPositions`

src/app/content/highlights/components/cardUtils.spec.ts
```diff
@@ -2,49 +2,97 @@ import { Highlight } from '@openstax/highlighter';
...
   });
-});
 
-describe('generateUpdatePayload', () => {
-  const highlightId = 'my-id';
-  const oldData = {
-    annotation: 'my annotation',
-    color: HighlightColorEnum.Blue,
-  };
+  describe('generateUpdatePayload', () => {
```
PiotrKozlowski, line 53: formatting shouldn't be changed if the tests wasn't updated

src/app/content/highlights/components/cardUtils.spec.ts
```diff
@@ -2,49 +2,97 @@ import { Highlight } from '@openstax/highlighter';
...
-    const update = {annotation: 'hi'};
-    const payload = generateUpdatePayload(oldData, {id: highlightId, ...update});
+      expect(payload.preUpdateData.highlight).toEqual(oldData);
+      expect(payload.updatePayload.highlight).toEqual({ ...update, color: oldData.color });
+    });
+  });
 
-    expect(payload.preUpdateData.highlight).toEqual(oldData);
-    expect(payload.updatePayload.highlight).toEqual({...update, color: oldData.color});
+  describe('updateStackedCardsPositions', () => {
```
PiotrKozlowski, line 85: I think it would make more sense to mock the actual behavior in CardWrapper.spec.tsx by passing for example 3 highlights and putting one of them into the collapsed container

PiotrKozlowski (changes requested):

src/app/content/highlights/components/cardUtils.ts
jarosik10: You are right, we don't need this check. I was thinking about a container which height in collapsed state > 0, but it `solution` s we don't have to handle this case.

src/app/content/highlights/components/cardUtils.spec.ts
PiotrKozlowski: there are tests that checks if cards positions were updated so you can use them as references

jarosik10:
```
jest.mock('./cardUtils', () => ({
  ...jest.requireActual('./cardUtils'),
  getHighlightOffset: jest.fn(() => ({ top: 100, bottom: 100 })),
}));
``` 
There is a `getHighlightOffset` mock in `CardWrapper.spec.tsx`.

jarosik10: I could just mock `getHighlightOffset` again but it won't let me test the actual behavior of group of highlights.

PiotrKozlowski: you could try to indetify which tests needs this mock and mock this function only in those tests

PiotrKozlowski (approved):

src/app/content/highlights/components/CardWrapper.spec.tsx
```diff
@@ -85,7 +81,37 @@ describe('CardWrapper', () => {
...
   });
 
+  it('returns hidden higlight position without adjusting the offset', () => {
+    const element = assertDocument().createElement('div');
+    const highlight = {
+      ...createMockHighlight(),
+      elements: [element],
+    };
+    const element2 = assertDocument().createElement('div');
+    const expandedAncestor = assertDocument().createElement('div');
```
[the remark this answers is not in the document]
jarosik10: fixed :D

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -66,23 +67,36 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+    const newOffsets = assertDefined(
+      getHighlightOffset(container, highlight),
+      `Couldn't get offsets for highlight with an id: ${highlight.id}`
+    );
+    setOffsets((state) => new Map(state).set(highlight.id, newOffsets));
+    return newOffsets;
+  }, [container]);
 
-  React.useEffect(() => {
+  const setNewCardsPositions = React.useCallback(() => {
```
TomWoodward, line 78: i think you can avoid one of your assertions if you change
```suggestion
  const setNewCardsPositions = setNewCardsPositionsRef.current = React.useCallback(() => {
```

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -66,23 +67,36 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
     const positions = updateCardsPositions(focusedHighlight, highlights, cardsHeights, getOffsetsForHighlight);
     setCardsPositions(positions);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [highlights, cardsHeights, focusedHighlight]);
+  }, [cardsHeights, focusedHighlight, getOffsetsForHighlight, highlights]);
+
+  React.useEffect(() => {
+    setNewCardsPositions();
+    setNewCardsPositionsRef.current = setNewCardsPositions;
```
TomWoodward, line 85: you also won't need to do this anymore

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -66,23 +67,36 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+  }, [setNewCardsPositions]);
```
TomWoodward, line 86: really probably you should just define the function into the ref as not-a-callback and then put the `[cardsHeights, focusedHighlight, getOffsetsForHighlight, highlights]` part in this useEffect

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -9,13 +9,26 @@ import { assertDefined } from '../../../utils/assertions';
...
+export const hiddenHighlightOffset = -9999;
+
 export const getHighlightOffset = (container: HTMLElement | undefined, highlight: Highlight) => {
   if (!container || !highlight.range || !highlight.range.getBoundingClientRect) {
     return;
   }
 
   const {top, bottom } = highlight.range.getBoundingClientRect();
 
+  const highlightElement = highlight.elements[0] as HTMLElement;
```
TomWoodward, line 21: i think this might be undefined, what is the type that you get out of the library? is it not already HTMLElement?

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -9,13 +9,26 @@ import { assertDefined } from '../../../utils/assertions';
...
+  const expandableAncestor = highlightElement
+    && highlightElement.closest('[data-type="solution"][aria-expanded="false"]');
```
TomWoodward, line 23: this seems super brittle, wouldn't it be just as good to check to see if the current highlight was hidden or not?

TomWoodward (changes requested):
i like the resizeObserver part maybe pending some re-org of that callback, but i'm not sure about the handling of the hidden highlights. i'll post some more thoughts once you answer the questions and i understand more what is going on

src/app/content/highlights/components/cardUtils.ts
jarosik10: it has `elements: HTMLElement[];` type.
However without the casting `HTMLElement` (imported from `'openstax/types/lib.dom`), ts throws error: `Property 'closest' does not exist on type 'HTMLElement'.`.

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -9,13 +9,26 @@ import { assertDefined } from '../../../utils/assertions';
...
+
+  if (expandableAncestor) {
+    return {
+      bottom: hiddenHighlightOffset,
```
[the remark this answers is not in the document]
jarosik10: Instead of calculating the offset including parent/scroll offesets, it just returns -9999, which will hide the highlight offscreen. 
Before this change, it would return the offsets of the highlight normally. Even tho, that highlight might have been hidden by collapsed solution wrapper.

src/app/content/highlights/components/cardUtils.ts
```diff
@@ -94,6 +107,11 @@ const updateStackedCardsPositions = (
   for (const [index, highlight] of highlightsElements.entries()) {
     const topOffset = getHighlightPosition(highlight).top;
 
+    if (topOffset === hiddenHighlightOffset) {
+      positions.set(highlight.id, topOffset);
```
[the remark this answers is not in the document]
PiotrKozlowski: Yes, that was actually my idea, but now I think it has one disadvantage - it will be confusing when using a keyboard to navigate between cards. We should probably hide it also with CSS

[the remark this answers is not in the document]
jarosik10: Yup, it shows it offscreen and it prevents changing the position through relationships among others highlights.
The side effect is that opening the collapsed wrapper makes the hidden highlights to fall from the top of the screen.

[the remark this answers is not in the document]
jarosik10: > We should probably hide it also with CSS

that could be the solution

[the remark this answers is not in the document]
TomWoodward: i would prefer that these cards be hide/shown and have the same animation coming in and out that they normally do

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -66,23 +67,42 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+      `Couldn't get offsets for highlight with an id: ${highlight.id}`
+    );
+    setOffsets((state) => new Map(state).set(highlight.id, newOffsets));
+    return newOffsets;
+  }, [container]);
+
+  const checkIfHiddenByCollapsedAncestor = (highlight: Highlight) => {
+    const highlightElement = highlight.elements[0] as HTMLElement;
+    const collapsedAncestor = highlightElement
+      ? highlightElement.closest('[data-type="solution"][aria-expanded="false"]')
```
TomWoodward, line 81: you could check in general for visibility, instead of specifically for collapsed solutions. there are a couple ways of doing that

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -66,23 +67,42 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
...
+  }, [cardsHeights, focusedHighlight, getOffsetsForHighlight, highlights]);
 
   React.useEffect(() => {
-    const positions = updateCardsPositions(focusedHighlight, highlights, cardsHeights, getOffsetsForHighlight);
-    setCardsPositions(positions);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [highlights, cardsHeights, focusedHighlight]);
+    const resizeObserver = new ResizeObserver(() => {
+      assertDefined(
+        setNewCardsPositionsRef.current,
```
TomWoodward, line 97: could the function itself be assigned to the ref outside of the effect? so that its guaranteed to be defined? i don't see why it needs to be defined in the effect

src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -98,6 +118,7 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
           onHeightChange={(ref: React.RefObject<HTMLElement>) => onHeightChange(highlight.id, ref)}
           zIndex={highlights.length - index}
           shouldFocusCard={focusThisCard}
+          isHidden={checkIfHiddenByCollapsedAncestor(highlight)}
```
TomWoodward, line 121: i think in the existing logic, the card position will end up as like 0 or undefined or something, you could check for that instead of making a new thing

TomWoodward (changes requested):

src/app/content/highlights/components/CardWrapper.tsx
jarosik10: I could recursively look for a parent with overflow: hidden and check if
```js
highlightSpan.getBoundingClientRect().bottom > containerWithOverflowHidden.getBoundingClientRect().bottom
```
but it seems kinda clanky.

TomWoodward: the problem is that we're styling the solution containers with `overflow: hidden; height: 0`, i'm not sure why we're doing that, if you use either `display: none` or `visibility: hidden` then `node.offsetParent` will be null when it isn't visible

jarosik10: @TomWoodward 
I've found the explanation:
```js
// hide text visually, but make rangy still search through it
// when highlighting a search result
// https://github.com/timdown/rangy/wiki/Text-Range-Module#visible-text
const initialSolutionStyles = 'display: block; overflow: hidden; height: 0px';
```

src/app/content/highlights/components/CardWrapper.tsx
jarosik10: @TomWoodward 
with the existing logic the card offset is normally calculated to make the cards appear normally (not jump from the top or anything) when become visible. So it'd say that `isHidden` is still needed.

jarosik10: This actually might need a change. I've just found a bug with wrong offset adjusting.

src/app/content/highlights/components/CardWrapper.tsx
TomWoodward: ugh... if we changed the solution expander to use details/summary block, would highlights still work then? this solution wouldn't require javascript and has been proposed to be moved upstream to content generation. at first glance i don't see display block or visibility hidden on the collapsed details elements

jarosik10:
@TomWoodward  I've made show solution to use summary+details elements. I've posted my thoughts [here](https://openstax.slack.com/archives/C8V1XET1R/p1627659282018400).

TomWoodward:
@jarosik10 i haven't looked at the code yet, but looking at the deployment this solution with the details/summary looks like it is working well, i don't see what you were talking about in slack with the slower response to opening/closing, can you give more details on that?

TomWoodward (changes requested):
tests failing, need more details on implementation concerns

jarosik10:
@TomWoodward 
> i haven't looked at the code yet, but looking at the deployment this solution with the details/summary looks like it is working well, i don't see what you were talking about in slack with the slower response to opening/closing, can you give more details on?

You are right, it seems to be working just fine. This slower response to opening/closing had to be a temporary, local issue.

src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -60,11 +60,10 @@ const selectResult = (services: Services, previous: HighlightProp, current: High
...
     // If selected result is in the collapsed solution then expand it
-    const solution = (firstSelectedHighlight.elements[0] as HTMLElement)
-      .closest('[data-type="solution"][aria-expanded="false"]');
-    if (solution) {
-      const button = solution.querySelector('button');
-      button?.click();
+    const details = (firstSelectedHighlight.elements[0] as HTMLElement)
+      .closest('[data-type="solution"][aria-expanded="false"]')?.querySelector('details');
+    if (details) {
+      details.open = true;
```
bethshook, line 66: You also need to add `.ui-solution-visible` to the solution so the show/hide text is correct.

src/app/content/highlights/components/CardWrapper.tsx
bethshook: seems like checking for general visibility should work now that you're using the details element

src/app/content/components/Page/searchHighlightManager.ts
jarosik10: @bethshook 
But `.ui-solution-visible` is getting added.
https://github.com/openstax/rex-web/blob/d0c424b002e226ebc23579b3de5cb08361add1a3/src/app/content/components/Page/contentDOMTransformations.ts#L19-L36
https://github.com/openstax/rex-web/blob/d0c424b002e226ebc23579b3de5cb08361add1a3/src/app/content/components/Page/PageComponent.tsx#L189-L193

I'll change the names from `button` to `summayElement` to make it less misleading.

But maybe i'm gonna refactor the toggling code to make it depended on the state of `open` prop of the `HTMLDetailsElement`.

bethshook: That class is added when details opens for highlighted search results? When I tested on the PR env the show/hide text was wrong.

jarosik10: This class was added on click of summary element (button). Not it is added on toggle event. 
`details.open = true;` should trigger that event, so it should work now.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -195,10 +195,10 @@ export default class PageComponent extends Component<PagePropTypes> {
...
     });
 
-    mapSolutions(this.container.current, (button) => {
-      const handler = toggleSolution(button, this.props.intl);
-      this.clickListeners.set(button, handler);
-      button.addEventListener('click', handler);
+    mapSolutions(this.container.current, (summary) => {
+      const handler = toggleSolution(summary, this.props.intl);
+      this.clickListeners.set(summary, handler);
+      summary.addEventListener('toggle', handler);
```
TomWoodward, line 201: the `listenersOff` behavior is looking for `click` only and won't properly remove this

TomWoodward: actually we shouldn't need any js behavior to handle this at all, details/summary naturally handles expand collapse

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -16,43 +16,29 @@ export const transformContent = (document: Document, rootEl: HTMLElement, intl:
...
-  if (shouldBeVisible) {
-    section.style.height = 'auto';
-    section.style.overflow = 'visible';
-  } else {
-    section.style.height = '0px';
-    section.style.overflow = 'hidden';
-  }
-};
-
 const toggleSolutionAttributes = (solution: HTMLElement, intl: IntlShape) => {
```
TomWoodward, line 19: details/solution is naturally accessible, we shouldn't need to do any of this stuff

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -141,10 +122,10 @@ function wrapSolutions(rootEl: HTMLElement, intl: IntlShape) {
     el.setAttribute('aria-expanded', 'false');
     const contents = el.innerHTML;
     el.innerHTML = `
-      <div class="ui-toggle-wrapper">
-        <button class="btn-link ui-toggle" title="${title}"></button>
-      </div>
-      <section class="ui-body" role="alert" style="${initialSolutionStyles}">${contents}</section>
+      <details class="ui-toggle-wrapper">
+        <summary class="btn-link ui-toggle" title="${title}"></summary>
```
TomWoodward, line 126: for intl we need to handle the contents of the summary here instead of the css we currently have, i think probably just putting the title in there would be good

src/app/content/highlights/components/CardWrapper.tsx
TomWoodward: i'm not even totally sure rangy will search in collapsed details elements, so thats still an open question

TomWoodward (changes requested):

src/app/content/highlights/components/cardStyles.ts
```diff
@@ -97,10 +97,11 @@ const fadeInAnimation = css`
 `;
 
 export const mainCardStyles = css`
-  ${(props: CardProps) => !props.isActive && (!props.data || !props.data.annotation)
-    ? 'display: none;'
-    : 'display: block;'}
+  ${(props: CardProps) => (!props.isActive && (!props.data || !props.data.annotation)) || props.isHidden
+    ? 'visibility: hidden;'
+    : 'visibility: visible;'}
```
jivey, line 102: @TomWoodward Looks like `visibility: hidden` will continue to work like `display: none` and prevent if from being read by a screenreader: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns

TomWoodward:
@jivey can you resolve the conflicts with master?

generic-styles/content-common.less
```diff
@@ -295,33 +295,21 @@ blockquote {
...
 
-  // Exercise Solutions contain a button to toggle showing or hiding the solution
+  // Exercise Solutions contain a summary to toggle showing or hiding the solution
   [data-type="solution"],
   .solution {
     border-top: 0.1rem solid @gray;
 
     // The Show/Hide Solution text
-    > .ui-toggle-wrapper {
+    > .ui-toggle {
```
TomWoodward, line 304: we have to update the baked styles to match this as a separate change

https://github.com/openstax/cnx-recipes/blob/main/styles/designs/webview/parts/_exercise-components.scss#L56

TomWoodward (approved):

generic-styles/content-common.less
jivey: Fixed conflicts 👍 For the recipe — now that the solution is also the ui-toggle-wrapper, maybe that gets removed? I'm realizing now that it may be unnecessary to have the wrapper class if it isn't being targeted by any CSS. I'll take a look at the recipe code and get a better understanding for how it's used.

TomWoodward: the recipe styles are not currently used, that is something that we've been recently developing and hope to move over to in the near future, so we're in the awkward phase where style changes need to be maintained in two places. unfortunately bartek was doing most of the work on that and has now left, so we'll need some other members of the team to spin up on how the "baked styles" are set up

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -130,26 +86,25 @@ function fixLists(rootEl: HTMLElement) {
...
-        <button class="btn-link ui-toggle" title="${title}" data-content="${showSolutionText}"></button>
-      </div>
-      <section class="ui-body" role="alert" style="${initialSolutionStyles}">${contents}</section>
+    const detailsEl = document!.createElement('details');
+
+    Array.from(el.attributes).forEach((attr) => {
+      detailsEl.setAttribute(attr.name, attr.value);
+    });
+
+    detailsEl.classList.add('ui-toggle-wrapper');
```
bethshook, line 102: Is this still necessary?

jivey: I think you're right, after all the changes, I thought some CSS remained that targeted that class, but I just searched and can't find anything. I'll remove it and re-test to make sure.

jivey:
Baked styles look good, testing I noticed something I didn't catch before: in Safari it renders a disclosure triangle for a `summary` element:

image

Added a commit to remove that so it matches the Chrome style.

PR #1201, opened 2021-06-02
Do not update highlight managers multiple times when it can be done once

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1169


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -216,4 +221,26 @@ export default class PageComponent extends Component<PagePropTypes> {
...
+
+  /**
+   * When a user navigates quickly between pages there are multiple calls to componentDidUpdate
+   * and since it is an async function there might be still unresolved promisses that would result
+   * in calling highlighter.update multiple times, see: https://github.com/openstax/unified/issues/1169
+   */
+  private shouldUpdateHighlightManagers(prevProps: PagePropTypes, props: PagePropTypes): boolean {
+    const runId = this.getRunId();
+    // Always update highlighters for the first run
+    if (runId === 1) { return true; }
```
PiotrKozlowski, line 239: This is for some reason required for the test (`scrolls to search result when selected before page navigation`) to pass(`Page.spec.tsx`)

src/app/content/components/Page.spec.tsx
```diff
@@ -912,12 +908,9 @@ describe('Page', () => {
...
 
     // page lifecycle hooks
-    await Promise.resolve();
-    // previous processing
-    await Promise.resolve();
-    // after images are loaded
-    await Promise.resolve();
+    await new Promise((resolve) => setImmediate(resolve));
 
+    expect(highlightResults).not.toHaveBeenCalledWith(expect.anything(), []);
```
TomWoodward, line 913: what would this have been called with? this assertion is confusing me

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -67,6 +68,12 @@ export default class PageComponent extends Component<PagePropTypes> {
   }
 
   public async componentDidUpdate(prevProps: PagePropTypes) {
+    const runId = this.getRunId();
+
+    if (prevProps.page !== this.props.page) {
+      this.postProcess();
```
TomWoodward, line 74: why are we adding this here? where was it done before? it looks like this was only called on mount before, i don't understand how math was working on navigation.

also, i see listenersOff being called in `getSnapshotBeforeUpdate` if the page is different, but nothing re-applying it in the previous code. it was working though, so was something applying it? are you going to end up with multiple listeners?

src/app/content/components/Page/highlightManager.ts
```diff
@@ -191,6 +193,58 @@ export default (container: HTMLElement, getProp: () => HighlightProp, intl: Intl
     }
   };
 
+  const updateHighlighter = async(prevProps: HighlightProp, options?: UpdateOptions): Promise<boolean> => {
+    const typsettingCounter = Number(container.getAttribute(typesettingCounterAttribute)) || 0;
```
TomWoodward, line 197: could use dataset and avoid the type conversions

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -75,12 +82,13 @@ export default class PageComponent extends Component<PagePropTypes> {
...
-    if (prevProps.page !== this.props.page) {
-      await this.postProcess();
+    if (!this.shouldUpdateHighlightManagers(prevProps, this.props, runId)) {
+      return;
     }
 
-    const highlightsAddedOrRemoved = this.highlightManager.update(prevProps.highlights, {
+    const highlightsAddedOrRemoved = await this.highlightManager.update(prevProps.highlights, {
       onSelect: this.onHighlightSelect,
+      waitForMathTypesetting: true,
```
TomWoodward, line 91: why is this necessary? mathjax returns a promise, we wait for those promises in the `postProcess` function. is the problem that you're not `await`ing that call on line 74?

TomWoodward (changes requested):
this logic was a mess

src/app/content/components/Page/PageComponent.tsx
PiotrKozlowski: Changing the page is causing props to change a few times - at least 3 - first time is because `page` is set to `undefined`, the second is because it is being set to the new page and the third is because we fetch highlights

Since react is not waiting for a `componentDidUpdate` when it is async, we are starting at least 2 promises when navigating (and even more when there are search results involved)

Then, when we stop (after the third rerender), we are waiting for the two mathjax promises to resolve - but when each of them resolves, they will try to update highlighters - this is why I've added `shouldUpdateHighlightManagers` check

> why is this necessary? mathjax returns a promise, we wait for those promises in the postProcess function. is the problem that you're not awaiting that call on line 74?

This is the part that I don't really understand, but without it sometimes when navigating quickly all highlights fail to load.
I was thinking that maybe it was related to `this.processing` being only a last promise so I've changed it to that array of promises and the `await Promise.all(this.processing)` + remove a promise from this array when it resolves, but it didn't help.
I think that might be related to mathjax resolving a milliseconds before the browser renders the math, but idk

TomWoodward: ok so i'm starting to understand all the stuff you did re-arranging the update function and adding the counter, i think that makes sense, i still don't understand why this `waitForMathTypesetting` needs to be passed in though, shouldn't it be finished because of the `await this.processing` on line 81?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1203, opened 2021-06-08
additional caching and performance reporting

TomWoodward:
adds additional caching to gateways that didn't have it before, modifies all caches in pre-rendering to use file caches that could theoretically be persisted between builds, and outputs some slightly more interesting stats at the end of the build

for: openstax/unified#1613


TomWoodward:
@bethshook i think this is ready for review now

bethshook (approved):

PR #1204, opened 2021-06-08
match layout language to book language

bethshook:
for: openstax/unified#1552


src/app/MessageProvider.tsx
```diff
@@ -18,17 +22,23 @@ async function polyfill(locale: 'en') {
 
 polyfill('en');
```
bethshook, line 23: @TomWoodward should this now be called from within the component with the appropriate locale?

TomWoodward: yea, and probably use state to wait until it finishes before you render

TomWoodward:
you probably need to update the fixture book right?

bethshook:
@TomWoodward it's in there

src/app/MessageProvider.tsx
```diff
@@ -18,17 +22,23 @@ async function polyfill(locale: 'en') {
...
+    <RawIntlProvider value={intl}>
+      {props.children}
+    </RawIntlProvider>
+  ) : null;
+};
 
-export default MessageProvider;
+export default connect(
+  (state: AppState) => ({
+    book: select.book(state),
```
TomWoodward, line 41: its easier to use `userSelector` rather than connect now

src/gateways/createIntl.ts
```diff
@@ -0,0 +1,18 @@
+import { createIntl, createIntlCache } from 'react-intl';
+import enMessages from '../app/messages/en';
+import plMessages from '../app/messages/pl';
+
+export default () => {
+  return {
+    getIntlObject: (locale: string = 'en') => {
```
TomWoodward, line 7: how much would it suck to make this async? it'd be nice to do a dynamic `import` rather than loading all the messages for every language

src/test/createTestServices.ts
```diff
@@ -1,9 +1,9 @@
 import { SearchApi } from '@openstax/open-search-client';
 import { createMemoryHistory } from 'history';
-import { intl } from '../app/MessageProvider';
 import config from '../config';
 import { BuyPrintResponse } from '../gateways/createBuyPrintConfigLoader';
 import createHighlightClient from '../gateways/createHighlightClient';
+import createIntl from '../gateways/createIntl';
```
TomWoodward, line 6: gateways are supposed to be proxies to remote or third party services, i'd put `createIntl` in `messages`, probably move `MessageProvider` in there too

TomWoodward (changes requested):

src/gateways/createIntl.ts
bethshook: so I have this working but the catch is now a whole lot of tests have to be made async (which is maybe what you were alluding to)

TomWoodward: yea i can see how a lot of tests would be broken... i'm struggling to think of a way to make this better without that problem.

i definitely don't want to be loading every language file and then picking one, and since we don't know the language until the book loads i'm not sure if there is any way around it

src/app/MessageProvider.tsx
```diff
@@ -1,34 +1,47 @@
...
-    await import('@formatjs/intl-pluralrules/polyfill');
-  }
+// tslint:disable-next-line:variable-name
+const MessageProvider = (props: { children?: React.ReactNode }) => {
+  const book = useSelector(bookSelector);
+  const currentPath = useSelector(pathSelector);
+  const [polyfillLoaded, setPolyfillLoaded] = useState(false);
+
+  // https://formatjs.io/docs/polyfills/intl-pluralrules/#dynamic-import--capability-detection
+  async function polyfill(locale: string | undefined) {
```
TomWoodward, line 16: this doesn't have to be defined in the function, you can `useEffect` with the locale as a dependency. something like:

```
useEffect(async () => {
  await polyfill(locale)
  setPolyfillLoaded(true);
}, [locale, polyfill])
```

src/app/MessageProvider.tsx
```diff
@@ -1,34 +1,47 @@
...
+    if ((Intl.PluralRules as (typeof Intl.PluralRules & { polyfilled?: boolean })).polyfilled) {
+      await import(`@formatjs/intl-pluralrules/locale-data/${locale}`);
+      setPolyfillLoaded(true);
+    }
   }
-}
 
-polyfill('en');
+  const lang = React.useMemo(() => {
+    return currentPath === '/' ? 'en' : book?.language;
```
TomWoodward, line 33: i thought you were gonna put something in the route or something?

src/app/MessageProvider.tsx
```diff
@@ -1,34 +1,47 @@
...
-  messages: enMessages,
-}, cache);
+  polyfill(lang);
 
-// tslint:disable-next-line:variable-name
-const MessageProvider: React.FC = (props) =>
-  <RawIntlProvider value={useServices().intl}>
-    {props.children}
-  </RawIntlProvider>;
+  return (lang && !shouldPolyfill()) || (lang && polyfillLoaded) ? (
```
TomWoodward, line 40: if the polyfill function was separate you wouldn't need the extra condition here

TomWoodward (changes requested):

src/app/messages/MessageProvider.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+      services = createTestServices();
+      (Intl as any).PluralRules.polyfilled = true;
+      store.dispatch(receiveBook(book));
+    });
+
+    afterEach(() => {
+      delete (Intl as any).PluralRules.polyfilled;
+    });
+
+    it('loads polyfill', async() => {
```
bethshook, line 61: @TomWoodward this is the test suite I'm updating. [here](https://github.com/openstax/rex-web/blob/master/src/app/MessageProvider.spec.tsx) you can see the original. I'm rendering the provider since `polyfill()` is now called from inside the component, but I think it's ignoring the polyfill module mocks.

PiotrKozlowski (dismissed):
The code looks good, conflicts need to be resolved and I think there will be some lint errors due to comments before `runHooks` util function, since they are not longer required

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import uuid from 'uuid/v4';
-import MessageProvider from '../../../../MessageProvider';
+import MessageProvider from '../../../../messages/MessageProvider';
 import { assertDocument, assertNotNull } from '../../../../utils';
 import ConfirmationModal from '../ConfirmationModal';
```
bethshook, line 6: @TomWoodward I will need Bartek's changes after all

src/app/messages/createIntl.ts
```diff
@@ -0,0 +1,26 @@
...
+    getIntlObject: async(locale: string) => {
+        const cache = createIntlCache();
+        let messages;
+
+        try {
+          messages = await require(`./${locale}`).default;
+        } catch (e) {
+          messages = enMessages;
+          Sentry.captureException(e);
+        }
```
bethshook, line 16: @TomWoodward is this what you had in mind?

src/app/content/components/BookBanner.tsx
```diff
@@ -213,7 +213,7 @@ const BookBanner = () => {
     }
     e.preventDefault();
 
-    if (!await showConfirmation(services)) {
+    if (!await showConfirmation(intl)) {
```
TomWoodward, line 216: this is gonna conflict with bartek's thing, leave it as services

src/app/content/components/Content.browserspec.tsx
```diff
@@ -9,8 +9,8 @@ const TEST_CASES: { [testCase: string]: (target: Page) => Promise<void> } = {
   Desktop: setDesktopViewport, Mobile: setMobileViewport,
 };
 const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
-  Desktop: [235, 90, 122, 235, 324, 554, 641, 1142, 1448],
-  Mobile: [209, 66, 96, 209, 418, 968, 1075, 1424, 1750],
+  Desktop: [236, 90, 122, 236, 324, 554, 641, 1143, 1449],
+  Mobile: [210, 66, 96, 210, 418, 968, 1075, 1425, 1751],
```
TomWoodward, line 13: is this misplaced from the other pr?

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -1,12 +1,12 @@
...
 import ReactDOM from 'react-dom';
 import { RawIntlProvider } from 'react-intl';
+import { IntlShape } from 'react-intl';
 import uuid from 'uuid/v4';
-import { AppServices } from '../../../../types';
 import { assertDocument, assertNotNull } from '../../../../utils';
 import ConfirmationModal from '../ConfirmationModal';
 
-export default async(services: AppServices) => {
+export default async(intl: IntlShape) => {
```
TomWoodward, line 9: so i think you can actually just call createIntl and not even bother taking the intl argument, but bartek is gonna need dispatch so the services are gonna end up still in here

src/app/messages/createIntl.ts
```diff
@@ -0,0 +1,27 @@
+import { createIntl, createIntlCache } from 'react-intl';
+import Sentry from '../../helpers/Sentry';
+import enMessages from '../messages/en/index';
+
+export default () => {
+  return {
+    getIntlObject: async(locale: string) => {
```
TomWoodward, line 7: i think i am with you that this doesn't have to be in services anymore. you can just export a function for `createIntl` and call it from anywhere, but its important that we not be re-loading the messages all the time, so you can check to see if the module system caches it, i can't remember if it does that automatically for inline imports. if it doesn't cache it by default you can wrap the whole definition in `lodash/fp/memoize`

TomWoodward (changes requested):

src/app/content/components/Content.browserspec.tsx
bethshook: no, tests are failing mysteriously. I'll make the other changes and then deal with them

src/app/content/components/BookBanner.tsx
bethshook: does it need services other than intl? can I pass them in separately since we're talking about removing intl from services?

bethshook: nvm, saw your other comment

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -2,13 +2,17 @@ import React from 'react';
...
 import uuid from 'uuid/v4';
+import createIntl from '../../../../messages/createIntl';
 import { AppServices } from '../../../../types';
 import { assertDocument, assertNotNull } from '../../../../utils';
 import ConfirmationModal from '../ConfirmationModal';
 
 export default async(services: AppServices) => {
   const document = assertDocument();
   const domNode = document.createElement('div');
+  const intl = await createIntl().getIntlObject('en');
```
jarosik10, line 13: `ConfirmationModal` will always be in English?

jarosik10 (changes requested):

src/app/types.ts
```diff
@@ -58,6 +57,11 @@ export interface AppServices {
   bookConfigLoader: ReturnType<typeof createBookConfigLoader>;
 }
 
+export interface AppServicesAndReduxMiddleware extends AppServices {
+  dispatch: Dispatch;
+  store: Store;
+}
+
```
jarosik10, line 64: I suppose it should be deleted.

jarosik10 (dismissed):

src/app/content/highlights/components/utils/showConfirmation.tsx
bethshook: just a placeholder, sorry. will ping you once the `showConfirmation` logic is ready.

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -2,13 +2,16 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { RawIntlProvider } from 'react-intl';
 import uuid from 'uuid/v4';
-import { AppServices } from '../../../../types';
+import createIntl from '../../../../messages/createIntl';
 import { assertDocument, assertNotNull } from '../../../../utils';
 import ConfirmationModal from '../ConfirmationModal';
 
-export default async(services: AppServices) => {
+export default async(services: any) => {
```
bethshook, line 9: @TomWoodward This is a placeholder as I'm confused about how to deal with typing given the change to `useServices`. Should I update create a new type that extends AppServices to include the redux middleware?

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -59,13 +59,14 @@ export default class PageComponent extends Component<PagePropTypes> {
     });
   };
 
-  public componentDidMount() {
+  public async componentDidMount() {
```
TomWoodward, line 62: this could be bad making this async, does it have to be async?

src/app/content/highlights/components/utils/showConfirmation.tsx
TomWoodward: you could just type this here as `AppServices & MiddlewareAPI`, pretty sure that is what the redux effect type does

src/app/content/utils/seoUtils.ts
```diff
@@ -110,7 +111,7 @@ export const getPageDescription = (services: Pick<AppServices, 'archiveLoader' |
   return pageDescription || intl.formatMessage({id: 'i18n:metadata:description'});
 };
 
-export const createTitle = (page: Page, book: Book, intl: IntlShape): string => {
+export const createTitle = async(page: Page, book: Book, intl: IntlShape): Promise<string> => {
```
TomWoodward, line 114: did you have to actually change these?

src/app/messages/MessageProvider.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+// tslint:disable-next-line:variable-name
+const MessageProvider = (props: { children?: React.ReactNode }) => {
+  const book = useSelector(bookSelector);
+  const route = useSelector(matchSelector)?.route;
+  const [polyfillLoaded, setPolyfillLoaded] = useState(false);
+  const [intl, setIntl] = useState<IntlShape | null>(null);
+
+  const bookLocale = React.useMemo(() => {
+    return route?.locale || book?.language;
```
TomWoodward, line 30: technically the concept of books is out of scope this high up in the app, but i'm not sure i care enough to fix it. the right encapsulation would be to defer to the route and have the content route do something with the book, but i would rather get this merged than refine this

src/app/messages/createIntl.ts
```diff
@@ -0,0 +1,24 @@
+import memoize from 'lodash/fp/memoize';
+import { createIntl, createIntlCache } from 'react-intl';
+import Sentry from '../../helpers/Sentry';
+import enMessages from '../messages/en/index';
```
TomWoodward, line 4: you're importing the english messages unconditionally here, you need to defer this to the catch and if that also fails just explode everywhere

src/app/types.ts
```diff
@@ -58,6 +57,11 @@ export interface AppServices {
   bookConfigLoader: ReturnType<typeof createBookConfigLoader>;
 }
 
+export interface AppServicesAndReduxMiddleware extends AppServices {
```
TomWoodward, line 60: this is the same as `AppServices & MiddlewareAPI`

TomWoodward (changes requested):

src/app/messages/MessageProvider.tsx
bethshook: ok, I'll make a note to create a ticket once this is merged

src/app/messages/createIntl.ts
```diff
@@ -0,0 +1,28 @@
...
+    messages = localeMessages.default;
+  } catch (e) {
+    const enMessages = await import(`./en/index`);
+    if (enMessages) {
+      messages = enMessages.default;
+    } else {
+      messages = null;
+    }
+    Sentry.captureException(e);
+  }
```
bethshook, line 20: @TomWoodward Is this what you had in mind for the catch? If `enMessages` are not imported then the app won't render.

src/test/MessageProvider.tsx
```diff
@@ -0,0 +1,19 @@
...
+// tslint:disable-next-line:variable-name
+const MessageProvider = ({children, ...props}: React.PropsWithChildren<Props>) => {
+  const intlObject = createIntl(props.locale, props.messages);
+
+  return <RawIntlProvider value={intlObject}>
+    {children}
+  </RawIntlProvider>;
+};
+
+export default MessageProvider;
```
bethshook, line 19: @TomWoodward I removed the polyfill stuff from this test component since I use the real MessageProvider when testing that

src/app/content/hooks/receiveContent.ts
```diff
@@ -31,7 +35,11 @@ const hookBody: ActionHookBody<typeof receivePage> = (services) => async() => {
     return;
   }
 
-  const title = createTitle(page, book, intl);
+  if (!intlObj) {
+    intlObj = await createIntl(locale);
+  }
```
bethshook, line 40: @TomWoodward this is what I mentioned re `services.intl.current` not being defined yet when this hook runs

TomWoodward (approved):

Bkblodget:
@lenka222 the "symbol" issue you found above has been resolved in the latest version of content which can be viewed [here](https://rex-web.herokuapp.com/books/728df0bb-e07f-489d-91e3-4734a5932f92@1.122/pages/skorowidz-rzeczowy?archive=https%3A%2F%2Fopenstax.org%2Fapps%2Farchive-preview%2F20210804.143519).

The CE Styles team is also opening up an issue to update the recipe to display "Symbol" in Polish, just in case if in the future there does need to be a Symbols section in the index.

PR #1208, opened 2021-06-11
1343 Store color filters for SG in the URL

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1343


src/app/content/studyGuides/reducer.ts
```diff
@@ -30,9 +31,17 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
-        && action.payload.action === 'PUSH';
+        && (action.payload.action === 'PUSH' || action.payload.action === 'REPLACE');
+      const colors = getColorFiltersFromQuery(action.payload.query) || state.summary.filters.colors;
 
-      return {...state, summary: {...state.summary, open: summaryShouldBeOpen}};
+      return {
+        ...state,
+        summary: {
+          ...state.summary,
+          filters: {...state.summary.filters, colors},
```
TomWoodward, line 41: we want to remove the reduction of `updateSummaryFilters` below, and rely on this instead

src/app/navigation/utils.ts
```diff
@@ -128,7 +128,7 @@ export const findPathForParams = (params: object, paths: string[]) => {
   });
 };
 
-export const getQueryForParam = (param: string, value: string, existingQuery?: string | OutputParams) => {
+export const getQueryForParam = (param: string, value: string | string[], existingQuery?: string | OutputParams) => {
```
TomWoodward, line 131: how does this library encode arrays? as i remember it there is no super standard way to do that

TomWoodward (changes requested):

src/app/navigation/utils.ts
PiotrKozlowski: it duplicates the params, so it looks like `colors=gree&colors=yellow`

PiotrKozlowski: This seems to be the default behaviour
```js
const search = new URLSearchParams('')
search.append('colors', 'green')
search.append('colors', 'yellow')
search.toString() // "colors=green&colors=yellow"
```

PiotrKozlowski: There is also `getAll`:
```js
search.getAll('colors') // Array [ "green", "yellow" ]
```

src/app/content/studyGuides/hooks/index.ts
```diff
@@ -28,4 +29,5 @@ export default [
   actionHook(openStudyGuides, openModal(modalUrlName)),
   actionHook(closeStudyGuides, closeModal),
```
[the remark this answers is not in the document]
PiotrKozlowski: `closeModal` is using `services.history.goBack` and changing filters is using `replace` so it should work as expected
I'll however update `openModal` hook to add current filters to the query when opening again

jarosik10 (changes requested):
Can't uncheck the only left chapter (it gets deleted from url).
To reproduce:
1. Open SG (for example: american-government-2e/pages/1-introduction)
2. My filters: 
- chapters:  American Government and Civic Engagement [checked]
- colors: Key Terms (yellow) [checked]
So my url is: `.../books/american-government-2e/pages/1-introduction?colors=purple&colors=yellow&locationIds=8e9fcde2-08ea-5a75-8cc6-c43711e2e0f3&modal=SG`
3. I click to uncheck American Government and Civic Engagemen chapter filter
It doesn't uncheck at all, but the url change to `.../books/american-government-2e/pages/1-introduction?colors=purple&colors=yellow&modal=SG`

And `purple` color is in the url even if that checkbox is disabled and it can't be clicked. Not sure if it is intended.

PiotrKozlowski:
omg, I also forgot about loading more highlights by scrolling that can change filters...

jarosik10 (changes requested):
What about "_And purple color is in the url even if that checkbox is disabled and it can't be clicked_"?

src/app/content/studyGuides/hooks/loadMore.ts
```diff
@@ -40,11 +41,14 @@ export const loadMore = async(services: MiddlewareAPI & AppServices, pageSize?:
...
-  typeof actions.setDefaultSummaryFilters |
+  typeof locationChange |
   typeof actions.setSummaryFilters |
-  typeof actions.updateSummaryFilters |
   typeof actions.loadMoreStudyGuides
 > = (services) => async() => {
+  const {book, page} = bookAndPage(services.getState());
+
+  if (!book || !page) { return; }
+
```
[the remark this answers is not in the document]
PiotrKozlowski: I think so, because it is now attached to the `locationChange` which happens before `book` or `page` is loaded
This probably is happening only on the testing env

PiotrKozlowski:
> What about "And purple color is in the url even if that checkbox is disabled and it can't be clicked"?

Thanks, fixed

jarosik10 (approved):

src/app/content/studyGuides/reducer.ts
```diff
@@ -29,10 +28,22 @@ export const initialState: State = {
 const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
   switch (action.type) {
     case getType(locationChange): {
-      const summaryShouldBeOpen = action.payload.query[modalQueryParameterName] === modalUrlName
-        && action.payload.action === 'PUSH';
+      const hasModalQuery = action.payload.query[modalQueryParameterName] === modalUrlName;
+      const summaryShouldBeOpen = hasModalQuery
+        && (action.payload.action === 'PUSH' || action.payload.action === 'REPLACE');
```
bethshook, line 33: @TomWoodward is there a way to differentiate between page reload and the back button here? we discussed this a few weeks ago and I think you had a suggestion for keeping the modal open on reload.

src/app/content/studyGuides/hooks/locationChange.ts
```diff
@@ -47,6 +47,7 @@ export const hookBody = (services: MiddlewareAPI & AppServices) => async() => {
   const totalCounts = extractTotalCounts(countsPerSource);
 
   services.dispatch(receiveStudyGuidesTotalCounts(totalCounts));
+  services.dispatch(loadMoreStudyGuides());
```
bethshook, line 50: @TomWoodward this ensures study guides are populated on load

src/app/content/studyGuides/selectors.ts
```diff
@@ -47,7 +47,10 @@ export const totalCountsPerPageOrEmpty = createSelector(
 
 export const studyGuidesOpen = createSelector(
   studyGuidesSummary,
-  (summary) => summary.open
+  studyGuidesEnabled,
+  parentSelectors.book,
+  parentSelectors.page,
+  (summary, flagEnabled, book, page) => summary.open && flagEnabled && !!book && !!page
```
bethshook, line 53: @TomWoodward initially I checked for a non-empty `summary.studyGuides` but that causes the modal to rerender when filters are changed

TomWoodward (changes requested):

src/app/content/studyGuides/hooks/locationChange.ts
bethshook: 1. when I call it directly it noops because the book hasn't loaded yet
2. if it stops listening to the app locationChange then study guides aren't loaded when the modal is first opened

TomWoodward: on 2, why does it not get run from here?

TomWoodward (changes requested):

src/app/content/studyGuides/hooks/loadMore.ts
```diff
@@ -60,7 +64,6 @@ export const hookBody: ActionHookBody<
   services.dispatch(actions.receiveSummaryStudyGuides(formattedHighlights, {pagination, filters}));
 };
 
+export const locationChangeHook = actionHook(locationChange, hookBody);
 export const loadMoreHook = actionHook(actions.loadMoreStudyGuides, hookBody);
 export const setSummaryFiltersHook = actionHook(actions.setSummaryFilters, hookBody);
-export const updateSummaryFiltersHook = actionHook(actions.updateSummaryFilters, hookBody);
-export const setDefaultSummaryFiltersHook = actionHook(actions.setDefaultSummaryFilters, hookBody);
```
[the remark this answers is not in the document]
bethshook: doesn't that apply to both of them ([here](src/app/content/studyGuides/hooks/updateQueryWithSummaryFiltersHookBody.ts))?

src/app/content/studyGuides/reducer.spec.ts
```diff
@@ -34,26 +74,6 @@ describe('study guides reducer', () => {
...
-      },
-    }, actions.setSummaryFilters({
-      locationIds: ['id'],
-    }));
-
-    expect(state.summary.filters.locationIds[0]).toEqual('id');
-    expect(state.summary.filters.locationIds.length).toEqual(1);
-    expect(state.summary.loading).toEqual(true);
-  });
-
```
bethshook, line 56: @TomWoodward removed this because it's now unnecessary (as you pointed out) but am I correct that any test that uses this to set the filters will now need to render the app in order to set them via locationChange?

src/app/content/studyGuides/hooks/loadMore.ts
```diff
@@ -60,7 +65,7 @@ export const hookBody: ActionHookBody<
...
+export const receiveStudyGuidesTotalCountsHook = actionHook(actions.receiveStudyGuidesTotalCounts, hookBody);
```
bethshook, line 71: @TomWoodward I removed the loadMore action from studyguides/hooks/locationChange and instead hook loadMore into this action which solved the issue I had with populating study guides on reload

TomWoodward: i thought loadMore was triggered by the dialog itself when it opened? was the issue that it was opening before `receiveStudyGuidesTotalCounts` comes in? that is a problem because that response may indicate that there are no study guides for this book. maybe the modal open indicator needs to include that the counts exist.

separately, just make sure loadmore is correctly called when changing filters and stuff

src/app/content/studyGuides/reducer.spec.ts
TomWoodward: what is the difference between setSummaryFilters and updateSummaryFilters.

in response to the user changing the selected filters, you can dispatch locationchange directly or via a helper, thats what i'm referring to in my comment. there might be complicating factors though idk

src/app/content/studyGuides/hooks/loadMore.ts
bethshook: Yes, loadMore is triggered with the dialog is opened manually after page load. The problem is loading the page with the modal in the query—the `studyGuidesOpen` selector evaluates as false because the book/page has not loaded.

src/app/content/studyGuides/hooks/updateQueryWithSummaryFiltersHookBody.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  if (!match) { return; }
+
+  const updatedFilters = action.type === getType(actions.setDefaultSummaryFilters)
+    ? action.payload
+    : updateSummaryFilters(filters, action.payload);
+  services.dispatch(replace(match, {
+    search: updateQuery(updatedFilters as any as Record<string, string[]>, existingQuery),
+  }));
+};
```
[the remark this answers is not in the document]
bethshook: I've replaced the setDefaultSummaryFilters action with a query update but I'm not sure the best way to do the same for updateSummaryFilters because of shared logic with highlights, eg the ColorFilter component. We can take a look if you want.

src/app/content/studyGuides/hooks/loadMore.ts
TomWoodward: thats good right?

bethshook: at first yes but once the book/page loads shouldn't it update?

src/app/content/studyGuides/selectors.ts
```diff
@@ -81,21 +88,57 @@ export const loadedCountsPerSource = createSelector(
...
+  navigationSelectors.query,
+  (query) => getFiltersFromQuery(query)
+);
+
+export const unloggedAndQueryMissingFirstChapter = createSelector(
+  authSelectors.loggedOut,
+  parentSelectors.firstChapter,
+  filtersFromQuery,
+  (notLoggedIn, firstChapter, queryFilters) =>
+    notLoggedIn && firstChapter && !queryFilters.locationIds.includes(firstChapter.id)
```
TomWoodward, line 112: a logged out user can't change filters right?

src/app/content/selectors.ts
```diff
@@ -93,6 +95,11 @@ export const bookAndPage = createSelector(
   (selectedBook, selectedPage) => ({book: selectedBook, page: selectedPage})
 );
 
+export const firstChapter = createSelector(
+  book,
+  (selectedBook) => selectedBook && findArchiveTreeNode(archiveTreeSectionIsChapter, selectedBook?.tree)
```
TomWoodward, line 100: does `findArchiveTreeNode` actually handle a null second argument?

src/app/content/studyGuides/selectors.ts
```diff
@@ -81,21 +88,57 @@ export const loadedCountsPerSource = createSelector(
...
+  unloggedAndQueryMissingFirstChapter,
+  loggedAndQueryMissingLocationIds,
+  parentSelectors.firstChapter,
+  highlightColorFiltersWithContent,
+  defaultLocationFilter,
+  (state, unlogged, logged, firstChapter, colorsWithSG, defaultFilter) => unlogged && firstChapter
+    ? { colors: Array.from(colorsWithSG.size ? colorsWithSG : colorfilterLabels), locationIds: [firstChapter.id] }
+    : (logged && defaultFilter
+      ? { colors: Array.from(colorsWithSG.size ? colorsWithSG : colorfilterLabels), locationIds: [defaultFilter.id] }
+      : state.summary.filters)
```
TomWoodward, line 134: you're not using the user selected filters ever if the user is logged in ?

i think this should be first and all the other logic only happen if its null/empty right?

could you simplify logic by combining into a "defaultFIlters" selector that combines both color and location, doing your auth check in there since the default filters are different if you're logged in or not, and then in summary filters basically just go `=> ({...defaultFilters, ...state.summary.filters})` (btw you should select `state.summary.filters` in another selector so that you aren't re-generating this filters value whenever the whole SG state changes)

TomWoodward (changes requested):

src/app/content/studyGuides/selectors.ts
bethshook: correct

bethshook: hmm ok, looks like they still can via the query. will adjust.

src/app/content/studyGuides/components/Filters.spec.tsx
```diff
@@ -161,20 +225,27 @@ describe('Filters', () => {
...
     });
 
-    expect(dispatch).toHaveBeenCalledWith(updateSummaryFilters({
-      colors: { new: [], remove: [HighlightColorEnum.Yellow] },
-    }));
+    expect(dispatch).toHaveBeenCalledWith(
+      replace(mockMatch, {
+        search: `colors=green&colors=blue&colors=purple&locationIds=${chapter.id}&modal=SG`,
+      })
+    );
```
bethshook, line 232: @TomWoodward what do I need to add in order to trigger a `locationChange` here rather than `callHistoryMethod`? In the browser a change to the filters dispatches a locationChange which in turn updates the filters in state.

src/app/content/studyGuides/hooks/loadMore.spec.ts
```diff
@@ -173,16 +197,37 @@ describe('loadMore', () => {
 
     expect(loadingSpy).toHaveBeenCalled();
     expect(highlightClient).toHaveBeenCalledTimes(3);
-    expect(dispatch).lastCalledWith(receiveSummaryStudyGuides(response2, {pagination: null, filters}));
+    expect(dispatch).lastCalledWith(receiveSummaryStudyGuides(response2, {pagination: null}));
```
bethshook, line 200: @TomWoodward This is the other failing test. When I log the actions dispatched this is the last one but it never happens according to the test.

src/app/content/studyGuides/components/Filters.spec.tsx
bethshook: @jarosik10 maybe you can advise here?

TomWoodward: do you really need it to trigger the location change?

i think what is happening is that our convention is to dispatch the `callHistoryMethod` actions, and those are not reduced, the middleware uses them to trigger the actual history method, and then the navigation middleware listens to history actions and dispatches corresponding `changeToLocation` actions

so if you could separately test that the filters dispatch the right history action and then separately that it is reduced properly, that might be easier to do in this setup

the issue is that the test setup doesn't have the navigation middleware, so you really need it to work the same way, you'll have to rig something up to translate the actions from one to the other

src/app/content/studyGuides/hooks/loadMore.spec.ts
TomWoodward: when you run this test it doesn't think `receiveSummaryStudyGuides` has been called here? i don't see how that could be possible other than throwing an error or hitting the noop you added for the open flag. have you tried running just this one test to make sure there is no strange output? what does the state look like?

TomWoodward (changes requested):

TomWoodward:
@bethshook i've forgotten if there was any ongoing conversation about this PR, i see that some of my previous comments are still there, particularly that one about the snapshot change, but i'm not sure if they're still relevant.

in testing on the heroku env i don't see any issues, and the code looks ok, you can go ahead and put the card in review unless you have any outstanding concerns

src/app/navigation/hooks/closeModalHook.ts
```diff
@@ -1,3 +1,11 @@
 import { AppServices, MiddlewareAPI } from '../../types';
 
-export const closeModal = (services: MiddlewareAPI & AppServices) => services.history.goBack;
+export const closeModal = () => (services: MiddlewareAPI & AppServices) => () => {
+  if (services.history.location.state === undefined) {
```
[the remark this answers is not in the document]
bethshook: It's undefined if the previous history record is not on rex—I'll add a comment. Maybe I'm forgetting some part of this but we only want to send them back if that keeps them on rex, otherwise we clear the query to close the modal.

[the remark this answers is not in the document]
TomWoodward: ok yea i remember wanting to replace instead if there is no previous record, but why is it setting search to an empty string? is that just to make it different at all?

TomWoodward (approved):

PR #1209, opened 2021-06-11
Issue 1349 - properly highlights search results

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1349


src/app/content/components/Page/searchHighlightManager.spec.ts
```diff
@@ -32,16 +33,49 @@ describe('searchHighlightManager', () => {
...
+          result: searchResults[0],
+        },
+      ]);
+
     onHighlightSelect = jest.fn();
   });
 
-  it('calls highlight select callback only when a new highlight is selected', () => {
+  it('calls highlight select callback only when a new highlight is selected + toggle solution if highlight is inside',
+    () => {
```
jarosik10, line 78: Is the indentation okay?

src/app/content/search/utils.ts
```diff
@@ -104,7 +104,13 @@ const getHighlightRanges = (element: HTMLElement, highlight: string): RangyRange
       return [];
     }
 
-    const [partRange] = findTextInRange(elementRange, part.replace(/<\/?strong>|\n/g, ''));
+    const ranges = findTextInRange(elementRange, part.replace(/<\/?strong>|\n/g, ''));
+    // sometimes there might be more fragments of the same text found in the element
+    // in this case get the highlight depending of it's index
+    // https://github.com/openstax/unified/issues/1349
+    const partRange = ranges.length >= index + 1
```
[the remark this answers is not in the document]
PiotrKozlowski: go to [this](https://rex-web-issue-1349-asnnhwxjhmb.herokuapp.com/books/calculus-volume-1/pages/1-5-exponential-and-logarithmic-functions?query=number&target=%7B%22index%22%3A1%2C%22type%22%3A%22search%22%7D#fs-id1170572550969) url and wait for search results to show up (~15 seconds)

As you can see there is the word `number` highlighted 3 times in a single paragraph. Usually, this is not a problem, because a word would be surrounded by different words, but if you look at the search results you can see that there are 3 identical fragments returned from the API for these search results (we display fragments separately, but they come from the single search hit)

This util is trying to find those fragments in the text. All of them are found and stored in `ranges` array.
All fragments have an index - they are stored in an array inside of a search hit.
Then we iterate over those fragments, find them, and highlight them.
`ranges.length >= index + 1` checks if there are more ranges found in the specific element that could match specific highlight and if the current index is less or equal the length of this array then it will use it.

Actually, I think the index of a highlight (+1) could be bigger than `ranges.length` only if `ranges` was empty, so maybe we should handle this differently.

PS. `+1` is because we are comparing an index to the length

src/app/content/components/Page/searchHighlightManager.spec.ts
```diff
@@ -74,8 +74,9 @@ describe('searchHighlightManager', () => {
     onHighlightSelect = jest.fn();
   });
 
-  it('calls highlight select callback only when a new highlight is selected + toggle solution if highlight is inside',
-    () => {
+  const testName = 'calls highlight select callback only when a new highlight is selected '
+    + '+ toggle solution if highlight is inside';
+  it(testName, () => {
```
jarosik10, line 79: Tbh I'd prefer disabling eslint in this line, but this is just my personal opinion.

jarosik10 (approved):

src/app/content/components/Page/searchHighlightManager.spec.ts
PiotrKozlowski: eslint wasnt failing at this line

jarosik10: But as I remember there was a newline between `'long-test-name'` and `()=>{`.

PR #1211, opened 2021-06-15
fixes related to confirmationModal

jarosik10:
for: https://github.com/openstax/unified/issues/1607  ~~& https://github.com/openstax/unified/issues/1304~~

- [x] fix confirmation modal
- [ ] ~~fix scrolling to the unsaved note~~


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -62,7 +62,10 @@ export default class PageComponent extends Component<PagePropTypes> {
       return;
     }
     this.searchHighlightManager = searchHighlightManager(this.container.current, this.props.intl);
-    this.highlightManager = highlightManager(this.container.current, () => this.props.highlights, this.props.intl);
+    this.highlightManager = highlightManager(
+      this.props.dispatch,
+      this.container.current, () => this.props.highlights,
+      this.props.intl);
```
PiotrKozlowski, line 68: I think you could just add it to the `props.highlights`

src/app/content/components/Page/connector.ts
```diff
@@ -29,6 +30,7 @@ export interface PagePropTypes {
   highlights: HighlightProp;
   services: AppServices;
   addToast: typeof addToast;
+  dispatch: Dispatch;
```
PiotrKozlowski, line 33: it could be a part of `HighlightProp`

src/app/content/highlights/actions.ts
```diff
@@ -64,3 +64,7 @@ export const receiveSummaryHighlights = createStandardAction('Content/Highlights
 export const receiveHighlightsTotalCounts = createStandardAction(
   'Content/receiveHighlightsTotalCounts'
 )<CountsPerSource, LocationFilters>();
+export const showConfirmationModal = createStandardAction('Content/Highlights/showConfirmation')<{
+  callback: (key: 'deny' | 'confirm') => any
```
PiotrKozlowski, line 68: [window.confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) is returning a `bool`. I think we could unify this. Also, this should probably return `void`

src/app/content/highlights/components/ConfirmationModal.tsx
```diff
@@ -1,47 +1,72 @@
...
+    return () => {
+      if (document.getElementById(modalNode.id)) {
+        root.removeChild(modalNode);
+      }
+    };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen, modalNode, root]);
+
+  return ReactDOM.createPortal(
```
PiotrKozlowski, line 42: Since now it is a normal component, I think you could just `return null` if it is not open and do not use portal

src/app/content/highlights/components/ConfirmationModal.tsx
```diff
@@ -1,47 +1,72 @@
...
-        {(msg) => <BodyHeading>{msg}</BodyHeading>}
-      </FormattedMessage>
-    </Body>
-    <ConfirmationFooter>
-      <FormattedMessage id='i18n:discard:button:discard'>
-        {(msg) => <Button
-          data-testid='discard-changes'
-          onClick={confirm}
-          variant='primary'
+const ConfirmationModal = () => {
```
PiotrKozlowski, line 19: I think it should be either named `DiscardChangesConfirmationModal` (actions, selectors, etc. should be the same) or it should be more generic by allowing to pass props that will handle displaying buttons and content

PiotrKozlowski: I think that making it more generic is actually better, since you would have to just allow 4 props: heading, body, ok and cancel i18n keys. You can store them in a constant as an object so it will not be much work and we will be able to use this modal for other messages.

src/app/content/highlights/reducer.ts
```diff
@@ -21,8 +21,14 @@ import {
 import { findHighlight } from './utils/reducerUtils';
 import updateSummaryFilters from './utils/updateSummaryFilters';
 
+const initialConfirmationModalState: State['confirmationModal'] = {
```
PiotrKozlowski, line 24: I think this should be somewhere higher in the structure

PiotrKozlowski (changes requested):

src/app/content/highlights/reducer.ts
jarosik10: What do you mean? Like put it from:
```
content: {
   ...,
   highlights: {
      ...,
      confirmationModal: {...}
   }
}
```
to
```
content: {
   ...,
   highlights: {...},
   confirmationModal: {...}
}
```
or even higher in state structure?

src/app/content/highlights/actions.ts
jarosik10: `resolve(boolean)` in `showConfirmation.tsx` is a `void` type callback. So we could leave it returning `void` or make `void | boolean` ?

PiotrKozlowski: I meant something like this `callback(confirmed: boolean) => void`

src/app/content/highlights/reducer.ts
PiotrKozlowski: you've put <ConfirmationModal> into the <Content> component so I think the second example would be accurate

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
PiotrKozlowski: this file should not be added to the repo, you can add it to the global .gitignore file: `/home/username/.gitignore`

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -62,7 +62,10 @@ export default class PageComponent extends Component<PagePropTypes> {
       return;
     }
     this.searchHighlightManager = searchHighlightManager(this.container.current, this.props.intl);
-    this.highlightManager = highlightManager(this.container.current, () => this.props.highlights, this.props.intl);
+    this.highlightManager = highlightManager(
```
PiotrKozlowski, line 65: you can reverse format change

src/app/content/highlights/components/ConfirmationModal.tsx
```diff
@@ -1,47 +1,55 @@
...
+  const {
+    callback,
+    headingi18nKey,
+    bodyi18nKey,
+    okButtoni18nKey,
+    cancelButtoni18nKey,
+  } = useSelector(confirmationModalOptions);
+
+  return isOpen
+    ? <Modal onModalClose={() => callback(false)} heading={headingi18nKey} data-test-id='confirmation-modal'>
```
PiotrKozlowski, line 27: if you do `if (!isOpen) { return null; }` you will have a lot of less diff

src/app/content/highlights/components/utils/showDiscardChangesConfirmation.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+    expect(await answer).toBe(true);
+  });
+
+  it('return false on denial', async() => {
+    const answer = showDiscardChangesConfirmation(store.dispatch);
+    const component = renderer.create(<TestContainer store={store}>
+      <ConfirmationModal/>
+    </TestContainer>);
+
+    const okButton = component.root.findByProps({ 'data-test-id': 'confirmation-modal-cancel-button' });
```
PiotrKozlowski, line 46:
```suggestion
    const cancelButton = component.root.findByProps({ 'data-test-id': 'confirmation-modal-cancel-button' });
```

src/app/content/highlights/components/utils/showDiscardChangesConfirmation.tsx
```diff
@@ -0,0 +1,17 @@
+import { MiddlewareAPI } from '../../../../types';
+import { closeConfirmationModal, showConfirmationModal } from '../../../actions';
+
+export default (dispatch: MiddlewareAPI['dispatch']): Promise<boolean> => new Promise((resolve) => {
+  dispatch(showConfirmationModal({
+    options: {
+      bodyi18nKey: 'i18n:discard:body',
+      callback: (confirmed: boolean) => {
+        resolve(confirmed);
+        dispatch(closeConfirmationModal());
```
PiotrKozlowski, line 10: This could be handled inside of `<ConfirmationModal>` component

src/app/content/highlights/reducer.ts
```diff
@@ -50,6 +51,7 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
       const currentPageId = state.currentPage.pageId;
       const actionPageId = action.payload.location.state && action.payload.location.state.pageUid;
       return {
+        ...state,
```
PiotrKozlowski, line 54: this isn't needed right? (in all of 5 places in this file)

src/app/content/reducer.ts
```diff
@@ -123,6 +135,23 @@ function reduceContent(state: State, action: AnyAction) {
...
+        },
+      };
+    }
+    case getType(actions.closeConfirmationModal): {
+      return {
+        ...state,
+        confirmationModal: {
+          ...initialConfirmationModalState,
+        },
+      };
```
PiotrKozlowski, line 153:
```suggestion
      return {
        ...state,
        confirmationModal: initialConfirmationModalState,
      };
```

src/app/content/types.ts
```diff
@@ -39,6 +39,16 @@ export interface State {
...
   references: Array<PageReferenceMap | PageReferenceError>;
   buyPrint: Pick<BuyPrintResponse['buy_urls'][number], 'url' | 'disclosure'> | null;
+  confirmationModal: {
+    open: boolean,
+    options: {
+      callback: (confimred: boolean) => void,
+      headingi18nKey: string | '',
+      bodyi18nKey: string | '',
+      okButtoni18nKey: string | '',
+      cancelButtoni18nKey: string | '',
```
PiotrKozlowski, line 49: do you need `| ''`?

PiotrKozlowski (changes requested):

PiotrKozlowski (approved):

TomWoodward (changes requested):
why did you choose to change this to pass a callback through redux state rather than maintain the existing pattern?

jarosik10:
@TomWoodward  In the process of creating this PR in one moment `ConfirmationModal` was getting created by `ReactDOM.createPortal` and at that state `callback` started to get passed by redux and it stayed that way.

I think that it's more reusable now, and in my opinion more readable.

TomWoodward:
while it is possible to pass functions through redux state, i don't really like doing that, i prefer to keep the state serializable. the idea behind this confirmation was that its essentially replacing `window.confirm` and we wanted to be able to use it without easy access to state middleware (dispatch/getState), or the react tree (use it from hooks). you can already see its getting pretty awkward to pass those things around in your implementation.

i still don't understand from your message why you changed from the way it was, mounting a separate react tree outside the root, to integrating it into the content layout with a redux flag, can you explain to me what issues you were dealing with that led you to use that approach?

jarosik10:
I was thinking about two different approaches to fix this issue: make `showConfirmation` to require `services` or use redux to handle the `confirmationModal`. At the beginning i've picked the first approach, but I've found out, that it would require more code refactoring than I thought, so I decided to use redux, additionally this solution seemed more intuitive.

jarosik10:
Do you want me to change it to bring back the previous approach?

TomWoodward:
to fix what issue?

jarosik10:
https://github.com/openstax/unified/issues/1607

TomWoodward:
`its broken` is not a very helpful description for me right now. what was the bug in the code that you were trying to fix? what was it doing that was wrong? how was it broken?

jarosik10:
showConfirmation was throwing: 
```
Error: [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.
```

TomWoodward:
i think @bethshook worked on this issue on her intl PR and addressed it by passing the app services into the function. the services are available when redux and react are not, but are easily available in react and hooks, so that is a pretty manageable solution.

beth can you comment on if you have a fix for this ready on your pr or if you were waiting fr @jarosik10 to do it?

TomWoodward:
ok so @jarosik10 can continue to work on the fix in this PR, but i'd like to rework it somewhat back more towards the original approach. i don't want to manage the confirmation callbacks through redux, i would rather pass the services object (or a reduced version of it, if all we need  is `{intl}`) into the confirmation

src/app/content/components/BookBanner.spec.tsx
```diff
@@ -37,39 +38,67 @@ describe('BookBanner', () => {
...
     beforeEach(() => {
-      ({React, renderer} = reactAndFriends());
-      BookBanner = require('./BookBanner').BookBanner;
+      ({React, renderer, TestContainer, renderToDom} = reactAndFriends());
+      BookBanner = require('./BookBanner').default;
     });
 
     it('renders empty state with no page or book', () => {
-      const component = renderer.create(<BookBanner bookTheme={defaultTheme} />);
+      const component = renderer.create(<TestContainer store={createTestStore()}><BookBanner /></TestContainer>);
```
TomWoodward, line 56: this is teh default value, you don't have to pass `createTestStore`

src/app/content/components/BookBanner.tsx
```diff
@@ -1,17 +1,17 @@
 import { HTMLAnchorElement, HTMLDivElement } from '@openstax/types/lib.dom';
-import React, { Component } from 'react';
```
TomWoodward, line 2: why did you refactor this whole component? its introducing a lot of potential for issues and i'm not sure there was a reason to do it

src/app/content/components/ContentLink.tsx
```diff
@@ -64,6 +65,7 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     ? createNavigationOptions({...search, ...systemQueryParams}, scrollTarget)
     : undefined;
   const URL = options ? relativeUrl + navigationOptionsToString(options) : relativeUrl;
+  const intl = useIntl();
```
TomWoodward, line 68: call `useServices` and pass the whole thing into showConfirmation, because pretty sure in future changes more services are needed

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
jarosik10: I thought that it would be easier to make make `BookBanner` a functional component and use `hook` as it's not as easy to pass the services a `prop`.

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,137 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
-    if (this.miniBanner.current && this.bigBanner.current && typeof(window) !== 'undefined') {
-      const miniRect = this.miniBanner.current.getBoundingClientRect();
-      this.setState({
-        scrollTransition: miniRect.top === 0 &&
-          this.bigBanner.current.offsetTop + this.bigBanner.current.clientHeight > window.scrollY,
-        tabbableBanner: miniRect.top === 0 ? 'mini' : 'big',
-      });
+const BookBanner = () => {
+  const [scrollTransition, setScrollTransition] = React.useState(false);
+  const [tabbableBanner, setTabbableBanner] = React.useState('big');
```
TomWoodward, line 182: this used to be an enum, now its a string.

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,137 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
+  const bigBanner = React.useRef<HTMLDivElement>();
+  const services = useServices();
+
+  const handleScroll = () => {
+    if (miniBanner.current && bigBanner.current && typeof(window) !== 'undefined') {
+      const miniRect = miniBanner.current.getBoundingClientRect();
+
+      setScrollTransition(miniRect.top === 0
+        && bigBanner.current.offsetTop + bigBanner.current.clientHeight > window.scrollY);
+      setTabbableBanner(miniRect.top === 0 ? 'mini' : 'big');
```
TomWoodward, line 197: i think if we're going to be setting these at the same time and they're related we should put them in the same state

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,137 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
-    this.handleScroll();
-  }
-
-  public render() {
-    const { pageNode, book, bookTheme } = this.props;
-
-    if (!book) {
-      return <BarWrapper colorSchema={undefined} up={false} />;
-    }
+    document.addEventListener('scroll', handleScroll);
```
TomWoodward, line 217: you gotta return the unbind to the teardown callback

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,137 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
-      up={this.state.scrollTransition}
-      ref={this.bigBanner}
+      up={scrollTransition}
+      ref={bigBanner}
       data-testid='bookbanner'
       data-analytics-region='book-banner-expanded'
     >
       <TopBar>
         {
-          bookUrl === undefined
```
TomWoodward, line 262: you could at least leave the variable names to reduce the diff

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,137 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
+    return <BarWrapper colorSchema={undefined} up={false} />;
+  }
+
+  const bookUrl = hasOSWebData(book)
+    ? book.book_state !== 'retired'
+      ? bookDetailsUrl(book)
+      : undefined
+    : undefined;
+
+  return <>{renderBars({theme: bookTheme, ...book}, bookUrl, pageNode)}</>;
```
TomWoodward, line 309: i'm not really feeling like this function is doing much right? might as well move the stuff all to just be down here

src/app/content/components/Page/contentLinkHandler.spec.ts
```diff
@@ -39,7 +42,7 @@ describe('contentLinkHandler', () => {
 
     beforeEach(() => {
       contentRoute = require('../../routes').content;
-      handler = require('./contentLinkHandler').contentLinkHandler(anchor, () => prop);
+      handler = require('./contentLinkHandler').contentLinkHandler(anchor, () => prop, intl);
```
TomWoodward, line 45: there is a testServices helper that would probably work better, i assume you're not using the spy for anything

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
```diff
@@ -176,158 +176,144 @@ export const BarWrapper = styled.div<BarWrapperProps>`
...
-  pageNode?: ArchiveTreeSection;
-  book?: Book;
-  hasUnsavedHighlight?: boolean;
-  bookTheme: BookWithOSWebData['theme'];
-}
-
 interface BookBannerState {
   scrollTransition: boolean;
-  tabbableBanner: 'mini' | 'big';
+  tabbableBanner: 'big' | 'mini';
```
TomWoodward, line 181: lol like this doesn't matter but please try to make an effort not to make meaningless changes, it makes the diff harder to read.

src/app/content/components/Page/highlightManager.ts
```diff
@@ -49,15 +48,21 @@ export type HighlightProp = ReturnType<typeof mapStateToHighlightProp>
...
+  highlight: Highlight | undefined,
+  appServices: AppServices
+) => defer(async() => {
+  if (!highlight || highlightManagerServices.getProp().focused === highlight.id) {
     return;
   }
-  if (services.getProp().focused && services.getProp().hasUnsavedHighlight && !await showConfirmation()) {
+  if (highlightManagerServices.getProp().focused
+    && highlightManagerServices.getProp().hasUnsavedHighlight
+    && !await showConfirmation(appServices)) {
```
TomWoodward, line 61:
```suggestion
    && !await showConfirmation(appServices)
  ) {
```

src/app/content/highlights/components/utils/showConfirmation.spec.ts
```diff
@@ -40,17 +43,18 @@ describe('ShowConfirmation', () => {
     document.body.appendChild(rootNode);
 
     rootNode.insertAdjacentElement = jest.fn().mockImplementation(() => {
-      document.body.appendChild(modalNode);
+        document.body.appendChild(modalNode);
```
TomWoodward, line 46: wrong indentation

TomWoodward (approved):

PR #1213, opened 2021-06-21
Add scrolling on denial discardChanges modal

jarosik10:
for: https://github.com/openstax/unified/issues/1304


PiotrKozlowski (dismissed):
There are at least 3 ways how this popup can be triggered:
First, you need to start editing the note and then:
- try creating a new highlight (you've added support for this case)
- click on another highlight
- click on a link on the page (or in toc, or page navigation, etc.)

You need to add a support for all of these cases. Maybe you could just pass a highlight elements to the `showDiscardChangesConfirmation` and the rest can be handled inside of a callback function

jarosik10:
@PiotrKozlowski 
It's hard to pass the highlight elements to `showDiscardChangesConfirmation`.
Finding the highlight with `highlights.selector.focused` is difficult as well, because it requires `highlighter.getHighlight(focusedId)`.

I'm thinking about a solution, which won't require so much data passing. Maybe closing `DiscardChangesConfirmationDialog` will dispatch some action, which `HighlightManager` will listen to.

PiotrKozlowski (commented):
Code looks okay and I think this is the simplest solution for this problem, however, I feel like we should do something bigger here, like storing the whole focused highlight in the state so we can scroll to the focused highlight from any place in the app cc @TomWoodward

TomWoodward:
can you explain to me what was going on that you had to add a "force scroll to highlight" action? also there are conflicts that need to be merged in

jarosik10:
@TomWoodward 
It necessary to pass highlight elements to `scrollIntoView` and it was hard to easily get access to this data. There were two different ideas: adding highlight object to the redux state or adding this "force scroll to highlight" action.

There is `useEffect` which has `isActive` prop in dependency array. `isActive` doesn't change after denial of the modal and dispatching actions to remove and add focus to the same highlight doesn't work, because it looks like it merges this redux state changes into one, which means no changes to the state at all. 

I'm going to start working on this PR today since https://github.com/openstax/rex-web/pull/1211 got approved.

TomWoodward:
@jarosik10 i don't understand, it seems like you only changed functionality in the card, but there are many places showConfirmation is called where a user might have staged changes. what are the times when its supposed to scroll to the highlight?

jarosik10:
Whenever user denials the discard changes modal, the previously focused highlight should be scrolled into view, so inside showConfirmation.tsx, deny callback  set `shouldForceScrollToHiglight` to `true` by `dispatch(setForceScrollToHiglight(true))`.

Previously, useEffect inside `Card.tsx` was scrolling into view newly focused highlights, but now I've added a new dependency to listen to - `shouldForceScrollToHiglight`.

As I see, there is still a bug to solve - after modal denial, the useEffect is called twice because the its dependency is changed inside of it.

TomWoodward:
does it not already scroll to the highlight when it becomes focused? should it be that the deny function calls focus highlight?

jarosik10:
@TomWoodward 
When confirmation modal shows up and user chooses to deny it,  the highlight doesn't get focused again, so the _selected highlight_ prop doesn't change at all. If the focused highlight would change, then `isActive` would be assigned again, what would trigger the srollIntoView useEffect.

TomWoodward:
is the bug really that the highlight doesn't get scrolled to _if it is already focused_?  or all the time?

if it is only if its already focused then the issue is the order in which clicking the link triggers unfocus vs the link's confirmation. maybe addressable by wrapping the link handler in `defer`, probably related to some of the effect refactoring we did in the card a bit ago

jarosik10:
@TomWoodward  It's just cant work with the current srollIntoView useEffect. Opening confirmation modal and denying doesn't change `higlhights.currentPage.focused` state. It only gets change if user confirms the confirmation dialog. new highlight gets focused and the srollIntoView useEffect scrolls the newly focused highlight into view.

So I wouldn't tell that it is a bug, scrolling on denying confirmation modal just isn't implemented yet.
Maybe we could clear the `higlhights.currentPage.focused` state whenever confirmation modal is opened and focus the previous highlight on denial. It should work too.

TomWoodward:
Yea I don't see what the deal is with the `shouldForce` that you added, it seems like just dispatching focus would do the same thing using the existing system

src/app/content/components/BookBanner.tsx
```diff
@@ -213,7 +215,7 @@ const BookBanner = () => {
     }
     e.preventDefault();
 
-    if (!await showConfirmation(services)) {
+    if (!await showConfirmation(services, dispatch, focsuedHighlight!)) {
```
TomWoodward, line 218: change `showConfirmation` to take `AppServices & MiddlewareAPI` as its first argument, that matches the hooks pattern, maybe also update `useServices` to return that to make life easier

TomWoodward (changes requested):

src/app/content/components/BookBanner.tsx
jarosik10: @TomWoodward I'm struggling with adding `MiddlwareAPI` to the context of `useServices`.

TomWoodward: [MiddlewareApi](https://github.com/reduxjs/redux/blob/a9181fc2f3230259dd36ba771bf6d2fb56081654/src/types/middleware.ts#L3) is just dispatch and getState, useServices is currently defined as:

```
export const useServices = () => React.useContext(servicesContext);   
```

you could change it to be:
```
export const useServices = () => ({
  ...React.useContext(servicesContext),
  dispatch: useDispatch(),
  getState: useStore().getState
```

or you could put that in a new hook, if you don't want to change the existing one

jarosik10:
@TomWoodward I have a problem with achieving 100% test coverage here. [Line 1](https://github.com/openstax/rex-web/blob/d6b7d62484e174dc3a3afa8c5ae892eb4b0a9be0/src/app/content/components/Page/highlightManager.ts#L66) [Line 2](https://github.com/openstax/rex-web/blob/d6b7d62484e174dc3a3afa8c5ae892eb4b0a9be0/src/app/content/components/Page/contentLinkHandler.ts#L117)

TomWoodward (changes requested):
now that the intl changes are merged this will have some conflicts with the integration, but once those are fixed i can review again

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -16,8 +18,14 @@ export default async(services: AppServices & MiddlewareAPI) => {
   root.insertAdjacentElement('afterend', domNode);
 
   const userChoice: boolean = await new Promise((resolve) => {
+    const focusedHighlight = focused(services.getState())!;
```
TomWoodward, line 21: not into the `!` in production code

TomWoodward (changes requested):
i agree this solution is a bit janky, i would prefer to be able to scroll to a highlight without unfocusing it and refocusing it, especially since it has pending changes which we have no guarantee of being saved and needing to re-trigger the editing state.

i think the problem is that the scrollto effect is happening in the react code in response to the prop change, i don't want to put complex objects into the redux state and while we could do some crazy context/appservices thing to communicate into a redux hook, i think keeping scrolling logic in react makes some sense because it is display layer logic. 

maybe the thing to do is add something to the state that can be bound to the redux props in a way that will allow us to trigger re-scroll to a highlight other than just a focused/unfocused boolean value.

off the top of my head all i can think of is replacing the focused boolean value with a timestamp, so that it becomes possible to _refocus_ a highlight, but i'm not sure if there would be unintended consequences around then needing to debounce focuses. if there are currently places that are repeatedly focusing a highlight during certain event chains, we wouldn't know or care about that now, but it would cause problems if we made that change.

open to suggestions, but i agree what is here may not be ideal.

src/app/content/highlights/components/utils/showConfirmation.tsx
```diff
@@ -16,8 +18,15 @@ export default async(services: AppServices & MiddlewareAPI) => {
...
   const userChoice: boolean = await new Promise((resolve) => {
+    const focusedHighlight = focused(services.getState());
     const confirm = () => resolve(true);
-    const deny = () => resolve(false);
+    const deny = () => {
+      resolve(false);
+      if (focusedHighlight) {
+        services.dispatch(focusHighlight(focusedHighlight));
+      }
+      services.dispatch(setAnnotationChangesPending(true));
```
TomWoodward, line 28: i don't think it should be necessary to set this anymore, it should still be true already. it used to be necessary because we were unfocusing/refocusing the highlight and it cleared the editing state

TomWoodward (changes requested):

TomWoodward (approved):

PR #1216, opened 2021-06-23
Issue 1612

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1612

### DIFF
--- a/src/app/content/components/Page.spec.tsx
+++ b/src/app/content/components/Page.spec.tsx
@@ -52,9 +52,8 @@ jest.mock('../../domUtils', () => ({
   scrollTo: jest.fn(),
 }));
 
-const makeEvent = (doc: Document) => {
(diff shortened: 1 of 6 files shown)


src/app/content/components/Page.spec.tsx
```diff
@@ -54,7 +54,7 @@ jest.mock('../../domUtils', () => ({
 
 const makeEvent = (doc: Document) => {
   const event = doc.createEvent('MouseEvents');
-  event.initEvent('click', true, false);
+  event.initEvent('click', true, true);
```
bethshook, line 57: This is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent) and I wonder if we should replace usage of it across tests with the [Event() constructor](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event).

bethshook (approved):

PR #1218, opened 2021-06-24
update rex build parameters

TomWoodward:
- full commit sha for concourse to be able to check it out
- identify the release by both code and book version information (prepare for content only releases)
- include books config in build args (prepare for content only releases)

for: openstax/unified#1622


concourse/create-build-configs/script.js
```diff
@@ -44,4 +63,7 @@ fs.readFile(versionFile, 'utf8', function(err, commit) {
 
   console.log(`Generating release id file with: ${releaseId}`);
   fs.writeFile(releaseFile, releaseId, handleErr);
+
+  console.log(`Generating books file...`);
+  fs.writeFile(booksConfigFile, JSON.stringify(books), handleErr);
```
bethshook, line 68: You're stringifying `books` twice, can you just do it once?

TomWoodward: it was actually worse than that, i wasn't parsing the json i was reading to begin with, so i was ending up with double encoded books

bethshook (approved):

PR #1221, opened 2021-06-25
Add minimum height for a td

PiotrKozlowski:
for: https://github.com/openstax/unified/issues/1615

### DIFF
--- a/generic-styles/index.less
+++ b/generic-styles/index.less
@@ -139,6 +139,10 @@ body {
       display: inline;
     }
   }
+
+  .os-table td {
+    height: 3rem;
+  }
 }
 
 .content-excerpt {


jarosik10 (approved):

generic-styles/index.less
```diff
@@ -139,6 +139,10 @@ body {
       display: inline;
     }
   }
+
+  .os-table td {
+    height: 3rem;
```
jarosik10, line 144: Shouldn't we use `min-height` instead?

PiotrKozlowski: nope, `height` for td is a `min-height` and `min-height` is not working :p
I was also confused :D

PR #1222, opened 2021-06-28
fix highlight area of images which are not inside figure element

jarosik10:
for: https://github.com/openstax/unified/issues/1312

I've changed the highlight area size of images which are not inside `<figure>` element to be the same as the the ones that are inside `<figure>`.

Image inside `<figure>` (highlighted):
obraz

Image **not** inside `<figure>` (highlighted) **BROKEN**:
obraz

Image **not** inside `<figure>` (highlighted) **FIXED**:
obraz

### DIFF
--- a/generic-styles/content-common.less
+++ b/generic-styles/content-common.less
@@ -32,8 +32,8 @@ a:not([role="button"]) {
 
 img {
   max-width: 100%;
-  padding-top: 1em;
-  padding-bottom: 1em;
+  margin-top: 1em;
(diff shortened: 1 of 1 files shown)


jarosik10:
@Bkblodget 
You [said](https://github.com/openstax/unified/issues/1312#issuecomment-868594541) that these images look the best with bigger margin. I think it should be considered as an additional change to the content styles, not as a fix to this issue.

PiotrKozlowski (approved):

PR #1238, opened 2021-07-09
content query params fixes

bethshook:
for: openstax/unified#1631
and a continuation of openstax/unified#1596


src/app/content/search/hooks/index.ts
```diff
@@ -98,10 +100,12 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
 
 export const clearSearchHook: ActionHookBody<typeof clearSearch | typeof openToc> = (services) => () => {
   const scrollTarget = selectNavigation.scrollTarget(services.getState());
+  const systemQueryParams = selectNavigation.systemQueryParameters(services.getState());
+
   if (scrollTarget && isSearchScrollTarget(scrollTarget)) {
     services.history.replace({
       hash: '',
-      search: '',
+      search: queryString.stringify(systemQueryParams),
```
bethshook, line 108: This fixes the "close search results" bug

src/app/content/components/ContentLink.spec.tsx
```diff
@@ -68,35 +68,14 @@ describe('ContentLink', () => {
       expect(event.preventDefault).toHaveBeenCalled();
     });
 
-    it('dispatches navigation action with search if there is a search', async() => {
```
bethshook, line 71: @TomWoodward am I right that this is no longer relevant since search params won't automatically be passed on navigation?

src/app/content/search/hooks/index.ts
```diff
@@ -97,13 +97,12 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
 };
 
 export const clearSearchHook: ActionHookBody<typeof clearSearch | typeof openToc> = (services) => () => {
-  const scrollTarget = selectNavigation.scrollTarget(services.getState());
-  if (scrollTarget && isSearchScrollTarget(scrollTarget)) {
-    services.history.replace({
-      hash: '',
-      search: '',
-    });
-  }
```
bethshook, line 106: Want to confirm that it's ok to remove the below test, which is now failing. I don't understand why the search string would not be cleared without a scrollTarget.
`hooks > clearSearchHook > noops if there was no scroll target or if it wasn't search scroll target`

src/app/content/search/hooks/index.ts
```diff
@@ -97,11 +97,14 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
+  const query = selectNavigation.persistentQueryParameters(state).query;
+  const systemQueryParams = selectNavigation.systemQueryParameters(state);
+
+  if (query) {
     services.history.replace({
       hash: '',
-      search: '',
+      search: queryString.stringify(systemQueryParams),
     });
   }
```
bethshook, line 109: @TomWoodward this fixes the bug that was initially [reported](https://github.com/openstax/unified/issues/1631) but I'm worried that I'm breaking something else. The only thing I can think of is that if you had a target set to a highlight and then made a search and closed the search results, that highlight query would be lost. That seems ok to me, though.

src/app/content/search/components/SearchResultsSidebar/SearchResultContainers.tsx
```diff
@@ -89,6 +94,7 @@ const SearchResult = (props: {
           page={props.page}
           result={thisResult}
           scrollTarget={target}
+          queryParams={{query: props.currentQuery}}
           onClick={() => props.selectResult(thisResult)}
```
bethshook, line 98: @TomWoodward to answer your question this handler is used to close mobile search results when one is selected.

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -39,24 +42,28 @@ const reducePageReferenceError = (reference: PageReferenceError, document: Docum
...
+  const options = createNavigationOptions(systemQueryParams);
   const a = assertNotNull(
     document.querySelector(`[href^='${reference.match}']`),
     'references are created from hrefs');
-  const href = assertNotNull(a.getAttribute('href'), 'it was found by href value')
-    .replace(reference.match, toRelativeUrl(currentPath, path) + systemQueryString);
-  a.setAttribute('href', href);
+  const href = assertNotNull(a.getAttribute('href'), 'it was found by href value');
+  // add hash to options if href only consists of hash, ie an anchor on the same page
+  options.hash = (a as HTMLAnchorElement).hash === href ? new URL(href, `https://openstax.org${currentPath}`).hash : '';
```
TomWoodward, line 53: i think its ok to unconditionally forward the hash, if its a link to a figure on another page we want that to work too

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -39,24 +42,28 @@ const reducePageReferenceError = (reference: PageReferenceError, document: Docum
 };
 
 // tslint:disable-next-line: max-line-length
-const reduceReference = (reference: PageReferenceMap, currentPath: string, document: Document, systemQueryString: string) => {
+const reduceReference = (reference: PageReferenceMap, currentPath: string, document: Document, systemQueryParams: SystemQueryParams) => {
   const path = content.getUrl(reference.params);
+  const options = createNavigationOptions(systemQueryParams);
   const a = assertNotNull(
     document.querySelector(`[href^='${reference.match}']`),
     'references are created from hrefs');
```
TomWoodward, line 50: you can cast this to `HTMLAnchorElement` or pass it into the querySelector generic so that you don't have to fuss with it below

src/app/content/components/Page/contentLinkHandler.ts
TomWoodward: i didn't realize the anchor element exposes `hash`, you probably don't need all that `URL` junk

src/app/content/search/hooks/index.ts
TomWoodward: i think really you want to take whatever the current params are and remove the search query, and remove the target only if it is for a search result, but otherwise leave everything else

src/app/content/utils.ts
```diff
@@ -26,18 +27,19 @@ export interface ContentPageRefencesType {
   pageId: string;
 }
 
-export function getContentPageReferences(htmlContent: string) {
-  const matches: ContentPageRefencesType[] = (htmlContent.match(/.\/([a-z0-9-]+(@[\d.]+)?):([a-z0-9-]+.xhtml)/g) || [])
+export function getContentPageReferences(book: ArchiveBook, page: ArchivePage) {
+  // tslint:disable-next-line: max-line-length
```
TomWoodward, line 31: you should reformat the code to be shorter, rather than disabling the linter everywhere

TomWoodward (changes requested):

src/app/content/search/hooks/index.ts
bethshook: updated and did the same for the hash

src/app/content/utils.ts
```diff
@@ -26,18 +27,22 @@ export interface ContentPageRefencesType {
...
+      const [bookId, bookVersion] = bookMatch && bookMatch.split('@') as [string, string | undefined];
+
+      console.log('book: ', book, book.id);
 
       return {
-        bookId: bookId.substr(2),
-        bookVersion,
-        match,
-        pageId: stripIdVersion(pageId),
+        bookId: (match.includes(':') && bookId && bookId.substr(3)) || book.id,
```
bethshook, line 42: @TomWoodward Everything was passing `bookId && bookId.substr(3)` so I added the first condition, but is there a better way to determine that we have what we need from the match?

TomWoodward: yea i mean the better way would be to use regex groups, but this is fine

TomWoodward (approved):

PR #1239, opened 2021-07-12
[WIP] Issue 996

jarosik10:
for: https://github.com/openstax/unified/issues/996

- [x] bug: Chrome when selecting text or clicking
- [ ] bug: Safari - image + partial caption highlighted with notes


jarosik10:
Related issues from MathJax repo:
- https://github.com/mathjax/MathJax/issues/1979
- https://github.com/QuantEcon/lecture-python.myst/issues/18
- https://github.com/mathjax/MathJax/issues/1282

bethshook (approved):

TomWoodward:
math appears very broken for a short time during typesetting 
image

src/app/content/components/Page/PageContent.tsx
```diff
@@ -89,7 +89,8 @@ export default styled(MainContent)`
         }
       }
 
-      &.first.text.has-note:after {
+      &.first.text.has-note [data-for-screenreaders]:first-child {
```
TomWoodward, line 92: why did you change this to from `:after`

TomWoodward (changes requested):

src/app/content/components/Page/PageContent.tsx
jarosik10: this was an unreverted change for my attempt to fix [this bug](https://drive.google.com/file/d/1ZsrkH0mDZbRK5PD9dU2YppcBG4st6KEz/view).

jarosik10:
> math appears very broken for a short time during typesetting, does not happen on master

yeah, i see..

@TomWoodward 
However, i'm no longer able to reproduce the ["math jumping" bug](https://drive.google.com/file/d/1LRJhAcdZrlLdczZM5eLuT2rf4A8xfD5t/view) on production. Maybe it got fixed by on the MathJax side.

@lenka222 Could you check it?

lenka222:
I'm also not able to reproduce "math jumping" issue.

TomWoodward:
@bethshook are we sure that this doesn't mess up block equations?

bethshook:
@TomWoodward I don't see any issue with block equations—`.MathJax` is always applied to spans whether the equation is inline or block and I'm not sure how its display property would get overwritten. I also can't reproduce the bug despite many attempts.

bethshook:
Closed per this discussion: https://github.com/openstax/unified/issues/996

PR #1246, opened 2021-07-16
Increase document font size

bethshook:
for: openstax/unified#1624


src/app/utils/increaseSize.ts
```diff
@@ -0,0 +1,4 @@
+// rem * 10px * 8/7, then convert back to rem
+const increaseSize = (size: number) => Math.round(size * 10 * 8 / 7) / 10;
```
bethshook, line 2: @TomWoodward I'm wondering if this should live somewhere else and if I should specify the ratio in the function name.

src/app/components/Typography/index.ts
```diff
@@ -62,7 +63,7 @@ export const bodyCopyRegularStyle = css`
 
 export const labelStyle = css`
   ${textStyle}
-  font-size: 1.4rem;
-  line-height: 1.6rem;
+  font-size: ${increaseSize(1.4)}rem;
+  line-height: ${increaseSize(1.6)}rem;
```
TomWoodward, line 67: i thought we were only changing the content font size, what is this for?

src/app/content/components/Page/PageContent.tsx
```diff
@@ -23,8 +24,8 @@ export const contentTextStyle = css`
 
 export default styled(MainContent)`
   ${contentTextStyle}
-  font-size: 1.4rem;
-  line-height: 2rem;
+  font-size: ${increaseSize(1.4)}rem;
+  line-height: ${increaseSize(2)}rem;
```
TomWoodward, line 28: really i think we should move these to the generic-styles sheet

src/app/utils/increaseSize.ts
TomWoodward: what even is all this math for?

TomWoodward (changes requested):

src/app/components/Typography/index.ts
```diff
@@ -62,7 +62,7 @@ export const bodyCopyRegularStyle = css`
 
 export const labelStyle = css`
```
TomWoodward, line 63: pull a list of places this is used so we can verify they're ok

TomWoodward (changes requested):
that test with the hardcoded heights needs to be updated for the new text sizes

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -128,7 +132,7 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
       & > ${NavItem} > ${ContentLink} {
         .os-number {
           width: ${numberWidth}rem;
-          overflow: hidden;
+          overflow: visible;
```
TomWoodward, line 135: this is worrying to me, why is it updated?

TomWoodward (changes requested):

src/app/content/components/TableOfContents/styled/index.tsx
bethshook: This happens otherwise: https://share.getcloudapp.com/wbu62gY4
I can leave as is and we can discuss since there is another visual issue (the dropdown arrows).

TomWoodward: there is some hideous math in the sidebar logic that computes the column size based on the text value of the number, that logic is probably wrong now with the new font size

generic-styles/index.less
```diff
@@ -21,8 +21,26 @@ body {
...
+    :hover {
+      color: '#0064A0';
+    }
+  }
+}
+
 #main-content:not([data-dynamic-style="true"]), .content-excerpt:not([data-dynamic-style="true"]) {
   @import "./content-baked.less";
+  font-size: 1.6rem;
+  line-height: 2.3rem;
```
bethshook, line 43: @TomWoodward I moved `#main-content` text styles here as well

TomWoodward (changes requested):

generic-styles/index.less
```diff
@@ -21,8 +21,26 @@ body {
   font-size: @font-size-base;
 }
 
+.page-content {
```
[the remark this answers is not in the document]
bethshook: they're the ones moved out of PageComponent.tsx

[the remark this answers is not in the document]
TomWoodward: what do these apply to that isn't covered by the `#main-content,.content-excerpt` selector? can these styles be moved in there?

TomWoodward (changes requested):

generic-styles/index.less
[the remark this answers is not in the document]
bethshook: probably... I haven't seen any other children the styles would apply to but could there be a book where that happens?

TomWoodward:
Rex renders the main-content div around where we inject the page, so I
think it's safe to just put everything in there

generic-styles/index.less
```diff
@@ -23,24 +23,36 @@ body {
...
+  a {
+    color: #027EB5;
+    cursor: pointer;
+    text-decoration: underline;
+  
+    &:hover {
+      color: #0064A0;
+    }
+  }
+
```
bethshook, line 41: @TomWoodward `.page-content` styles moved here. Everything that I've checked looks good.

TomWoodward (approved):

PR #1254, opened 2021-07-22
Set order of TOC button to initial value in page not found text

jarosik10:
for: https://github.com/openstax/unified/issues/1640

### DIFF
--- a/src/app/content/components/Page/PageNotFound.tsx
+++ b/src/app/content/components/Page/PageNotFound.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import styled, { css } from 'styled-components/macro';
 import theme from '../../../theme';
-import { SidebarControl } from '../Toolbar/styled';
+import { OpenSidebarControl } from '../SidebarControl';
 
 // tslint:disable-next-line: variable-name
 const PageNotFoundWrapper = styled.div`
@@ -46,7 +46,7 @@ const PageNotFound = () => <PageNotFoundWrapper>
         {(msg) => msg}
       </FormattedMessage>
     </span>
-    <SidebarControl />
+    <OpenSidebarControl hideMobileText={false}/>
   </PageNotFoundText>
 </PageNotFoundWrapper>;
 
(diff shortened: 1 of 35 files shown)


bethshook (approved):

src/app/content/components/Page/PageNotFound.tsx
```diff
@@ -31,6 +31,10 @@ const PageNotFoundText = styled.div`
   span {
     margin-right: 0.5rem;
   }
+
+  ${SidebarControl} {
+    order: initial;
```
TomWoodward, line 36: instead of setting the order in the control, and overriding it here, it'd be better to only set it where its needed

TomWoodward (changes requested):

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -376,3 +361,20 @@ export const SidebarControl = styled(OpenSidebarControl)`
 export const NudgeElementTarget = styled.div`
   display: contents;
 `;
+
+// tslint:disable-next-line:variable-name
```
jarosik10, line 365: I've moved it down because `SidebarControl` has to be defined before being used.

src/app/content/components/Toolbar/styled.tsx
```diff
@@ -376,3 +361,20 @@ export const SidebarControl = styled(OpenSidebarControl)`
...
+export const TopBar = styled.div`
```
jarosik10, line 366: I've moved it down because SidebarControl has to be defined before being used.

TomWoodward (changes requested):
ok so i'm sorry the situation was worse initially than i thought and i think i led you astray. the page not found is using `toolbar/SidebarControl` which is itself a customization of the styles from `components/SidebarControl`. it makes sense for `toolbar/SidebarControl` to include all the custom styling for being in the toolbar, including the order attribute. i think the issue is that the page not found shouldn't be using this customized style, it should be using the base component from `components/SidebarControl`

src/app/content/components/Page/PageNotFound.tsx
```diff
@@ -46,7 +46,7 @@ const PageNotFound = () => <PageNotFoundWrapper>
         {(msg) => msg}
       </FormattedMessage>
     </span>
-    <SidebarControl />
+    <OpenSidebarControl hideMobileText={false}/>
```
jarosik10, line 49: @TomWoodward 
Do we want to always keep the text or make the sidebar control to behave and look same as in toolbar?

TomWoodward (approved):

PR #1256, opened 2021-07-23
Make noteish elements scrollable

jarosik10:
for: https://github.com/openstax/unified/issues/1637

I've also added `overflow-y: hidden` because of https://stackoverflow.com/questions/6421966/css-overflow-x-visible-and-overflow-y-hidden-causing-scrollbar-issue/6433475#6433475

@TomWoodward I've been trying to find some workaround, but still can't find a good solution.
The problem is about overflows-x: auto that disables the margin collapsing.


bethshook:
Why is `overflow-y: hidden` better than `auto` if no height is set on the container? What issue did you have before adding that line?

jarosik10:
@bethshook there was a place where overflow-y: auto created a scrollbar, but I can't find that page now :P. Well.. i guess it'd better if I remove that overflow-y: hidden, because it might cause some trouble in the future.

bethshook:
Yeah, maybe we can just flag that for QA

bethshook (approved):

TomWoodward (changes requested):
this note now has a vertical scrollbar that isn't needed 
image

TomWoodward:
also we want to try to change the scrollbar to include the padding https://openstax.slack.com/archives/C8V1XET1R/p1627481352313100

bethshook:
If you add `box-sizing: border-box` to the parent `[data-type="note"]` does that eliminate the vertical scroll?

jarosik10:
@TomWoodward on what browser do you have that vertical scrollbar? I'm looking at the same element on chrome a firefox on different screen sizes and I don't see that scrollbar.

bethshook:
I reproduced it on Chrome too (mobile size). But I do think it must be an issue with box-sizing because there's no height on the parent.

TomWoodward:
i find it very odd that while the scroll container works for the math, the rest of the text is wrapping at the visible width and not the element width, so when you scroll over you just see the math and the text is awkwardly pushed left

why is this happening?

TomWoodward (changes requested):
previous comment

src/app/content/components/Content.browserspec.tsx
```diff
@@ -9,8 +9,8 @@ const TEST_CASES: { [testCase: string]: (target: Page) => Promise<void> } = {
   Desktop: setDesktopViewport, Mobile: setMobileViewport,
 };
 const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
-  Desktop: [242, 90, 122, 242, 365, 668, 761, 1262, 1606],
-  Mobile: [239, 66, 96, 239, 523, 1263, 1402, 1751, 2118],
+  Desktop: [227, 90, 122, 227, 365, 668, 761, 1262, 1606],
+  Mobile: [224, 66, 96, 224, 523, 1263, 1402, 1751, 2118],
```
TomWoodward, line 13: why are these modified?

TomWoodward (changes requested):

src/app/content/components/Content.browserspec.tsx
jarosik10: I'm not sure what is causing this test to fail. Sizes of the noteish elements looks the same as in `master`.

TomWoodward: it looks like you removed 15px of bottom padding, i think this test is legitimately failing

TomWoodward (changes requested):

jarosik10:
@TomWoodward 
The test fails because of the fact that `overflow-x: auto` disabled margin collapsing. So the paragraph inside _DEFINITION_ creates additional margin. It's not so easy to fix. 
Zrzut ekranu z 2021-08-31 15-14-16
Zrzut ekranu z 2021-08-31 15-10-16

generic-styles/content-baked.less
```diff
@@ -190,7 +190,6 @@ span[data-type="media"] img {
 #blockish {
   .style(note) {
     margin: 3rem 0;
-    padding: 1.5rem;
```
TomWoodward, line 193: the top of this is replaced on line 207, the sides on line 206/216, the remaining amounts on those lines are carrying over what was already there, i don't see the bottom ever being replaced in the other changes, and the test is off by 15px, are you sure about your assessment of the margin collapsing being the problem? maybe you're just missing 1.5rem on the bottom from your changes?

TomWoodward (changes requested):

generic-styles/content-baked.less
jarosik10: Production:
Zrzut ekranu z 2021-09-14 14-42-51

After changes:
Zrzut ekranu z 2021-09-14 14-43-24

jarosik10:
@TomWoodward  I've been trying to find some workaround, but still can't find a good solution. 
The problem is about `overflows-x: auto` that disables the margin collapsing.

generic-styles/content-baked.less
```diff
@@ -636,9 +639,13 @@ figure {
     text-transform: uppercase;
   }
   p {
-    margin: 0 0 1rem 1.5rem;
+    margin: 1rem 0 1rem 1.5rem;
     color: #555;
   }
+
+  &[data-type="note"] {
```
TomWoodward, line 646: what is this and how is what its selecting different from the note styling above

generic-styles/content-baked.less
```diff
@@ -366,6 +366,10 @@ h1.example-title .text {
   img {
     display: block;
   }
+
+  [data-type="exercise"]:first-of-type {
```
TomWoodward, line 370: you said you don't like this, why is it necessary and why don't you like it?

TomWoodward (changes requested):

generic-styles/content-baked.less
jarosik10: It's about Section Learning Objectives: https://openstax.org/books/physics/pages/15-1-the-electromagnetic-spectrum.
Wrapper `<div class="learning-objectives" data-type="note">:  `learning-objectives` are overriding the styling by `[data-type="note"]`, so it has `1.5rem` padding
Title: `[data-type="note"] > .os-title` overrides `.learning-objectives [data-type="title"] `, so it has the _new_ styling - additional padding. 

So I've added this selector to make normal learning objectives and noteish learning objectives look the same as before.

generic-styles/content-baked.less
jarosik10: I'm looking through the books for different noteish elements and find some workaround to fix the disabled margin collapsing problem.
This selector is fixing the `Checkpoint 1.18 ` in https://rex-web-issue-1637-ugrznxqolyj.herokuapp.com/books/calculus-volume-3/pages/1-5-conic-sections

jivey (commented):
I also cannot see a clear solution to the lack of margin collapse, I think it's because the overflow creates a new block formatting context. One fix could be making the entire container scrollable, but then the header underline also scrolling may be undesirable:

image

I don't think it was introduced with this change, but should we also look at changing the margins on the os-numbers in this? They may need some alignment/margin tweaks:

image

Other than that this looks good to me :)

TomWoodward:
@jivey the failure in the build tests indicate that the height of some elements has probably changed. if we can't avoid the change to the margin collapsing, can we update the styles to compensate somehow?

TomWoodward (approved):

PR #1258, opened 2021-07-26
Add labs-cta component

jarosik10:
for: https://github.com/openstax/unified/issues/1628


TomWoodward:
there is already a pr for this but it needs some help. jp made the backend and plugged some stuff into the FE but didn't really know what he was doing. i think there is also something going on with swagger bindings that i'll want to look at

jivey (approved):
My knowledge of Redux isn't the best, but this looks good to me, and the variants look right in local testing. The only thing I wasn't sure about is how the practiceQuestions feature flag changes are related, maybe that was part of the scoped work though.

jivey:
In case it's helpful, I noticed the Selenium test seems to be failing from a hardcoded path https://github.com/openstax/rex-web/blob/master/pytest-selenium/utils/utility.py#L214

Malar-Natarajan:
> In case it's helpful, I noticed the Selenium test seems to be failing from a hardcoded path https://github.com/openstax/rex-web/blob/master/pytest-selenium/utils/utility.py#L214

yes...its my bad. I will update it

TomWoodward (approved):

PR #1300, opened 2021-08-06
New intl init hook and updates to services and MessageProvider

bethshook:


src/app/content/hooks/locationChange/index.ts
```diff
@@ -1,4 +1,5 @@
 import googleAnalyticsClient from '../../../../gateways/googleAnalyticsClient';
+import initializeIntl from '../../../messages/intlHook';
```
TomWoodward, line 2: move the hook file into the content module

TomWoodward (approved):

PR #1315, opened 2021-08-20
exit update-redirects script if page was removed from book

bethshook:
for: openstax/unified#1674

### DIFF
--- a/script/utils/update-redirects-data.ts
+++ b/script/utils/update-redirects-data.ts
@@ -44,7 +44,7 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
       countNewRedirections++;
     } else {
       // tslint:disable-next-line: no-console
-      console.error(`section ${section} not found in new book`);
+      console.error(`page with ID ${section.id} not found in new book`);
       process.exit(1);
     }
   }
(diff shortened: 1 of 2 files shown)


script/utils/update-redirects-data.ts
```diff
@@ -42,6 +42,10 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
     ) {
       redirects.push(formatSection(section));
       countNewRedirections++;
+    } else if (!slug) {
+      // tslint:disable-next-line: no-console
+      console.error(`page with ID ${section.id} not found in new book`);
```
TomWoodward, line 47: lets put a comment in here like "remove this to enable legitimately removing pages from books only once uuids are guaranteed to be consistent"

script/utils/update-redirects-data.ts
```diff
@@ -42,6 +42,10 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
+      process.exit(1);
```
TomWoodward, line 48: process.exit will kill the whole script, you should throw an error like above to allow the parent script to aggregate errors from multiple books or whatever it does

TomWoodward (changes requested):

TomWoodward (approved):

PR #1318, opened 2021-08-24
FE selenium test: search results

Malar-Natarajan:
https://openstax.testrail.com/index.php?/cases/view/543252

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1684


gregfitch (approved):
The library dataset is getting pretty big. You may want to break that out into its own YAML or JSON file (end result is the same) then read that file into the Library object.

PR #1328, opened 2021-08-31
add a way to pin archive code versions

TomWoodward:
for: openstax/unified#1686

### DIFF
--- a/script/update-archive-version.ts
+++ b/script/update-archive-version.ts
@@ -27,12 +27,16 @@ async function updateArchiveVersion() {
   const osWebLoader = createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`);
 
   const currentBookLoader = makeUnifiedBookLoader(
-    createArchiveLoader(`${ARCHIVE_URL}${REACT_APP_ARCHIVE_URL}`),
+    createArchiveLoader(REACT_APP_ARCHIVE_URL, {
+      archiveHost: ARCHIVE_URL,
+    }),
     osWebLoader
   );
 
   const newBookLoader = makeUnifiedBookLoader(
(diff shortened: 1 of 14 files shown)


src/gateways/createArchiveLoader.ts
```diff
@@ -1,41 +1,45 @@
...
-  const {pageCache, bookCache, appUrl} = {...defaultOptions(backendUrl), ...options};
+export default (archiveBasePath: string, options: Options = {}) => {
+  const {pageCache, bookCache, appHost, archiveHost, archiveOverride} = {
+    ...defaultOptions(),
+    ...options,
+  };
 
-  const contentUrl = (base: string, ref: string) => `${base}/contents/${ref}.json`;
+  const contentUrlBase = (host: string, bookId: string) => archiveOverride ||
+    `${host}${BOOKS[bookId]?.archiveOverride || archiveBasePath}`;
```
bethshook, line 28: The issue here is that `archiveOverride` has to be an absolute URL if no host is passed in, which is the case with the update-archive-version script.

TomWoodward:
Doesn't it set a host? How could it work if it doesn't?

script/update-archive-version.ts
```diff
@@ -27,12 +27,16 @@ async function updateArchiveVersion() {
...
+      archiveHost: ARCHIVE_URL,
+    }),
     osWebLoader
   );
 
   const newBookLoader = makeUnifiedBookLoader(
-    createArchiveLoader(`${ARCHIVE_URL}${args.newArchiveUrl}`),
+    createArchiveLoader(args.newArchiveUrl, {
+      archiveHost: ARCHIVE_URL,
+    }),
```
bethshook, line 39: @TomWoodward this should do the trick

bethshook (approved):

Malar-Natarajan (approved):

PR #1333, opened 2021-09-01
Add pipeline to search query

bethshook:
for: openstax/unified#1567


script/update-archive-version.ts
```diff
@@ -13,13 +13,14 @@ import createOSWebLoader from '../src/gateways/createOSWebLoader';
...
+const { REACT_APP_ARCHIVE_URL_BASE } = ArchiveUrlConfig;
 
 const args = argv as any as {
-  newArchiveUrl: string
+  newArchive: string
 };
 
 async function updateArchiveVersion() {
-  if (args.newArchiveUrl === REACT_APP_ARCHIVE_URL) {
+  if (`${REACT_APP_ARCHIVE_URL_BASE}${args.newArchive}` === REACT_APP_ARCHIVE_URL) {
```
TomWoodward, line 23: Can you compare the old against the new without making it into url shape?

TomWoodward (changes requested):

src/config.test.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   CODE_VERSION: 'test',
   DEPLOYED_ENV: 'test',
   REACT_APP_ARCHIVE_URL: '/apps/archive/codeversion',
+  SEARCH_URL: 'https://search-os-f42d77d.sandbox.openstax.org',
```
bethshook, line 6: @TomWoodward search isn't working in the browser tests. I thought adding this might help but no dice

src/app/content/utils/idUtils.ts
```diff
@@ -1,2 +1,3 @@
 export const stripIdVersion = (id: string): string => id.split('@')[0];
 export const getIdVersion = (id: string): string | undefined => id.split('@')[1];
+export const stripArchiveVersion = (id: string): string => id.split('__')[1];
```
TomWoodward, line 3: i think the underscore formatting is search specific, in archive its still formatted `id@version`

src/config.test.js
```diff
@@ -3,6 +3,8 @@ module.exports = {
   CODE_VERSION: 'test',
   DEPLOYED_ENV: 'test',
   REACT_APP_ARCHIVE_URL: '/apps/archive/codeversion',
+  REACT_APP_ARCHIVE: 'codeversion',
+  SEARCH_URL: 'https://search-os-f42d77d.sandbox.openstax.org',
```
TomWoodward, line 7: i think this is unnecessary

TomWoodward (changes requested):

src/app/content/utils/idUtils.ts
TomWoodward: since this formatting is specific to search, i think it would make more sense to leave it embedded in that function that is like `getIndexInfo` or whatever

TomWoodward (changes requested):

TomWoodward (approved):

PR #1345, opened 2021-09-16
Cross book refs version bug

bethshook:
for: openstax/unified#1704

### DIFF
--- a/src/app/content/utils.spec.ts
+++ b/src/app/content/utils.spec.ts
@@ -116,7 +116,7 @@ describe('getContentPageReferences', () => {
       },
       {
         bookId: '13ac107a-f15f-49d2-97e8-60ab2e3b519c',
-        bookVersion: undefined,
+        bookVersion: 29.7,
         match: './13ac107a-f15f-49d2-97e8-60ab2e3b519c:99d38770-49c7-49d3-b567-88f393ffb4fe.xhtml',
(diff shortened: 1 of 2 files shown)


src/app/content/utils.spec.ts
```diff
@@ -116,7 +116,7 @@ describe('getContentPageReferences', () => {
       },
       {
         bookId: '13ac107a-f15f-49d2-97e8-60ab2e3b519c',
-        bookVersion: undefined,
+        bookVersion: 29.7,
```
bethshook, line 119: @TomWoodward is there a reason we would ever want this to be undefined?

src/app/content/utils.spec.ts
```diff
@@ -116,7 +116,7 @@ describe('getContentPageReferences', () => {
       },
       {
         bookId: '13ac107a-f15f-49d2-97e8-60ab2e3b519c',
-        bookVersion: undefined,
+        bookVersion: '29.7',
```
bethshook, line 119: @TomWoodward is there a scenario where a book would exist in the config but we would want this to be undefined?

TomWoodward: yea i think we want this to be undefined. i think it'll get populated downstream. with this test here i'm not sure how we were getting the wrong version though

TomWoodward (changes requested):

TomWoodward (approved):

PR #1366, opened 2021-10-20
Vertical navigation

jarosik10:
for: https://github.com/openstax/unified/issues/1632


TomWoodward (changes requested):
remove vscode config please

also i tried to print and then the print button went away

jarosik10:
@TomWoodward 
> also i tried to print and then the print button went away

What exactly do you mean? The print button from the vertical navigation?

bethshook:
Confirming that sometimes the print button disappears. It happens to me if I reload the page.

jarosik10:
@TomWoodward 
Test (build) fails on prerender of `SidebarControl`:
```
BUG: Window is undefined Error: BUG: Window is undefined
    at assertWindow (/__w/rex-web/rex-web/src/app/utils/browser-assertions.ts:3:11)
    at useMatchMediaQuery (/__w/rex-web/rex-web/src/app/reactUtils.ts:160:22)
    at useMatchMediaQuery (/__w/rex-web/rex-web/src/app/reactUtils.ts:180:42)
    at /__w/rex-web/rex-web/src/app/content/components/SidebarControl.tsx:108:20
    at processChild (/__w/rex-web/rex-web/node_modules/react-dom/cjs/react-dom-server.node.development.js:2888:14)
    at resolve (/__w/rex-web/rex-web/node_modules/react-dom/cjs/react-dom-server.node.development.js:2812:5)
    at ReactDOMServerRenderer.render (/__w/rex-web/rex-web/node_modules/react-dom/cjs/react-dom-server.node.development.js:3202:22)
    at ReactDOMServerRenderer.read (/__w/rex-web/rex-web/node_modules/react-dom/cjs/react-dom-server.node.development.js:3161:29)
    at renderToString (/__w/rex-web/rex-web/node_modules/react-dom/cjs/react-dom-server.node.development.js:3646:27)
    at renderHtml (/__w/rex-web/rex-web/script/prerender/contentPages.tsx:92:11)
```
Do you know what would be a good approach to handle it?

TomWoodward:
you can't use window in rendering components because it won't be defined when rendering in nodejs.

the original TOC implementation addresses this by having a "default" state which is before the user has pressed open/close, when in this state css media queries will render the bar as either open or closed, so the server can render this layout without caring if it is open or closed, its handled in css.

src/app/components/NavBar/styled.tsx
```diff
@@ -140,9 +140,22 @@ export const Link = styled.a`
...
-export const DropdownToggle = styled.span`
-  ${navElementStyle}
+export const DropdownToggle = styled.div`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 35px;
+  height: 35px;
+  border-radius: 50%;
+  line-height: 16px;
```
bethshook, line 150: Any reason not to use rem here?

src/app/components/NavBar/styled.tsx
```diff
@@ -286,7 +299,7 @@ export const BarWrapper = styled.div`
   z-index: ${theme.zIndex.navbar}; /* above book nav */
   background: ${theme.color.neutral.base};
   position: relative; /* drop shadow above notifications */
-  padding: 0 ${theme.padding.page.desktop}rem;
+  padding: 0 3.2rem;
```
bethshook, line 302: Why this change?

src/app/content/components/BookBanner.tsx
```diff
@@ -141,7 +141,7 @@ export const BarWrapper = styled.div<BarWrapperProps>`
   ${disablePrint}
 
   top: 0;
-  padding: 0 ${theme.padding.page.desktop}rem;
+  padding: 0 3.2rem;
```
bethshook, line 144: Isn't this the same thing?

src/app/content/components/Page/PageContent.tsx
```diff
@@ -40,7 +40,7 @@ export default styled(MainContent)`
      */
     #${MAIN_CONTENT_ID} > [data-type="page"],
     #${MAIN_CONTENT_ID} > [data-type="composite-page"] {
-      margin-top: ${theme.padding.page.desktop}rem;
+      margin-top: 3.2rem;
```
bethshook, line 43: Same q here as above

src/app/content/components/SidebarControl.tsx
```diff
@@ -80,17 +109,27 @@ const connector = connect(
...
     {...props}
     data-testid='toc-button'
-    message={isOpen ? openMessage : closedMessage}
-    data-analytics-label={isOpen ? 'Click to close the Table of Contents' : 'Click to open the Table of Contents'}
-    onClick={isOpen ? props.closeToc : props.openToc}
-  />);
+    message={isToCOpened === false ? closedMessage : openMessage}
+    data-analytics-label={isToCOpened === false
+      ? 'Click to open the Table of Contents'
+      : 'Click to close the Table of Contents'}
```
bethshook, line 122: These should be internationalized

src/app/content/components/Content.tsx
```diff
@@ -50,21 +49,28 @@ const Background = styled.div`
...
+  grid-column: 1 / -1;
+  grid-row: 1;
+  justify-self: center;
+  width: 100%;
   overflow: visible; /* so sidebar position: sticky works */
   margin: 0 auto;
   max-width: ${contentWrapperMaxWidth}rem;
+  ${theme.breakpoints.mobile(css`
+    grid-column-start: 2;
+  `)}
```
bethshook, line 73: I think the grid here is breaking the layout on very small screens

IMG_9001
IMG_9003

src/app/content/components/Page/PageNotFound.tsx
```diff
@@ -33,6 +34,17 @@ const PageNotFoundText = styled.div`
...
+// tslint:disable-next-line: variable-name
+export const StyledOpenSidebarControl = styled(OpenSidebarControl)`
+  min-height: unset;
+  flex-direction: row;
+  justify-content: start;
+
+  ${ToCButtonText} {
+    ${textRegularSize};
+  }
+`;
```
bethshook, line 46: I think it would make more sense to store this in SidebarControl in case we want to reuse it elsewhere

src/app/messages/es/messages.json
```diff
@@ -159,6 +159,8 @@
   "i18n:toolbar:buy-book:aria-label:text": "Comprar libro",
   "i18n:toolbar:buy-book:text": "Solicitar una copia impresa",
   "i18n:practice-questions:popup:answers:choice": "Elección {choiceIndicator}",
+  "i18n:toolbar:header:title": "Menu",
+  "i18n:toolbar:mobile-menu:open": "Click to open menu",
```
bethshook, line 163: Menú
Haga clic para abrir el menú

bethshook: (we are rex's internal translation team 😂))

src/app/components/NavBar/styled.tsx
jarosik10: good catch

src/app/content/components/BookBanner.tsx
jarosik10: At some point in my work, I changed this `theme.padding.page.desktop`, hence the discrepancy. Good catch :)

src/app/content/components/SidebarControl.tsx
jarosik10: @bethshook Do we really want to internationalize the `data-analytics-label`? In a lot of places it is hard coded

src/app/content/components/BookBanner.tsx
bethshook: I figured!

src/app/content/components/SidebarControl.tsx
bethshook: Whoops, yeah, I misread what this was

jarosik10:
@bethshook @TomWoodward 
I've found one bug related to the default state of ToC. The ToC Toggle button in vertical navigation gets a different prop if the sidebar is closed or not - like `aria-label` and `onClick` handler. So `isTocOpen === null` on desktop means that ToC is opened and the button should close ToC, but for smaller devices ToC is closed in its default state and the button should open the ToC.

In the previous implementation there was no "toggle" toc button, but two different buttons, what was a good workaround. But using two different button for vertical navigation seems a little bit odd.

Previously I've been checking if the device size is mobile using js, but here comes the issue:
> you can't use window in rendering components because it won't be defined when rendering in nodejs.

Do you have any idea how to handle it? Is is true that I can only deal with using two buttons + css?

src/app/content/components/Page/PageNotFound.tsx
jarosik10: @bethshook 
You mean only the different ToCButtonText styling, am i right? Or the whole component that makes the toc button icon and the text to be next to each other?
obraz

TomWoodward:
there is a bug with the margins in this medium screen size 
image

TomWoodward:
and also at larger sizes the content is not centered in the white area 
image

TomWoodward:
i see what you mean about the default open/close behavior, it is a little weirder in the vertical nav especially because in the future there will be more buttons and more panel states.

i think its still ok to have the table of contents button be special and have two buttons with media queries to handle this in css, that seems less gross to me than having a css media query in the click handler

src/app/content/components/Page/PageNotFound.tsx
bethshook: I was thinking the whole component

jarosik10:
In `Content.browserspec.tsx` we are making the scroll tests for desktop viewport which is 1250x874 and mobile viewport 411x731. At the width=1250px, the vertical navigation sticks to left side of the window, so it makes the width of the content a little bit smaller than it used to be.
Zrzut ekranu z 2021-12-13 14-42-11
Zrzut ekranu z 2021-12-13 14-42-30

Do we want to adjust these scrollTop values to pass the tests?

TomWoodward:
@jarosik10 so all the mobile offsets should stay the same but the desktop ones will be a little different because of the addition of the space occupied by the navbar right? that sounds reasonable i think you can go ahead and update the offsets

TomWoodward:
also why are your images broken

jarosik10:
@TomWoodward I've manually injected the html of the `test-page-for-generic-styles`. It's fine;)

src/app/messages/es/messages.json
```diff
@@ -161,6 +161,7 @@
   "i18n:practice-questions:popup:answers:choice": "Elección {choiceIndicator}",
   "i18n:toolbar:header:title": "Menú",
   "i18n:toolbar:mobile-menu:open": "Haga clic para abrir el menú",
+  "i18n:toolbar:mobile-menu:close": "Click to close menu",
```
jarosik10, line 164: @bethshook 
Needs translation:)

bethshook: Haga clic para cerrar el menú

TomWoodward (changes requested):

src/app/components/NavBar/index.tsx
```diff
@@ -63,14 +63,10 @@ export const Dropdown: React.FunctionComponent<{user: User, currentPath: string}
...
-const DropdownToggle: SFC<{user: User}> = ({user}) =>
-  <FormattedMessage id='i18n:nav:hello:text' values={{name: user.firstName}}>
-    {(msg) => <Styled.DropdownToggle tabIndex='0' data-testid='user-nav-toggle'>
-      {msg}
-      <Styled.DownIcon aria-hidden='true' />
-      <Styled.HamburgerIcon ariaHidden='true' />
-    </Styled.DropdownToggle>}
-  </FormattedMessage>;
+const DropdownToggle: SFC<{user: User}> = ({user: { firstName }}) => {
+  const initials = (firstName[0] + firstName.slice(-1)).toUpperCase();
```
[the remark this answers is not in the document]
TomWoodward: https://openstax.slack.com/archives/C8V1XET1R/p1641918768045500?thread_ts=1641918206.044000&cid=C8V1XET1R

> 8px below so that there's enough tap room

src/app/content/utils/domUtils.ts
```diff
@@ -88,6 +90,28 @@ export const setSidebarHeight = (sidebar: HTMLElement, window: Window) => {
   };
 };
 
+export const setContentAdditionalPadding =  (content: HTMLElement, window: Window) => {
```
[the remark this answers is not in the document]
bethshook: I think it's ensuring the content area doesn't get covered up by the sidebar between 1200 and 1480px ([video](https://share.getcloudapp.com/bLuxjr6G)). Here is my attempt to handle it with CSS: https://github.com/openstax/rex-web/compare/issue-1632...issue-1632-content-row

src/app/content/components/Content.spec.tsx
```diff
@@ -268,7 +272,7 @@ describe('content', () => {
 
     renderer.act(() => {
       component.root
-        .findByProps({ 'aria-label': 'Click to close the Table of Contents' })
+        .findAllByProps({ 'aria-label': 'Click to close the Table of Contents' })[0]
```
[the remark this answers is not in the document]
bethshook: no, `findByProps` throws an error if more than one is found

src/app/content/components/TableOfContents/styled/wrapper.tsx
```diff
@@ -97,35 +106,50 @@ export const SidebarBody = styled.div<{isOpen: State['tocOpen']}>`
...
+  color: ${toolbarIconColor.base};
+  flex: 1 1 0%;
+  text-align: left;
+  ${theme.breakpoints.mobileMedium(css`
+    text-align: center;
+  `)}
+`;
+
+// tslint:disable-next-line: variable-name
+export const CloseToCAndMobileMenuButton = styled((props) => {
```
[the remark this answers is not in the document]
bethshook: We need to be able to close the TOC without closing the mobile menu because that's what the back arrow is for. But I think it's ok to close the TOC whenever we close the mobile menu. In that case I think I could just move these styles over to SidebarControl and rename some things.

src/app/content/components/SidebarControl.tsx
```diff
@@ -1,51 +1,51 @@
 import React from 'react';
```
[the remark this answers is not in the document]
bethshook: This file is back to only dealing with the TOC except for the button that controls both TOC and mobile menu.

src/app/components/ScrollLock.tsx
```diff
@@ -51,13 +60,13 @@ const fadeIn = keyframes`
 
 interface OverlayProps extends HTMLAttributes<HTMLDivElement> {
   className?: string;
-  mobileOnly?: boolean;
+  mediumOnly?: boolean;
```
bethshook, line 63: Replaced because the overlay was serving no purpose on mobile and was just being hidden via z-index. Now the overlay is only used between mobile and mediumMobile breakpoints.

src/app/content/actions.ts
```diff
@@ -14,6 +14,9 @@ export const openToc = createStandardAction('Content/openToc')<void>();
 export const closeToc = createStandardAction('Content/closeToc')<void>();
 export const resetToc = createStandardAction('Content/resetToc')<void>();
 
+export const openMobileMenu = createStandardAction('Content/openMobileMenu')<void>();
+export const closeMobileMenu = createStandardAction('Content/closeMobileMenu')<void>();
```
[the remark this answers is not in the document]
bethshook: I don't think that will work because there's no child or sibling relationship between this menu and the button that opens it.

src/app/components/Layout.tsx
```diff
@@ -25,7 +25,13 @@ export const wrapperPadding = css`
 // tslint:disable-next-line:variable-name
 export const LayoutBody = styled.div`
```
TomWoodward, line 26: these changes are breaking the dev homepage, in this place, `/app/components`, we shouldn't be making assumptions that the content will follow the content layout, the grid described here is content specific and should be defined  as the content layout

TomWoodward:
display issue with the sidebar covering the toolbar on larger screens 
image

src/app/components/ScrollLock.tsx
```diff
@@ -13,11 +13,11 @@ const ScrollLockBodyClass = createGlobalStyle`
       ${theme.breakpoints.mobile(css`
         @media print {
           #root {
-            display: none;
+            display: block;
```
TomWoodward, line 16: i'm not sure what is up with this.. when i look in heroku the body is still getting `display: none` and it does seem necessary for making the highlights print screen display right. is there something else applying this style? is it necessary here?

src/app/content/actions.ts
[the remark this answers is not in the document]
TomWoodward: but the mobile menu is full screen, can it be moved to be a child of the button that opens it?

TomWoodward (changes requested):

bethshook:
@TomWoodward I think the visual bug is actually because max-width was removed from SearchPrintWrapper - you shouldn't be able to see the toolbar go beyond the sidebar

src/app/components/ScrollLock.tsx
bethshook: I think this was done in error after I reverted a change to the breakpoints.

src/app/content/actions.ts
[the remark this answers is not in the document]
bethshook: It's the same component as the desktop menu - is it worth splitting them up?

TomWoodward (changes requested):

src/app/content/components/Page/MinPageHeight.tsx
```diff
@@ -5,18 +5,19 @@ import { desktopAttributionHeight, mobileAttributionHeight } from '../Attributio
...
-  navDesktopHeight + bookBannerDesktopBigHeight + toolbarDesktopHeight + desktopAttributionHeight;
+  navDesktopHeight + bookBannerDesktopBigHeight + topbarDesktopHeight + desktopAttributionHeight;
 
 const minMobileContentSize =
-  navMobileHeight + bookBannerMobileBigHeight + toolbarMobileHeight + mobileAttributionHeight;
+  navMobileHeight + bookBannerMobileBigHeight + topbarMobileHeight + mobileAttributionHeight;
 
 export default styled.div`
   @media screen {
+    width: 100%;
```
bethshook, line 20: This fixes a bug where error toast messages were being covered up by the nav

src/app/components/ScrollLock.tsx
```diff
@@ -77,6 +77,10 @@ export const Overlay = styled(({ mobileOnly, zIndex, ...props}: OverlayProps) =>
     ${theme.breakpoints.mobile(css`
       display: block;
     `)}
+
+    ${theme.breakpoints.mobileMedium(css`
+      display: none;
+    `)}
```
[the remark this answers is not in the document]
bethshook: I renamed the prop - hopefully this is clearer.

src/app/content/components/Page/PageToasts.tsx
```diff
@@ -4,32 +4,37 @@ import styled, { css } from 'styled-components/macro';
...
 export const getMobileSearchFailureTop = ({mobileToolbarOpen}: {mobileToolbarOpen: boolean}) => mobileToolbarOpen
-  ? bookBannerMobileMiniHeight + toolbarMobileHeight + toolbarMobileSearchWrapperHeight
-  : bookBannerMobileMiniHeight + toolbarMobileHeight;
+  ? bookBannerMobileMiniHeight + topbarMobileHeight + toolbarMobileSearchWrapperHeight
+  : bookBannerMobileMiniHeight + topbarMobileHeight;
 
 // tslint:disable-next-line:variable-name
-export const ToastContainerWrapper = styled.div`
-  width: 100%;
+export const ToastContainerWrapper = isVerticalNavOpenConnector(styled.div`
```
bethshook, line 24: @TomWoodward this all works unless the page is reloaded with a sidebar open because in that case `content.tocOpen` is null. Is there a way around that?

src/app/components/ScrollLock.tsx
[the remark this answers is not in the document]
bethshook: and yes, mobile is bigger than mobileMedium which I agree is weird. we could rename that as well.

src/app/content/components/constants.ts
```diff
@@ -8,30 +8,35 @@ export const bookBannerMobileMiniHeight = 6;
...
 
-export const toolbarDesktopHeight = 5;
-export const toolbarMobileHeight = 4;
-export const toolbarMobileExpandedHeight = toolbarMobileHeight
+export const topbarDesktopHeight = 5;
+export const topbarMobileHeight = 4;
+export const toolbarMobileExpandedHeight = topbarMobileHeight
   + toolbarMobileSearchWrapperHeight;
 
+export const verticalNavbar = 8;
```
TomWoodward, line 38: vertical navbar... what? width?

src/app/content/components/utils/sidebar.ts
```diff
@@ -3,12 +3,25 @@ import { FlattenSimpleInterpolation } from 'styled-components';
...
 import { AppState } from '../../../types';
-import * as selectors from '../../selectors';
+import * as searchSelectors from '../../search/selectors';
+import * as contentSelectors from '../../selectors';
 import { State } from '../../types';
 
-export const isOpenConnector = connect((state: AppState) => ({isOpen: selectors.tocOpen(state)}));
+export const isVerticalNavOpenConnector = connect((state: AppState) => ({
+  isTocOpen: contentSelectors.tocOpen(state),
+  isVerticalNavOpen: searchSelectors.searchResultsOpen(state) || contentSelectors.tocOpen(state),
```
TomWoodward, line 12: we're gonna have to rearrange this in the future, but this is fine for now

TomWoodward (changes requested):
the initial state of the toc toggle is wrong on desktop, it is initially unselected, when i click it once it becomes selected, when i click it again the toc closes

src/app/content/components/Content.tsx
```diff
@@ -165,51 +144,63 @@ const OuterWrapper = styled.div`
...
-                </MainContentWrapper>
-              </UndoPadding>
-            </ContentPane>
-          </CenteredContentRow>
-        </Wrapper>
-      </OuterWrapper>
-    </ErrorBoundary>
-  </Background>
-</Layout>;
+  }, []);
```
bethshook, line 154: @TomWoodward this is so `tocOpen` doesn't stay set to null when Content renders on desktop

src/app/content/components/constants.ts
bethshook: max width. I've updated it.

src/app/content/components/Content.tsx
TomWoodward: ok, but this produces a mis-alignment. if we're going to give up on css and set the state with js, then we should do that for the whole sidebar. the way this is set up, if i load the page on mobile, then enlarge the window, the toc does appear but the button state becomes wrong.

i prefer the css approach to the default state, but i'm not sure what the issue you're having is with the button

src/app/content/components/Page/PageToasts.tsx
TomWoodward: this component doesn't even seem to be using any of the properties provided by this conector, am i missing something?

src/app/content/components/Topbar/styled.tsx
```diff
@@ -0,0 +1,315 @@
...
+  text-align: right;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: visible;
+  background-color: ${theme.color.neutral.base};
+  transition: padding-left ${sidebarTransitionTime}ms;
+  ${(props) => (props.isVerticalNavOpen === null || props.isVerticalNavOpen) && `
+    padding-left: ${sidebarDesktopWidth}rem;
+  `}
```
TomWoodward, line 242: there used to be helpers for managing the sidebar like `styleWhenSidebarOpen`, not only did this contain the `null` checking logic so that that was handled in one place, it also handled adding multiple media queries so that the null condition was handled properly at different screen sizes.

some of your comments indicated some trouble managing the permutations of sidebar state and null state, you might find it helpful to bring that pattern back

TomWoodward (changes requested):

src/app/content/components/Content.tsx
bethshook: I removed the state logic and fixed the issue in css

src/app/content/components/Content.browserspec.tsx
```diff
@@ -9,7 +9,7 @@ const TEST_CASES: { [testCase: string]: (target: Page) => Promise<void> } = {
   Desktop: setDesktopViewport, Mobile: setMobileViewport,
 };
 const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
-  Desktop: [242, 90, 122, 242, 365, 668, 761, 1262, 1606],
+  Desktop: [242, 90, 122, 242, 365, 668, 761, 1253, 1597],
```
bethshook, line 12: @TomWoodward I don't know why this continues to be an issue but it may be because the content pane max-width is slightly narrower (80px) than before. I assume we want to fix this but it's tricky because of the complex layout. Wondering if we want to fix before or after merging.

src/app/messages/en/messages.json
```diff
@@ -179,13 +182,13 @@
...
   "i18n:toolbar:highlights:popup:help-info": "Need help? Visit the <a data-testid='support-link' data-analytics-label='support-link' target=\"_blank\" href=\"https://openstax.secure.force.com/help/articles/FAQ/Using-the-highlighting-and-note-taking-feature-on-mobile-or-tablet?search=highlighting\">Support Center</a>",
-  "i18n:toolbar:highlights:text": "My highlights",
+  "i18n:toolbar:highlights:text": "Highlights",
   "i18n:toolbar:practice-questions:button:text": "Practice",
   "i18n:toolbar:print:aria-label": "print",
   "i18n:toolbar:print:text": "Print",
   "i18n:toolbar:search:placeholder": "Search this book",
   "i18n:toolbar:search:toggle": "Search this book",
-  "i18n:toolbar:studyguides:button:text": "Study guides",
+  "i18n:toolbar:studyguides:button:text": "Guides",
```
bethshook, line 191: I made some educated assumptions about Polish for this change.

TomWoodward (approved):

PR #1371, opened 2021-10-29
remove local server from prerender

TomWoodward:
for: openstax/unified#1746


jarosik10 (changes requested):
build test is failing.

TomWoodward:
yea its unstable right now, i had to run it like 6 times on the other PR to get it to pass

jarosik10 (approved):

TomWoodward:
oh i see there is actually an issue here, the proxy server was necessary for the tests to hit the fixtures, i didn't hit it locally because of the pre-rendering cache

i'll convert back to draft and work on it

TomWoodward:
i'm thinking about abandoning this because of the issues with the fixtures that i had forgotten about happening in the proxy server.

when running the app, fixture resolution must happen in the dev server because the running app doesn't have access to the file system

so if we were to remove the dev server from the pre-rendering machinery, we'd have to change the app to do fixture resolution itself through some override of the fetch behavior, this is possible because pre-rendering is running in node-js not a browser

but the differences between these two things, the data provided as part of the fetch call vs the inbound request in the dev server proxy middleware, are really different. it seems like more trouble than its worth to try to shove some fetch middleware into the app and then reconcile the input, plus it would introduce a lot of different logic/code in the pre-render vs runtime environments, even if only in testing that still reduces the value of the tests.

PR #1379, opened 2021-11-09
Update key term results logic; add check for accepted search doc types

bethshook:
for: openstax/unified#1751


src/app/content/search/utils.ts
```diff
@@ -222,7 +236,9 @@ export const getFilteredResults = (searchResults: SearchResult) => ({
   ...searchResults,
   hits: {
     ...searchResults.hits,
-    hits: searchResults.hits.hits.filter((hit) => !matchKeyTermHit(hit) || !!hit.highlight.title)},
+    hits: searchResults.hits.hits.filter((hit) =>
+      matchAcceptedDocTypeHit(hit) && (!matchKeyTermHit(hit) || !!hit.highlight.title)
```
bethshook, line 240: In the future, when we're no longer receiving key terms results we don't want, we can just rely on the first condition

TomWoodward: leave this in a code comment

src/app/content/search/utils.ts
```diff
@@ -152,7 +158,12 @@ export const highlightResults = (
       return {result: hit, highlights: {}};
     }
 
-    const hitHighlights = hit.highlight.visibleContent.map((highlightText, index) => {
+    // necessary for correct search highlighting of key term results
+    if (isKeyTermHit(hit) && hit.highlight.title) {
```
TomWoodward, line 162: long term i don't think this is the right way to do this, but we can do it for now. if we were to develop different element types that we search each might have different logic for how it becomes highlighted.

TomWoodward (approved):

PR #1387, opened 2021-11-23
Writing guide content icons

bethshook:
for: openstax/unified#1724


jarosik10 (commented):
@bethshook 
So we want the icons the be in the next, not next to the text?
obraz
obraz

TomWoodward:
Yea we iterated on the layout because our margins are smaller than the pdf.
Larissa approved the design with them inline at the beginning of the
paragraph

generic-styles/index.less
```diff
@@ -79,14 +79,18 @@ body {
   }
 
   span {
+    &.os-icons {
+      display: flex;
+      float: left;
+    }
```
jarosik10, line 85: These styles are no longer applied, because `.os-icons` wrapper is a `div`.

jarosik10 (changes requested):

generic-styles/index.less
bethshook: That's right - CE is going to update the structure so it will work once we have a new pipeline.

bethshook: New structure ([per this PR](https://github.com/openstax/recipes/pull/165/files)):
```
<p class="annotation dreaming-icon" id="auto_88eb0fc4-6144-4c92-84ff-56f8f811e4f7_para-00005">
   <span class="os-icons">
      <span aria-label="Generating and capturing ideas icon" class="dreaming-icon" role="img" title="Generating and capturing ideas icon"></span>
   </span>
   <span class="os-text">In other words, the more you write, the more confidence you will have in your voice and your writing.</span>
</p>
```

jarosik10 (approved):

PR #1388, opened 2021-11-30
Fine tune content reference matching

bethshook:
for: openstax/unified#1759

### DIFF
--- a/src/app/content/utils.ts
+++ b/src/app/content/utils.ts
@@ -36,8 +36,8 @@ const referenceRegex = `(?<matchPath>((${pathRegex})|(${hashRegex})))`;
 export function getContentPageReferences(book: ArchiveBook, page: ArchivePage) {
   const domParser = new DOMParser();
   const domNode = domParser.parseFromString(page.content, 'text/html');
-  const references = Array.from(domNode.links)
-    .map((link) => (link instanceof HTMLAnchorElement && link.getAttribute('href')) || '');
+  const references: string[] = Array.from(domNode.querySelectorAll('a'))
+    .map((link) => (link as HTMLAnchorElement).getAttribute('href') || '');
 
   const matches: ContentPageRefencesType[] = (
     references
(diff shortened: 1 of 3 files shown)


src/app/content/utils.ts
```diff
@@ -30,11 +31,17 @@ export interface ContentPageRefencesType {
...
+  const domNode = domParser.parseFromString(page.content, 'text/html');
+  const references = Array.from(domNode.links)
+    .map((link) => (link as HTMLAnchorElement).getAttribute('href') || '')
+    .filter((href) => href && href.match(new RegExp(referenceRegex, 'g')));
+
   const matches: ContentPageRefencesType[] = (
-    page.content.match(new RegExp(referenceRegex, 'g')) || []
+    references
   )
     .map((match) => match.match(new RegExp(referenceRegex))?.groups)
```
TomWoodward, line 46: you can combine these method chains, you're just staring with the result of the previous one here

src/app/content/utils.ts
```diff
@@ -30,11 +31,17 @@ export interface ContentPageRefencesType {
...
 const pathRegex = `\\./((?<bookId>[a-z0-9-]+)(@(?<bookVersion>[^/]+))?):(?<pageId>[a-z0-9-]+)\\.xhtml(${hashRegex})?`;
-const referenceRegex = `['"]{1}(?<matchPath>((${pathRegex})|(${hashRegex})))`;
+const referenceRegex = `(?<matchPath>((${pathRegex})|(${hashRegex})))`;
 
 export function getContentPageReferences(book: ArchiveBook, page: ArchivePage) {
+  const domParser = new DOMParser();
+  const domNode = domParser.parseFromString(page.content, 'text/html');
+  const references = Array.from(domNode.links)
+    .map((link) => (link as HTMLAnchorElement).getAttribute('href') || '')
+    .filter((href) => href && href.match(new RegExp(referenceRegex, 'g')));
```
TomWoodward, line 41: you can remove this `'g'` and combine this with the next match below, you're already doing them one at a time now so you don't need the two step identification and then group extraction. you can just extract groups and then filter undefined

TomWoodward (changes requested):

TomWoodward (approved):

PR #1398, opened 2021-12-17
Prerender workers

Dantemss:
For: https://github.com/openstax/unified/issues/1797

Instructions:
1. Deploy the infrastructure stack in https://github.com/openstax/unified-deployment/pull/239 (already done in the sandbox)
2. `cd rex-web`
3. Export the following environment variables (you can save this as .env.local for example, then source it):
```
export REACT_APP_CODE_VERSION="test"
export REACT_APP_RELEASE_ID="test/2022-01-12"
export PUBLIC_URL="/rex/releases/$REACT_APP_RELEASE_ID"
export REACT_APP_ENV="production"
export PRERENDER_IMAGE_TAG=$(tr / - <<< $REACT_APP_RELEASE_ID)
```
4. Build and tag an image for the prerender target: `docker build . --target prerender -t 373045849756.dkr.ecr.us-east-2.amazonaws.com/rex:$PRERENDER_IMAGE_TAG`
5. Login to ECR: `aws ecr get-login-password --region us-east-2 | docker login --username AWS --password-stdin 373045849756.dkr.ecr.us-east-2.amazonaws.com`
6. To assume the Builder and Manager roles, you'd need to modify the openstax-users account, but you can use `sandbox:full-admin` instead.
7. As the Builder role (or full-admin), push the image: `docker push 373045849756.dkr.ecr.us-east-2.amazonaws.com/rex:$PRERENDER_IMAGE_TAG`
8. As the Manager role (or full-admin), run: `yarn prerender:fleet`


Dantemss:
Not entirely sure where the `src/app/navigation` changes are from... maybe I need to rebase something?
Also this changes one thing in the actual build if merged which is `verify-production-urls`. It should work, but I haven't tested it with Concourse so we'd have to test.

Dantemss:
The navigation changes were actually in the lambda PR... if they don't belong let me know and I can take them out.

script/prerender-workers/work.ts
```diff
@@ -0,0 +1,262 @@
...
+    return worker;
+  } else {
+    // console.log('All worker threads busy; main thread waiting for available worker thread');
+
+    // Return a promise that will be resolved when pushWorker is called
+    return new Promise<SQSWorker>((resolve) => { resolveWorkerPromises.push(resolve); });
+  }
+}
+
+async function sqsHeartbeat(messages: Message[]) {
```
TomWoodward, line 139: what is sqs heartbeat?

script/server/index.ts
```diff
@@ -6,6 +6,8 @@ import serveStatic from 'serve-static';
 import config from '../../src/config';
 import setupProxy from '../../src/setupProxy';
 
+export { Server } from 'http';
```
TomWoodward, line 9: whats this for?

TomWoodward (changes requested):

Dockerfile
```diff
@@ -26,11 +26,25 @@ RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | b
...
+
+# Docker trickery so we can reuse the yarn install layer until package.json or yarn.lock change
+COPY package.json yarn.lock /code/
+WORKDIR /code
+RUN yarn install
+
+COPY . /code
+RUN yarn build
+
+ENTRYPOINT ["yarn"]
```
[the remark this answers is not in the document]
Dantemss: My intention was for the ENTRYPOINT and CMD to apply only to the prerender target, which is not used in the CI, but you are right that I have not tested if that is true. I'll see if I can test that this is the case.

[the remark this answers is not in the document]
Dantemss: I have tested and confirmed that ENTRYPOINT and CMD apply only to the prerender target:

- When building an image with `docker build .` then running `docker run image-id` I get no message and exit code 0.
- When building an image with `docker build .` then running `docker run image-id echo Hello` I get `Hello`.
- When building an image with `docker build . --target prerender` then running `docker run image-id`, I get a process running `yarn prerender:work`
- When building an image with `docker build . --target prerender` then running `docker run image-id help`, I get yarn's help message

concourse/prerender-workers/script.bash
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -eux
```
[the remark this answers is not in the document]
Dantemss: I'll take it out and make a new PR

concourse/verify-production-urls/script.bash
```diff
@@ -6,14 +6,29 @@ production_url=https://openstax.org
 work_dir=$(pwd)
 failed=0
 
-release_id=$(curl -s "${production_url}/rex/environment.json" | jq .release_id -r)
```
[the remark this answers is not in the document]
Dantemss: I separated this into another PR and I'll continue to refactor it to move it after the uploads and to look at the buckets only. But for the purposes of this PR, it's been removed, so done.

script/prerender-workers/cfn.yml
```diff
@@ -0,0 +1,209 @@
+AWSTemplateFormatVersion: 2010-09-09
```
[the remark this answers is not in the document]
Dantemss: I pinged him, waiting for an answer.

script/prerender-workers/manage.ts
```diff
@@ -0,0 +1,396 @@
...
+// This is insurance in case this process gets stuck or crashes without deleting the workers stack
+const PRERENDER_TIMEOUT_SECONDS = 1800;
+
+// Abort the build if the workers stack is not ready after this many seconds
+const WORKERS_DEPLOY_TIMEOUT_SECONDS = 120;
+
+const BUCKET_NAME = process.env.BUCKET_NAME || 'sandbox-unified-web-primary';
+const BUCKET_REGION = process.env.BUCKET_REGION || 'us-east-1';
+const PUBLIC_URL = process.env.PUBLIC_URL || `/rex/releases/${RELEASE_ID}`;
+const RELEASE_ID_SUFFIX = RELEASE_ID.split('/', 2)[1];
```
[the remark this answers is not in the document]
Dantemss: Yeah the ECR repo doesn't like the slashes

script/prerender-workers/manage.ts
```diff
@@ -0,0 +1,396 @@
...
+        ParameterKey: 'CodeVersion',
+        ParameterValue: CODE_VERSION,
+      },
+      {
+        ParameterKey: 'PublicUrl',
+        ParameterValue: PUBLIC_URL,
+      },
+      {
+        ParameterKey: 'ReleaseIdSuffix',
+        ParameterValue: RELEASE_ID_SUFFIX,
```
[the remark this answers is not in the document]
Dantemss: The worker uploads using BUCKET_REGION, BUCKET_NAME and PUBLIC_URL, so I guess that knowledge is included in the PUBLIC_URL.

script/server/index.ts
Dantemss: Probably a leftover from when I was using the proxy. I'll take it out.

script/prerender-workers/work.ts
Dantemss: I should probably write this as a comment in there :)

script/prerender-workers/contentPages.tsx
```diff
@@ -0,0 +1,254 @@
...
+import { Link, Meta } from '../../src/app/head/types';
+import * as navigationSelectors from '../../src/app/navigation/selectors';
+import { AnyMatch } from '../../src/app/navigation/types';
+import { matchPathname } from '../../src/app/navigation/utils';
+import { AppState } from '../../src/app/types';
+import { assertDefined } from '../../src/app/utils';
+import FontCollector from '../../src/helpers/FontCollector';
+import { readAssetFile } from './fileUtils';
+
+export type BookMatch = {
```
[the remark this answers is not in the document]
Dantemss: I combined the prerender-workers into the old prerender folder, hopefully without modifying any of the old behavior.

script/prerender-workers/thread.ts
```diff
@@ -0,0 +1,146 @@
...
+        break;
+      }
+      case 'sitemap': {
+        const payload = task.payload as SitemapPayload;
+        const pages = payload.pages.map((page: SerializedPageMatch) => deserializePage(page));
+        const items = await asyncPool(MAX_CONCURRENT_CONNECTIONS, pages, async(page) => {
+          const archivePage = await getArchivePage(page);
+          return getSitemapItemOptions(archivePage, matchPathname(page));
+        });
+        await renderSitemap(payload.slug, items);
```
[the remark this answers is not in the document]
Dantemss: Moved them into separate functions and changed to a map of task types to functions instead of a switch statement.

script/prerender/fileUtils.ts
```diff
@@ -42,3 +47,34 @@ export const createDiskCache = <K extends string, V>(prefix: string): Cache<K, V
     },
   };
 };
+
+let s3Client: S3Client | undefined;
+
+async function writeS3File(contentType: string, filepath: string, contents: string) {
+  let basePath = assertDefined(process.env.PUBLIC_URL, 'PUBLIC_URL environment variable not set');
+  if (basePath[0] === '/') { basePath = basePath.slice(1); }
+  const key = `${basePath}${filepath}`;
```
TomWoodward, line 56: the presence of this logic makes the function more than just writing files to s3, there aren't a lot of rexy files outside of the release directories but there are some, it would be nice to separate this logic into like... `writeS3File` and `prefixReleasePath` or something

script/prerender/fileUtils.ts
```diff
@@ -42,3 +47,34 @@ export const createDiskCache = <K extends string, V>(prefix: string): Cache<K, V
...
+  return s3Client.send(new PutObjectCommand({
+    Body: contents,
+    Bucket: process.env.BUCKET_NAME,
+    CacheControl: 'max-age=0',
+    ContentType: contentType,
+    Key: key,
+  }));
+}
+
+export const writeS3HtmlFile = writeS3File.bind(null, 'text/html');
```
TomWoodward, line 79: i'd have done this with function scope, with a semi curried function, rather than bind and messing around with the whole `this` context

i think that is just personal preference though i don't know if there is necessarily anything wrong with doing it this way

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
+// tslint:disable:no-console
+
+/*
+  Manages a fleet of spot instances
+  Creates a cloudformation stack with a spot fleet and SQS queues, preloads all the book ToCs,
+  queues up each page in the queue to be prerendered, and deletes the stack at the end
+*/
+
+import './setup';
```
TomWoodward, line 9: if this file isn't rendering any react code it probably doesn't need this setup thing

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+// This is insurance in case this process gets stuck or crashes without deleting the workers stack
+const PRERENDER_TIMEOUT_SECONDS = 1800;
+
+// Abort the build if the workers stack is not ready after this many seconds
+const WORKERS_DEPLOY_TIMEOUT_SECONDS = 120;
+
+const BUCKET_NAME = process.env.BUCKET_NAME || 'sandbox-unified-web-primary';
+const BUCKET_REGION = process.env.BUCKET_REGION || 'us-east-1';
+const PUBLIC_URL = process.env.PUBLIC_URL || `/rex/releases/${RELEASE_ID}`;
+const RELEASE_ID_SUFFIX = RELEASE_ID.split('/', 2)[1];
```
TomWoodward, line 62: this doesn't really have any value when split out, i think it would make more sense to string replace the blocked characters rather than just using the last bit

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+let bookLoader: ReturnType<typeof makeUnifiedBookLoader>;
+
+let resolveWorkQueuePromise: (workQueueUrl: string) => void;
+const workQueuePromise = new Promise<string>((resolve) => { resolveWorkQueuePromise = resolve; });
+
+let timeoutDate: Date;
+let numBooks = 0;
+let numPages = 0;
+
+async function renderManifest() {
```
TomWoodward, line 80: this is duplicated from `script/prerender/prerender.tsx`

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+        ParameterKey: 'PublicUrl',
+        ParameterValue: PUBLIC_URL,
+      },
+      {
+        ParameterKey: 'ReleaseIdSuffix',
+        ParameterValue: RELEASE_ID_SUFFIX,
+      },
+      {
+        ParameterKey: 'ValidUntil',
+        ParameterValue: `${timeoutDate.toISOString().slice(0, -5)}Z`,
```
TomWoodward, line 123: what are you slicing off the end here? the zone? why?

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  const numBookPages = pages.length;
+  numPages += numBookPages;
+
+  // Wait until the work queue is ready
+  const workQueueUrl = await workQueuePromise;
+
+  console.log(`[${book.title}] Queuing ${numBookPages} book pages in batches of 10`);
+
+  for (let pageIndex = 0; pageIndex < pages.length; pageIndex += 10) {
+    const pageBatch = pages.slice(pageIndex, pageIndex + 10);
```
TomWoodward, line 161: you can use `'lodash/chunk'`

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+
+    // If the entire request fails, this command will throw and be caught at the end of this file
+    // However, we also need to check if only some of the messages failed
+    const sendMessageBatchResult = await sqsClient.send(new SendMessageBatchCommand({
+      Entries: pageBatch.map((page, batchIndex) => ({
+        Id: batchIndex.toString(),
+        MessageBody: JSON.stringify({ payload: { page }, type: 'page' }),
```
TomWoodward, line 168: why the wrapping page key? is there something wrong with the payload just being the `{state, params}` object?

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+        JSON.stringify(successfulMessages)}`);
+    }
+  }
+
+  console.log(`[${book.title}] All ${numBookPages} book pages queued`);
+
+  // console.log(`[${book.title}] Queuing sitemap`);
+
+  await sqsClient.send(new SendMessageCommand({
+    MessageBody: JSON.stringify({ payload: { pages, slug: book.slug }, type: 'sitemap' }),
```
TomWoodward, line 194: you don't have any typing for these queue bodies, it might be helpful to create some aliases for encoding them and some guards that match the type on read so that you can get strong typing

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  return {
+    params: { book: { slug: book.slug } },
+    state: { bookUid: book.id, bookVersion: book.version },
+  };
+}
+
+async function manage() {
+  console.log('Preloading route components');
+
+  await Loadable.preloadAll();
```
TomWoodward, line 210: this is for the rendering, its not necessary here

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  archiveLoader = createArchiveLoader(REACT_APP_ARCHIVE_URL, {
```
TomWoodward, line 212: you should just define these above, there is no point in letting them be undefined until now

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+
+  const stacks = assertDefined(
+    describeStacksResult.Stacks, '[CFN] [DescribeStacks] Unexpected response: missing Stacks key'
+  );
+  const stack = assertDefined(stacks[0], `${WORKERS_STACK_NAME} stack not found`);
+  const outputs = assertDefined(
+    stack.Outputs, '[CFN] [DescribeStacks] Unexpected response: missing stack Outputs key'
+  );
+
+  let workQueueUrl: string | undefined;
```
TomWoodward, line 249: i get that its less efficient to loop over the outputs twice but since there are only like two outputs i feel like it would be fine to do a `find` here and it would be much more readable

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  console.log('Waiting for all jobs to be queued');
+
+  // Wait for all book pages to be queued
+  const books = await allPagesQueuedPromise;
+
+  numBooks = books.length;
+
+  console.log(`All ${numPages} page prerendering jobs and all ${numBooks} sitemap jobs queued`);
+
+  // console.log('Queuing sitemap index job');
```
TomWoodward, line 290: commented log?

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  console.log('1 sitemap index job queued');
+
+  const numJobs = numPages + numBooks + 1;
+
+  // Ensure the work queue is empty
+
+  console.log('Waiting 1 minute for the work queue attributes to stabilize');
+
+  // First wait 1 minute after sending the last message for the queue attributes to stabilize
+  // This is required according to SQS docs
```
TomWoodward, line 306: link to docs? some of this comment is redundant with the log above and can be removed

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  await new Promise((resolve) => setTimeout(resolve, 60000));
+
+  console.log('Waiting for the work queue to be empty');
+
+  // Now check that all the NumberOfMessages attributes are 0
+  const getQueueAttributesCommand = new GetQueueAttributesCommand({
```
TomWoodward, line 312: this manage function is pretty big, i feel like you could extract this whole waiting for queue to be empty thing into a helper

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  }
+
+  console.log(`All ${numJobs} prerendering jobs finished`);
+
+  console.log('Ensuring that the dead letter queue is empty');
+
+  // Ensure that the dead letter queue is also empty
+  // Since we are the only consumer of the dead letter queue, we use long polling to check
+  // that it is empty rather than waiting another minute and checking the queue attributes
+  const receiveDLQMessageResult = await sqsClient.send(new ReceiveMessageCommand({
```
TomWoodward, line 350: this removes the messages from the queue right? wouldn't that prevent us from being able to redrive them?

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+            We can do it to get notified and prevent having a bunch of failed delete stacks,
+            but for the purposes of the build itself we are already done */
+}
+
+// Start creating the worker stack first since it takes a while
+// Do not wait for the stack creation to complete since we have other tasks to do first
+createWorkersStack().then(() => manage().catch(async(e) => {
+  // catch for manage() only
+  // Do not wait for the stack deletion to complete since we can't handle a delete failure anyway
+  await deleteWorkersStack();
```
TomWoodward, line 390: this should be in a finally so it doesn't need to be called in two places

TomWoodward (changes requested):
i only made it through the manager code

script/prerender/cfn.yml
```diff
@@ -0,0 +1,209 @@
...
+    Type: String
+    ConstraintDescription: Must be an existing S3 bucket
+
+  BucketRegion:
+    Description: AWS Region of the above bucket
+    Type: String
+    ConstraintDescription: Must be a valid AWS region name
+
+  CodeVersion:
+    Description: Code version to use, e.g. "v4"
```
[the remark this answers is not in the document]
Dantemss: I got rid of the CodeVersion, ReleaseIdSuffix and instead added ReleaseId and SanitizedReleaseId.

script/prerender/cfn.yml
```diff
@@ -0,0 +1,209 @@
...
+      ClusterName: !Sub ${AWS::StackName}-cluster
+
+  DeadLetterQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      KmsMasterKeyId: alias/aws/sqs
+      QueueName: !Sub rex-${CodeVersion}-${ReleaseIdSuffix}-prerender-dlq
+      ReceiveMessageWaitTimeSeconds: 20
+      VisibilityTimeout: 0
+      # RedriveAllowPolicy: # leads to circular dependency errors with the worker queue
```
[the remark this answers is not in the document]
Dantemss: This would limit which queues can use this DeadLetterQueue. I don't see a way to set it on creation though, due to the circular dependency issue, only edit it in later.

script/prerender/cfn.yml
```diff
@@ -0,0 +1,209 @@
...
+      #     - !GetAtt WorkQueue.Arn
+
+  WorkQueue:
+    Type: AWS::SQS::Queue
+    DependsOn: Cluster # workaround for Cluster deletion to work properly
+    Properties:
+      KmsMasterKeyId: alias/aws/sqs
+      QueueName: !Sub rex-${CodeVersion}-${ReleaseIdSuffix}-prerender-work-queue
+      ReceiveMessageWaitTimeSeconds: 20
+      RedriveAllowPolicy: '{"redrivePermission": "denyAll"}'
```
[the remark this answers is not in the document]
Dantemss: Can't use the WorkQueue as another queue's dead letter queue.

script/prerender/cfn.yml
```diff
@@ -0,0 +1,209 @@
...
+      #   redrivePermission: byQueue
+      #   sourceQueueArns:
+      #     - !GetAtt WorkQueue.Arn
+
+  WorkQueue:
+    Type: AWS::SQS::Queue
+    DependsOn: Cluster # workaround for Cluster deletion to work properly
+    Properties:
+      KmsMasterKeyId: alias/aws/sqs
+      QueueName: !Sub rex-${CodeVersion}-${ReleaseIdSuffix}-prerender-work-queue
```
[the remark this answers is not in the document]
Dantemss: Sure, I guess. Just need to also randomize the stack name and the image tag in the ECR.

script/prerender/contentPages.tsx
```diff
@@ -31,12 +31,39 @@ export const stats = {
...
+  state: { bookUid: string, bookVersion: string },
+};
+export type SerializedBookMatch = Omit<BookMatch, 'route'>;
+
+export function deserializeBook(book: SerializedBookMatch) {
+  return {...book, route: content};
+}
+
+/* Note: PageMatch can be made more restrictive if single-instance prerender code is removed:
+export type PageMatch = {
```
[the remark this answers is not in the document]
Dantemss: I used to have this type but had to take it out because of the single-instance prerender code.
I'll remove the comment.

script/prerender/fileUtils.ts
Dantemss: MDN docs claim that bind(null) doesn't change `this` but I guess they lied:
`what()` is a function that logs `this`.

```
> obj.what = what
[Function: what]
> obj.what()
{ what: [Function: what] }
undefined
> obj.what = what.bind(null)
[Function: bound what]
> obj.what()
<ref *1> Object [global] {
...
```

I'll rewrite this not to use bind()...

script/prerender/manage.ts
Dantemss: Done

script/prerender/manage.ts
Dantemss: Have to slice off the milliseconds since AWS only accepts seconds

script/prerender/manage.ts
Dantemss: I did it for consistency with the sitemap type which needs them but you are right, it is not necessary so I'll remove it.

script/prerender/manage.ts
Dantemss: I did already add the guards on read, but I'll add the types as well.

script/prerender/manage.ts
Dantemss: I thought they needed the preload, but if that is not the case yeah I'll move them up.

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+    );
+    const outputValue = assertDefined(
+      output.OutputValue,
+      '[CFN] [DescribeStacks] Unexpected response: missing output OutputValue key'
+    );
+
+    switch (outputKey) {
+      case 'WorkQueueUrl':
+        workQueueUrl = outputValue;
+        resolveWorkQueuePromise(workQueueUrl);
```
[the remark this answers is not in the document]
Dantemss: 0 seconds since the workers take some time to boot even after the stack is deployed and the manager can start queuing up some messages before they are ready. :) Fine, I'll rewrite this.

script/prerender/manage.ts
Dantemss: Normally this would make the messages invisible, but the DLQ has the VisibilityTimeout set to 0 which may or may not mean messages can never be successfully deleted from it.
However, the DLQ is deleted when the workers stack is deleted at the end of the build, even if it failed, which happens right after the manager detects any messages in the DLQ.
If you're thinking about wanting to retry failed builds (not from the beginning), then we'd maybe have to split the DLQ into a separate stack.

script/prerender/cfn.yml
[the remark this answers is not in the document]
Dantemss: I think I'll leave the name and just add a random suffix so we can still tell what each of the resources belongs to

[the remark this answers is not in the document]
Dantemss: I guess the image doesn't need the random suffix, I can just allow it to be pushed again. In theory the code should be the same, just the internal docker ids may be different.

script/prerender/cfn.yml
```diff
@@ -0,0 +1,209 @@
...
+          - !ImportValue RexPrerenderWorkersSecurityGroupID
+        TagSpecifications:
+          - ResourceType: instance
+            Tags:
+              - Key: Name
+                Value: !Sub ${AWS::StackName}-instance
+              - Key: Project
+                Value: Unified
+              - Key: Application
+                Value: Rex
```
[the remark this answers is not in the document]
Dantemss: Changed to "Rex Prerender"

script/prerender/contentPages.tsx
```diff
@@ -31,12 +31,39 @@ export const stats = {
   renderHtml: 0,
 };
 
-export async function prepareContentPage(
+export type BookMatch = {
```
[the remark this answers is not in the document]
Dantemss: Done, changed to use `Match<Route<>>`

script/prerender/contentPages.tsx
```diff
@@ -31,12 +31,39 @@ export const stats = {
...
+  params: { book: { slug: string } },
+  route: typeof content,
+  state: { bookUid: string, bookVersion: string },
+};
+export type SerializedBookMatch = Omit<BookMatch, 'route'>;
+
+export function deserializeBook(book: SerializedBookMatch) {
```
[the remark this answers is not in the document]
Dantemss: I did this at first but then it turned out to not be necessary. I could just delete `deserializeBook()`, since the sitemapIndex stuff did not need the `route: content`.

script/prerender/contentPages.tsx
```diff
@@ -31,12 +31,39 @@ export const stats = {
   renderHtml: 0,
 };
 
-export async function prepareContentPage(
+export type BookMatch = {
+  params: { book: { slug: string } },
+  route: typeof content,
+  state: { bookUid: string, bookVersion: string },
+};
+export type SerializedBookMatch = Omit<BookMatch, 'route'>;
```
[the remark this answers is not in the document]
Dantemss: Moved these types and getArchivePage/Book to prerender/contentRoutes

script/prerender/contentPages.tsx
```diff
@@ -118,41 +145,77 @@ export const minuteCounter = () => {
...
+  if (!route.state || !('bookUid' in route.state)) {
+    throw new Error('match state wasn\'t defined, it should have been');
+  }
 
-const makeRenderPage: MakeRenderPage = (services) => async({code, route}) => {
+  const {bookUid, bookVersion} = route.state;
+
+  return assertDefined(
+    await services.archiveLoader.book(bookUid, bookVersion).load(),
+    'book wasn\'t cached, it should have been'
```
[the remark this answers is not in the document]
Dantemss: I removed them and typescript did not complain so I guess they were not necessary.

script/prerender/contentPages.tsx
```diff
@@ -171,15 +234,20 @@ export const prepareBooks = async(
   }));
 };
 
-export type Pages = Array<{code: number, route: Match<typeof content>}>;
+// Note: Could return Array<SerializedPageMatch> if single-instance prerender code is removed
```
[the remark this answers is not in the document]
Dantemss: Removed the "code" argument and changed most of these functions to use SerializedPageMatch, only adding the content route when needed.

script/prerender/manage.ts
TomWoodward: it may become convenient to have this script try to redrive the dlq once or twice before exiting if the build becomes flaky, not necessarily right now but it would be nice if we didn't have to redo this part of the logic if we decide to do that

script/prerender/manage.ts
TomWoodward: we have a date uitll library with a [format](https://date-fns.org/v2.28.0/docs/format) function that you could use to construct the right thing instead of mangling the ISO string

script/prerender/manage.ts
TomWoodward: no `Loadable.preloadAll();` is necessary for the react components in each route, thats how we avoid bundling all the route components at once

TomWoodward:
i looked at the comments but most of them aren't tagged as outdated so i think you still have some changes to make? you can press the request re-review when you're ready for me to look again

script/prerender/fileUtils.ts
Dantemss: The have been rewritten not to use `bind()`.

script/prerender/fileUtils.ts
Dantemss: The function has been split into these 2.

script/prerender/manage.ts
Dantemss: Just removing it leads to an error, but I figured it's actually contentPages that's loading the React stuff so I moved this import there and it works.

script/prerender/manage.ts
Dantemss: Split `renderManifest()` into its own file

script/prerender/manage.ts
Dantemss: It seems date-fns 1.x does not support printing ISO strings if your local time is not GMT.
However, I can update it to 2.x and add the date-fns-tz library.

Dantemss: I added date-fns-tz to convert the Date to UTC. I had to update date-fns because date-fns-tz fails to import with date-fns 1.x

script/prerender/manage.ts
Dantemss: Done

script/prerender/manage.ts
Dantemss: Done

script/prerender/manage.ts
Dantemss: I added types when sending messages. You may want to check the receiving side too though. I have some types there, but it mostly relies on runtime checks.

script/prerender/manage.ts
Dantemss: Removed

script/prerender/manage.ts
Dantemss: Moved this code up to the spot where they are first defined

script/prerender/manage.ts
Dantemss: Removed

script/prerender/manage.ts
Dantemss: I think that is handled in the normal queue. If the worker doesn't delete the message it'll be retried by another worker when the visibility timeout expires. The DLQ only comes into play when the message has already failed some (configurable) number of times. To retry more, we can just increase that number instead.

script/prerender/manage.ts
Dantemss: Reorganized the comments and added links

script/prerender/manage.ts
```diff
@@ -0,0 +1,397 @@
...
+  // We can start fetching book ToCs while the stack is created,
+  // but we need a limit so we don't use up all the memory
+  console.log(`Loading books in batches of ${MAX_CONCURRENT_BOOKS}`);
+
+  const allPagesQueuedPromise = asyncPool(
+    MAX_CONCURRENT_BOOKS, Object.entries(BOOKS), prepareAndQueueBook
+  );
+
+  // Wait for the workers stack to be ready
+  await waitUntilStackCreateComplete(
```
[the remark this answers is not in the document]
Dantemss: I couldn't move this into createWorkersStack() because I want a failure on the create command to not trigger a stack delete command, but a failure on waitUntilStackCreateComplete() to trigger a stack delete command.

However, I extracted this and the part that gets queue urls into a separate function and rewrote a lot of this file to hopefully make everything clearer.

script/prerender/manage.ts
Dantemss: Done, replaced with 2 calls to `find()`

script/prerender/manage.ts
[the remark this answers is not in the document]
Dantemss: Done, it now waits for the stack creation at the beginning and gets the queue urls before it starts loading any of the books.

script/prerender/manage.ts
Dantemss: Split it into smaller pieces and as a bonus got rid of all non-const global variables in this file.

script/prerender/manage.ts
Dantemss: Moved to a try-finally with await instead of then().finally() since that's probably more readable.

script/prerender/contentPages.tsx
```diff
@@ -118,47 +119,63 @@ export const minuteCounter = () => {
...
-  const {bookUid, bookVersion, pageUid} = route.state;
-  const archivePage = assertDefined(
-    await services.archiveLoader.book(bookUid, bookVersion).page(pageUid).load(),
-    'page wasn\'t cached, it should have been'
-  );
+  const html = await renderHtml(styles, app, state);
 
-  const {app, styles, state, url} = await prepareApp(services, route, code);
-  console.info(`rendering ${url}`); // tslint:disable-line:no-console
+  await savePage(url, html);
```
TomWoodward, line 140: do you want to type this on line 130 to return a promise?

script/prerender/fileUtils.ts
```diff
@@ -42,3 +47,46 @@ export const createDiskCache = <K extends string, V>(prefix: string): Cache<K, V
...
+  if (!s3Client) {
+    // We explicitly pass credentials to the S3 Client only once
+    // so it'll not try to load them over and over
+
+    console.log('Fetching container credentials');
+    const credentials = await fromContainerMetadata();
+
+    console.log('Initializing S3 client');
+    s3Client = new S3Client({ credentials, region: process.env.BUCKET_REGION });
+  }
```
TomWoodward, line 70: there is some danger of doing this twice concurrently, you could also do it like
```
const getS3Client = once(async () => {
    console.log('Fetching container credentials');
    const credentials = await fromContainerMetadata();

    console.log('Initializing S3 client');
    return new S3Client({ credentials, region: process.env.BUCKET_REGION });
})

async function writeS3File(key: string, contents: string, contentType: string) {
  const s3Client = await getS3Client();
}


```

this way even if its called concurrently both usages will get the same promise

script/prerender/sitemap.ts
```diff
@@ -7,25 +7,54 @@ import max from 'lodash/fp/max';
...
+export const renderAndSaveSitemap = async(
+  saveFile: (path: string, contents: string) => void,
+  slug: string,
+  urls: SitemapItemOptions[]
+) => {
   const bookSitemap = sitemap.createSitemap({ hostname: 'https://openstax.org', urls });
+
+  const filePath = sitemapPath(slug);
+
+  await saveFile(filePath, bookSitemap.toString());
```
TomWoodward, line 21: do you want to type this as returning a promise?

src/app/utils/assertions.ts
```diff
@@ -27,3 +27,15 @@ export const assertString = <X>(x: X, message: string): string => {
...
+// When the argument is "any", we don't want to cast it to object
+// or we'll be unable to assign it to more specific types
+// So we just exclude all primitive types instead
+export const assertObject = <X>(x: X, message: string) => {
+  // typeof null === 'object'
+  if (x === null || typeof x !== 'object') {
+    throw new Error(message);
+  }
+
+  return x as Exclude<X, number | string | boolean | bigint | symbol | null | undefined>;
```
TomWoodward, line 40: i have a different version of this https://github.com/openstax/ancillary-service/blob/main/utils/src/guards.ts#L6 i'm wondering if one or the other type conversion is better?

TomWoodward (changes requested):

script/prerender/contentPages.tsx
Dantemss: Ah good idea

script/prerender/fileUtils.ts
Dantemss: Ah interesting... ok I'll add this

script/prerender/sitemap.ts
Dantemss: writeAssetFile() is not async but I suppose I can wrap it

src/app/utils/assertions.ts
Dantemss: I think yours is more strict, since it only allows non-class non-array non-function objects as far as I can tell. I use mine on instances of some class so I need the more lax type.

TomWoodward (approved):

PR #1401, opened 2021-12-21
update search highlights on page mount if results already populated

bethshook:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1743

### DIFF
--- a/src/app/content/components/Page/PageComponent.tsx
+++ b/src/app/content/components/Page/PageComponent.tsx
@@ -68,7 +68,7 @@ export default class PageComponent extends Component<PagePropTypes> {
     this.scrollToTopOrHashManager = scrollToTopOrHashManager(this.container.current);
 
     if (this.props.searchHighlights.selectedResult) {
-      this.searchHighlightManager.update({searchResults: [], selectedResult: null}, this.props.searchHighlights, {
+      this.searchHighlightManager.update(null, this.props.searchHighlights, {
         forceRedraw: true,
         onSelect: this.onSearchHighlightSelect,
       });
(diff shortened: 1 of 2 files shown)


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -66,6 +66,13 @@ export default class PageComponent extends Component<PagePropTypes> {
     // tslint:disable-next-line: max-line-length
     this.highlightManager = highlightManager(this.container.current, () => this.props.highlights, this.props.services, this.props.intl);
     this.scrollToTopOrHashManager = scrollToTopOrHashManager(this.container.current);
+
+    if (this.props.searchHighlights.selectedResult) {
+      this.searchHighlightManager.update({searchResults: [], selectedResult: null}, this.props.searchHighlights, {
```
TomWoodward, line 71: maybe we should tweak the manager to allow the prev structure to just be null

TomWoodward (changes requested):
other than that one comment this looks fine

src/app/content/components/Page/PageComponent.tsx
bethshook: this is the update handler so I don't know if it's worth tweaking:
```
const handleUpdate = (services: Services) => (
  previous: HighlightProp,
  current: HighlightProp,
  options: UpdateOptions
) => {
  updateResults(services, previous, current, options);
  selectResult(services, previous, current, options);
};
```

TomWoodward: yea you'd have to change it to be nullable all the way down, which sucks, but also manually constructing the empty state here is pretty jank, so i think that might be a good option

TomWoodward (approved):

PR #1405, opened 2022-01-05
Add data analytics region to key terms results

bethshook:
for: openstax/unified#1773

### DIFF
--- a/pytest-selenium/regions/search_sidebar.py
+++ b/pytest-selenium/regions/search_sidebar.py
@@ -22,7 +22,7 @@ class SearchSidebar(Region):
     _search_result_locator = (By.CSS_SELECTOR, "[data-testid$=result]")
     _search_results_sidebar_locator = (
         By.XPATH,
-        "//div[@data-testid = 'search-results-sidebar']/ol",
+        "//div[@data-testid = 'search-results-sidebar']/div[2]",
     )
 
     # fmt: off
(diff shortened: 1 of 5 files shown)


bethshook:
@TomWoodward currently there's a region `search-results` on the parent of all results. do we want a `content-results` region for non-kt results?

TomWoodward (changes requested):
pretty sure it just goes up to the nearest region tag, so you may want to make your kt one more specific, like `kt-search-results`. you could also add the `content-search-results` if you wanted.... that is more specific, and specific is good, but it would also change the category on the event and cause a disconnect in the historical data... maybe thats a good thing since we're causing a disconnect anyway by moving the kt results out of that category

i think you can go ahead and add the new label for the content results section and definitely leave the existing one on the whole sidebar so it catches things like the sidebar close button

bethshook:
I'm realizing that the markup for the search results is screwed up such that adding a region for non-KT results is not currently possible. Should I attempt to fix this here or create a separate issue? Alternatively I could add it to [this list](https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1750).

```
<ol class="NavOl">
   <div class="RelatedKeyTerms">
      Key Term
      Key Term
   </div>
   <span>Chapter Results</span>
   <li>Result</li>
   <li>Result</li>
</ol>
```

TomWoodward (approved):

PR #1406, opened 2022-01-11
FE selenium test: move book library to json

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1796

FE selenium test: move book library to json


Malar-Natarajan:
@gregfitch please review when you get a chance

pytest-selenium/utils/utility.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from enum import Enum
+import json
```
gregfitch, line 6: switch the import from the full package to just the `loads` function (`from json import loads`), unless you plan to use more from the `json` library (then the import should be moved above `enum` as full packages should be before select function imports (`from ... import ...` )

gregfitch (changes requested):

pytest-selenium/utils/utility.py
Malar-Natarajan: @gregfitch updated

pytest-selenium/utils/utility.py
```diff
@@ -209,291 +210,10 @@ def page_highlights(cls, driver) -> int:
...
-        "us-history": {
-            "default_page": "1-introduction",
-            "search_term": "PATRIOTS",
-            "chapter_search_results_total": 20,
-            "rkt_search_results_total": 0,
-        },
-    }
+    # Read the books details from books.json file
+    book_details = open("/Users/malar/Rex/rex-web/pytest-selenium/utils/books.json", "r")
+    books = json.loads(book_details.read())
```
gregfitch, line 215: Since you modified the import, this should be `loads(` instead of `json.loads(`; I expect this to throw a `NameError` unless JSON is imported elsewhere before it gets to the utilities file.

Malar-Natarajan: yes my bad. corrected now @gregfitch

gregfitch (approved):

PR #1411, opened 2022-01-18
Faster index page prerender

Dantemss:
For: openstax/unified#1795
Speeds up prerendering of index pages by making a single reference pass instead of 1 pass per reference
This can be rebased to master and merged independently of the rest of the prerendering speedup work.


Dantemss:
Going to revert so we can merge this into master instead

PR #1418, opened 2022-01-25
Combine content version and archive version update scripts (WIP)

bethshook:
for: openstax/unified#1771


script/update-archive-version.ts
```diff
@@ -35,22 +59,26 @@ async function updateArchiveVersion() {
   );
 
   const newBookLoader = makeUnifiedBookLoader(
-    createArchiveLoader(args.newArchive, {
+    createArchiveLoader(`${REACT_APP_ARCHIVE_URL_BASE}${args.pipelineVersion}`, {
```
TomWoodward, line 62: since the other script is no longer callable, shouldn't this pipeline version default to the current version to handle the situation where you just want to update content with no pipeline change?

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -1,81 +1,50 @@
...
-
   // tslint:disable-next-line: no-console
-  console.log(`updated ${title} and added ${newRedirectionsCounter} new redirections`);
+  console.log(`updated ${title}`);
 }
 
-processBook().catch(() => {
-  process.exit(1);
-});
+export default processBookVersionUpdate;
```
TomWoodward, line 50: if this script is no longer callable, please move it to the utils directory

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -1,81 +1,50 @@
...
+export interface SimpleBook {
+  bookId: string;
+  versionNumber: string;
+}
 
 const booksPath = path.resolve(__dirname, '../src/config.books.json');
 
-const bookLoader = makeUnifiedBookLoader(
-  createArchiveLoader(REACT_APP_ARCHIVE_URL, {
+const bookLoader = (newArchiveUrl?: string) => makeUnifiedBookLoader(
```
TomWoodward, line 16: oh i see maybe this is the defaulting i was asking about in the other place

script/update-content-versions-and-check-for-archived-slugs.ts
```diff
@@ -1,81 +1,50 @@
...
+  createArchiveLoader((newArchiveUrl || REACT_APP_ARCHIVE_URL), {
     archivePrefix: ARCHIVE_URL,
   }),
   createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`)
 );
 
-async function updateRedirections(bookId: string, currentVersion: string, newVersion: string) {
```
TomWoodward, line 25: oh i see you removed this from here, never mind the other comment

script/update-archive-version.ts
```diff
@@ -35,22 +59,26 @@ async function updateArchiveVersion() {
...
-  const bookEntries = Object.entries(BOOKS);
-
-  console.log('Preparing books...');
+  for (const book of booksToUpdate) {
+    await processBookVersionUpdate(book, `${REACT_APP_ARCHIVE_URL_BASE}${args.pipelineVersion}`);
+  }
+  const bookEntries = Object.entries(BOOKS_CONFIG);
 
+  console.log('Preparing redirects...');
   for (const [bookId, { defaultVersion }] of bookEntries) {
```
TomWoodward, line 75: shouldn't it skip everything not specified in `booksToUpdate` if there is no updated pipeline?

TomWoodward (changes requested):

script/update-archive-version.ts
bethshook: this doesn't run if there's no updated pipeline (see lines 39–47)

script/update-archive-version.ts
```diff
@@ -35,22 +59,26 @@ async function updateArchiveVersion() {
...
     }),
     osWebLoader
   );
 
   const updateRedirectsPromises: Array<() => Promise<[BookWithOSWebData, number]>> = [];
-  const bookEntries = Object.entries(BOOKS);
-
-  console.log('Preparing books...');
+  for (const book of booksToUpdate) {
+    await processBookVersionUpdate(book, `${REACT_APP_ARCHIVE_URL_BASE}${args.pipelineVersion}`);
```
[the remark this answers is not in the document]
bethshook: I removed the redirect logic from the old script so it only runs here

script/update-archive-version.ts
```diff
@@ -2,27 +2,52 @@
...
+
+  if (args.pipelineVersion === REACT_APP_ARCHIVE && !booksToUpdate.length) {
+    console.log('Current and new archive url are the same. No books need version updates. Skipping...');
     return;
+  } else if ((!args.pipelineVersion || args.pipelineVersion === REACT_APP_ARCHIVE) && booksToUpdate.length) {
+    console.log('Current and new archive url are the same. Processing content version updates...');
+    for (const book of booksToUpdate) {
+      await processBookVersionUpdate(book);
+    }
+    return;
```
TomWoodward, line 48: but.. then this doesn't hit the redirect saving right?

TomWoodward (changes requested):

script/update-archive-version.ts
```diff
@@ -34,17 +53,26 @@ async function updateArchiveVersion() {
...
 
   console.log('Preparing books...');
+  for (const book of booksToUpdate) {
+    await processBookVersionUpdate(book, updatePipeline ? newArchiveUrl : undefined);
+  }
+  const bookEntries = updatePipeline
+    ? Object.entries(BOOKS_CONFIG)
+    : Object.entries(BOOKS_CONFIG).filter((entry) =>
+      booksToUpdate.map((b) => b.bookId).find((id) => id === entry[0])
+    );
```
bethshook, line 75: @TomWoodward maybe you can suggest a cleaner way to filter the config for books on the booksToUpdate list

TomWoodward (changes requested):
don't forget to update the concourse job that calls update-content-versions-and-check-for-archived-slugs to do the new script

script/update-archive-version.ts
```diff
@@ -2,29 +2,48 @@
...
+  const booksToUpdate = booksReceived.length
+    ? getBooksToUpdate(booksReceived).filter(isDefined)
+    : [];
+
+  if (!updatePipeline && !booksToUpdate.length) {
+    console.log('Current and new archive url are the same. No books need version updates. Skipping...');
     return;
   }
 
+  const newArchiveUrl = `${REACT_APP_ARCHIVE_URL_BASE}${args.pipelineVersion}`;
```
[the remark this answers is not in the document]
TomWoodward: oh i see because for the non-pipeline update you construct a different one [here](https://github.com/openstax/rex-web/pull/1418/files#diff-36084040e5187727e1bf828bd23b4a032f998045a260bbb8f630306b0a38c9d5R16)... hmm... you could probably just move that here and pass it in, its almost duplicated code

TomWoodward (changes requested):

script/update-archive-version.ts
```diff
@@ -2,26 +2,45 @@
...
-async function updateArchiveVersion() {
-  if (args.newArchive === REACT_APP_ARCHIVE) {
-    console.log('Current and new archive url are the same. Skipping...');
+const getBooksToUpdate = (books: string[]) => books.map((book) => {
+  const bookId = book.split('@')[0];
+  const versionNumber = book.split('@')[1];
+  const { defaultVersion } = BOOKS_CONFIG[bookId] || {};
+  return defaultVersion === versionNumber
+    ? undefined
+    : [bookId, {defaultVersion: versionNumber}] as [string, {defaultVersion: string}];
```
TomWoodward, line 31: this is fine, you can also do `tuple(bookId, {defaultVersion: versionNumber})` but.... oh we don't have that helper in this project... its like this

```
 export const tuple = <T extends any[]>(...args: T): T => args
```

its just a quick thing to get out of sprinkling `as [the same thing i just wrote]` all over

script/update-archive-version.ts
```diff
@@ -2,26 +2,45 @@
...
+async function updateArchiveAndContentVersions() {
+  const updatePipeline = args.pipelineVersion && args.pipelineVersion !== REACT_APP_ARCHIVE;
+  const newArchiveUrl = updatePipeline ? `${REACT_APP_ARCHIVE_URL_BASE}${args.pipelineVersion}` : REACT_APP_ARCHIVE_URL;
+  const booksReceived = args.contentVersion
+    ? (typeof args.contentVersion === 'string' ? [args.contentVersion] : args.contentVersion)
+    : [];
+  const booksToUpdate = booksReceived.length ? getBooksToUpdate(booksReceived).filter(isDefined) : [];
+
+  if (!updatePipeline && !booksToUpdate.length) {
+    console.log('Current and new archive url are the same. No books need version updates. Skipping...');
```
TomWoodward, line 43: it wouldn't suck to have this exit a non-zero code... seems like it would be a bug if something tried to run this with no args

script/update-archive-version.ts
```diff
@@ -35,16 +54,24 @@ async function updateArchiveVersion() {
...
   );
 
   const updateRedirectsPromises: Array<() => Promise<[BookWithOSWebData, number]>> = [];
-  const bookEntries = Object.entries(BOOKS);
 
   console.log('Preparing books...');
+  for (const book of booksToUpdate) {
+    await processBookVersionUpdate({
+        bookId: (book[0] as string),
+        versionNumber: (book[1] as {defaultVersion: string}).defaultVersion,
```
TomWoodward, line 69: do these need the cast since you added the typing on line 31?

script/utils/update-content-versions.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const booksPath = path.resolve(__dirname, '../../src/config.books.json');
+
+const bookLoader = (newArchiveUrl: string) => makeUnifiedBookLoader(
+  createArchiveLoader(newArchiveUrl, {
+    archivePrefix: ARCHIVE_URL,
+  }),
+  createOSWebLoader(`${ARCHIVE_URL}${REACT_APP_OS_WEB_API_URL}`)
+);
```
TomWoodward, line 21: i think you were going to move this and use the newBookLoader from the other file passed in right? maybe you just haven't gotten to that change yet

script/utils/update-content-versions.ts
```diff
@@ -0,0 +1,50 @@
...
+    console.log(`${bookId} already at desired version.`);
+    return;
+  }
+
+  const { title, version } = await bookLoader(newArchive)(bookId, newVersion)
+    .catch((error) => {
+      // tslint:disable-next-line: no-console max-line-length
+      console.log(`error while loading book ${bookId} with version ${newVersion} using pipeline ${newArchive || REACT_APP_ARCHIVE_URL}`);
+      throw error;
+    });
```
TomWoodward, line 39: actually do you even need to load this? isn't the loaded version the same as the input version? and the title is just for debug output.

we don't necessarily have to validate that the book exists, because the PR checks will get it. maybe you can actually remove this?

TomWoodward (changes requested):

TomWoodward (approved):

PR #1436, opened 2022-02-07
Escape the title and description attributes in meta tags

Dantemss:
For: https://github.com/openstax/unified/issues/1778


bethshook (approved):

Dantemss:
This change requires pre-rendering to be visible so we'll have to merge it before UX/QA review so it goes through the pre-rendering process for some future release.

PR #1444, opened 2022-02-10
catch errors that happen outside of the normal error boundary

jivey:
For: https://github.com/openstax/unified/issues/1785


src/app/messages/SimpleMessageProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+import createIntl from './createIntl';
+
+// tslint:disable-next-line:variable-name
+const SimpleMessageProvider = (props: { children?: React.ReactNode }) => {
+  const [intl, setIntl] = useState<IntlShape | null>(null);
+
+  useEffect(() => {
+    const setUpIntl = async() => {
+      const languageWithCode = window!.navigator.language || 'en-US';
+      const language = new (Intl as any).Locale(languageWithCode).language;
```
jivey, line 12: The type for this is available in Typescript 4.5, so I figured it was desirable to wait to upgrade rather than pull in a package like `@formatjs/intl-locale`. But I can still pull that in, or regex it out instead.

TomWoodward: if its on `window.Intl` you may just need to change `src/typings/global.d.ts` to expose it

we have some awkward typing going on around the browser globals to avoid them being defined when they shouldn't in ssr

TomWoodward (changes requested):

src/app/messages/en/modalErrorText.ts
```diff
@@ -1,4 +1,4 @@
 export const bodyWithLink = `
-  We're not quite sure what went wrong. Refresh the page. If this doesn't solve
+  We're not quite sure what went wrong. Restart your browser. If this doesn't solve
```
bethshook, line 2: For Spanish, replace "Actualice la página." with "Reinicia el navegador."

jivey: Thanks!

src/app/errors/components/ErrorBoundary.spec.tsx
```diff
@@ -41,4 +54,17 @@ describe('ErrorBoundary', () => {
     );
     expect(Sentry.captureException).toHaveBeenCalled();
   });
+
+  it('captures unhandled rejected promises', () => {
+    const tree = renderer.create(<TestContainer>
```
bethshook, line 59: maybe this doesn't matter here but <ErrorBoundary> would not normally be wrapped in this container since TestContainer includes Services and MessageProvider

jivey: I think the ErrorBoundary needs the MessageProvider, I'm not sure if that's always been the case, but it complains if I remove it

bethshook: Yeah, it's probably fine for the purposes of this test - it just technically should be using your new SimpleMessageProvider since ErrorBoundary falls outside of the normal MessageProvider (which gets its language from whatever book is loaded)

src/app/errors/components/ErrorBoundary.tsx
```diff
@@ -56,6 +60,18 @@ class ErrorBoundary extends React.Component<Props, State> {
     this.props.recordError(error);
   }
 
+  public componentDidMount() {
+    if (windowDefined()) {
+      window?.addEventListener('unhandledrejection', this.handleRejection);
```
TomWoodward, line 65: if you just do the `typeof` check here in the if statement, you won't need the `?`, i don't love the creation of the extra code path

TomWoodward (changes requested):

src/helpers/PromiseCollector.spec.ts
```diff
@@ -58,4 +58,14 @@ describe('PromiseCollector', () => {
       done();
     });
   });
+
+  it('swallows errors when removing promises with .finally', async() => {
+    // The "fail" case of this test is with a process exit code
```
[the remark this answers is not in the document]
jivey: I had wondered the same, and found out something interesting: jest creates its own process so it can attach handlers, which makes it difficult/awkward to work around. I found some interesting looking suggestions on the jest issue, like https://github.com/facebook/jest/issues/5620#issuecomment-761879230 and https://github.com/facebook/jest/issues/5620#issuecomment-998716759 but wasn't sure we'd want to complicate the test suite like that. I could explore it more though and try to find something that feels like a good fit.

TomWoodward (approved):
this looks good, but needs conflicts resolved

Malar-Natarajan:
@jivey looks like this is having merge conflicts

PR #1485, opened 2022-02-14
FE selenium test: verify attribution text and copyright name

Malar-Natarajan:
https://openstax.testrail.com/index.php?/cases/view/623701

https://openstax.testrail.com/index.php?/cases/view/623700

https://openstax.testrail.com/index.php?/cases/view/623699

Add new books tot the book.json library

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/rex-web/1485

pull requests:
- [ ] openstax/rex-web#1485

### DIFF
--- a/pytest-selenium/tests/test_attribution.py
+++ b/pytest-selenium/tests/test_attribution.py
@@ -181,13 +181,11 @@ def test_license_details(selenium, base_url, page_slug):
         license_name = attribution.citation_builder.get_attribute("text").strip()
         license_url = attribution.citation_builder.get_attribute("href").strip()
 
-        if license_name == license_name_expected:
-            pass
-        elif license_name == "CC BY":
+        if license_name == "CC BY":
             print(
                 f"{book_slug} has license name as {license_name}. But expected name is {license_name_expected}"
             )
-        else:
+        elif license_name != license_name_expected:
             pytest.xfail(
                 f"{book_slug} has license name as {license_name}. But expected name is {license_name_expected}"
             )
--- a/pytest-selenium/tests/test_attribution.py
+++ b/pytest-selenium/tests/test_attribution.py
@@ -1,6 +1,6 @@
 from pages.content import Content
 from tests import markers
-from utils import utility
+from utils.utility import Library, get_default_page
 
 
 @markers.test_case("C476302")
@@ -105,7 +105,7 @@ def test_book_url_in_citation_text_shows_url_for_default_page(
     attribution.click_attribution_link()
 
     # THEN: The book url in the the citation section should reference the default page of the book
-    default_page_slug = utility.get_default_page(book_slug)
+    default_page_slug = get_default_page(book_slug)
     attribution_book_url_expected = (
         "https://openstax.org/books/" + book_slug + "/pages/" + default_page_slug
     )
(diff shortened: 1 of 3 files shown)


pytest-selenium/tests/test_attribution.py
```diff
@@ -128,9 +128,138 @@ def test_access_free_url_in_citation_text_shows_url_for_default_page(
...
+        attribution.click_attribution_link()
+
+        # THEN: License name & license link address is Creative Commons
+        license_name = attribution.citation_builder.get_attribute("text").strip()
+        license_url = attribution.citation_builder.get_attribute("href").strip()
+
+        try:
+            assert (
+                license_name == license_name_expected
+            ), f"{book_slug} has license name as {license_name}. But expected name is {license_name_expected}"
```
gregfitch, line 186: You're not likely to see this message because the `try` will immediately catch the error. Instead, remove the `try` block and move both boolean statements into a single `assert`.

gregfitch (changes requested):

pytest-selenium/tests/test_attribution.py
Malar-Natarajan: I want the statement to be printed only if `license_name == "cc by"`. The test can still pass but I just need the list of books with that name. So is this ok?

gregfitch: In that case just remove the failure string from the first assertion as you won't see it - it won't print if it passes and it won't print if it fails because the assertion error will be caught and handled.

pytest-selenium/tests/test_attribution.py
```diff
@@ -128,9 +129,142 @@ def test_access_free_url_in_citation_text_shows_url_for_default_page(
...
+        if license_name == license_name_expected:
+            pass
+        elif license_name == "CC BY":
+            print(
+                f"{book_slug} has license name as {license_name}. But expected name is {license_name_expected}"
+            )
+        else:
+            pytest.xfail(
+                f"{book_slug} has license name as {license_name}. But expected name is {license_name_expected}"
+            )
```
gregfitch, line 193: Consider reworking by removing the first part of the `if`, changing the `elif` to be:
`if license_name == "CC BY":`
and removing the `else` clause because you're testing this same case in the next `assert`

Malar-Natarajan: By `because you're testing this same case in the next assert` do you mean lines 195 - 197? It is a different parameter (license_url) that it checks. It fails straight if the assertion fails.

I am not quite sure I understand the removing else clause. 
I want to make sure the license name is equal to the value stored in `license_name_expected` . Thats the ideal scenario.

But some books are hanging on with `cc by`. That is still ok to pass but I need a list to make sure no new books are added to this group with `cc by` till CE team fixes this issue.

The elif condition, checks if any unknown value/placeholder other than the above two appears (negative testing).

Please let me know if I am still not getting what you are trying to explain.

gregfitch: Ah...I misread the assert vars.

I would still remove the first `if` clause (as it doesn't do anything), change the `elif` to an `if` and change the final `else` statement to an `elif` so you end up with:
`if license_name == "CC BY":`
and
`elif license_name != license_name_expected:`

gregfitch (approved):

PR #1493, opened 2022-02-16
Exclude modal query param from PQ LinkToSection

bethshook:
for: openstax/unified#1713

### DIFF
--- a/src/app/content/components/ContentLink.tsx
+++ b/src/app/content/components/ContentLink.tsx
@@ -38,6 +38,7 @@ interface Props {
   myForwardedRef: React.Ref<HTMLAnchorElement>;
   systemQueryParams?: SystemQueryParams;
   persistentQueryParams?: ContentQueryParams;
+  ignoreModal?: boolean;
 }
 
 // tslint:disable-next-line:variable-name
@@ -56,6 +57,7 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
     hasUnsavedHighlight,
     systemQueryParams,
     persistentQueryParams,
+    ignoreModal,
     ...anchorProps
   } = props;
   const {url, params} = getBookPageUrlAndParams(book, page);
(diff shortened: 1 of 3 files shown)


src/app/content/components/ContentLink.tsx
```diff
@@ -64,7 +64,7 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
   const bookUid = stripIdVersion(book.id);
   // Add options only if linking to the same book
   const options = currentBook && currentBook.id === bookUid
-    ? createNavigationOptions({...persistentQueryParams, ...systemQueryParams}, scrollTarget)
+    ? createNavigationOptions({...persistentQueryParams, modal: null, ...systemQueryParams}, scrollTarget)
```
TomWoodward, line 67: this is going to completely disable linking to content pages with a modal open... which... may not be really what we want... i think it might be better to set the `{modal: undefined}` (or whatever) in the implementation [here](https://github.com/openstax/rex-web/blob/master/src/app/content/practiceQuestions/components/LinkToSection.tsx#L45)

the only other thing would be to re-evaluate the definition of `modal` as a member of `persistentQueryParams` but... that seems like a can of worms... it seems to me that if you're rendering links to places (like to one of the other steps of practice questions) you're gonna want to keep the modal, and if you want to eject from the modal as we do in this case, you're gonna know that and you can specify it.

the really tricky thing i'm trying to wrap my head around is if there are ever any differences between the way we'd want it to work when doing simple click vs open in new tab.... i... think.... that its safe to assume they would always act the same

TomWoodward (changes requested):

src/app/content/components/ContentLink.tsx
bethshook: I've added an `ignoreModal` prop. In this case the component is hardcoded to open a new tab so I don't think we have to worry about how it should work with a simple click unless we start using this prop elsewhere.

TomWoodward (approved):

PR #1500, opened 2022-02-22
fix expanding details element when scrolling to search result or highlight

jivey:
For: https://github.com/openstax/unified/issues/1847


bethshook (approved):

jivey:
Added another bug fix: when scrolling to a highlight inside a solution (like from My Highlights) the solution wasn't expanding.

PR #1505, opened 2022-02-24
Revert Issue 1304 changes

bethshook:
for: openstax/unified#1304


TomWoodward:
How did you generate this revert? The closed pr has a revert button or you can git revert the commit on master, but I'm worried that I see two commits here

TomWoodward (approved):

bethshook:
I was in the files to see if I could fix the bug so reverted each file using VS Code. Didn't realize there was a button, but I did check it over.

TomWoodward:
when i got to a computer i was able to flip back and forth and see it was a complete revert, but if the commit had been made with `git revert` instead of file by file its easier to have more confidence in its completeness. this is one of the reasons we squash our prs, so there is only one commit to deal with.

PR #1506, opened 2022-02-25
Discard note modal bugs

bethshook:
for: openstax/unified#1304


src/app/content/highlights/reducer.ts
```diff
@@ -194,7 +194,14 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
       };
     }
     case getType(actions.focusHighlight): {
-      return {...state, currentPage: { ...state.currentPage, hasUnsavedHighlight: false, focused: action.payload }};
```
bethshook, line 197: @TomWoodward removing `hasUnsavedHighlight: false` from here fixes the bug Malar found when testing the discard changes modal. That property was added when the discard modal [was first implemented](https://github.com/openstax/rex-web/pull/485/files). Do you know why this would have been desirable? No tests are failing but I feel like this must break something.

src/app/content/highlights/reducer.ts
```diff
@@ -194,7 +194,16 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
+      const isRefocusing = state.currentPage.focused === action.payload;
+      return {
+        ...state,
+        currentPage: {
+          ...state.currentPage,
+          focused: action.payload,
+          hasUnsavedHighlight: isRefocusing,
```
TomWoodward, line 203: i think what you want to do is pull forward the previous value unless the id is different, in which cases set it to false. in this change you're assuming that a refocus happens when there are pending changes, which may be true now but seems like a weird assumption.

TomWoodward (changes requested):

src/app/content/highlights/reducer.ts
```diff
@@ -194,7 +194,16 @@ const reducer: Reducer<State, AnyAction> = (state = initialState, action) => {
...
     }
     case getType(actions.focusHighlight): {
-      return {...state, currentPage: { ...state.currentPage, hasUnsavedHighlight: false, focused: action.payload }};
+      const hasNewFocus = state.currentPage.focused !== action.payload;
+      return {
+        ...state,
+        currentPage: {
+          ...state.currentPage,
+          focused: action.payload,
+          hasUnsavedHighlight: hasNewFocus ? false : state.currentPage.hasUnsavedHighlight,
```
bethshook, line 203: @TomWoodward good catch on my assumption here. I also don't think it makes sense to dispatch clearFocusedHighlight before this action - it removes the `focused` property entirely from `state.currentPage` which I don't think we want when shifting focus from one highlight to another.

TomWoodward (approved):

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -98,6 +98,9 @@ const EditCard = React.forwardRef<HTMLElement, EditCardProps>((props, ref) => {
     if (element.current) {
       props.onHeightChange(element);
     }
+    if (!props.isActive && !props.data?.annotation) {
+      setPendingAnnotation(defaultAnnotation());
+    }
```
bethshook, line 103: @TomWoodward new change here - Malar caught a bug where unsaved note text was being stored in the EditCard's local state even after focus leaves the card. I think it predates all of the recent work.

Issue #1322, opened 2021-08-25
Service worker upgrade bug in Safari


bethshook:

Malar-Natarajan:
Regression testing only

PR #1517, opened 2022-03-07
Potentially fix service worker update error in Safari

Dantemss:
For: https://github.com/openstax/rex-web/issues/1322

This one may be hard to test... Maybe we need to create a release with the old code, load it on Safari, update it to something else and see if we get the error?

Here's my reasoning for this change:

The error seems to be coming from step 3 of `sw.update()` described in https://www.w3.org/TR/service-workers/#service-worker-registration-update

This means step 2 must have returned null, which means the registration has neither an active, waiting or installing service worker, but at the same time it did not previously error.

I believe this could be either a new registration that is not ready yet, or an old registration that has been overwritten.
If it was an old registration, this means we registered the same scope url but did not update `app.services.serviceWorker`, since this is the serviceWorker that is passed to the update check methods. The only place that calls `serviceWorker.register()` seems to set `app.services.serviceWorker`, so I think it's more likely to be a new registration. There's also a comment near the `register()` call saying that workers are null for some time after `register()` is called.

In either case, instead of storing the registration in `app.services.serviceWorker`, there is a `ready` promise in the ServiceWorkerContainer that we can use instead. This promise will resolve when the service worker is active and return the service worker registration, so `update()` should not error. The promise is described in https://www.w3.org/TR/service-workers/#serviceworkercontainer-ready-promise


src/app/notifications/components/UpdatesAvailable.tsx
```diff
@@ -16,16 +16,17 @@ import { Body, Group, Header, P } from './Card';
...
-  const [readyToReload, setReadyToReload] = useState<boolean>(!serviceWorkerNeedsUpdate(sw));
+  const serviceWorkerContainer = assertWindow().navigator.serviceWorker;
+  const [sw, setSw] = useState<ServiceWorkerRegistration | undefined>(undefined);
 
-  useEffect(() => {
-    findAndInstallServiceWorkerUpdate(sw, () => setReadyToReload(true));
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  if (serviceWorkerContainer) {
+    serviceWorkerContainer.ready.then((registration) => {
```
TomWoodward, line 23: you shouldn't be binding a promise handler outside of useEffect, or it'll be re-bound every time the component renders

TomWoodward (changes requested):
i have an alternate theory, that the updateAvailable in pollUpdates is being triggered before the promise resolution in index.tsx, causing services.serviceWorker to be undefined when the UpdatesAvailable component is rendered. when the promise completes, it doesn't update react, and the `undefined` is still passed into activateSwAndReload, causing it to just reload without calling skipUpdates

in any case, i think your fix should still work. i'm trying to think if there is any benefit to saving the reference to the registration returned in the index.tsx promise, but it seems overall better to use your approach that is more stateless.

src/app/notifications/components/UpdatesAvailable.tsx
Dantemss: Added useEffect()

TomWoodward (approved):

PR #1518, opened 2022-03-07
Limit when page views are registered

bethshook:
for: openstax/unified#1846


bethshook:
@TomWoodward last week we discussed ignoring same-page hash changes when registering a page view. a problem I ran into with that is that sometimes when navigating through same-page search results, results will have the same hash but different indices. I opted instead to only limit page view registration to when there is a change to the pathname or to the `modal` param. do you foresee issues with this? would it be better to filter out changes to hash or search index?

src/app/navigation/types.ts
```diff
@@ -58,12 +58,13 @@ export interface LocationChange<M = AnyMatch> {
...
   match?: M;
   action: Action;
+  prevLocation?: State;
 }
 
 export type AnyRoute = typeof routes[number];
 export type AnyMatch = UnionRouteMatches<AnyRoute>;
 
-export type RouteHookBody<R extends AnyRoute> = (helpers: MiddlewareAPI & AppServices) =>
+export type RouteHookBody<R extends AnyRoute> = (helpers: MiddlewareAPI & AppServices, prevLocation?: string) =>
```
TomWoodward, line 67: i think this is unused and wrong?

TomWoodward (changes requested):

src/app/content/hooks/registerPageView.ts
```diff
@@ -0,0 +1,22 @@
+import googleAnalyticsClient from '../../../gateways/googleAnalyticsClient';
+import * as selectNavigation from '../../navigation/selectors';
+import { RouteHookBody } from '../../navigation/types';
+import { AppServices, MiddlewareAPI } from '../../types';
+import { content } from '../routes';
+
+export const hookBody: RouteHookBody<typeof content> = (services: MiddlewareAPI & AppServices) => async(action) => {
```
TomWoodward, line 7: i'm not sure about having prevLocation in the state, did you have a reason for doing that in redux? i was trying to suggest doing it here, by creating a cache layer like this:

```
export const hookBody: RouteHookBody<typeof content> = (services: MiddlewareAPI & AppServices) => {
  let lastTrackedLocation;

  return async(action) => {
      const state = services.getState();
      const pathname = selectNavigation.pathname(state);
      const query = selectNavigation.query(state);
      const prevPath = lastTrackedLocation?.pathname;
      const prevQuery = lastTrackedLocation?.query;

      const pathOrModalChanged =
      pathname !== prevPath || prevQuery?.modal !== query.modal;

      if (action.action !== 'REPLACE' && pathOrModalChanged) {
        lastTrackedLocation = {query, pathname};
        googleAnalyticsClient.trackPageView(pathname, query);
     }
  }
}
```

TomWoodward (dismissed):

src/app/content/hooks/registerPageView.ts
bethshook: I tried something like this initially, but this hook gets called on every locationChange so you can't cache the previous location here. Instead I cache it in routeHook and then on each locationChange it's added to the payload. I wasn't sure how else to make it accessible from within the registerPageView hook.

src/app/content/hooks/registerPageView.ts
```diff
@@ -0,0 +1,27 @@
+import googleAnalyticsClient from '../../../gateways/googleAnalyticsClient';
+import * as selectNavigation from '../../navigation/selectors';
+import { RouteHookBody } from '../../navigation/types';
+import { AppServices, MiddlewareAPI } from '../../types';
+import { content } from '../routes';
+
+export const hookBody: RouteHookBody<typeof content> = (services: MiddlewareAPI & AppServices) => {
+  let lastTrackedLocation: any;
```
bethshook, line 8: I implemented the changes you proposed and everything looks good, except that I'm not sure how to mock this for testing. Currently `pathOrModalChanged` always evaluates to true in testing because lastTrackedLocation always starts out undefined.

TomWoodward: if you mean for a unit test for this hook specifically you can do something like 

```
const theHook = registerPageViewHook(mockServices)

theHook(mockLocationAction)
expect(pageviewtobetracked)
theHook(mockLocationActionThatShouldntBeTracked)
expect(pageviewnottracked)
theHook(mockLocationActionThatShouldBeTracked)
pageviewtobetracked
```

does that makes sense?

bethshook: Hmm ok I had already tried that but I think I left out the first `expect`. What's weird is when I call the hook with any action I get a tracked page view, even when the action fails the `action.action !== 'REPLACE' && pathOrModalChanged` condition.

bethshook: I pushed the spec so you can see.

bethshook: @jivey maybe you will have thoughts about this failing spec

jivey: @bethshook I think this is due to some odd behavior on Jest's part. The GA client calls `.ga` twice here: https://github.com/openstax/rex-web/blob/1d43b5861dc02e6091a97706bba393b3b89262c6/src/gateways/googleAnalyticsClient.ts#L196 for each `trackPageView` call, and `toHaveBeenCalledWith/toHaveBeenCalledTimes` just looks at the `.mock.calls` array to decide if multiple calls were made, even though one call results in this:

```
[
  [ 'tfoo.set', 'queueTime', 0 ],
  [ 'tfoo.send', { hitType: 'pageview', page: '/' } ]
]
```

This was a little surprising to me, I always thought Jest tracked call results more granularly. Anyway, I think the fix is a mockReset between, this passes for me:

```
await hook(action);
expect(mockGa).toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/'});
mockGa.mockReset();
await hook(action);
expect(mockGa).not.toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/'});
```

bethshook: Thanks @jivey - there is always some jest idiosyncrasy that hasn't occurred to me

src/app/content/hooks/registerPageViews.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    it('does not register a page view if pathname and modal query are unchanged', async() => {
+      await hook(action);
+      expect(mockGa).toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/'});
+      mockGa.mockReset();
+      await hook(action);
+      expect(mockGa).not.toHaveBeenCalledWith('tfoo.send', {hitType: 'pageview', page: '/'});
+    });
+  });
```
jivey, line 79: Strange, I thought tslint would complain about > 2 spaces in the describe block but maybe not because the statements are aligned?

jivey (approved):

PR #1555, opened 2022-04-06
allow deselecting all study guide filters

bethshook:
for: openstax/unified#1839


bethshook:
Because the state is now updated based on the query, I made it possible for the filter query parameters to have no value. This way in the reducer we can check whether a parameter is undefined (because the modal has been opened for the first time) or empty (because all filters have been deselected).

src/app/content/studyGuides/components/Filters.spec.tsx
```diff
@@ -243,7 +243,7 @@ describe('Filters', () => {
 
     expect(dispatch).toHaveBeenCalledWith(
       replace(mockMatch, {
-        search: `colors=yellow&colors=green&colors=blue&colors=purple&modal=${modalUrlName}`,
+        search: `colors=yellow&colors=green&colors=blue&colors=purple&locationIds&modal=${modalUrlName}`,
```
Dantemss, line 246: Is it ok to have a query value without an `=` here? It's not common to see this. I guess it's a valid thing though.

Dantemss (commented):
Since you asked for Tom's review I'll let him approve instead :)

TomWoodward (approved):

PR #1561, opened 2022-04-12
Introduction to Sociology 3e updates

staxbot:


theresamasciale:
This doesn't look right. It looks like the PR is reversing to the old commit sha instead of the new one. The sha should be: `433bfcd11e4294a704a871c1ffdb29d0ed6c8d41` I checked the ABL and it looks correct. 

This was added to the ABL because we had a high prio image change in chapter 8, figure 8.2 should look like this (without a part "C"): 
Screen Shot 2022-04-12 at 8 54 35 AM

theresamasciale (changes requested):
This doesn't look right. It looks like the PR is reversing to the old commit sha instead of the new one. The sha should be: `433bfcd11e4294a704a871c1ffdb29d0ed6c8d41` I checked the ABL and it looks correct.

This was added to the ABL because we had a high prio image change in chapter 8, figure 8.2 should look like this (without a part "C"):

Screen Shot 2022-04-12 at 8 54 35 AM

Malar-Natarajan:
looks like released[ last sprint](https://openstax.slack.com/archives/CH5514N57/p1650980570301749?thread_ts=1650980327.383439&cid=CH5514N57)

PR #1564, opened 2022-04-13
Update canonicalUrl util to check for deeper canonical maps and fix Chemistry AF mapping

bethshook:
for: openstax/unified#1885


TomWoodward:
@Dantemss can you take over reviewing this?

src/app/content/utils/canonicalUrl.ts
```diff
@@ -14,29 +15,50 @@ export async function getCanonicalUrlParams(
...
-    ).defaultVersion;
-    const canonicalBook = book.id === id  && hasOSWebData(book) ? book : await getBook(id, version);
-    const mappedPageId = CANONICAL_PAGES_MAP[pageId] || pageId;
-    const treeSection = findArchiveTreeNodeById(canonicalBook.tree, mappedPageId);
+  const getCanonicalMap = (bookId: string) => {
+    const bookDefaultMap = [[bookId, {}]] as Array<[string, ObjectLiteral<undefined>]>;
+    const bookVersionFromConfig = getBookVersionFromUUIDSync(bookId);
+    return ([
+      ...(CANONICAL_MAP[bookId] || []),
+      ...(bookVersionFromConfig && bookVersion === bookVersionFromConfig.defaultVersion ? bookDefaultMap : []),
```
Dantemss, line 25: If a non-default bookVersion of a book that is not in the CANONICAL_MAP  is used here, getCanonicalMap will return []. The same can happen if none of the books returned are in BOOKS. This can be a problem for the loop below.

src/app/content/utils/canonicalUrl.ts
```diff
@@ -14,29 +15,50 @@ export async function getCanonicalUrlParams(
...
-      const pageInBook = assertDefined(treeSection.slug, 'Expected page to have slug.');
-      return {book: {slug: canonicalBook.slug}, page: {slug: pageInBook}};
+  let canonicalMap = getCanonicalMap(book.id);
+  let canonicalPageId = pageId;
+  let done = false;
+  let canonicalBook;
+  let treeSection;
+
+  while (!done) {
+    for (const [id, CANONICAL_PAGES_MAP] of canonicalMap) {
```
Dantemss, line 37: If canonicalMap is [], the loop will hang since it becomes `while (True) {}`

src/app/content/utils/canonicalUrl.ts
```diff
@@ -14,29 +15,50 @@ export async function getCanonicalUrlParams(
...
+      const version = assertDefined(
+        getBookVersionFromUUIDSync(id),
+        `We've already filtered out books that are not in the BOOK configuration`
+      ).defaultVersion;
+      const useCurrentBookAsCanonical = book.id === id  && hasOSWebData(book);
```
Dantemss, line 42: Might have to check that the book.version also matches the default version?

src/app/content/utils/canonicalUrl.ts
```diff
@@ -14,29 +15,50 @@ export async function getCanonicalUrlParams(
...
+      canonicalBook = useCurrentBookAsCanonical ? book : await getBook(id, version);
+      canonicalPageId = CANONICAL_PAGES_MAP[canonicalPageId] || canonicalPageId;
+      treeSection = findArchiveTreeNodeById(canonicalBook.tree, canonicalPageId);
+
+      if (!useCurrentBookAsCanonical) {
+        const newMap = getCanonicalMap(canonicalBook.id);
+        // stop when we run out of canonical maps to check
+        done = !newMap.length || isEqual(canonicalMap, newMap);
+        canonicalMap = newMap;
```
Dantemss, line 51: It's possible there would be a cycle in the CANONICAL_MAP and this would run forever, alternating between them. Not sure how important it is to prevent this.

src/canonicalBookMap.ts
```diff
@@ -45,6 +45,11 @@ export const CANONICAL_MAP: CanonicalBookMap = {
   /* Astronomy */ '2e737be8-ea65-48c3-aa0a-9f35b4c6a966': [
     /* Astronomy 2e */ ['4c29f9e5-a53d-42c0-bdb5-091990527d79', {}],
   ],
+  /* Biology for AP Courses */ '6c322e32-9fb0-4c4d-a1d7-20c95c5c7af2': [
+    /* Biology 2e */ ['8d50a0af-948b-4204-a71d-4826cba765b8', {
+      // needs page mapping
```
Dantemss, line 50: Need to fill this in?

src/canonicalBookMap.ts
```diff
@@ -291,7 +296,69 @@ export const CANONICAL_MAP: CanonicalBookMap = {
...
+      '37feaf08-2a93-4dbe-84c4-140d213f9b5c': '353391e0-9913-4452-9f83-38f57233275a',
+      /* 17.0 Introduction to the same module in 2e */
+      '9d111d7d-7cb8-4688-9cd2-46774c477da9': 'b451b1e5-82ab-48c4-8275-37628dd79df7',
+      /* 17.7 Catalysis to the same module in 2e */
+      '40651eeb-28e7-4577-9bf7-d42282ec6ae5': 'a6400a57-a040-4768-8222-61705d4fff83',
+    }],
+  ],
+  /* College Physics for AP Courses */ '8d04a686-d5e8-4798-a27d-c608e4d0e187': [
+    /* College Physics */['031da8d3-b525-429c-80cf-6c8ed997733a', {
+      // needs page mapping
```
Dantemss, line 360: Same

Dantemss (changes requested):

src/app/content/utils/canonicalUrl.ts
```diff
@@ -14,29 +15,50 @@ export async function getCanonicalUrlParams(
...
+
+  while (!done) {
+    for (const [id, CANONICAL_PAGES_MAP] of canonicalMap) {
+      const version = assertDefined(
+        getBookVersionFromUUIDSync(id),
+        `We've already filtered out books that are not in the BOOK configuration`
+      ).defaultVersion;
+      const useCurrentBookAsCanonical = book.id === id  && hasOSWebData(book);
+      canonicalBook = useCurrentBookAsCanonical ? book : await getBook(id, version);
+      canonicalPageId = CANONICAL_PAGES_MAP[canonicalPageId] || canonicalPageId;
```
Dantemss, line 44: If a page is missing in the map, the previous book's page id is used, which means this is only valid if the book's canonical book is itself or if the 2 books are using the same page with the same id (is this even possible anymore?), otherwise we'll end up returning a null here. Not necessarily a problem, just need to ensure the canonical maps are complete.

src/canonicalBookMap.ts
bethshook: just placeholders - I'll remove them

src/app/content/utils/canonicalUrl.ts
bethshook: Added array length as a condition

src/app/content/utils/canonicalUrl.ts
bethshook: Hmm thinking through this... If we add this check and they don't match, then on line 43 we'll get the latest version of the book and check for the current page in that book's tree. If the page isn't in the tree then no canonical URL is used. I think that's probably what we want? What do you think?

Dantemss: I think with the current code if we got into a situation where the book is in its own `CANONICAL_MAP[bookId]` (will this happen?) then this check would be true for any version of the book so they would all have themselves as canonical links. This doesn't happen with the bookDefaultMap simply because the bookDefaultMap is not added if the version doesn't match.

I think the desired behavior in the case you described would be to find the latest version of the book that has the removed page, which would be a pain...

Dantemss: I think it would be a good idea to add a few more tests with the expected behavior for some of the different cases so we can make sure they all work:

- Book has no CANONICAL_MAP and is defaultVersion
- Book has no CANONICAL_MAP and is not defaultVersion. Page is in the defaultVersion
- Book has no CANONICAL_MAP and is not defaultVersion. Page is missing from the defaultVersion
- Book has itself as CANONICAL_MAP and is defaultVersion
- Book has itself as CANONICAL_MAP and is not defaultVersion. Page is in the map.
- Book has itself as CANONICAL_MAP and is not defaultVersion. Page is missing from the map.
- Book has another book as CANONICAL_MAP and is defaultVersion
- Book has another book as CANONICAL_MAP and is not defaultVersion. Page is in the map.
- Book has another book as CANONICAL_MAP and is not defaultVersion. Page is missing from the map.

src/app/content/utils/canonicalUrl.ts
bethshook: Per today's conversation it sounds like that would be the expected behavior.

src/app/content/utils/canonicalUrl.ts
bethshook: I've updated the specs. Did not need to add much since we decided we're not concerned with loaded page version.

src/app/content/utils/canonicalUrl.ts
bethshook: Now this throws if it's about to run through a map that's already been checked.

Dantemss (approved):
Approved but see question

PR #1565, opened 2022-04-13
add mathjax 3 as experimental toggle via query param

jivey:
For: https://github.com/openstax/unified/issues/1807


bethshook (approved):
Tested on Fizyka vol. 3 and looks good

PR #1570, opened 2022-04-20
fix old safari

TomWoodward:
For: https://github.com/openstax/unified/issues/1890

Safari versions < 14 do not support Intl.Locale so replace it with a simple substring to get the two letter language code if it's inside a tag like 'es-MX' or 'en-US'.

MediaQueryList on older Safari needs to use `addListener/removeListener` methods instead of `addEventListener/removeEventListener`, and don't take a type param.


jivey:
@TomWoodward Tested on Safari 13.1 and confirmed both issues are resolved.

bethshook (approved):

PR #1571, opened 2022-04-20
Fix Intl

jivey:
For: https://github.com/openstax/unified/issues/1890
Safari versions < 14 do not support `Intl.Locale` so replace it with a simple substring to get the two letter language code if it's inside a tag like 'es-MX' or 'en-US'.


jivey:
Closing in favor of #1570

PR #1576, opened 2022-04-25
add logic to get latest pipeline version

bethshook:
for: openstax/unified#1866


jivey (approved):

jivey:
Looks good, just needs a snapshot update I think

src/gateways/createBookConfigLoader.ts
```diff
@@ -1,40 +1,52 @@
...
+      .then((response) => response && {books: response.books, archiveUrl: response.archiveUrl})
       .catch((e) => {
         Sentry.captureException(e, Severity.Warning);
         return Promise.resolve(undefined);
       });
   };
 
   return {
+    getArchiveUrl: (): Promise<string | undefined> => {
+      return cachedArchiveUrl ? Promise.resolve(cachedArchiveUrl) : loadRemoteBookConfig().then((config) => {
```
bethshook, line 33: @TomWoodward `cachedArchiveUrl` has to be initialized to `REACT_APP_ARCHIVE_URL` in order for `getArchiveUrlSync()` to work off the bat, but then the remote config is never loaded here. I think instead this should receive a book UUID and version and check if it loads using the cached pipeline and load the remote config if not. Does that make sense?

TomWoodward: i think... that maybe thats fine... because if the original `cachedBooks` is never updated, then the original `cachedArchiveUrl` will also continue to work and not need to be updated. the problem is that if the `cachedBooks ` is updated, the archive url needs to be kept in sync or things won't work. i think the fix is to add on line 43 when the cache is updated you always have to update both things

TomWoodward (changes requested):

src/index.tsx
```diff
@@ -37,7 +37,7 @@ if (window.top === window.self) {
 }
 
 const archiveUrl = config.REACT_APP_ARCHIVE_URL_OVERRIDE ||
-  assertDefined(config.REACT_APP_ARCHIVE_URL, 'REACT_APP_ARCHIVE_URL must be defined');
+  assertDefined(getArchiveUrlSync(), 'archive url must be defined');
```
TomWoodward, line 40: when this changes later it won't be used, something else needs to be changed to make this work

TomWoodward (changes requested):

src/gateways/createArchiveLoader.ts
```diff
@@ -44,12 +44,13 @@ const defaultOptions = () => ({
...
 });
 
-export default (archivePath: string, options: Options = {}) => {
+export default (getArchivePath: () => string, options: Options = {}) => {
   const {pageCache, bookCache, appPrefix, archivePrefix, disablePerBookPinning} = {
     ...defaultOptions(),
     ...options,
   };
 
+  const archivePath = getArchivePath();
```
TomWoodward, line 53: this part of the function is run when the gateway is created, so you're still not picking up changes in the cached value over time

TomWoodward (changes requested):

src/gateways/createArchiveLoader.ts
```diff
@@ -44,53 +44,55 @@ const defaultOptions = () => ({
...
-export default (archivePath: string, options: Options = {}) => {
+export default (getArchivePath: () => string, options: Options = {}) => {
   const {pageCache, bookCache, appPrefix, archivePrefix, disablePerBookPinning} = {
     ...defaultOptions(),
     ...options,
   };
 
-  const contentUrlBase = (host: string, bookId: string) => disablePerBookPinning
+  const contentUrlBase = (host: string, archivePath: string, bookId: string) => disablePerBookPinning
     ? `${host}${archivePath}`
```
TomWoodward, line 54: you don't have to pass it all around, you can just call it here, so that it gets called at the time of usage, that way it'll get the most recent value

TomWoodward (changes requested):

src/gateways/createArchiveLoader.ts
bethshook: I've updated this

TomWoodward (changes requested):

TomWoodward (approved):

PR #1577, opened 2022-04-26
don't rerender edit card once highlight removed

bethshook:
for: openstax/unified#1868


bethshook:
The showCreate event bug was due to the fact that the event is logged via `useEffect` every time EditCard is rendered without a highlight. We want it to render when a new highlight is being created but not when one is being deleted.

jivey (approved):

PR #1585, opened 2022-05-05
Added optimized highlights CTA images of different sizes

Dantemss:
For: https://github.com/openstax/unified/issues/1845

The images were resized using imagemagick's default settings (which seemed to produce the best balance of keeping the text readable and avoiding jagged edges).
Then the resized images and the original images were fed into ImageOptim to minimize their disk size (lossless compression).
The CTA img tag asks for 360x256 images but retina displays actually display 720x512 images instead, so a srcset attribute was added.

If desired, we could also save a further ~50% image size by passing them through ImageOptim in lossy mode with PNG quality set to 90% (the maximum).


bethshook (approved):

TomWoodward (dismissed):
i understand there are some updated assets that need to be added as a result of this thread https://openstax.slack.com/archives/C8V1XET1R/p1653404920027699

Dantemss:
> i understand there are some updated assets that need to be added as a result of this thread https://openstax.slack.com/archives/C8V1XET1R/p1653404920027699

Well do you want some lossy compression then? The panda thing is lossy, but you can't really tell IMO.

TomWoodward:
Pick something you think is reasonable and ask Victoria if it's ok, I think
a little loss is fine

PR #1592, opened 2022-05-13
Updated search sidebar for large mobile screens

bethshook:
for: openstax/unified#1906

This shifts some breakpoints so that mobile search sidebar behavior begins below 1200px instead of 800px. The search bar had to be adapted accordingly. This will likely need to change again after further UX review.


bethshook:
@jivey this is updated with changes per [today's conversation](https://github.com/openstax/unified/issues/1632#issuecomment-1129055604) with Victoria. Search sidebar between 800 and 1200px should now match desktop, but search toolbar will include the "back to results" link when sidebar is hidden and results have not been cleared.

jivey (approved):
Looks good!

PR #1598, opened 2022-05-25
Text Resizer

jivey:
For: https://github.com/openstax/unified/issues/1676


src/app/messages/es/messages.json
```diff
@@ -210,5 +210,9 @@
   "i18n:toolbar:studyguides:button:text": "Guías",
   "i18n:toolbar:studyguides:popup:button:back-to-top": "Volver al inicio",
   "i18n:toolbar:studyguides:popup:close-button:aria-label": "Cerrar guía de estudio",
-  "i18n:toolbar:studyguides:popup:heading": "Guía de estudio"
+  "i18n:toolbar:studyguides:popup:heading": "Guía de estudio",
+  "i18n:toolbar:textresizer:button:aria-label": "Cambiar el tamaño del texto",
+  "i18n:toolbar:textresizer:button:decrease:aria-label": "Disminuir el tamaño del texto",
+  "i18n:toolbar:textresizer:button:increase:aria-label": "Aumentar el tamaño del texto",
+  "i18n:toolbar:textresizer:popup:heading": "Tamaño del texto"
```
jivey, line 217: @bethshook added some placeholder translations, hoping google didn't steer me too far off

src/app/messages/pl/messages.json
```diff
@@ -210,5 +210,9 @@
   "i18n:toolbar:studyguides:button:text": "Kompendia",
   "i18n:toolbar:studyguides:popup:button:back-to-top": "Idź na górę",
   "i18n:toolbar:studyguides:popup:close-button:aria-label": "Zamknij Kompendia",
-  "i18n:toolbar:studyguides:popup:heading": "Kompendium"
+  "i18n:toolbar:studyguides:popup:heading": "Kompendium",
+  "i18n:toolbar:textresizer:button:aria-label": "Zmień rozmiar tekstu",
+  "i18n:toolbar:textresizer:button:decrease:aria-label": "Zmniejsz rozmiar tekstu",
+  "i18n:toolbar:textresizer:button:increase:aria-label": "Zwiększ rozmiar tekstu",
+  "i18n:toolbar:textresizer:popup:heading": "Rozmiar tekstu"
```
jivey, line 217: @firadecki 👋 Would you have some time to look at these and let me know if they work? I took them from google translate so maybe they need some changes.

src/app/messages/es/messages.json
bethshook: Looks good!

generic-styles/index.less
```diff
@@ -38,8 +68,8 @@ body {
     src: url('../../static/fonts/ibm-plex-sans-v9-latin-regular.woff2') format('woff2'), url('../../static/fonts/ibm-plex-sans-v9-latin-regular.woff') format('woff');
   }
 
-  font-size: 1.6rem;
-  line-height: 2.3rem;
+  .content-font-size(1.6rem);
+  line-height: calc(2.3rem * var(--content-text-scale));
```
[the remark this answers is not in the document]
bethshook: Another place I'm noticing existing issues that will be exacerbated by large text: the [writing guide handbook](https://openstax.org/books/writing-guide/pages/18-5-writing-process-create-a-multimodal-advocacy-project). Something (tables?) is causing the content to overflow its container and it gets worse as you increase font size. But maybe both of these things should just be separate bug cards. I will approve and leave it to you.

bethshook (approved):

src/app/messages/pl/messages.json
firadecki: @jivey Translation looks fine, grammarly correct. There is more than one translation possible for some positions but it's hard to specify without seeing text resizer. Translation is correct and good for now, I'll check if anything can be perfected during testing.

generic-styles/index.less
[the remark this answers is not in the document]
jivey: @victoria-cooke @bethshook I updated the line-heights for h2s and h3s, picked an arbitrary 125% of the font-size, could see wanting to tweak it some more, maybe slightly tighter for h2? Let me know your thoughts and I can update them:

Before | After
--- | ---
image | image

Before | After
--- | ---
image | image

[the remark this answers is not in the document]
jivey: I forgot to label these images better: top is the default font size, bottom is the largest.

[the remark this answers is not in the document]
jivey: And here's a single line version at default font size that helps show the slight space change, for example an h2 would get around 10px more visible space between the header and paragraph. Might be a lot easier to test on staging and adjust in a future release?

Before | After
--- | ---
image | image |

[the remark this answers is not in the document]
bethshook: I defer to @victoria-cooke

jivey:
@bethshook Made one last change that might be significant, fixes a bug with using dynamic styles by ensuring the data-attribute rules always get applied: https://github.com/openstax/rex-web/pull/1598/commits/ed666bf274c6be30fdc65e9225d85c5ee551584b

src/app/content/hooks/storeTextSize.ts
```diff
@@ -0,0 +1,14 @@
+import { ActionHookBody } from '../../types';
+import { setTextSize } from '../actions';
+import { textSize } from '../selectors';
+
+const hookBody: ActionHookBody<typeof setTextSize> = (services) => async() => {
```
TomWoodward, line 5: you can add a body to the first function level here, which is run when the app is wired up and dispatch the initial `setTextSize` there, that way everything would all be in the same place

src/app/content/reducer.ts
```diff
@@ -18,6 +18,8 @@ import searchReducer, {initialState as initialSearchState } from './search/reduc
 import studyGuidesReducer, {initialState as initialStudyGuidesState } from './studyGuides/reducer';
 import { State } from './types';
 
+const storedTextSize = typeof window !== 'undefined' && window.localStorage.getItem('textSize');
```
TomWoodward, line 21: you may need to have this here if my other suggestion causes a flash of the stuff at a different size, but even then the key should at least be stored in a constant

src/app/content/components/Topbar/index.tsx
```diff
@@ -121,18 +151,29 @@ class Topbar extends React.Component<Props, State> {
...
             <Styled.SearchButton desktop colorSchema={this.props.searchButtonColor} data-experiment />
           }
         </Styled.SearchInputWrapper>
+        <TextResizer
+          bookTheme={this.props.bookTheme}
+          onChangeTextSize={onChangeTextSize}
+          onDecreaseTextSize={onDecreaseTextSize}
+          onIncreaseTextSize={onIncreaseTextSize}
+          textSize={this.props.textSize}
+          onToggle={this.props.clearSearch}
```
TomWoodward, line 171: i don't think we want to do this

TomWoodward (changes requested):

src/app/content/components/popUp/Filters.tsx
```diff
@@ -76,23 +80,31 @@ const Toggle = styled(React.forwardRef<HTMLButtonElement, ToggleProps>(
...
+      toggleChildren={toggleChildren}
+      ariaLabelId={ariaLabelId}
+      data-analytics-label={dataAnalyticsLabel}
+      showAngleIcon={props.showAngleIcon} />}
+    transparentTab={false}
+    {...props}
+  >
+    {children}
+  </Dropdown>
+);
```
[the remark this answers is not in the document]
jivey: good call

src/app/content/hooks/storeTextSize.ts
jivey: I tried that but redux complains about dispatching while middleware is constructing, maybe I set it up wrong though

TomWoodward: No you're probably right, you might need another hook in this file listening to locationChange for the setup

src/app/content/components/Content.browserspec.tsx
```diff
@@ -9,8 +9,8 @@ const TEST_CASES: { [testCase: string]: (target: Page) => Promise<void> } = {
   Desktop: setDesktopViewport, Mobile: setMobileViewport,
 };
 const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
-  Desktop: [242, 90, 122, 242, 365, 668, 761, 1263, 1607],
-  Mobile: [239, 66, 96, 239, 523, 1263, 1402, 1751, 2118],
+  Desktop: [249, 90, 122, 249, 372, 675, 768, 1270, 1614],
+  Mobile: [246, 66, 96, 246, 530, 1270, 1409, 1758, 2125],
```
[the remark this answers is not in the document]
TomWoodward: Ok that makes sense, did Victoria ever get back to you? You may want to check with Larissa too.

Also did you check the math based on the number of headings in the test page to make sure you're setting the expected values?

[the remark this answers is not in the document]
jivey: After another round of line height updates, these do not change.

generic-styles/index.less
```diff
@@ -143,14 +151,16 @@ body {
...
-  h2{
-    font-size: 3.2rem;
+  h2 {
+    .content-font-size(3.2rem);
+    line-height: calc(3.5rem * var(--content-text-scale));
   }
   h3 {
-    font-size: 2.4rem;
+    .content-font-size(2.4rem);
+    line-height: calc(2.3rem * var(--content-text-scale));
```
TomWoodward, line 160: you wanna make a similar helper for line height?

src/app/content/actions.ts
```diff
@@ -32,3 +32,5 @@ export const receivePageNotFoundId = createStandardAction('Content/receivePageNo
 
 export const openNudgeStudyTools = createStandardAction('Content/openNudgeStudyTools')();
 export const closeNudgeStudyTools = createStandardAction('Content/closeNudgeStudyTools')();
+
+export const setTextSize = createStandardAction('Content/setTextSize')<number>();
```
TomWoodward, line 36: you should type this and `textResizerValueMap` with a union of const values, or replace `textResizerValueMap` with some kind of math function, otherwise you're setting yourself up to have invalid numbers go in the top and `undefined` printed in the styles

src/app/content/components/Topbar/index.tsx
```diff
@@ -85,28 +91,52 @@ class Topbar extends React.Component<Props, State> {
...
+      this.props.setTextSize(newValue);
+    };
+
     const showBackToSearchResults = !this.props.searchSidebarOpen && this.props.hasSearchResults;
 
     const newButtonEnabled = !!this.props.searchButtonColor;
 
     return <Styled.TopBarWrapper data-testid='topbar'>
+      {/* mobileMedium is smaller (800px) than mobile (1200px) */}
+      {/* Only hides on .mobile breakpoint, shows on desktop and .mobileMedium */}
```
TomWoodward, line 125: what hides on mobile? does this comment conflict with the one on line 175?

src/app/content/hooks/storeTextSize.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (typeof window !== 'undefined' && window.localStorage) {
+    storedTextSize = window.localStorage.getItem(textResizerStorageKey);
+  }
+
+  if (storedTextSize) {
+    value = parseInt(storedTextSize, 10);
+  }
+
+  dispatch(setTextSize(value));
```
TomWoodward, line 24: you wanna check that value is not NaN

src/app/content/hooks/locationChange/index.ts
```diff
@@ -24,6 +25,7 @@ const hookBody: RouteHookBody<typeof content> = (services) => {
       loadStudyGuides(services)(),
       loadPracticeQuestions(services)(),
       initializeIntl(services)(),
+      loadStoredTextSize(services)(),
```
TomWoodward, line 28: you only have to do this here if your hook relies on the book data being loaded, i think in your case you could compose the hook directly in `src/app/content/hooks/index.ts`

TomWoodward (changes requested):

src/app/content/components/Topbar/index.tsx
jivey: Styled.SearchPrintWrapper below, but these were more to keep it straight for myself while working on it, might just remove them because they feel prone to being out of date soon

generic-styles/index.less
jivey: yeah good idea, at first there were only 2 entries and didn't think about it after we updated a few other places

src/app/content/hooks/storeTextSize.ts
jivey: yeah this really did need better NaN and bounds checking, added that here and also in the event callback

src/app/content/actions.ts
jivey: Added better typing here with `as const`, does require a few `as TextResizerValue` because of the narrowing when using `.includes` but overall seems to work nicely

jivey: Ended up going with some minor duplication to explicitly type the values https://github.com/openstax/rex-web/pull/1598/files#diff-51c9e59c9ff2ef74713ad6f6f68e468bc12c90fd9778beab1fa3f710c2733553R24 but maybe there is a better way, I thought a union approach wouldn't narrow it enough? Let me know if you have a suggestion!

TomWoodward (changes requested):

src/app/content/constants.ts
```diff
@@ -20,3 +20,12 @@ export const highlightStyles: Array<{label: HighlightColorEnum, passive: string,
   {label: HighlightColorEnum.Purple, passive: '#cbcfff', focused: '#545ec8'},
   {label: HighlightColorEnum.Pink, passive: '#ffc5e1', focused: '#de017e'},
 ];
+
+export type TextResizerValue = -2 | -1 | 0 | 1 | 2 | 3;
+export const textResizerValues: TextResizerValue[] = [-2, -1, 0, 1, 2, 3];
```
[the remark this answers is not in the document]
jivey: No cause `as const` doesn't get transformed, babel thinks it's a typo.

src/app/content/components/Page/connector.ts
```diff
@@ -30,6 +30,7 @@ export interface PagePropTypes {
   services: AppServices & MiddlewareAPI;
   addToast: typeof addToast;
   systemQueryParams: SystemQueryParams;
+  textSize: number;
```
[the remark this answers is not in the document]
jivey: Oops I shouldn't not have relied on tslint complaining to me about these, good catch...

TomWoodward (approved):

PR #1603, opened 2022-06-01
Fix safari background

jivey:
For: https://github.com/openstax/unified/issues/1914

In Safari only, this rule was causing the white background to have a cut off height instead of filling 100% of the container. Removing the rule does not seem to impact other browsers, so it seems safe to remove.

Before | After
--- | ---
image | image


Dantemss (commented):
I think if there's a page where the content doesn't even fill one screen, then we may get some grey at the bottom. Could you check? A solution for that which hopefully doesn't cause this issue in Safari could be `min-height: 100vh;`

jivey:
The MinPageHeight component keeps it from being too short:

image

So I'm not sure if there's a need for the outer wrapper to worry about it. But maybe we should do a `100vh` in case some content ends up getting rendered outside of the Page component. I just need to test it on some older browsers.

Dantemss (approved):
Well if there's a component that sets the page height elsewhere, maybe this rule is not even needed here. So I'll approve.

PR #1607, opened 2022-06-06
Added Polish translation fixes for the Keyboard Shortcuts menu

Dantemss:
For: https://github.com/openstax/unified/issues/1764


Dantemss:
@firadecki when I pasted your last translations in, a few of the lines had a diff, so if you could take a look that would be great

firadecki (approved):
I've just checked, looks fine.

PR #1608, opened 2022-06-07
use self-referencing url for canonical if page ID missing in canonical book

bethshook:
for: openstax/unified#1862

this fixes the bug outlined [here](https://github.com/openstax/unified/issues/1862#issuecomment-1147679099). in some books that map to newer editions, eoc/eob page IDs were changed. pages that don't have a new mapped ID should use self-referencing canonicals (as is currently the case on production).


Dantemss (approved):

bethshook:
@Malar-Natarajan this is ready for review - it fixes the issue with canonicals on eob/eoc pages. I am trying to test other books and pages to make sure it doesn't cause new regressions.

PR #1613, opened 2022-06-10
upgrade sentry to 7.2.0

jivey:
For: https://github.com/openstax/unified/issues/1928

Upgrades Sentry to 7.2.0

`redux-sentry-middleware` is no longer updated, `@sentry/react` has Redux integration built-in. 

Severity enums were replaced with string literals to save on bundle size:
https://github.com/getsentry/sentry-javascript/blob/9ca911bdadfef236950cdbb2c2dcc48433bdaba1/MIGRATION.md#severity-severitylevel-and-severitylevels


src/app/index.tsx
```diff
@@ -101,16 +102,23 @@ export default (options: AppOptions) => {
...
+    enhancers,
     initialState,
     middleware,
     reducer,
   });
 
+  if (Sentry.shouldCollectErrors) {
+    Sentry.initialize(store);
+  }
+
```
jivey, line 121: It would be nice if I didn't need both `if`s here, we could alternatively leave the initializer as a no-op middleware and push it in the same spot, if that's preferred: https://github.com/openstax/rex-web/pull/1613/commits/965785d87989192bff5b5b9d571ebb78df46d5a3#diff-8a43619f80571a2033826efebd666b14d7d578f88af0f252421fe9266face636L45

bethshook (approved):

PR #1619, opened 2022-06-21
Add dynamic style loading for Organizational Behavior

Dantemss:
For: https://github.com/openstax/unified/issues/1896


Dantemss:
A few questions I had while making the PR:

1. Should we list allowed books by id/uuid or slug? Does it matter?
2. Do we keep style_href and style_name optional or say all archive books will have them and update the mocks to match?
3. Should style_name even be in the code if it's unused?

TomWoodward:
> A few questions I had while making the PR:
> 
> 1. Should we list allowed books by id/uuid or slug? Does it matter?
> 2. Do we keep style_href and style_name optional or say all archive books will have them and update the mocks to match?
> 3. Should style_name even be in the code if it's unused?

1. id, because corgi preview uses the id routes not slugs
2. we should probably make it a required field as long as we have no books running on older pipelines that don't have it
3. i think we should ignore the style_name field

Dantemss:
> > A few questions I had while making the PR:
> > 
> > 1. Should we list allowed books by id/uuid or slug? Does it matter?
> > 2. Do we keep style_href and style_name optional or say all archive books will have them and update the mocks to match?
> > 3. Should style_name even be in the code if it's unused?
> 
> 1. id, because corgi preview uses the id routes not slugs
> 2. we should probably make it a required field as long as we have no books running on older pipelines that don't have it
> 3. i think we should ignore the style_name field

I checked and Sociology 1e is pinned to a pipeline version that doesn't have style_href so I left it optional for now.
Removed style_name

src/app/components/DynamicContentStyles.tsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import { useSelector } from 'react-redux';
 import styled from 'styled-components';
+import { book } from '../content/selectors';
 import { query } from '../navigation/selectors';
 import { assertDefined } from '../utils/assertions';
 
+// Currently only Organizational Behavior
+const bookIdsWithDynamicStyles = [ '2d941ab9-ac5b-4eb8-b21c-965d36a4f296' ];
```
bethshook, line 9: could we add a property to the books config instead?

src/app/components/DynamicContentStyles.spec.tsx
```diff
@@ -50,6 +54,30 @@ describe('DynamicContentStyles', () => {
...
+    store.dispatch(receiveBook(book));
+    const spyFetch = jest.spyOn(globalThis, 'fetch')
+      .mockImplementation(async() => ({ text: async() => '.cool { color: red; }' }) as any);
+
+    const componenet = renderer.create(<TestContainer store={store}>
+      <Component />
+    </TestContainer>);
+
+    // tslint:disable-next-line: no-empty
+    await renderer.act(async() => {});
```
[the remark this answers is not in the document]
Dantemss: Replaced with that method call

bethshook (approved):

TomWoodward (changes requested):
in the review environment i'm not getting any css, whats going on with that? https://rex-web-organizational--xpf5y3.herokuapp.com/books/organizational-behavior/pages/1-introduction

src/app/components/DynamicContentStyles.tsx
```diff
@@ -22,27 +24,35 @@ const DynamicContentStyles = React.forwardRef<HTMLElement, DynamicContentStylesP
...
       setStyles('');
       return;
     }
 
-    const cssfileUrl = queryParams['content-style'];
+    let cssfileUrl = queryParams['content-style'];
+    if (!cssfileUrl && currentBook) {
+      const bookConfig = BOOKS[currentBook.id];
+      if (bookConfig && bookConfig.dynamicStyles) {
+        cssfileUrl = currentBook.style_href;
```
TomWoodward, line 39: you gotta run this through that content reference resolver helper thing don't you?

TomWoodward (changes requested):

src/app/components/DynamicContentStyles.tsx
TomWoodward: like this, to rebase the relative url based on the content url? https://github.com/openstax/rex-web/blob/master/src/app/content/utils/contentManipulation.ts#L30

Dantemss: I got into several rabbit holes trying to get the book archive url from the book object, but in the end building it from REACT_APP_ARCHIVE_URL seems to have worked...

Dantemss:
Not sure how to get Heroku to update. GitHub is saying the deployment is inactive, but it's still there...

src/app/components/DynamicContentStyles.tsx
```diff
@@ -20,29 +26,44 @@ const DynamicContentStyles = React.forwardRef<HTMLElement, DynamicContentStylesP
...
-    const cssfileUrl = queryParams['content-style'];
+    let cssfileUrl = queryParams['content-style'];
+
+    if (!cssfileUrl && currentBook && currentBook.style_href) {
+      const bookConfig = BOOKS[currentBook.id];
+      if (bookConfig && bookConfig.dynamicStyles) {
+        cssfileUrl = currentBook.style_href;
+
+        if (!isAbsoluteUrl(cssfileUrl)) {
+          const archiveUrl = `${bookConfig.archiveOverride || REACT_APP_ARCHIVE_URL}/`;
```
bethshook, line 47: Just looked at the same thing with @jivey. You can use [createBookConfigLoader.ts](https://github.com/openstax/rex-web/blob/main/src/gateways/createBookConfigLoader.ts#L44) for this and maybe some of the previous logic. `getBookVersionFromUUIDSync` returns the latest config for a given book.

Dantemss: I added calls to this method and getArchiveUrl() to get the config and the default archive url.
It seems the method can return undefined if the book config is not loaded, so I added it to the useEffect array to hopefully get it to re-render properly...

bethshook: I don't see how either would return undefined - the local book config and archive URL are used until you attempt to load a new book content version, in which case the remote config is loaded and they're updated.

TomWoodward:
i haven't been reviewing this since the href isn't in its final form yet, is that still the case?

Dantemss:
> i haven't been reviewing this since the href isn't in its final form yet, is that still the case?

I think that is correct, since changes would show up in a new archive pipeline

Dantemss:
This should be ready for re-review

src/app/components/DynamicContentStyles.tsx
```diff
@@ -12,37 +16,51 @@ export const WithStyles = styled.div`
...
+
     if (cssfileUrl && typeof cssfileUrl === 'string') {
       if (cacheStyles.has(cssfileUrl)) {
         setStyles(assertDefined(cacheStyles.get(cssfileUrl), `we've just checked for this`));
       } else {
         fetch(cssfileUrl)
           .then((res) => res.text())
           .then((data) => {
-            cacheStyles.set(cssfileUrl, data);
+            // we are inside an "if" that checks that cssfileUrl is defined and a string
```
TomWoodward, line 57: it doesn't trust you because you initialized it with `let` and there might be async logic that changes it. if you change your initialization to use `const` it'll stop complaining. if you don't want to re-write your logic you can move it to a function so that you can use a return value in the `const cssfileUrl = figureOutTheCssUrl()` line

TomWoodward (changes requested):
really minor comment, this is mostly good to go

src/app/components/DynamicContentStyles.tsx
Dantemss: Alright, I moved the cssFileUrl logic to a separate function

src/app/components/DynamicContentStyles.tsx
```diff
@@ -11,25 +17,46 @@ export const WithStyles = styled.div`
...
 const DynamicContentStyles = React.forwardRef<HTMLElement, DynamicContentStylesProps>((
-  { children, disable, ...otherProps }: React.PropsWithChildren<DynamicContentStylesProps>,
+  { book, children, disable, ...otherProps }: React.PropsWithChildren<DynamicContentStylesProps>,
   ref
 ) => {
-  const queryParams = useSelector(query);
   const [styles, setStyles] = React.useState('');
+  const queryParams = useSelector(query);
+  const { archiveLoader } = useServices();
+  const cssfileUrl = getCssFileUrl(queryParams, book, archiveLoader);
```
TomWoodward, line 52: you probably want to wrap this in React.useMemo or it could result in the following effect running on every render

TomWoodward: or move the call inside the useEffect

TomWoodward (changes requested):

src/app/components/DynamicContentStyles.tsx
Dantemss: I don't know why I had moved it out of useEffect... anyway I put it back in.

TomWoodward (approved):

PR #1623, opened 2022-06-22
simplify canonicalUrl logic and add test

bethshook:
for: openstax/unified#1935


src/app/content/utils/canonicalUrl.ts
```diff
@@ -38,23 +38,18 @@ export async function getCanonicalUrlParams(
         getBookVersionFromUUIDSync(id),
         `We've already filtered out books that are not in the BOOK configuration`
       ).defaultVersion;
-      const useCurrentBookAsCanonical = book.id === id  && hasOSWebData(book);
-      canonicalBook = useCurrentBookAsCanonical ? book : await getBook(id, version);
-      canonicalPageId = CANONICAL_PAGES_MAP[pageId] || canonicalPageId;
```
Dantemss, line 43: This looks like a bug fix. pageId was not updated in the loop.

Dantemss (approved):
OK I see 2 main changes from the previous version of the code:
1. The pageId variable was not updated in the loop but it was used to look into the CANONICAL_PAGES_MAP, this was replaced with the canonicalPageId, which does get updated in the loop.
2. The branches of code that had special handling for useCurrentBookAsCanonical were removed, only the ternary assignment remains. I think this leads to one more lookup in the CANONICAL_PAGES_MAP but then it stops as normal, as the canonicalMap should stop changing at this point. Unless there's a cycle in the canonical maps that used to be handled by the termination in useCurrentBookAsCanonical. But maybe we have cycle detection code elsewhere? I forget...
Anyway, looks good to me.

TomWoodward (approved):

bethshook:
Thanks for reviewing. Line 51 should catch any loops in the mapping. Maybe removing that conditional was not necessary but it cuts down a few lines of code.

PR #1628, opened 2022-06-24
script for importing canonical page maps

bethshook:
for: openstax/unified#1948


script/add-page-map.ts
```diff
@@ -0,0 +1,50 @@
...
+  } = argv.string('mapPath').argv as any as {
+    mapPath: string,
+    bookSlug: string,
+    bookTitle: string,
+    bookId: string,
+    canonicalBookTitle: string,
+    canonicalBookId: string,
+  };
+
+  if (!mapPath || !bookSlug || !bookId || !canonicalBookId) {
```
jivey, line 26: Maybe `bookTitle` and `canonicalBookTitle` should be required?

jivey (approved):
Looks good, left a thought about required args.

script/add-page-map.ts
bethshook: I think it's ok - they're not required for the map to work and you can always look them up.

TomWoodward (approved):

PR #1633, opened 2022-06-28
Target all numbered .os-problem-container and .os-solution-container with display: inline

Dantemss:
For: https://github.com/openstax/unified/issues/1952
Also made .mixin-exercise target both exercise types to fix the indentation and removed duplicate styles

Screen Shot 2022-06-28 at 2 02 08 PM

Screen Shot 2022-06-28 at 1 10 55 PM

### DIFF
--- a/generic-styles/content-baked.less
+++ b/generic-styles/content-baked.less
@@ -273,27 +273,21 @@ h1.example-title .text {
   padding: 0;
 }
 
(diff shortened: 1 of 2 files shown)


jivey (approved):
Looks good, may need a PR in `ce-styles` as well?

PR #1634, opened 2022-06-29
fix seo issues

TomWoodward:
for: openstax/unified#1955

- fix issue where meta tags in head are cleared from prerendered pages after initial locationChange is dispatched
- override to produce fewer javascript files in the build
- fix issue where intl provider is not populated on first render of hydrated pages
- fix formatting of sitemap index
- fix a link with no href
- fix an issue where images were not being processed by pre-rendering correctly

i'm not like... the most happy with some of these changes, i left comments where improvements could be made

### DIFF
--- a/script/prerender/setup.ts
+++ b/script/prerender/setup.ts
@@ -12,7 +12,6 @@
  */
 import * as fs from 'fs';
 import ignoreStyles, { DEFAULT_EXTENSIONS, noOp } from 'ignore-styles';
-import md5File from 'md5-file';
 import mime from 'mime';
 import * as path from 'path';
 
@@ -27,14 +26,15 @@ ignoreStyles(DEFAULT_EXTENSIONS, (mod, filename) => {
     // If we find a style
     return noOp();
   } else {
-    // If we find an image
-    const hash = md5File.sync(filename).slice(0, 8);
-    const bn = path.basename(filename).replace(/(\.\w{3})$/, `.${hash}$1`);
-    const filePath = `/static/media/${bn}`;
+    const bn = path.basename(filename).replace(/(\.\w{3})$/, '');
+    const dir = '/static/media';
+    // TODO - make this better
+    const fileName = fs.readdirSync(path.resolve(__dirname, '../../build' + dir))
+      .find((name) => name.indexOf(bn) !== -1);
 
-    if (fs.existsSync(path.resolve(__dirname, '../../build', filePath.replace(/^\//, '')))) {
+    if (fileName) {
       // file exists in build folder, refrence it by url here
-      mod.exports = `${process.env.PUBLIC_URL || ''}${filePath}`;
+      mod.exports = `${process.env.PUBLIC_URL || ''}${dir}/${fileName}`;
     } else {
       // file doesn't exist in build folder, assume it is an inlined image
       // and inline it again here
--- a/script/prerender/setup.ts
+++ b/script/prerender/setup.ts
@@ -26,15 +26,24 @@ ignoreStyles(DEFAULT_EXTENSIONS, (mod, filename) => {
     // If we find a style
     return noOp();
   } else {
-    const bn = path.basename(filename).replace(/(\.\w{3})$/, '');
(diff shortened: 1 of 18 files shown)


script/prerender/setup.ts
```diff
@@ -27,14 +26,24 @@ ignoreStyles(DEFAULT_EXTENSIONS, (mod, filename) => {
...
 
-    if (fs.existsSync(path.resolve(__dirname, '../../build', filePath.replace(/^\//, '')))) {
+    /*
+     * CRA configures these media files be output with the [hash] token, which is a content
+     * hash mixed in with a build id, so they're busted on every build. it may be better to
+     * override the config to use [contenthash], but thats not really straightforward even using
+     * craco, so we're hacking it here by ignoring the hash. the hack will work as long as we don't
+     * have any assets with the same name (that would normally be differentated by the hash), so we
+     * bail if that is the case.
+     * */
```
TomWoodward, line 41: this must have been broken since an update to CRA or something

src/app/content/components/Content.browserspec.tsx
```diff
@@ -13,6 +14,14 @@ const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
   Mobile: [239, 66, 96, 239, 523, 1263, 1402, 1751, 2118],
 };
 
+beforeAll(async() => {
+  await page.setCookie({
```
TomWoodward, line 18: idk what change i made but the nudge became a bigger problem in some of these tests, the cookie makes it not obscure the rest of the test behavior

src/index.tsx
```diff
@@ -84,9 +86,19 @@ app.services.promiseCollector.calm().then(() => {
...
 if (window.__PRELOADED_STATE__) {
-  Loadable.preloadReady().then(() => {
-    ReactDOM.hydrate(<app.container />, document.getElementById('root'), doneRendering);
-  });
+  Loadable.preloadReady()
+    .then(() => {
+      // during pre-rendering this happens in src/app/content/hooks/intlHook.ts
+      // it would be nice to consolodate this logic, but in hydration we don't necessarily
+      // want to wait for promiseCollector.calm() before rendering _anything_, so there are some
+      // discrepancies in the flow that make the logic annoying.
```
Dantemss, line 94: This refers to the `if` branches right? I see you could combine the 2 `then`s in this branch of the `if` by awaiting on createIntl() if needed, but you'd have to make the arrow function async. In the end I don't know if it would be better or worse or matter at all.

Dantemss (changes requested):
Only the comment about the sitemapPath really matters

script/prerender/sitemap.ts
```diff
@@ -8,7 +8,7 @@ import sitemap, { SitemapItemOptions } from 'sitemap';
 import { SerializedPageMatch } from './contentRoutes';
 import { writeAssetFile } from './fileUtils';
 
-export const sitemapPath = (pathName: string) => `/rex/sitemaps/${pathName}.xml`;
+export const sitemapPath = (pathName: string) => `https://openstax.org/rex/sitemaps/${pathName}.xml`;
```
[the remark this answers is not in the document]
TomWoodward: ah... ok i'll take another look... it needs to be the full url in the index

script/prerender/thread.ts
```diff
@@ -100,7 +100,7 @@ function makeSitemapIndexTask(services: AppOptions['services']) {
     );
     const items = await asyncPool(MAX_CONCURRENT_CONNECTIONS, books, async(book) => {
       const archiveBook = await getArchiveBook(services, book);
-      return getSitemapItemOptions(archiveBook, sitemapPath(book.params.book.slug));
+      return getSitemapItemOptions(archiveBook, `https://openstax.org${sitemapPath(book.params.book.slug)}`);
```
TomWoodward, line 103: this was apparently a discrepancy that was introduced when switching to the distributed pre-render, the old logic is here https://github.com/openstax/rex-web/blob/main/script/prerender/sitemap.ts#L60

PR #1636, opened 2022-06-30
Send pipeline version with highlights

jivey:
For: https://github.com/openstax/unified/issues/1957


src/app/content/utils.ts
```diff
@@ -137,3 +139,7 @@ export const loadPageContent = async(loader: ReturnType<AppServices['archiveLoad
   const page = await loader.page(pageId).load();
   return page.content;
 };
+
+export const getBookPipelineVersion = (book: Book): string => {
```
bethshook, line 143: we have this getter to make sure the latest pipeline is being used: https://github.com/openstax/rex-web/blob/main/src/gateways/createBookConfigLoader.ts#L44

jivey: Ah okay so it's probably safe then to also make this search hook use it? https://github.com/openstax/rex-web/pull/1636/files#diff-a3aea54c89e3ebfc1fdca9da914d39cf889cc9b3d2de9574a1b207fd243fce6cL31

bethshook: I can't see why not. In the hook I think you'd want to do something like 

```
const archiveUrl = getBookVersionFromUUIDSync(bookId)?.archiveOverride || getArchiveUrl();
const pipeline = archiveUrl.replace(/^\/apps\/archive\//, '');
```

jivey: Thanks makes sense, for some reason thought it might be handled specially there, but forgot we just added that method recently

src/app/content/utils.ts
```diff
@@ -137,3 +139,8 @@ export const loadPageContent = async(loader: ReturnType<AppServices['archiveLoad
...
+  const url = BOOKS[book.id]?.archiveOverride || getArchiveUrl();
```
bethshook, line 144: Is there a reason to not use the book config loader to get the archive override as well?

bethshook (approved):

PR #1638, opened 2022-07-01
College Physics 2e updates

staxbot:


TomWoodward:
i believe this isn't needed because it was added with the canonical maps

PR #1641, opened 2022-07-05
Fix URL double-encoding in sitemaps

Dantemss:
For: https://github.com/openstax/unified/issues/1958
Seems to have been introduced with the prerender workers.


jpslav (approved):
I think this is ok.  Tried to look to see if the URL from `matchPathname` is expected to be encoded already but couldn't find where it comes from.

bethshook (approved):

PR #1642, opened 2022-07-06
Remove map redundancies

bethshook:
for: openstax/unified#1948

Fixes canonical URL error in Algebra & Trigonometry. I did two things here:
- per our [master list of canonical books](https://docs.google.com/spreadsheets/d/1Hj5vm2AbEiLgxgcbNRi550InzFTLvBOv16-Fw6P-c5E/edit#gid=0) I removed redundancies from math book mapping
- I terminated the loop for checking maps once a page is found in a canonical book that has no map of its own. The idea is that if Book A canonicals to Book B and Book C, we don't need to keep looking for pages once a match is found in Book B.

### DIFF
--- a/src/app/content/utils/canonicalUrl.ts
+++ b/src/app/content/utils/canonicalUrl.ts
@@ -48,7 +48,7 @@ export async function getCanonicalUrlParams(
         break;
       }
 
-      // use map for canonical book if it exists
+      // use canonical book's map
       const newMap = getCanonicalMap(canonicalBook.id);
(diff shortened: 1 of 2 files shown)


Dantemss (approved):
I'm not 100% sure what that new termination condition is, but I'll trust you on this one

jivey (approved):

Dantemss (changes requested):
Hold since we are going to revert 1623. This will probably need to re-include those changes.

PR #1644, opened 2022-07-06
Fix canonical mapping logic bugs

bethshook:
for: openstax/unified#1935


src/app/content/utils/canonicalUrl.spec.ts
```diff
@@ -217,4 +218,76 @@ describe('getCanonicalURL', () => {
...
+    const pageId = page.id;
+
+    CANONICAL_MAP[bookId] = [['testbook2-uuid', { [pageId]: 'testbook2-page' }]];
+    CANONICAL_MAP['testbook2-uuid'] = [['testbook3-uuid', {'testbook2-page': 'testbook3-page'}]];
+    CANONICAL_MAP['testbook3-uuid'] = [['testbook3-uuid', {}]];
+
+    const node = archiveUtils.findArchiveTreeNodeById(book.tree, pageId);
+    const spy = jest.spyOn(archiveUtils, 'findArchiveTreeNodeById')
+      .mockReturnValueOnce({ ...node, slug: 'testbook2-page' } as LinkedArchiveTreeSection)
+      .mockReturnValue({ ...node, slug: 'testbook3-page' } as LinkedArchiveTreeSection);
```
TomWoodward, line 236: i'm not sure about these `findArchiveTreeNodeById` mocks, is it possible to work with stuff that is in the test tree? or set up the mock test tree? 

i feel like we may rearrange this logic again, it would be nice for these tests to not be coupled to the implementation logic

TomWoodward (changes requested):

src/app/content/utils/canonicalUrl.ts
```diff
@@ -38,34 +38,36 @@ export async function getCanonicalUrlParams(
...
+      canonicalMap = newMap;
     }
   }
 
-  // use current page if page not found in canonical book
-  if (!treeSection && canonicalMap.length) {
-    treeSection = findArchiveTreeNodeById(book.tree, pageId);
-    canonicalBook = book;
+  // use current page as canonical if page not found in canonical book
+  if (!canonicalBookWithPage && canonicalMap.length) {
```
jivey, line 64: This isn't a new change, but what is the canonicalMap existence check here a signal for? I'm curious if these two ifs can be collapsed but I'm not sure what the check does.

bethshook: Historically this function only generated a canonical URL if a canonical page was successfully found. If the book wasn't found in the config (see [line 24](https://github.com/openstax/rex-web/pull/1644/files#diff-a7b7fc51715f392a83e1785cc3c82834c33371291b07eebced8573f006f25d21R24)) OR if the book was found but a canonical page was not, we returned null instead. Now we've decided any page that exists should get a canonical URL (if there's no canonical page it just references its own URL). So line 64 handles that case and line 68 handles the case where an actual `canonicalBookWithPage` was found.

It's possible that we never want to return null and just want to return a URL or throw an error instead. In that case I think we could remove the `canonicalMap.length` check.

src/app/content/utils/canonicalUrl.ts
```diff
@@ -29,7 +29,7 @@ export async function getCanonicalUrlParams(
   let canonicalPageId = pageId;
   let done = false;
   let canonicalBook;
-  let treeSection;
+  let canonicalBookWithPage;
 
   while (canonicalMap.length && !done) {
```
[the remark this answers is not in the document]
bethshook: `canonicalMap` is an array either of maps that have been specified in canonicalBookMap.ts or an empty default map. it's only undefined if the loaded book doesn't exist in the books config. `canonicalPageId` is the current page id and `canonicalBookWithPage` is undefined.

while there is a map array, loop through each map. set `canonicalBook` to the book id of the map being checked and update `canonicalPageId` to the page specified in the page map (if found). set `treeSection` to the tree node corresponding to the canonical page.

if no tree section was found but a `canonicalBookWithPage` was previously set, end the loop and use that `canonicalBookWithPage` to generate the url (think I was seeing this happen for pages with multiple canonical books specified but where not all those books had a matching page). else if a tree section was found, update `canonicalBookWithPage` to use it.

next check to see if the `canonicalBook` has its own map (happens with precalc 2e for example). if it does not or if it's the same map we just checked, stop the loop. throw an error if we've encountered a loop within the maps themselves. otherwise, set `canonicalMap` to the new map and start over.

once we're past the loop, return a canonical URL using `canonicalBookWithPage` if defined or using the current page's URL. return null if no map was ever defined (see discussion below on this).

TomWoodward (approved):

PR #1691, opened 2022-07-13
Fix build stuff

TomWoodward:
for: openstax/unified#1972

### DIFF
--- a/src/app/errors/components/OuterErrorBoundary.tsx
+++ b/src/app/errors/components/OuterErrorBoundary.tsx
@@ -6,8 +6,8 @@ import { assertWindow } from '../../utils/browser-assertions';
 import ErrorBoundary from './ErrorBoundary';
 
 // tslint:disable-next-line:variable-name
-const OuterErrorBoundary = (props: { children?: React.ReactNode }) => {
-  const [intl, setIntl] = useState<IntlShape | null>(null);
+const OuterErrorBoundary = (props: React.PropsWithChildren<{ intl: IntlShape | null }>) => {
+  const [intl, setIntl] = useState<IntlShape | null>(props.intl);
 
   useEffect(() => {
     const setUpIntl = async() => {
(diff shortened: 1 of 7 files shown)


src/app/errors/components/OuterErrorBoundary.tsx
```diff
@@ -18,11 +18,13 @@ const OuterErrorBoundary = (props: { children?: React.ReactNode }) => {
       setIntl(intlObject);
     };
 
-    setUpIntl();
-  }, []);
+    if (!intl) {
+      setUpIntl();
```
Dantemss, line 22: OK with no await here?

src/app/messages/MessageProvider.tsx
```diff
@@ -25,7 +25,7 @@ const MessageProvider = (props: { children?: React.ReactNode }) => {
     };
 
     setUpIntl();
```
Dantemss, line 27: I guess this one doesn't await either so it's probably ok?

src/app/messages/MessageProvider.tsx
```diff
@@ -7,13 +7,13 @@ import createIntl from './createIntl';
...
 // tslint:disable-next-line:variable-name
-const MessageProvider = (props: { children?: React.ReactNode }) => {
+const MessageProvider = (props: React.PropsWithChildren<{}>) => {
   const services = useServices();
   const [intl, setIntl] = useState<IntlShape | null>(services.intl.current);
   const bookLocale = useSelector(currentLocale);
 
   useEffect(() => {
-    if (!bookLocale) {
+    if (!bookLocale || intl) {
```
Dantemss, line 16: Would this be a problem when switching to a book with a different locale?

Dantemss (commented):
I'm not 100% sure so I have some questions.

src/app/messages/MessageProvider.tsx
TomWoodward: Yea, that's not a thing that should ever happen but I can add some logic to test for it

src/app/errors/components/OuterErrorBoundary.tsx
TomWoodward: Yea it shouldn't matter, this is the way it was set up already

Dantemss (approved):

PR #1692, opened 2022-07-13
Save a screenshot on scroll test failure

Dantemss:
For: https://github.com/openstax/unified/issues/1974
In case it happens again...
I put it in the diff_output dir so it gets uploaded as a GH action artifact.


Dantemss:
I think when I got the scroll failure the screenshot was not very useful so I'm going to close this...

PR #1693, opened 2022-07-14
redirect retired books

bethshook:
for: openstax/unified#1953


script/retire-book.ts
```diff
@@ -0,0 +1,65 @@
...
+import updateRedirectsData from './utils/update-redirects-data';
+
+const booksPath = path.resolve(__dirname, '../src/config.books.json');
+
+const args = argv.string('bookId').argv as any as {
+  retiredBook: string,
+  redirectBook: string
+};
+
+const retireBook = async() => {
```
bethshook, line 19: @TomWoodward my thought here is that we will manually run this until we get to a place where we want to trigger it via ABL changes

script/utils/update-redirects-data.ts
```diff
@@ -60,10 +69,20 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
       countNewRedirections++;
     // remove `else` to enable legitimately removing pages from books
     // only once uuids are guaranteed to be consistent
-    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException) {
+    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException && !allowBookRedirect) {
       throw new Error(
         `updateRedirects prohibits removing pages from a book, `
         + `but neither section with ID ${section.id} nor slug ${section.slug} was found in book ${newBook.id}`);
+    } else if (allowBookRedirect && archiveTreeSectionIsPage(section)) {
```
bethshook, line 76: @TomWoodward I probably need more conditions here such as checking for existing redirects like on line 51/60. I'm also wondering if there's a better way to exclude chapter and unit slugs than `archiveTreeSectionIsPage`.

script/utils/update-redirects-data.ts
```diff
@@ -60,10 +69,20 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
       countNewRedirections++;
     // remove `else` to enable legitimately removing pages from books
     // only once uuids are guaranteed to be consistent
-    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException) {
+    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException && !allowBookRedirect) {
```
TomWoodward, line 72: Once you get past the initial validation, I'd change these checks to compare the book ids, I think that is more clear. Like in this case this check is only used if the book ids are the same

script/utils/update-redirects-data.ts
```diff
@@ -23,9 +32,9 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
   const flatCurrentTree = flattenArchiveTree(currentBook.tree).filter((section) => section.id !== currentBook.id);
   const flatNewTree = flattenArchiveTree(newBook.tree).filter((section) => section.id !== currentBook.id);
```
TomWoodward, line 33: Idk what exactly this filter is for, but in this case I think the I'd might be wrong

script/utils/update-redirects-data.ts
TomWoodward: I....I feel like we aren't expecting the array to have non page entries in it, so maybe that would be a better change... Maybe

TomWoodward (changes requested):
I think we want the earlier checks, like the one on like 51/60 to fire even if the slug is the same when the book id is different.

That will catch any of the book retirement redirects that don't depend on a page mapping 

I'm a little on the fence about using this script for this. looking at it, it makes sense if you frame it as a script that generates redirects from one book to another if it can correlate the pages and the URL doesn't match already. That works in both cases, but really we have two kinda different use cases here so if you start to feel like you're building a hole that accepts both square and round pegs, you can switch to two scripts if you want.

script/utils/update-redirects-data.ts
```diff
@@ -21,14 +31,18 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
-  const formatSection = (section: LinkedArchiveTreeNode) => ({
-    bookId: currentBook.id,
-    pageId: section.id,
+  const formatSection = (section: LinkedArchiveTreeNode, newSection?: ArchiveTreeNode) => ({
+    bookId: allowBookRedirect ? newBook.id : currentBook.id,
+    pageId: newSection?.id ? stripIdVersion(newSection.id) : section.id,
     pathname: decodeURI(
       content.getUrl({ book: { slug: currentBook.slug }, page: { slug: section.slug } })
     ),
+    ...(allowBookRedirect && {query: `?${messageQueryParameterName}=retired`}),
```
bethshook, line 45: @TomWoodward where would be the best place to should store possible `message` values?

src/app/content/hooks/receivePageNotFoundId.ts
```diff
@@ -16,6 +21,9 @@ export const receivePageNotFoundIdHookBody: ActionHookBody<typeof receivePageNot
       return;
     }
   }
+  if (bookConfig?.retired) {
+    return Promise.reject(new BookNotFoundError(`Could not resolve uuid: ${book?.id}`));
+  }
```
bethshook, line 26: @TomWoodward this seems to work for pages of retired books where no redirect is specified. There's a bit of a delay though so you see the old book for a moment.

script/utils/update-redirects-data.ts
```diff
@@ -60,10 +75,20 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
       countNewRedirections++;
     // remove `else` to enable legitimately removing pages from books
     // only once uuids are guaranteed to be consistent
-    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException) {
+    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException && !booksAreDifferent) {
       throw new Error(
         `updateRedirects prohibits removing pages from a book, `
         + `but neither section with ID ${section.id} nor slug ${section.slug} was found in book ${newBook.id}`);
+    } else if (booksAreDifferent && !redirects.find(matchRedirect(section))) {
```
TomWoodward, line 82: lets do the matchException check here too, like if we didn't throw on the previous condition because of that then also don't go in this logic

script/utils/update-redirects-data.ts
```diff
@@ -60,10 +75,20 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
+      // if redirecting a book but no page match found, check for a canonical page
+      const canonicalPageMap = CANONICAL_MAP[currentBook.id]?.find((pageMap) => pageMap[0] === newBook.id) || [];
+      const canonicalPageId = canonicalPageMap[1] && canonicalPageMap[1][section.id];
+      // if no canonical page, fall back to default book page
+      const redirectSection = (canonicalPageId && findArchiveTreeNodeById(newBook.tree, canonicalPageId))
```
TomWoodward, line 87: lets also add a code comment in here that this condition case is specific to the retirement logic

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -132,8 +132,9 @@ const resolvePage = async(
   const pageId = match.state && 'pageUid' in match.state
     ? match.state.pageUid
     : getPageIdFromUrlParam(book, match.params.page);
+  const bookConfig = getBookVersionFromUUIDSync(book.id);
 
-  if (!pageId) {
+  if (!pageId || bookConfig?.retired) {
```
TomWoodward, line 137: i think we should move this into resolveBook, you'll have to change `getIdFromPageParam` to not need a specific page, which isn't used anyway

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -36,9 +36,19 @@ const getBookResponse = async(
...
   bookSlug?: string
 ): Promise<[Book, ReturnType<AppServices['archiveLoader']['book']>]>  => {
   const osWebBook = bookSlug ? await osWebLoader.getBookFromSlug(bookSlug) : undefined;
+
+  const bookConfig = getBookVersionFromUUIDSync(bookId);
+  // if (bookConfig?.retired && !UNLIMITED_CONTENT) {
+  if (bookConfig?.retired) {
+    dispatch(receivePageNotFoundId(null));
+    return Promise.reject();
+  }
```
bethshook, line 50: @TomWoodward however I try to bail here I end up with a "Book not loaded" error from the intlHook

src/app/content/hooks/receivePageNotFoundId.ts
```diff
@@ -1,11 +1,16 @@
...
-import { actionHook } from '../../utils';
+import { actionHook, BookNotFoundError } from '../../utils';
 import { receivePageNotFoundId } from '../actions';
 
 export const receivePageNotFoundIdHookBody: ActionHookBody<typeof receivePageNotFoundId> = (
   services
 ) => async() => {
+  const state = services.getState();
+  const book = contentSelect.book(state);
+  const bookConfig = book ? await services.bookConfigLoader.getBookVersionFromUUID(book.id) : undefined;
```
bethshook, line 12: this no longer works if the book hasn't been loaded because it's not in state

src/app/content/hooks/locationChange/index.ts
```diff
@@ -14,7 +14,10 @@ const hookBody: RouteHookBody<typeof content> = (services) => {
   const boundRegisterPageView = registerPageView(services);
 
   return async(action) => {
-    await resolveContent(services, action.match);
+    const response = await resolveContent(services, action.match);
+    if (!response) {
+      return;
+    }
```
bethshook, line 20: @TomWoodward I'm unsure of the best way to do this but this is to prevent the intlHook from running.

src/app/content/actions.ts
```diff
@@ -29,7 +29,7 @@ export const requestPage = createStandardAction('Content/requestPage')<Params['p
 export const receivePage = createStandardAction('Content/receivePage')<
   ArchivePage & {references: State['references']}
 >();
-export const receivePageNotFoundId = createStandardAction('Content/receivePageNotFoundId')<string>();
+export const receivePageNotFoundId = createStandardAction('Content/receivePageNotFoundId')<PageNotFoundParams>();
```
bethshook, line 32: Including book id in this payload so it can be used to check the config.

script/utils/update-redirects-data.ts
```diff
@@ -21,14 +32,18 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
+  const flatNewTree = booksAreDifferent
+    ? flattenArchiveTree(newBook.tree).filter((section) => section.id !== newBook.id)
+    : flattenArchiveTree(newBook.tree).filter((section) => section.id !== currentBook.id);
 
-  const formatSection = (section: LinkedArchiveTreeNode) => ({
-    bookId: currentBook.id,
-    pageId: section.id,
+  const formatSection = (section: LinkedArchiveTreeNode, newSection?: ArchiveTreeNode) => ({
+    bookId: booksAreDifferent ? newBook.id : currentBook.id,
+    pageId: newSection?.id ? stripIdVersion(newSection.id) : section.id,
```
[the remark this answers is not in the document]
TomWoodward: i do think we should still make this change, we just need to make matchRedirect only care about the pathname. maybe make a separate `formatRedirectSource` that can be shared between matchRedirect and `formatSection`. it wouldn't suck to rename `formatSection` to `formatRedirect` too

src/app/content/hooks/locationChange/index.ts
TomWoodward: the only potentially cleaner thing would be to switch your resolveContent approach to throwing an error instead of dispatching and nooping the reset of the logic. then you could catch it here and return

src/config.books.json
```diff
@@ -157,7 +158,8 @@
   },
   "afe4332a-c97f-4fc4-be27-4e4d384a32d8": {
     "defaultVersion": "18.2",
-    "archiveOverride": "/apps/archive/20210713.205645"
+    "archiveOverride": "/apps/archive/20210713.205645",
```
TomWoodward, line 161: is there any value in keeping the archiveOverride if the book is retired? it'll never be used right?

TomWoodward (changes requested):

src/app/content/hooks/receivePageNotFoundId.ts
bethshook: yeah, it's already been removed

src/app/content/hooks/receivePageNotFoundId.ts
```diff
@@ -16,6 +19,9 @@ export const receivePageNotFoundIdHookBody: ActionHookBody<typeof receivePageNot
       return;
     }
   }
+  if (bookConfig?.retired) {
+    return Promise.reject(new BookNotFoundError(`Could not resolve uuid: ${bookId}`));
```
TomWoodward, line 23: why do you need to do this here?

src/config.books.json
TomWoodward: no, if we're on dev or heroku we should be able to load any archive content, production has `UNLIMITED_CONTENT=false` meaning you can only see content that is specified in `config.books.json`

TomWoodward (changes requested):

src/app/content/hooks/receivePageNotFoundId.ts
bethshook: because we only know to 404 a page from an old book once we've checked it for redirects.

src/app/content/hooks/locationChange/index.ts
bethshook: @TomWoodward I added a new error type above because BookNotFoundError would trigger a 404

src/app/content/hooks/locationChange/resolveContent-in-development.spec.ts
```diff
@@ -279,6 +278,28 @@ describe('locationChange', () => {
       expect(referenceBook).toEqual(undefined);
     });
 
+    it('noops if book is retired', async() => {
+      BOOKS['13ac107a-f15f-49d2-97e8-60ab2e3abcde'] = { defaultVersion: '1.0', retired: true };
```
bethshook, line 282: @TomWoodward trying to modify the config but this does not seem to do anything

script/utils/update-redirects-data.ts
```diff
@@ -21,49 +32,66 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
-    pathname: decodeURI(
-      content.getUrl({ book: { slug: currentBook.slug }, page: { slug: section.slug } })
-    ),
-  });
+  const currentSections = flatCurrentTree.filter(archiveTreeSectionIsPage);
+  const flatNewTree = flattenArchiveTree(newBook.tree).filter((section) => section.id !== newBook.id);
 
-  const matchRedirect = (section: LinkedArchiveTreeNode) => isEqual(formatSection(section));
+  const matchSlug = (currentPageSlug: string) => flatNewTree.find((newPage) => newPage.slug === currentPageSlug);
+  const matchRedirect = (section: LinkedArchiveTreeNode | ArchiveTreeNode) => isEqual(formatRedirectSource(section));
```
TomWoodward, line 39: we can't use isEqual here anymore, because a whole redirect is being passed in, and we're comparing it to a string. you could do like
```suggestion
  const matchRedirect = (section: LinkedArchiveTreeNode | ArchiveTreeNode) => (redirect: {pathname: string}) => redirect.pathname === formatRedirectSource(section);
```

script/utils/update-redirects-data.ts
```diff
@@ -21,49 +32,66 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
+      // fall back to default book page
+      return (canonicalPageId && findArchiveTreeNodeById(newBook.tree, canonicalPageId))
+        || findDefaultBookPage(newBook);
+    }
+  };
 
   let countNewRedirections = 0;
-  for (const section of flatCurrentTree) {
+  for (const section of currentSections) {
     const newSection = flatNewTree.find(matchSection(section));
```
TomWoodward, line 83: could put this at the top of the `getRedirectTargetSection` function right?

src/app/content/hooks/locationChange/index.ts
TomWoodward: it won't if you catch it

TomWoodward (changes requested):

src/app/content/hooks/locationChange/index.ts
bethshook: I'm catching it on line 21 - I don't understand how to get around this: https://github.com/openstax/rex-web/blob/main/src/app/utils.ts#L53

src/app/content/constants.ts
```diff
@@ -9,6 +9,7 @@ export const maxResourcesPerFetch = 10;
 export const summaryPageSize = 20;
 export const loadMoreDistanceFromBottom = 50;
 
+export const messageQueryParameterName = 'message';
```
bethshook, line 12: @TomWoodward you deleted this before - did you mean to move it elsewhere?

TomWoodward: no i didn't mean to delete it

script/utils/update-redirects-data.ts
```diff
@@ -21,49 +31,66 @@ const updateRedirectsData = async(currentBook: BookWithOSWebData, newBook: BookW
...
+      return newSection;
     // remove `else` to enable legitimately removing pages from books
     // only once uuids are guaranteed to be consistent
-    } else if (!newSection && matchSlug(section.slug) === undefined && !matchException) {
+    } else if (!newSection && matchSlug(section.slug) === undefined && !isAllowedDeletion && !booksAreDifferent) {
       throw new Error(
         `updateRedirects prohibits removing pages from a book, `
         + `but neither section with ID ${section.id} nor slug ${section.slug} was found in book ${newBook.id}`);
+    // below condition is specific to retiring books
+    } else if (booksAreDifferent && !isAllowedDeletion && !redirects.find(matchRedirect(section))) {
```
TomWoodward, line 71: `isAllowedDeletion` is a function right? whats going on here?

TomWoodward (changes requested):

src/app/content/hooks/locationChange.spec.ts
```diff
@@ -168,6 +172,15 @@ describe('locationChange', () => {
     expect(helpers.osWebLoader.getBookIdFromSlug).not.toHaveBeenCalled();
   });
 
+  it('noops if no book is loaded', async() => {
+    const intlSpy = jest.spyOn(intlHook, 'default');
+    const contentHookSpy = jest.spyOn(contentHook, 'default');
+    contentHookSpy.mockReturnValue(Promise.resolve({book: undefined, page: undefined}));
+
+    await hook(payload);
+    expect(intlSpy).not.toHaveBeenCalled();
```
bethshook, line 181: @TomWoodward there is maybe a better way to do this?

TomWoodward (approved):

PR #1694, opened 2022-07-14
Retired notification

TomWoodward:
for: openstax/unified#1951


src/app/messages/es/messages.json
```diff
@@ -98,6 +98,9 @@
   "i18n:nav:profile:text": "Perfil de la cuenta",
   "i18n:notification:appmessage:dismiss": "Está bien",
   "i18n:notification:appmessage:header": "Notificación",
+  "i18n:notification:retired-book:body": "Intentaste acceder a una edición anterior de este recurso. Ha sido redirigido a la edición más reciente.",
```
bethshook, line 101: The first word should be `Intentó`. Otherwise looks fine.

Dantemss (approved):
Looks good to me

src/app/notifications/components/RetiredBookRedirect.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { htmlMessage } from '../../components/htmlMessage';
+import { Dispatch } from '../../types';
+import { dismissNotification, retiredBookRedirect } from '../actions';
+import { Body, Group, Header, P } from './Card';
+
+// tslint:disable-next-line:variable-name
+const Content = htmlMessage('i18n:notification:retired-book:body', P);
+
+// tslint:disable-next-line:variable-name
+const RetiredBookRedirect = ({dismiss, className}:
```
bethshook, line 15: Is there a reason not to use a shared component for all of the notifications, or at least for this and AcceptCookies? The only difference is the message name.

TomWoodward: Yea I mean it's pretty subjective, but I feel that unnecessary globbing of things together is just creating structure that you'll eventually have to do more work to dig your way out of.

Yea in this case the components are very similar, but it's also only like 5 lines. If anything I would instead work on more helpers to make those elements with text in them more terse like the html one.

The design terminology is "favor composition over inheritance" which basically says be real careful about forcing different things into the same logical flow. Sometimes you do that, but it's much safer to instead create helpers and stuff that allow your different implementations to be super terse and simple while maintaining your ability to mix and match.

In this case there is a difference in the click handler, and having separate components made it easy to do that, and in other notifications there might be some that don't have buttons or whatever and we have the flexibility to do that easily too.

bethshook: This is helpful, thanks.

PR #1704, opened 2022-07-25
Remove unused Tex, LaTex, Asciimath packages from MathJax

jpslav:


TomWoodward:
there is an error loading some follow up script that matjax is triggering. i think... i can't quire figure out the url format, it works if you specify `AM_HTMLorMML-full`

by searching https://cdnjs.com/libraries/mathjax/2.7.5, i don't see any obviously better options

TomWoodward:
these are necessary for exercises

PR #1712, opened 2022-08-10
fix algebra & trig canonical mapping

bethshook:
for: openstax/unified#1935

seems like `getCanonicalUrlParams` was doing what it was supposed to when it threw the loop error because it noticed some weird mapping for math books. I've updated the maps to reflect [the current spreadsheet](https://docs.google.com/spreadsheets/d/1Hj5vm2AbEiLgxgcbNRi550InzFTLvBOv16-Fw6P-c5E/edit#gid=0)


src/canonicalBookMap.ts
```diff
@@ -37,9 +37,6 @@ export const CANONICAL_MAP: CanonicalBookMap = {
   /* Algebra & Trigonometry */ '13ac107a-f15f-49d2-97e8-60ab2e3b519c' : [
     /* College Algebra 2e */ ['35d7cce2-48dd-4403-b6a5-e828cb5a17da', {}],
     /* Precalculus 2e */ ['f021395f-fd63-46cd-ab95-037c6f051730', {}],
-    /* College Algebra */ ['9b08c294-057f-4201-9f48-5d6ad992740d', {}],
-    /* Precalculus */ ['fd53eae1-fa23-47c7-bb1b-972349835c3c', {}],
-    /* Algebra & Trigonometry 2e */ ['eaefdaf1-bda0-4ada-a9fe-f1c065bfcc4e', {}],
```
TomWoodward, line 42: Don't we want this one?

TomWoodward (changes requested):

src/canonicalBookMap.ts
bethshook: hmm, yeah I suppose so. it's not on the spreadsheet though - do we just assume a new edition is a fallback?

TomWoodward: Yea I think so

bethshook: I realized the order of these maps needed to be reversed because the loop uses the last matching page found.

bethshook: for example on production [this page](https://openstax.org/books/college-algebra-corequisite-support/pages/preface) canonicals to itself, whereas on this PR branch it correctly points to the 2e page

bethshook: for example, without reversing the order [this page](https://openstax.org/books/algebra-and-trigonometry/pages/8-1-graphs-of-the-sine-and-cosine-functions) will canonical to alg & trig 2e, whereas once reordered it correctly points to precalc 2e

TomWoodward (approved):

PR #1714, opened 2022-08-16
add styles to suppress print only content

jivey:
for: openstax/unified#1998

### DIFF
--- a/generic-styles/content-baked.less
+++ b/generic-styles/content-baked.less
@@ -6,6 +6,14 @@
 @import "./content-common.less";
 @import "./_make-block.less";
 
+.visual-hide() {
+  position: absolute;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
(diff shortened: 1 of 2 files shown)


generic-styles/content-baked.less
```diff
@@ -641,3 +649,11 @@ figure {
...
   visibility: hidden;
 }
+
+@media screen {
+  [data-media~=screenreader] {
+    .visual-hide();
+  }
+
+  [data-media]:not([data-media~=screen],[data-media~=screenreader]) { display: none }
+}
```
bethshook, line 659: don't you also need the PDF styles here? otherwise the iframe is visible in print view.

jivey: glad you caught that, I totally missed that in the card. Updated!

bethshook: the card was not the clearest

bethshook (approved):

PR #1716, opened 2022-08-18
Content loading refactor

TomWoodward:
for: openstax/unified#2014
basically the idea here is to include with the book data the parameters under which the data was loaded, then when you load data _associated_ with the book, you're sure to use the same parameters. this cleans up any issues where the book config was async updated after the book content was loaded. 

there are a ton of gotchas in the process, like using an explicit version when loading content is still special even if you're explicitly specifying the same version as the config has. this is to maintain the same url format the user indicated and handle any cross-dependencies if you're testing specific content from archive.

i also removed the location state optimization, the book version was in there and it _isn't_ cleared on reload, even hard reload, so we've seen this in production where folks are browsing the book and suddenly the urls start having versions in them, because the config changed and the routing logic suddenly started thinking that they had used a version override. i did some jiggery in the prerender logic to keep passing the ids around   because i think that helps with the prerender speed.

in the new logic the behavior of the routing will not change once the data has been loaded, because its now a pure function of the resulting book data.

as part of this refactor, i also introduced a separation of the routing between the static routing matching the pre-rendered pages and the dynamic routing. this is to fix an issue where if you have a service worker, you can use the dynamic routing in production, but then if you hard reload (or share the url with someone else) you'll get a 404. in order to make all the urls consistently work, we'll separate them into two prefixes. for now this means the dynamic routing will no longer work on production, and thats fine because its never been explicitly supported there, but next we'll add a cloudfront function to redirect all the requests to the new url prefix to the SPA index and then the dynamic routing will work on production too.


bethshook:
There's a conflict. Think you just need to add the `!book.retired` filter to `findBooks`.

src/app/content/utils/archiveTreeUtils.ts
```diff
@@ -154,7 +156,18 @@ export const getTitleFromArchiveNode = (book: Book, node: ArchiveTree | ArchiveT
 
   if (extra) { extra.remove(); }
 
-  return container.innerHTML;
+  return container;
+};
+
+export const getTitleStringFromArchiveNode = (book: ArchiveBook, node: ArchiveTree | ArchiveTreeSection): string => {
```
bethshook, line 162: do we need this? I don't see it used anywhere

src/app/content/components/PrevNextBar.tsx
```diff
@@ -86,7 +86,7 @@ interface PropTypes {
 }
 
 // tslint:disable-next-line:variable-name
-const PrevNextBar = ({book, prevNext}: PropTypes) => {
+export const PrevNextBar = ({book, prevNext}: PropTypes) => {
```
bethshook, line 89: ?

script/update-archive-version.ts
```diff
@@ -50,51 +53,48 @@ async function updateArchiveAndContentVersions() {
...
-    osWebLoader
-  );
-
-  const getNewArchiveUrl = () => newArchiveUrl;
-
-  const newBookLoader = makeUnifiedBookLoader(
-    createArchiveLoader(getNewArchiveUrl, {
+    createArchiveLoader({
       archivePrefix: ARCHIVE_URL,
-      disablePerBookPinning: true,
```
bethshook, line 64: can you explain this? I feel like there was a case that led to us including it here.

bethshook (approved):

PR #1719, opened 2022-08-23
map spanish books

jivey:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1909


bethshook:
are you going to add the other maps to this PR or make separate ones?

jivey:
Didn't mean to actually open this yet, yeah I can add the rest here, probably easier to test all at once?

jivey:
@bethshook okay I went through the list and I think just the statistics books were missing, here's my understanding: Calculus volumes are already mapped, Precalculus is supposed to point to itself, and this PR updates the Statistics and Chem AF 2e mappings. Are there any other mappings I'm missing?

jivey:
@bethshook Though maybe something missing is that Calculus says "shares two chapters with volume two" so maybe that needs to be mapped, but it doesn't say which 🤔

bethshook:
Re Calculus, believe that should be working automatically, assuming IDs are the same. I think you're just missing Física Universitaria Volumen 1-3.

bethshook (approved):
I haven't checked the canonical links themselves but the maps look good

PR #1731, opened 2022-08-29
Added retries for EPROTO errors

Dantemss:
For: https://github.com/openstax/unified/issues/2021

An EPROTO error generally looks like this:

```
Unhandled Rejection at: Promise {
  <rejected> Error: write EPROTO
      at WriteWrap.onWriteComplete [as oncomplete] (internal/stream_base_commons.js:94:16) {
    errno: -71,
    code: 'EPROTO',
    syscall: 'write',
    '$metadata': { attempts: 1, totalRetryDelay: 0 }
  }
} reason: Error: write EPROTO
    at WriteWrap.onWriteComplete [as oncomplete] (internal/stream_base_commons.js:94:16) {
  errno: -71,
  code: 'EPROTO',
  syscall: 'write',
  '$metadata': { attempts: 1, totalRetryDelay: 0 }
}
```

`$metadata` is added by the AWS SDK for JS.
`attempts: 1` indicates the AWS clients are not retrying those errors, since they don't match anything in here: https://github.com/aws/aws-sdk-js-v3/blob/fa60bf375c0a142263e4a67a330a79862cfa4102/packages/middleware-retry/src/retryDecider.ts#L14
https://github.com/aws/aws-sdk-js-v3/blob/fa60bf375c0a142263e4a67a330a79862cfa4102/packages/service-error-classification/src/index.ts#L10-L21
So this PR implements our own retrier.

Don't worry about the branch name :)

### DIFF
--- a/script/prerender/fleet.ts
+++ b/script/prerender/fleet.ts
@@ -9,17 +9,24 @@
 import {
   CloudFormationClient,
   CreateStackCommand,
+  CreateStackCommandOutput,
   DeleteStackCommand,
+  DeleteStackCommandOutput,
   DescribeStacksCommand,
+  DescribeStacksCommandOutput,
   Output,
   waitUntilStackCreateComplete,
   waitUntilStackDeleteComplete,
 } from '@aws-sdk/client-cloudformation';
 import {
   GetQueueAttributesCommand,
+  GetQueueAttributesResult,
   ReceiveMessageCommand,
+  ReceiveMessageResult,
   SendMessageBatchCommand,
+  SendMessageBatchResult,
   SendMessageCommand,
+  SendMessageResult,
   SQSClient,
 } from '@aws-sdk/client-sqs';
 import { randomBytes } from 'crypto';
@@ -55,6 +62,9 @@ assertDefined(RELEASE_ID, 'REACT_APP_RELEASE_ID environment variable must be set
 // Increasing this too much can lead to connection issues and greater memory usage in the manager
 const MAX_CONCURRENT_BOOKS = 5;
 
+// Retry EPROTO errors in requests this many times
+const MAX_ATTEMPTS = 5;
+
 // The worker fleet is automatically terminated after this many seconds
 // This is insurance in case this process gets stuck or crashes without deleting the workers stack
 const PRERENDER_TIMEOUT_SECONDS = 1800;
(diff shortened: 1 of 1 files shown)


src/app/content/utils.ts
```diff
@@ -57,13 +58,14 @@ export function getContentPageReferences(book: ArchiveBook, page: ArchivePage) {
   return matches;
 }
 
-export const parseContents = (book: Book, contents: Array<ArchiveTree | ArchiveTreeNode>) => {
-  contents.map((subtree) => {
```
Dantemss, line 61: This `map` used to return the (modified) subtree, but the result was not used. So it must have worked by modifying the arguments in place during execution.

src/app/content/utils.ts
```diff
@@ -57,13 +58,14 @@ export function getContentPageReferences(book: ArchiveBook, page: ArchivePage) {
...
+export const parseContents = async(book: Book, contents: Array<ArchiveTree | ArchiveTreeNode>) => {
+  // we use 1 here to prevent the number of queued parseContents from growing exponentially
+  // if there are multiple levels in the tree
+  await asyncPool(1, contents, async(subtree: ArchiveTree | ArchiveTreeNode) => {
     subtree.title = getTitleFromArchiveNode(book, subtree);
     if (isArchiveTree(subtree)) {
-      subtree.contents = parseContents(book, subtree.contents);
```
Dantemss, line 64: `parseContents` just returns the `contents` argument, so once again that works by the argument being modified in place.

script/prerender/fleet.ts
```diff
@@ -98,7 +171,7 @@ async function createWorkersStack() {
 
   console.log(`Creating ${workersStackName} stack...`);
 
-  await cfnClient.send(new CreateStackCommand({
+  await sendWithRetries(cfnClient, new CreateStackCommand({
```
Dantemss, line 174: I tried using `callWithRetries` with `cfnClient.send.bind(cfnClient)` but typescript always got mad at some type deep within the AWS libraries, probably because `send` returns different types depending on the argument.

Dantemss, line 174: I tried using `callWithRetries` with `cfnClient.send.bind(cfnClient)` but that runs into some error deep within the AWS libraries, likely due to the fact that `send()` returns different types for different arguments. I could only solve it by making the overloaded `sendWithRetries` using the AWS-provided types.

Dantemss:
It may be a good idea to deploy this to the test pipeline first? Since there are no tests for prerendering.

Dantemss:
I ran this through the unified pipeline and prerendering seemed to still work fine

jivey (approved):

Dantemss:
Not verifiable in a review app because the review app doesn't pre-render... maybe @TomWoodward wants to take a look to make sure I didn't do anything horrible?

PR #1745, opened 2022-09-27
Pre-rendered baked styles

Dantemss:
For: https://github.com/openstax/unified/issues/2052
- Added resource loading methods and resource cache to archiveLoader
- dynamicStyles hook that caches style data
- Use cached style data in DynamicContentStyles component

Should also prevent styles from flashing (I think), since the pre-rendered content now has the correct styles embedded.

### DIFF
--- a/src/gateways/createArchiveLoader.ts
+++ b/src/gateways/createArchiveLoader.ts
@@ -6,7 +6,7 @@ import {
   VersionedArchiveBookWithConfig
 } from '../app/content/types';
 import { stripIdVersion } from '../app/content/utils';
-import { isAbsolutePath, isAbsoluteUrl } from '../app/content/utils/urlUtils';
+import { fromRelativeUrl, isAbsoluteUrl } from '../app/content/utils/urlUtils';
 import { ifUndefined } from '../app/fpUtils';
 import { ArchiveBookMissingError, BookNotFoundError, tuple } from '../app/utils';
 import { REACT_APP_ARCHIVE_URL_OVERRIDE } from '../config';
@@ -91,7 +91,7 @@ export default (options: Options = {}) => {
    * handle that
    *
    */
-  const getArchivePathAndVersion = (loadOptions: ArchiveLoadOptions, bookId?: string) =>
+  const getArchivePathAndVersion = (bookId: string, loadOptions: ArchiveLoadOptions) =>
     REACT_APP_ARCHIVE_URL_OVERRIDE
       ? loadOptions.archiveVersion
         ? tuple(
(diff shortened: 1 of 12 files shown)


Dantemss:
With this PR the prerendered content gets data-dynamic-style="true" but the styles also get embedded so they shouldn't flash... hopefully.

TomWoodward (changes requested):

src/gateways/createArchiveLoader.ts
```diff
@@ -192,11 +205,19 @@ export default (options: Options = {}) => {
     };
   };
 
-  const bookGetter = (bookId: string, loadOptions: ArchiveLoadOptions) => {
-    return bookInterface(
-      bookId,
-      loadOptions
-    );
+  const resourceGetter = (
```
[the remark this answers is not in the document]
Dantemss: Done!

src/gateways/createArchiveLoader.ts
```diff
@@ -114,33 +117,45 @@ export default (options: Options = {}) => {
   const contentUrl = (host: string, archivePath: string, ref: string) =>
     `${host}${archivePath}/contents/${ref}.json`;
 
+  // Allow loading absolute and relative paths as those could be passed to the content-styles param
+  const makeResourceUrl = (bookRef: string) => (
+    host: string, archivePath: string, resourceRef: string
+  ) => isAbsoluteUrl(resourceRef) ?
+         resourceRef :
+         /^\.{0,2}\//.test(resourceRef) ? // Absolute path or relative path starting with ./ or ../
```
TomWoodward, line 125: I'm not really liking the handling of the non resource version of the stylesheet here, but I see what you set it up this way, let me think about it for a little bit

TomWoodward (changes requested):

src/gateways/createArchiveLoader.ts
TomWoodward: @Dantemss can you move only the query handling part of it back to the react component? I really like the way all this is set up for the book data and book stylesheet, but the query is a special case and we don't care about it being prerendered or anything

Sorry about the rework on this

Dantemss: So you want the loader to only be called when not using the param, and I'm guessing just assume all urls passed to the loader are relative to the book url?

TomWoodward: Yes

And its ok for the query param to not be integrated into prerender and/or cause style flash, that is just a development tool

Dantemss: Alright, I restored the query param logic and the useEffect for that case.

TomWoodward (approved):

PR #1746, opened 2022-09-29
Assignable toolbar

jivey:
for: https://github.com/openstax/unified/issues/2061


src/app/content/styles/PopupStyles.tsx
```diff
@@ -11,15 +11,15 @@ import { BookWithOSWebData } from '../types';
...
+  topBottomMargin
 } from './PopupConstants';
 
-export const desktopPopupWidth = 74.4;
-export const popupPadding = 3.2;
-export const popupBodyPadding = 2.4;
-export const headerHeight = 7.2;
-export const topBottomMargin = headerHeight + popupBodyPadding;
-export const popupHeaderZIndex = 5;
+export * from './PopupConstants';
```
jivey, line 22: These two files are changed to prevent a circular import

src/app/content/components/AssignedTopBar.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import styled, { css } from 'styled-components';
+import theme from '../../theme';
+import { setTextSize } from '../actions';
+import * as selectContent from '../selectors';
+import { LinkedArchiveTreeSection } from '../types';
+import { TextResizer } from './Topbar/TextResizer';
+
+// tslint:disable-next-line:variable-name
+const TopBar = styled.div`
```
TomWoodward, line 11: would it be better to re-use the existing bar? isn't the design the same? we could at least use the `Styled.TopBarWrapper` component

TomWoodward (changes requested):

src/app/content/components/AssignedTopBar.tsx
jivey: Yeah good call to use TopBarWrapper here, updated. I didn't want to use SearchInputWrapper/MobileSearchWrapper because of how much is specific to the complex layout on the main app, and they do a lot that I'm hoping we refactor at some point (maybe that's what we do now, extract these styles into a generic TopBar component?)

TomWoodward: Yea idk, I agree on avoiding all the complicated logic we don't need about the search results but if seems like there should be a shared component just for the border style and size

src/app/content/components/Topbar/styled.tsx
```diff
@@ -357,7 +357,10 @@ export const TextResizerDropdown = styled(FilterDropdown)`
     }
   }
 
-  ${theme.breakpoints.mobileMedium(css`
+  ${(props: {
+    mobileVariant: boolean,
+    mobileToolbarOpen: boolean
+  }) => props.mobileVariant !== false && theme.breakpoints.mobileMedium(css`
     margin-left: 0;
```
TomWoodward, line 364: What's this about?

src/app/content/components/Topbar/styled.tsx
```diff
@@ -53,7 +53,7 @@ const closeIconStyles = css`
   color: #cdcdcd;
 `;
 
-const shadow = css`
+export const shadow = css`
```
TomWoodward, line 56: How is Shadow currently applied?

src/app/content/components/AssignedTopBar.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { LinkedArchiveTreeSection } from '../types';
+import { shadow, TopBarWrapper } from './Topbar/styled';
+import { TextResizer } from './Topbar/TextResizer';
+
+// tslint:disable-next-line:variable-name
+const StyledTopBarWrapper = styled(TopBarWrapper)`
+  ${shadow}
+
+  && {
+    top: 0;
```
TomWoodward, line 16: Did Victoria specify if this should be sticky?

src/app/content/components/AssignedTopBar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  background: #fff;
```
TomWoodward, line 19: Isn't this white in the regular view too?

TomWoodward (changes requested):

src/app/content/components/Topbar/styled.tsx
jivey: There are mobile styles that get applied because of the main toolbar collapses a bit on mobile: 
image
That ends up looking broken on this new toolbar because it doesn't collapse... maybe we want that? I don't see any mobile designs but I can make the styles more flexible.

src/app/content/components/Topbar/styled.tsx
jivey: SearchPrintWrapper and MobileSearchContainer use `shadow` for the `box-shadow` styles, and those wrappers really contain most of the styles that make the toolbar work, so this change dry-s it up and replaces my hard-coded `box-shadow`. Maybe we want this to be a styled wrapper with the single style rule instead?

src/app/content/components/AssignedTopBar.tsx
jivey: I thought it would match the current toolbar, @victoria-cooke thoughts on if the assignable toolbar should be sticky or inline?

src/app/content/components/AssignedTopBar.tsx
jivey: It is, updated to use the same theme constant for white.

src/app/content/components/AssignedTopBar.tsx
jivey: Oops just saw this one after I made the other comments... This makes sense to me as well. It looks simple enough to make a wrapper component for the shadow/border/sticky/background styles, I'll spend a few extracting those.

TomWoodward (approved):

Dantemss (approved):

PR #1749, opened 2022-10-03
filter outlook errors

TomWoodward:
for: openstax/unified#2075

i waited a long time for a better solution to this problem and i guess its just not coming. the best we can do is this very specific ignore rule, so that we don't also ignore correctly broken things like these:
image


Dantemss (approved):
Yeah I guess the Object Not Found part is what makes it different, since it comes from deep within some .NET binding library: https://github.com/cefsharp/CefSharp/blob/2ead42338062ec60c9bfc8564b88fb4307dc7bc3/CefSharp/Internals/JavascriptObjectRepository.cs#L298

(for future reference)

PR #1751, opened 2022-10-03
clean up some test output

TomWoodward:
for: openstax/unified#2079


Dantemss (approved):
I'm not entirely sure the acts are needed because these promises don't modify state but sure, the code should work

TomWoodward:
The awaits are clearing react effects that modify the state

PR #1752, opened 2022-10-04
Fix page specs

TomWoodward:
for: openstax/unified#2081

these specs were using the wrong act, the right act wasn't available in async until a later version, and the mocks were getting a bit polluted


Dantemss (approved):

Dantemss:
There are a few failing tests on unrelated files, maybe due to the package updates?

package.json
```diff
@@ -24,8 +24,8 @@
     "prop-types": "^15.7.2",
     "query-string": "^6.2.0",
     "rangy": "^1.3.0",
-    "react": "^16.11.0",
-    "react-dom": "16.8.6",
+    "react": "^16",
+    "react-dom": "^16",
```
TomWoodward, line 28: needed the newest react-dom for their `act` bugfix, needed to keep react in sync, needed to kep @types/react in sync, ended up causing a few issues downstream

i have no idea what part of that caused CRA to start modifying the jsx handling config in tsconfig.json but the `DISABLE_NEW_JSX_TRANSFORM` is necessary to stop it. the new behavior only works on react 17

PR #1755, opened 2022-10-06
Attempt to fix dynamic style hydration

Dantemss:
For: https://github.com/openstax/unified/issues/2052


Dantemss:
Testing on https://release-2038.sandbox.openstax.org/books/organizational-behavior/pages/references, it seems to work. The styles stay loaded even when you refresh.

src/app/components/DynamicContentStyles.tsx
```diff
@@ -9,8 +11,12 @@ import { AppServices } from '../types';
 import { assertDefined } from '../utils/assertions';
 
 // tslint:disable-next-line: variable-name
-export const WithStyles = styled.div`
-  ${(props: { styles: string }) => props.styles}
+export const ScopedGlobalStyle = createGlobalStyle`
```
TomWoodward, line 14: if you're using the styled global style, can you use the styled css processor here instead of postcss?

TomWoodward (changes requested):

src/app/components/DynamicContentStyles.tsx
Dantemss: good point

jivey (approved):
Looks good to me but not sure if @TomWoodward has some feedback

TomWoodward (approved):

PR #1762, opened 2022-10-20
update info script to support book subject

TomWoodward:
for: openstax/unified#2385


jivey (changes requested):
I think the specs are complaining because this is missing in some of the OSWeb mocks, otherwise looks good

jivey (approved):

jivey:
Hmm that failure is interesting, does the prerender fetch from osweb?

PR #1763, opened 2022-10-20
Upgrade to gtag

jivey:
For: https://github.com/openstax/unified/issues/2107


Dantemss (approved):
Code looks fine to me. There was some issue with a test though. And we'll probably want someone who uses GA to verify that they are getting what they need.

jivey:
One thing that could be a blocker with this is that custom dimensions aren't sending to 4, though they are sending to 3. Based on the docs this doesn't seem possible but that's the behavior I'm seeing when inspecting the request.

TomWoodward (approved):
This looks good, but I want to check out the ga data before merging

PR #1764, opened 2022-10-21
adding interacted event

TomWoodward:
for: openstax/unified#1969

here is the first pass over the collected data

image

had to go to college-physics to find an example problem, not sure if we have them in bio

image


src/helpers/analytics/utils.ts
```diff
@@ -1,4 +1,88 @@
...
+        state = config.getState && target ? config.getState(target) as typeof state : state;
+        const stateChange = target ? config.getStateChange(target, state) : undefined;
+
+        if (reportedTarget) {
+          handler(reportedTarget, stateChange);
+        }
+      });
+    }
+  }
+};
```
nathanstitt, line 88: seeing the passed in `window` makes me wonder if this is called when pages are pre-rendered?  Is it so the `window` instance can be stubbed in that case?

src/helpers/analytics/utils.ts
```diff
@@ -1,4 +1,88 @@
...
+  getState?: (element: T) => S,
+  getStateChange: (element: T, state: S) => string | undefined;
+  getTarget?: (element: T, state: S) => HTMLElement | undefined,
+  match: (element: any) => element is T;
+  state?: S;
+};
+
+// stupid thing to make TS infer the genric type of each config record
+const makeInteractableConfig = <T, S = undefined>(config: InteractableConfig<T, S>) => config;
+
```
nathanstitt, line 27: I tried a few ideas to get rid of the empty function call, but didn't come up with anything better.  Mainly I tried to explicitly type the `interactableElementEvents` array but that got unwieldy.

nathanstitt (approved):
LGTM Tom, your use of TS continues to be illuminating 😁

reedstrm (approved):
Hey, did you manage to solve the PhET sym's eating the events?

reedstrm (commented):
Hmm, are the failed tests concerning?

src/helpers/analytics/utils.ts
TomWoodward: Yea it does suck, but it makes me feel better knowing webpack-terser will remove the empty function call

src/helpers/analytics/utils.ts
TomWoodward: It's kinda for testing but mostly actually laziness because of the prerendering rex changes the global window type to be possibly undefined, and sometimes passing a reference around reduces the nullcheck logic

TomWoodward:
> Hey, did you manage to solve the PhET sym's eating the events?

I did not, I'm ignoring the problem for now

> Hmm, are the failed tests concerning?

Yea the selenium one fails sometimes but the build tests are concerning, I'll have to fix those

jivey (approved):
Looks good, interesting spec failure but I saw your comment

PR #1768, opened 2022-11-09
Attempt to make Heroku use node 14

Dantemss:
For: https://github.com/openstax/unified/issues/2126

I considered Node 16 but we should probably do that when we are ready to switch `.nvmrc` as well.

Which maybe we could just do right now, I think Node 16 just works for Rex.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "rex-web",
   "version": "0.1.0",
   "private": true,
+  "engines": {
+    "node": "14.x"
+  },
   "dependencies": {
     "@craco/craco": "^6.1.2",
     "@formatjs/intl-pluralrules": "^4.0.6",


Dantemss:
I unchecked the card so we can delay merging until the current release is locked (the code here doesn't affect the build but merging would cause a new search build which takes a while).

Dantemss:
Nvm, commented on the wrong PR.

Dantemss:
Heroku deployment seems to have worked, and is running Node 14.

jivey (approved):
For what it's worth, I've been using 16 with rex for a while and haven't seen issues. One other reason to consider it sooner too is that 2023-04-30 is end of life for 14 (https://github.com/nodejs/release#release-schedule).

PR #1769, opened 2022-11-10
fix issue where like 400 script tags are added to pre-rendered pages

TomWoodward:
for: openstax/unified#2127

the main issue was the missing chunk names causing all those data and locale chunks to be numbered, there is still some unknown issue where the numbered chunks have to be preloaded or the hydration fails, now there are just less numbered chunks

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -25,12 +25,6 @@ module.exports = {
                   }
                 }
 
-                webpackConfig.plugins.push(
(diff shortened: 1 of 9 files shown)


jivey (approved):
I saw briefly saw a hiccup on heroku for proxying manifest.json but that may have been unrelated, and seemed to test fine afterwards.

PR #1770, opened 2022-11-14
fix footer cutting off in some breakpoints

jivey:
for: https://github.com/openstax/unified/issues/2015

### DIFF
--- a/src/app/content/components/CenteredContentRow.tsx
+++ b/src/app/content/components/CenteredContentRow.tsx
@@ -1,12 +1,26 @@
 import React from 'react';
+import { connect } from 'react-redux';
 import styled, { css } from 'styled-components/macro';
 import theme from '../../theme';
+import { AppState } from '../../types';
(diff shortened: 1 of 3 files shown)


src/app/content/components/CenteredContentRow.tsx
```diff
@@ -1,37 +1,69 @@
...
+    isDesktopSearchOpen={isDesktopSearchOpen}
+  >
     {children}
   </Wrapper>;
 };
 
-export default CenteredContentRow;
+export default connect(
+  (state: AppState) => ({
+    isDesktopSearchOpen: !!searchSelectors.query(state),
```
Dantemss, line 65: Shouldn't the right hand side use `searchSelectors.searchResultsOpen(state)` instead?

jivey:
Closing in favor of https://github.com/openstax/rex-web/pull/1782 as I found out an almost upsettingly simpler solution...

PR #1771, opened 2022-11-15
Update scripts

TomWoodward:
for: openstax/unified#2138
i'm working to update our projects that need local ssl to using a single shared CA so we don't have to trust them all separately.

the script/make-certificate.bash script is currently duplicated in all the projects, idk where a better place to put it would be but i'll figure that out later

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ node_modules
 
 # production
 /build
+/data/certs
 
 # misc
 .DS_Store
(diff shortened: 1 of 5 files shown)


jivey (commented):
After running start, I still see the cert untrusted. Same thing when running `trust-localhost` afterwards. I've skipped the cert for months so I could be forgetting a step here...

jivey (approved):
I can't really test if this is working (still shows as untrusted to me) but the change itself seems good

PR #1772, opened 2022-11-16
Ga page view false

jivey:


src/environment.development.json
```diff
@@ -2,6 +2,10 @@
   "release_id": "development",
   "messages": [],
   "configs": {
+    "google_analytics": [
+      "UA-73668038-11",
+      "G-0D68920QV9"
+    ],
```
Dantemss, line 8: I was going to say to put a comment here so we remember to remove it later but it's a json file :/

Dantemss (approved):
Code looks good to me but I did not check if the GA views are as they should be, as I have not requested access to GA yet.

TomWoodward:
@jivey i think all these changes are included in your other PR, can we close this one?

PR #1777, opened 2022-11-25
remove flex display style from stepwise content

jivey:
for: https://github.com/openstax/unified/issues/486

There are two changes here [to solve a problem with stepwise list items](https://github.com/openstax/cnx-recipes/issues/4939). A quick overview:

1. **Stepwise list items**. Remove flex from these to prevent the overlap render cases below. This seems like a safe change; when looking through the [Sifter](https://openstax.github.io/sifter/#v=1&q=.stepwise&sourceFormat=xhtml) results, I didn't see anything that required the flex style. I could be missing some context for that rule though.

| Before | After |
---|---
| image | image |
| image | image |

2. **Inline media**. The media inline change is related to the first, when removing the flex, the stepwise media content drops to the next line, and non-stepwise media had some existing alignment issues, so I'm trying to resolve both of these with the media inline rule. 

| Stepwise Before | Stepwise After |
---|---
| image | image |

| Non-Stepwise Before | Non-Stepwise After |
---|---
| image | image |


TomWoodward:
don't forget to check the baked styles to see if this bug exists there too

jivey:
related pr: https://github.com/openstax/ce-styles/pull/148

Dantemss (approved):
No idea if anything else requires flex, but ce-styles approved it so I see no reason not to approve it.

PR #1778, opened 2022-11-25
don't draw focus when selecting text while unauthed

jivey:
for: https://github.com/openstax/unified/issues/1178


src/app/content/highlights/components/Confirmation.tsx
```diff
@@ -53,7 +51,7 @@ const Confirmation = React.forwardRef<HTMLElement, Props>((
   ref
 ) => {
   return <Overlay
-    ref={mergeRefs(ref, useDrawFocus())}
+    ref={ref}
```
TomWoodward, line 54: maybe we could test this, but i think when deleting a highlight we still want this behavior. maybe you could add a drawFocus prop to the Confirmation component and do
```suggestion
    ref={drawFocus ? mergeRefs(ref, useDrawFocus()) : ref}
```

if the keyboard tab order naturally falls into the confirmation window when deleting a highlight, then it doesn't matter and we can leave it as you have it here

TomWoodward (changes requested):

src/app/content/highlights/components/Confirmation.tsx
jivey: In testing, the tab order did fall into order, but without the focus call, the menu overlapped it until you hit tab, so I think the change you proposed is a good one, updated the PR

src/app/content/highlights/components/Confirmation.tsx
```diff
@@ -45,15 +45,16 @@ interface Props {
...
 
 // tslint:disable-next-line:variable-name
 const Confirmation = React.forwardRef<HTMLElement, Props>((
-  {message, confirmMessage, confirmLink, always, onCancel, onConfirm, ...props}: Props,
+  {message, confirmMessage, confirmLink, always, onCancel, onConfirm, drawFocus = true, ...props}: Props,
   ref
 ) => {
   return <Overlay
-    ref={mergeRefs(ref, useDrawFocus())}
+    ref={mergeRefs(drawFocus ? ref : null, useDrawFocus())}
```
TomWoodward, line 57: This feels weird... Shouldn't it be always using the passed in ref and sometimes using the drawfocus ref? This is doing it backwards?

Also is drawfocus a hook? It might complain about being conditional, you could extract it to a variable and conditionally compose the variable (I think that'd work anyway)

TomWoodward (changes requested):

src/app/content/highlights/components/Confirmation.tsx
jivey: Hmm yeah maybe I flipped this around and didn't notice in testing, I've simplified it as you suggested

TomWoodward (approved):

PR #1779, opened 2022-11-27
Fix some analytics issues

jivey:
for: https://github.com/openstax/unified/issues/2132
Fixes a timing bug with analytics firing before the page title has been updated. (based off #1772)


src/app/content/hooks/receiveContent.ts
```diff
@@ -66,4 +66,8 @@ const hookBody: ActionHookBody<typeof receivePage | typeof locationChange> = (se
   dispatch(setHead({links, meta, title}));
 };
 
+const hookBody: ActionHookBody<typeof receivePage | typeof locationChange> = (services) => {
+  return async() => updateHead(services);
+};
+
```
jivey, line 72: I'm not certain if there is a better way to accomplish this, I needed to extract the body due to the types expecting the action to be passed, but I need to call it directly in locationChange without having the page info

TomWoodward: where you're calling this here https://github.com/openstax/rex-web/pull/1779/files#diff-5c521d5b56bf1fc8410db4d851fabef38392a7ee950c8d6d314d7a317d8d5336R25 you have the locationChange action passed in on line 17

TomWoodward (changes requested):
looks good, i think you can undo that change to the hook body, but other than that everything looks fine

src/app/content/hooks/receiveContent.ts
jivey: Thank you I totally missed that :(

jivey:
@TomWoodward Re-requesting one more review after discovering an existing analytic bug with highlight card: https://github.com/openstax/rex-web/pull/1779/commits/d169577d1e0cb0a939e4f5633a4d06d2bdbbc703

For testing what was happening, I did this:

1. Open a page and start to make a highlight
2. Confirm the analytic event fires for "show create" or "show login" depending on auth status (using the debug extension or network logs)
3. Navigate to another page via TOC
4. See the analytic event fires again without doing anything

I think all that is needed here is the isActive check but this also seems like a good target for a future refactor, it feels like it maybe should skip the entire Card render unless it's active 🤔

TomWoodward (approved):
Looks fine but you have to remove that dev config

Dantemss (approved):
Tom already reviewed but it really wants me to review too so I'll approve :)

PR #1782, opened 2022-12-01
fix footer cutting off in some breakpoints

jivey:
for: https://github.com/openstax/unified/issues/2015

`clamp` won't be able to fix it for Safari 12 and under but it doesn't make it worse. Ideally I think a refactor of the nav and content body CSS would be good to not need this workaround, it was a kinda strange bug to fix, and clamp here seems the simplest solution. The `calc` calculations were copied from `ContentPane` instead of importing them due to bug (something with the import?) I couldn't figure out: https://github.com/openstax/rex-web/actions/runs/3596314873/jobs/6056840427#step:6:1368


src/app/theme.ts
```diff
@@ -107,13 +107,20 @@ const color = {
...
 const mobileBreak = 75; // 1200px
+const desktopBreak = mobileBreak + .0625; // 1201px
 const mobileSmallQuery = `(max-width: ${mobileSmallBreak}em)`;
 const mobileMediumQuery = `(max-width: ${mobileMediumBreak}em)`;
 const mobileQuery = `(max-width: ${mobileBreak}em)`;
 const touchDeviceQuery = `not all and (pointer: fine), (hover: none)`;
 
 export default {
   breakpoints: {
+    desktop: (style: FlattenSimpleInterpolation) => css`
```
jivey, line 118: May not be used in this PR but feels right to bundle for future use.

jivey:
Hmm actually this needs more work... it seems the highlight cards get misaligned this way.
image

Dantemss (approved):
For the `calc` issue: from the error, I'm guessing global style components don't work without a window, but the test is specifically testing that this thing works outside of the browser so it makes sense to avoid global styles... I think

PR #1783, opened 2022-12-02
Added a modal context that provides a function to focus the modal

Dantemss:
For: https://github.com/openstax/unified/issues/1933
Fixes ESC after removing a filter from the MH modal or editing/canceling the edit of a highlight or deleting/canceling the delete of a highlight
Fixes ESC after removing a filter from the SG modal or closing the "Using this Guide" submodal


jivey (approved):
Nice use of context here, my initial thought was to pass a ref all over the place and this is much nicer. Codecov is complaining but everything else looks good

PR #1794, opened 2022-12-06
Add canonical maps for the old econ editions and the new macro/micro

Dantemss:
For: https://github.com/openstax/unified/issues/2100
I added maps for macro/micro 3e to econ 3e but 2e didn't have that so now I'm wondering...


TomWoodward (commented):
its not necessarily a problem, but its a little confusing that some of the econ books are in these separate files and others are listed directly in https://github.com/openstax/rex-web/blob/dabb223e0adcee93bd1978e9744f7a4b5ecfa262/src/canonicalBookMap.ts

Dantemss:
Maybe this is causing the test failure? Because now that I look at it, those books being in that other file means I added some duplicates. I'll remove/move them and see if things improve.

Dantemss:
OK I moved them and replaced a couple of the ones I had added since they'll chain anyway.

Dantemss:
Checked micro 3e, micro 2e and econ 2e on Heroku and they all link to econ 3e.

TomWoodward (approved):

PR #1798, opened 2022-12-08
Activity integration config

TomWoodward:
for: openstax/unified#2173

i sorta need the icons, been sitting on a branch for a while now


jivey (approved):
I think this is only used in assignable so nothing to really test yeah?

TomWoodward:
> I think this is only used in assignable so nothing to really test yeah?

Right

PR #1799, opened 2022-12-10
FE Playwright test: Make multiple highlights

Malar-Natarajan:
Make 2 highlights in a page on random paragraph

for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/2171


playwright/src/utilities/utilities.ts
```diff
@@ -28,9 +28,15 @@ function randomChoice(list: ElementHandle[]): ElementHandle {
   return list[option]
 }
 
-function randomNumber(count: number) {
-  const randomnum = Math.floor(Math.random() * count)
-  return randomnum
+function randomNum(count: number, excludenum?: number) {
```
Malar-Natarajan, line 31: At some later point I will work on making the `excludenum` parameter an array so that more than 2 highlights can be made on a page

playwright/tests/rex-test/rex.behaviorspec.ts
```diff
@@ -90,3 +90,34 @@ test('signup and highlight', async ({ page, isMobile }) => {
...
+  let highlightcount = await BookPage.highlightCount()
+  expect(highlightcount).toBe(1)
+
+  // AND: Highlight another random paragraph
+  await BookPage.scrolltotop()
+  const randomparanumber2 = randomNum(await paracount, randomparanumber)
+  await BookPage.highlightText('yellow', randomparanumber2)
+
+  // THEN: Text is highlighted
+  highlightcount = await BookPage.highlightCount()
```
Malar-Natarajan, line 121: `highlightcount` works only if we create an instance again (first one created in line 112). Is that how it works?

gregfitch: Yes, that would be expected.

gregfitch (approved):

PR #1802, opened 2022-12-16
Auth Tokens

Dantemss:
For: https://github.com/openstax/unified/issues/2074
- Added browserAuthProvider to Rex
- Use tokens when accessing Highlights

There's some behavior that's different between the browerAuthProvider and what was here before. I left some comments pointing out the differences. We can choose to leave it or try to reconcile the behaviors.


src/gateways/createUserLoader.spec.ts
```diff
@@ -28,36 +28,31 @@ describe('userLoader', () => {
         (global as any).fetch = mockFetch(200, testAccountsUser);
 
         const user = await userLoader.getCurrentUser();
-        expect(fetch).toHaveBeenCalledWith('url/api/user', {
+        expect(fetch).toHaveBeenCalledWith('url/accounts/api/user', {
```
Dantemss, line 31: Different behavior here (browserAuthProvider always appends `/accounts`) so we'll have to remove `/accounts` from Rex urls to avoid `/accounts/accounts` (or change browserAuthProvider to, for example, not append `/accounts` if the url already ends in `/accounts`

src/gateways/createUserLoader.spec.ts
```diff
@@ -28,36 +28,31 @@ describe('userLoader', () => {
...
           credentials: 'include',
         });
         expect(user).toEqual(undefined);
       });
     });
 
-    it('throws for unexpected errors', async() => {
+    it('gets undefined for unexpected errors', async() => {
```
Dantemss, line 48: Different behavior here: browserAuthProvider returns undefined for all non-200 status codes, not just 403.

src/typings/ts-utils.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@openstax/ts-utils/dist/services/authProvider/browser';
```
Dantemss, line 1: I think this type and the path having `dist` are required because Rex uses Node 14? It wouldn't work without these.

src/gateways/createUserLoader.spec.ts
TomWoodward: i think the rex behavior of throwing for other statuses is probably more correct, we could update the browserProvider to match

src/gateways/createHighlightClient.ts
```diff
@@ -8,19 +9,23 @@ const formatError = (response: Response) => {
     });
 };
 
-export default (url: string) => {
+export default (
+  basePath: string,
+  getAuthorizedFetchConfig: () => Promise<FetchConfig> = () => Promise.resolve({credentials: 'include'})
```
TomWoodward, line 14: does this default value ever get used?

src/gateways/createHighlightClient.ts
```diff
@@ -8,19 +9,23 @@ const formatError = (response: Response) => {
...
-          return formatError(response);
-        } else if (response.status === 401) {
-          return Promise.reject(new UnauthenticatedError());
-        }
-        return Promise.resolve(response);
-      }),
+    basePath,
+    fetchApi: async(url: string, fetchConfig: FetchConfig) => {
+      const authorizedFetchConfig = await getAuthorizedFetchConfig();
+      const response = await fetch(url, {...fetchConfig, ...authorizedFetchConfig});
```
TomWoodward, line 20: this'll prevent the given `fetchConfig` from specifying headers sometimes, you may want to use one of the deep merge functions either from lodash or ts-utils

src/gateways/createUserLoader.spec.ts
TomWoodward: this is fine... i think... but don't you have to update some configs because of this?

TomWoodward (changes requested):

src/gateways/createUserLoader.spec.ts
Dantemss: Yeah, I'll go hunt down the configs

src/gateways/createUserLoader.spec.ts
Dantemss: Cool, I'll update the provider

src/gateways/createHighlightClient.ts
Dantemss: In tests, I think. Maybe prerendering. I could probably just update them all though.

Dantemss: All those places are fine with the default being `() => Promise.resolve({})`. I think it's a fine default for prerendering etc but let me know if you want it taken out.

TomWoodward (changes requested):
@Dantemss please look at the commit i added and lmk if it looks ok to you

TomWoodward:
@Dantemss don't forget about wrapping this up

src/gateways/createUserLoader.ts
```diff
@@ -7,21 +8,19 @@ export interface AccountsUser {
...
+  is_administrator: boolean;
   is_not_gdpr_location: boolean;
+  contact_infos: any[];
 }
 
 export default (url: string) => {
+  const authProvider = browserAuthProvider({window: assertWindow()})({auth: {accountsUrl: url}});
+
+  // Note: previously getCurrentUser would return undefined for 403 and reject non-200 statuses,
+  //       but the browserAuthProvider always returns undefined for non-200 statuses instead
```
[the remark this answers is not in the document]
Dantemss: Yeah it's in https://github.com/openstax/typescript-service-template/pull/32
I'll remove the comment

Dantemss:
The commit you added looks OK. I think this ready for another look.

Dantemss:
Except there's some yarn.lock conflict I gotta solve now

Dantemss:
OK ready now

TomWoodward (approved):

PR #1803, opened 2022-12-21
Async/defer scripts

Dantemss:
For: https://github.com/openstax/unified/issues/1964

- Made googleoptimize async
- Made craco-generated script tags deferred

Googleoptimize has instructions suggesting it works with async.
react-scripts 5.0.0+ also defaults to "defer", but upgrading that requires updating to node v14.17.0 and something breaks along the way. As far as I could tell, Rex seems to work fine with defer.


Dantemss:
Pulse Insights was already async even before this PR

Dantemss:
I thought I was issues in the dev home page on Heroku and a bunch of console warnings in a book, but maybe that happened because it was still being deployed, they seem to be gone now...

TomWoodward (changes requested):
you may also need to add here 
https://github.com/openstax/rex-web/blob/main/script/prerender/contentPages.tsx#L247

script/prerender/contentPages.tsx
```diff
@@ -268,7 +268,7 @@ function injectHTML(html: string, {body, styles, state, fonts, meta, links, modu
     `<script>window.__PRELOADED_STATE__ = ${JSON.stringify(state).replace(/</g, '\\u003c')}</script>`
   );
 
-  html = html.replace(/(<script src="[^"]+\.chunk\.js"><\/script>)/, scripts.join('') + '$1');
+  html = html.replace(/(<script defer="defer" src="[^"]+\.chunk\.js"><\/script>)/, scripts.join('') + '$1');
```
Dantemss, line 271: The htmlWebpackPluginInstance config in the craco config renders the `defer="defer"` but seems like that is standard or something, though the inconsistency of the boolean attributes here bothers me a little.

Dantemss:
Ah right no prerendering on Heroku to test that part...

jivey (approved):
Not sure if @TomWoodward had some more feedback, but this seems good to me. I wasn't able to test locally but can take a close look when it's on staging.

TomWoodward (approved):

PR #1836, opened 2023-01-12
default to loading baked styles unless set to false

jivey:
for: https://github.com/openstax/unified/issues/2154

Flips the default to fetch styles if the url is there, unless explicitly set to false.


jivey:
tagging @TomWoodward on this because we were discussing recently and I want to make sure I didn't misunderstand

src/app/content/hooks/dynamicStyles.ts
```diff
@@ -10,7 +10,7 @@ const hookBody: ActionHookBody<typeof receiveBook> = (services) => async() => {
   const loadingBook = loadingBookSelector(state);
 
   if (loadingBook || !book || !book.style_href ||
-      !book.loadOptions.booksConfig.books[book.id]?.dynamicStyles) {
+      book.loadOptions.booksConfig.books[book.id]?.dynamicStyles === false) {
```
Dantemss, line 13: There is a potential behavior change here if the book config is completely missing but book and book.style_href are present somehow, but I guess that fits with the intent of the PR.

Just a little surprising it does not complain about coverage due to the `?` here. I'm guessing there's an unrelated spec somewhere that covers this case but doesn't check if the dynamic styles are loaded.

Dantemss (approved):

jivey:
@Dantemss I just updated the book config to turn them off for everything but Org Behavior and Principles of Marketing, looks like we'll expand to all books in the future

Dantemss (approved):

Malar-Natarajan (approved):

PR #1844, opened 2023-01-24
remove obsolete content parsing functions

bethshook:
for: openstax/unified#2212

these markup changes are now handled on the recipes side


TomWoodward (approved):
what about the extra html stripping regex we talked about? that removes the stuff around the body from the content?

PR #1861, opened 2023-01-25
trigger highlights modal via query

bethshook:
for: openstax/unified#1722


jivey (approved):
Looks good to me. I did notice one thing that's probably super edge case: like production, the search query still wins out over the modal if they are both in the url, but now it flashes the highlights modal briefly (example url `https://localhost:3001/books/american-government-2e/pages/2-4-the-ratification-of-the-constitution?modal=MH&query=test&target=%7B%22type%22%3A%22search%22%2C%22index%22%3A0%7D`) I wonder if the modal should be prioritized over the search query 🤔

bethshook:
@jivey so that is relatively easy to fix because we just need to keep the existing modal query in the navigation options in the `receiveSearchHook`. but once that is working, if you load a URL with both modal and search in the query and then you close the modal, some wacky things happen because of the `closeModalHook`. I've added my attempt at a fix there - would be great if @TomWoodward could take a look too.

src/app/content/search/hooks/index.ts
```diff
@@ -77,6 +78,7 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
     ? {
       hash: selectedResult.result.source.elementId,
       search: queryString.stringify({
+        modal: persistentQueryParams.modal,
```
jivey, line 81: @bethshook Fix working for me too, though I think you also want it for the other branch for when the query doesn't have results

PR #1880, opened 2023-01-27
FE playwright test: Make edits in MH page

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/2223

1. Update book library with search results changes
2. Navigate to next page
3. Click on context menu in MH modal


playwright/tests/rex-test/rex.behaviorspec.ts
```diff
@@ -103,21 +103,55 @@ test('multiple highlight', async ({ page, isMobile }) => {
...
+  expect(MHhighlightcount).toBe(4)
+
+
+  // let x = await Edithighlight.highlightlist()
+  // console.log(x[1])
+  // const highlight = x[1].toString().replace(/\D/g, "")
+  // console.log(highlight)
+
+  await Edithighlight.clickContextMenu(1)
+  sleep(5)
```
Malar-Natarajan, line 149: @gregfitch Please review lines 137 - 148. Is this the right way to open the context menu of a highlight?
So the number we are passing as a parameter in 148 is what determines the highlight in the modal.

gregfitch: I may be reading the structure incorrectly (MHmodal.ts), but it looks like you're doing something different than I would expect. Your comment lists the class EditHighlights as a highlight, but it's extending the whole modal; why can't this live in one of the other classes?

gregfitch: As to the code, yes, this should get the nth element as long as you adjust `n` to be one less (1st element is n=0); clickContextMenu(1) should open the second highlight toggle menu.

Malar-Natarajan: @gregfitch can you please review again. I removed the toolbar class

gregfitch (approved):

PR #1881, opened 2023-01-31
Fix ordering of book maps and fix Alg & Trig canonical bug

bethshook:
for: openstax/unified#2009

- allows us to list a book's canonical maps in order of priority
- fixes a bug that prevented Alg & Trig 2e from being listed as a canonical for Alg & Trig
- removes the error thrown when a loop is discovered in the mappings, which no longer seemed necessary or helpful now that canonicals are listed in order of priority


bethshook:
@jivey the easiest way to test is by exporting book csv's from the dev homepage

jivey (commented):
At first was confused about the ordering but after looking through the code I think I get it. `/books/algebra-and-trigonometry/pages/6-1-exponential-functions` ends up with `/books/precalculus-2e/pages/4-1-exponential-functions` because it looks as deep as possible with in the ordering, right? I guess I'm trying to better understand the original bug.

bethshook:
@jivey I think the original bug was that because books were ordered from last-to-first in terms of priority, alg & trig would first check alg & trig 2e which canonicals to college alg 2e; finding no page match there it would check precalc 2e, which also canonicals to college alg 2e. so the loop error was going off when there wasn't actually a loop. so a) that error was not distinguishing between loops and expected redundancies in the maps and b) it didn't really make sense to check the books in that order because once we find a match in the preferred book we should just use it and stop traversing.

jivey (approved):

bethshook:
@jivey reopening this after addressing some regressions in math books. I think the function is a bit easier to follow now and infinite loops are prevented by stopping as soon as we hit a book we've already checked.

based on the script I'm using to compare canonicals before and after (I can push that branch if you want to try it) the only differences between this and what's on prod are in Algebra & Trig, where some chapters will now correctly point to Algebra & Trig 2e instead of self ([example](https://rex-web-issue-2009-pt4q3z6nyy5.herokuapp.com/books/algebra-and-trigonometry/pages/7-introduction-to-the-unit-circle-sine-and-cosine-functions)), and Algebra & Trig 2e, where some chapters will now correctly point to College Algebra 2e instead of Precalc 2e ([example](https://rex-web-issue-2009-pt4q3z6nyy5.herokuapp.com/books/algebra-and-trigonometry-2e/pages/6-introduction-to-exponential-and-logarithmic-functions)).

jivey:
@bethshook Makes sense, retesting my earlier comment now points to college algebra 2e as well. Side note, that script sounds really helpful!

jivey (approved):

PR #1887, opened 2023-02-03
Ignore page view event if tag id is GA4

jivey:
for: https://github.com/openstax/unified/issues/2225

Probably don't want to release this until https://github.com/openstax/unified/issues/2226 is figured out.


src/gateways/googleAnalyticsClient.ts
```diff
@@ -196,18 +213,20 @@ class GoogleAnalyticsClient {
       return;
     }
 
-    for (const tagId of this.tagIds) {
+    for (const tag of this.tags) {
       command.payload = {...command.payload, queue_time: queueTime };
+
       if (command.name === 'event') {
-        this.gtag('event', command.eventName, { ...command.payload, send_to: tagId });
+        if (tag.ignoredEventNames.includes(command.eventName)) { return; }
```
Dantemss, line 220: I think this should be `continue` instead?

Dantemss (changes requested):

src/gateways/googleAnalyticsClient.ts
Dantemss: But then maybe you need some code to abort if everything got skipped

Dantemss: Or maybe no need to check for the abort condition, since there's nothing after the loop.

jivey: Whoops good catch, that was definitely a typo and should have been `continue`

src/gateways/googleAnalyticsClient.ts
```diff
@@ -196,18 +213,20 @@ class GoogleAnalyticsClient {
       return;
     }
 
-    for (const tagId of this.tagIds) {
+    for (const tag of this.tags) {
       command.payload = {...command.payload, queue_time: queueTime };
+
       if (command.name === 'event') {
-        this.gtag('event', command.eventName, { ...command.payload, send_to: tagId });
+        if (tag.ignoredEventNames.includes(command.eventName)) { continue; }
```
TomWoodward, line 220: you could invert the condition and put the gtag call inside it, that feels a bit nicer than using the continue to short circuit

TomWoodward (changes requested):

TomWoodward:
> Probably don't want to release this until https://github.com/openstax/unified/issues/2226 is figured out.

i don't think we have to wait, ga4 reporting won't be significantly more broken than it is already

TomWoodward (approved):

Dantemss (approved):

jivey:
@Malar-Natarajan I've also temporarily added ga3 and ga4 testing ids to this PR. For testing, we want to make sure page_view events get sent to ga3 but not ga4.

jivey:
@Malar-Natarajan I forgot to also mention that in the GA4 debug view, it's possible that the automatic, enhanced tracking pageview event will still be there, which we expect. You can distinguish our custom pageview event, which should not be there, by it having a `queue_time` param:

Malar-Natarajan:
> in the GA4 debug view, it's possible that the automatic, enhanced tracking pageview event will still be there, which we expect. You can dis

ok, so this is why there are 2 page views

PR #1889, opened 2023-02-09
change gtm container

TomWoodward:
for: openstax/unified#2238


jivey (approved):

TomWoodward:
@jivey i'm pretty sure this is working now, can you take another look?

jivey (approved):
Seems to work for all normal methods of loading (I occasionally saw the missing title when initially navigating from the dev home page, not sure why.) Nice approach with the race as well 👍

PR #1891, opened 2023-02-10
scroll to top of assignment section

bethshook:
for: openstax/unified#2090

@TomWoodward  I think what's happening here is there is no locationChange clearing the page before the new one is requested, so when the scroll manager checks if the page has changed it doesn't see it right away. It seems like it will be ok to just send the page to the top on load, since the scroll manager would still scroll to any scroll target that it finds (is this even relevant in assignable?), but let me know if I'm missing something.


src/app/content/components/Assigned.tsx
```diff
@@ -36,6 +37,10 @@ const useLoadSection = (currentSection: ArchiveTreeSection | undefined) => {
     if (!book || !currentSection) {
       return;
     }
+
+    const window = assertWindow();
+    window.scrollTo(0, 0);
```
TomWoodward, line 42: can we try clearing the `content.page` state when `requestPage` (with a different page than is in state already) is dispatched and see if that helps?

TomWoodward (changes requested):

bethshook:
@TomWoodward your suggestion seems to work and is passing tests but I'm nervous there's going to be some side effect I haven't identified

src/app/content/reducer.ts
```diff
@@ -88,7 +88,7 @@ function reduceContent(state: State, action: AnyAction) {
       return reduceReceiveBook(state, action);
     }
     case getType(actions.requestPage):
-      return {...state, loading: {...state.loading, page: action.payload}};
+      return {...omit(['page'], state), loading: {...state.loading, page: action.payload}};
```
TomWoodward, line 91: similar to the way this is done down on line 118, i think we only want to omit the page if the requested page is different from the current one

TomWoodward (changes requested):

src/app/content/reducer.ts
```diff
@@ -88,7 +88,10 @@ function reduceContent(state: State, action: AnyAction) {
       return reduceReceiveBook(state, action);
     }
     case getType(actions.requestPage):
-      return {...state, loading: {...state.loading, page: action.payload}};
+      return {
+      ...(isEqual(action.payload, state.page) ? state : omit(['page'], state)),
```
TomWoodward, line 92: the condition you have here will always be false, i think maybe you wanted to do 
```suggestion
      ...(isEqual(action.payload, state.params.page) ? state : omit(['page'], state)),
```
but i'm realizing that won't work either, because on a location change, the state.params will have just been updated to be the new ones.

i apologize apparently the logic of how all this works has fallen out of my head, digging back around again i see that the case of navigating to another location on the same page is short circuited [here](https://github.com/openstax/rex-web/blob/main/src/app/content/hooks/locationChange/resolveContent.ts#L159), and in normal rex behavior `requestPage` is only ever called when we _really_ want to request the page. in the assignable layout we also know for sure that we want to request the page

so long story short i think you were right the first time, you can remove the extra conditional logic. let me know if this is making sense to you, its possible i'm still confused...

TomWoodward (changes requested):

src/app/content/reducer.ts
bethshook: ugh actually what I meant to do was check the page params against each other - something like `isEqual(action.payload.uuid, state.page.id)` but I just realized the payload could be of type SlugParams. also when loadPage is called from resolveContent, there is already a check to ensure a different page is being loaded. so I think you're right that the conditional is not necessary.

TomWoodward (approved):

TomWoodward:
@bethshook i just noticed this PR is missing some code coverage, could you look into that?

PR #1841, opened 2023-01-18
update PQ button and remove search query from (most) rendered hrefs

bethshook:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/1895

PR #1892, opened 2023-02-10
restore behavior of search result opened in new tab

bethshook:
for: openstax/unified#2222

reverting a change from [this pr](https://github.com/openstax/rex-web/pull/1841/files) so that any queryParams passed as a prop will be included in the ContentLink href


jivey (approved):
Link behavior look good to me

PR #1895, opened 2023-03-01
send visibility change events to event-capture

TomWoodward:
for: openstax/unified#2265

### DIFF
--- a/src/helpers/analytics/bindEvents.ts
+++ b/src/helpers/analytics/bindEvents.ts
@@ -1,3 +1,4 @@
+import debounce from 'lodash/debounce';
 import { AppState } from '../../app/types';
 import { captureEvent } from '../../gateways/eventCaptureClient';
 import googleAnalyticsClient from '../../gateways/googleAnalyticsClient';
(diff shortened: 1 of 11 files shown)


reedstrm (approved):
Looks good, but I only see focus and unfocus events - and their hard coded as logical opposites. What about initial page view, and page destruction? Was hoping to get those with a "none, focus" and "focus, none" for previous, current state.

reedstrm (changes requested):
As discussed, let's incorporate visibility as well as focus, and get initialization and teardown in there as well.  We can keep the state_type as visibility.

src/helpers/analytics/events/unload.ts
```diff
@@ -13,6 +14,10 @@ export const track = (
   {pathname}: ReturnType<typeof selector>
 ): AnalyticsEvent | void => {
   return {
+    getEventCapturePayload: () => stateChange({
+      current: 'unload', // TODO - the type only allows string, should it allow null?
```
TomWoodward, line 18: @reedstrm we wanted to represent the unloaded state as a null right? the types coming out of swagger are set to only string values -- is that on your end or a swagger-codegen issue?

src/helpers/analytics/bindEvents.ts
```diff
@@ -70,7 +80,12 @@ export const events = {
   openNudgeStudyTools: mapEventType(openNudgeStudyTools),
   openStudyGuides: mapEventType(openStudyGuides),
   openUTG: mapEventType(openUTG),
-  pageFocus: mapEventType(pageFocus),
+  // debounce prevents common case of focusout immediately followed by move to background
+  // note that `unload` also triggers the visibility state change but it must be executed
+  // immediately without a debounce for the beacon to work. i don't think it will, but if
+  // this results in events representing unload followed by focusout we may have to
+  // re-visit this.
```
TomWoodward, line 87: > i don't think it will, but if this results in events representing unload followed by focusout we may have to re-visit this.

this is annoying to test in dev because the console closes before unload is fired, i'd like to look at the reported data from the heroku environment and make sure everything looks ok

src/helpers/analytics/events/unload.ts
reedstrm: Probably me, I think to allow it to be null, I'd need to  make it optional. But the semantics are actually that at least one-of-two (current or previous) is required. How about we use a string-nonce to indicate "not existing" - How about "none" since it's not null (which to me means not-known, could be anything - i.e. SQL NULL) but in fact means "there is no previous state" or "there is no current value for this state"

TomWoodward: ts differentiates between a null or undefined value and having the key not be present in the hash, but in any case using a value of `none` works for me

reedstrm (approved):

src/helpers/analytics/events/pageFocus.ts
```diff
@@ -1,19 +1,28 @@
...
+import { AppState } from '../../../app/types';
 import { AnalyticsEvent } from './event';
 
-// const eventName = 'REX page focus';
-
-export const selector = createSelector(
-  selectNavigation.pathname,
-  (pathname) => ({pathname})
-);
+export const selector = (_state: AppState) => ({});
```
jivey, line 6: I'm probably overlooking something obvious but do we need to keep this and the unused param in track?

jivey (approved):

src/helpers/analytics/events/pageFocus.ts
TomWoodward: yea it needs to match the signature for the other events

PR #1901, opened 2023-03-16
Be quiet about third party cookie errors

RoyEJohnson:
for: openstax/unified#2289
https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/2289


src/index.tsx
```diff
@@ -119,23 +119,39 @@ startMathJax();
...
+    window.localStorage.setItem('securityTest', '1');
+    window.localStorage.removeItem('securityTest');
+  } catch (e) {
+    lsDisabled = true;
+  }
+
+  return inAnIframe && lsDisabled;
+}
+
+if (!securityIsTooTight) {
```
Dantemss, line 136: Missing `()` here to call the method

src/index.tsx
```diff
@@ -119,23 +119,39 @@ startMathJax();
...
+  // Learn more about service workers: http://bit.ly/CRA-PWA
+  serviceWorker.register()
```
Dantemss, line 138: Do the service works in general not work if localStorage is disabled? I thought it was just going to be the text resizing that was affected

Dantemss (changes requested):

src/index.tsx
RoyEJohnson: Service workers won't initialize if two things are true: running in an iframe _and_ localstorage is not working. Or at least that's how I read the issue.

src/index.tsx
RoyEJohnson: D'oh!

src/index.tsx
TomWoodward: i'm not sure if cookies specifically is the problem, but the service worker registration is getting a permission denied message, i think its happening when third party cookies are disabled

package.json
```diff
@@ -172,8 +172,8 @@
     "serve-static": "^1.13.2",
     "sitemap": "^3.2.2",
     "source-map-explorer": "^2.5.2",
-    "stylelint": "^10.1.0",
-    "stylelint-config-standard": "^18.3.0",
+    "stylelint": "^13.13",
+    "stylelint-config-standard": "^22.0.0",
```
TomWoodward, line 176: the stylelint change doesn't seem like it has to be in this PR, why did you need to update it? you can make a separate PR or roll it into your big ol updates PR if you want

src/app/content/hooks/storeTextSize.ts
```diff
@@ -38,7 +38,11 @@ const hookBody: ActionHookBody<typeof setTextSize> = (services) => async() => {
   const value = textSize(state);
 
   if (typeof window !== 'undefined' && value !== null) {
-    window.localStorage.setItem(textResizerStorageKey, value.toString());
+    try {
```
TomWoodward, line 41: pretty sure you have to do this up on line 22 as well

src/index.tsx
TomWoodward: reading the code -- i would try to recognize the security error in the catch before reporting to sentry rather than assume that the security issue is related to localStorage (which may be true but also might be coincidental)

TomWoodward (changes requested):

TomWoodward:
remember to update tests to cover those new logic branches. for the ones in the text size hook it may be helpful to make a helper that obfuscates the branch

src/app/content/hooks/storeTextSize.ts
RoyEJohnson: I wasn't able to get an error from line 22 in Firefox or Safari, but I put the try/catch around it just in case. The setItem call was causing the page to blow up when localStorage was disabled, and now fonts change normally.

src/app/content/hooks/storeTextSize.ts
```diff
@@ -19,7 +19,11 @@ export const loadStoredTextSize = (services: MiddlewareAPI & AppServices) => asy
   }
 
   if (typeof window !== 'undefined' && window.localStorage) {
-    storedTextSize = window.localStorage.getItem(textResizerStorageKey);
+    try {
+      storedTextSize = window.localStorage.getItem(textResizerStorageKey);
+    } catch {
+      // They have blocked access to localStorage; ignore it
+    }
```
TomWoodward, line 26: trying to parse this function in my head, it looks like what its gonna do now is reset your text size back to the default every time you navigate, is that the case?

src/index.tsx
```diff
@@ -137,5 +151,7 @@ serviceWorker.register()
     }
   })
   .catch((e) => {
-    Sentry.captureException(e);
+    if (!securityIsTooTight()) {
```
TomWoodward, line 154: i was thinking more like actually look at the exception and determine if it is the access denied error. doing a little testing in chrome i see that the exception doesn't look super standardized, i get a `DOMException` with message `"Failed to register a ServiceWorker for scope ('https://rex-web.herokuapp.com/books/') with script ('https://rex-web.herokuapp.com/books/service-worker.js'): The user denied permission to use Service Worker."`, so we'd probably have to do a check like `if (e instanceof DOMException && e.message.includes('user denied permission to use Service Worker'))`, which is a little sloppy but does what it says on its hat and is pretty straightforward. we'd need to test firefox/edge/safari to see if their error messages are similar though 

if we need to have a blocklist of exact error messages, like separate ones for each browser, it may become easier to put them [here](https://github.com/openstax/rex-web/blob/main/src/helpers/Sentry.ts#L34) so they don't clutter up the sentry code

TomWoodward (changes requested):

src/app/content/hooks/storeTextSize.ts
RoyEJohnson: Yes, that is what will happen.

TomWoodward: oh no i see now there is a short-circuit on line 17 that will noop if the state has a value, so we should be ok right? (i haven't tested, does it work or no?)

TomWoodward: why doesn't it load if you block all cookies?

TomWoodward: suppressing the error that causes the `"Uh-oh, there's been a glitch"` dialog is the major acceptance criteria for this issue

RoyEJohnson: I was able to suppress that (you can try it with the recent deployment). The book still doesn't load, though, because the failing line is [here](https://github.com/openstax/rex-web/blob/main/src/app/utils.ts#L43). I put a check to bypass `catchError` (which does the glitch dialog).

TomWoodward: thats just where its being re-thrown. the error is coming from the redux middleware somewhere, and then the error handler sends to sentry here https://github.com/openstax/rex-web/blob/main/src/app/utils.ts#L70, once the error is recorded in sentry it shows the "uh oh" dialog. you need to make the error not get as far as sentry.

what you're looking at here doesn't really indicate where the error is coming from, the error's stacktrace should have that, maybe if you go there you can handle the error where it happens.

alternatively you could add some more generic error handling in this error handler that you're linking to (the handling is a few lines down in the actual error handler function), but that would be for handling any instance of `SecurityError` no matter where it comes from, and we might not want to be that aggressive with suppressing the errors

RoyEJohnson: The stack trace takes the error to [here](https://github.com/openstax/rex-web/blob/main/src/app/utils.ts#L37) from [here](https://github.com/openstax/rex-web/blob/main/src/app/navigation/middleware.ts#L17). I can alter `catchError` or intercede before calling `catchError`. Either way, the dialog is prevented, but the book content doesn't get loaded.

src/app/content/hooks/storeTextSize.ts
```diff
@@ -18,8 +18,10 @@ export const loadStoredTextSize = (services: MiddlewareAPI & AppServices) => asy
     return;
   }
 
-  if (typeof window !== 'undefined' && window.localStorage) {
-    storedTextSize = window.localStorage.getItem(textResizerStorageKey);
+  try {
+    storedTextSize = window?.localStorage.getItem(textResizerStorageKey);
```
TomWoodward, line 22: pretty sure the `window?.` will throw in prerendering, you may need to keep that first part of the condition using `typeof window !== 'undefined'`

src/index.tsx
```diff
@@ -141,6 +141,8 @@ serviceWorker.register()
       });
     }
   })
-  .catch((e) => {
-    Sentry.captureException(e);
+  .catch((e: Error) => {
+    if (e) {
```
TomWoodward, line 145: whats this all about? is there a case where its passing something falsey?

TomWoodward (changes requested):

src/app/content/hooks/storeTextSize.ts
RoyEJohnson: If it throws, it will be caught, though, right? For that matter, I could leave the `?` out.
I ran the prerender locally, and it seemed ok to me, but it's possible I don't know what to look for.

src/index.tsx
RoyEJohnson: Yes, on initial load, I get a throw where the error is undefined. Maybe that's just a local thing? Doesn't seem to happen on the Heroku version. I can take it back out.

TomWoodward (approved):

TomWoodward:
does the service worker registration error also need to be ignored? or does it not matter?

src/index.tsx
RoyEJohnson: The service worker error doesn't happen locally because SW is disabled outsided production. The SecurityError does happen on the Heroku app. I have put the ignore block back in.

TomWoodward (changes requested):

src/index.tsx
```diff
@@ -142,5 +142,9 @@ serviceWorker.register()
     }
   })
   .catch((e) => {
-    Sentry.captureException(e);
+    if (e as Error instanceof DOMException && e.name === 'SecurityError') {
```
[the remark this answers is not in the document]
RoyEJohnson: Firefox: SecurityError
Chrome: NotSupportedError
Edge: NotSupportedError
Safari: does not trigger the catch block

src/index.tsx
```diff
@@ -124,6 +124,10 @@ startMathJax();
 // load optimize
 loadOptimize(window, app.store);
 
+function cookiesBlocked(e: DOMException) {
+  return ['SecurityError', 'NotSupportedError'].includes(e.name);
```
TomWoodward, line 128: you're accessing e.name here assuming e is a DOMException and its not complaining because the e coming into the catch on line 148 is `any`, but you observed some errors coming into there that were not subclass of `Error`, i think you saw `undefined` coming in ? anyway you probably want some additional checks here to prevent errors accessing props of nonobjects or whatever the error is.

if you're gonna leave this function with the `DOMException` type, maybe change the condition below like

```diff
-    if (cookiesBlocked(e)) {
+    if (e instanceof DOMException  && cookiesBlocked(e)) {
```

or you could change this function here to accept `unknown` or something for that type and do the `instanceof` check here

TomWoodward (changes requested):
wrong button on the last one

src/index.tsx
RoyEJohnson: Yes, it had caught `undefined` when running locally. I have it check for `Error` first (so it won't try to report undefined), and for `DOMException` in `cookiesBlocked`
The serviceWorker still reports an error in the console, but it is not an uncaught error.

TomWoodward (approved):
ok this is looking good now. thank you for working through all those revisions roy

Dantemss (approved):

PR #1911, opened 2023-04-10
Change styling of MathJax so it doesn't overflow

RoyEJohnson:
for: https://github.com/openstax/unified/issues/1887


jivey (approved):
Looks good, just needs the issue link added to the pr description`(for: https://github.com/openstax/unified/issues/1887`)

TomWoodward (changes requested):
i think we definitely still want the content to scroll if necessary, the overflow should only be changed on the actual math elements right?

RoyEJohnson:
I don't think we want individual questions to scroll, nor individual answer options. The questions seem to be constructed so that they have enough room, but the browsers are adding some junk whitespace, making them a little larger than expected.
I can change visibility to `visible`, but for the question, I need to also change it on the surrounding div.

TomWoodward:
The question may contain figures and tables and stuff, we definitely want it to be "auto" so it scrolls when necessary, the answers I'm not so sure about

RoyEJohnson:
The question height is dynamic. It displays whatever it contains. It's just that on some browsers, that height isn't calculated quite right: 
`height: calc(100% - 7.2rem)`
I can add lots of junk tot he question and it doesn't mind. It just makes room for it and pushes the answer block down, and the dialog scrolls.
image

TomWoodward:
the reason for the `overflow: auto` is mostly for content that is too wide, but i also wouldn't want to preclude ourselves from putting a max height on the question area. i think that it would be better to fix this on the mathjax elements rather than modifying our layout

could you try adding 
```css
.MathJax {
  display: inline-block;
  overflow: hidden;
  vertical-align: middle;
}
```

and see if that causes any issues? i see that getting the math to align at baseline is hard to do with `inline-block`, but i don't think it looks bad aligned at middle

RoyEJohnson:
It actually doesn't need `vertical-align` adjusted. It does if you're fiddling with it in the browser console, but it's properly baseline-aligned when it loads normally.

TomWoodward (approved):
looks good roy, thanks for humoring me on the revisions there

RoyEJohnson:
I have reverted to making the overflow visible on question and answer content. If actual content is larger than available space, containers scroll reasonably: the window scrolls vertically, while individual answers are horizontally-scrollable. (Images in questions automatically scale to fit).
image

TomWoodward (changes requested):
would the stuff have been moving around if we had the `vertical-align: middle;` part in from before?

RoyEJohnson:
I thought I checked that, but I looks steady now. We can do it that way if you want. The scrolling behavior when questions/answers are large is the same either way.
image

RoyEJohnson:
Question 4 with the overflow change vs with the inline-block change
image
image

TomWoodward:
The bottom screenshot is inline-block? It definitely looks kinda weird

TomWoodward:
What is different about the math in the pq dialog vs the main content? Why does this issue not come up there?

RoyEJohnson:
I have not been able to find what is actually changing when the question is the activeElement (MathJax aligns properly) vs when it is not (non-MathJax shifts down). If a MathJax item is the activeElement, it shifts down with the rest of the text.

RoyEJohnson:
If I change a MathJax item in the main text to `display: inline-block; overflow: hidden`, it also pushes its line down.
image

RoyEJohnson:
If, instead of `.MathJax`, I apply the styling to `.math-rendered`, the line-shifting doesn't happen. It still requires a magic negative margin-bottom (-5px is generally pretty good, though it's maybe a pixel off for taller glyphs like theta and b-vector.
I've created a [draft PR](https://github.com/openstax/rex-web/pull/1924) with this change if you want to try it out when it deploys to Heroku.

RoyEJohnson:
I just noticed that when it goes from desktop to mobile, the scrollbar goes away on the question. It turns out the difference is padding. Mobile has top and bottom padding, while desktop sets it to 0.
While there is no breakpoint change for the answers, I do find that adding padding to the AnswerAlignment object makes the scrollbars go away there.

TomWoodward (changes requested):
this update seems to be working VERY well, thank you for finding a less invasive version of the fix. i am noticing that the answer alignment is a little off now though

image

RoyEJohnson:
Ok, _this_ looks like the minimal change set. No scrollbars, questions stay at the baseline.

TomWoodward:
ok this is the second PR to start complaining about lighthouse, either we broke something and merged it or maybe the library changed

RoyEJohnson:
Closed PRs don't have errors on Test (build) phase, so I guess the library changed.

TomWoodward (approved):

PR #1919, opened 2023-04-21
Use default navigation for links in Assigned, and open them in new tabs

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2354
Two issues at play:
1. When in Assigned, all links should have `target="_blank"`
2. The navigation handler does not honor that target


RoyEJohnson:
I have bypassed `contentLinkHandler` when in Assigned. Equivalently, could bypass it if `target` is `_blank` (though this would apply outside Assigned).
Might need to push the decision-making into `contentLinkHandler`, depending on how it might handle non-figure links.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -130,9 +139,17 @@ export default class PageComponent extends Component<PagePropTypes> {
 
   public render() {
     const pageIsReady = this.props.page && this.props.textSize !== null;
+    const PT = this.props.inAssigned ?
```
TomWoodward, line 142: this fix is for a different card, and i think we need to work on how to communicate the nav height a bit more, can you move it to a different pr?

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -47,6 +48,14 @@ export default class PageComponent extends Component<PagePropTypes> {
 
       transformContent(parsedContent, parsedContent.body, this.props.intl);
 
+      if (this.props.inAssigned) {
+        const links = parsedContent.body.querySelectorAll('a[href]');
```
TomWoodward, line 52: it shouldn't be all links, but all links to another page. a link to an anchor on the same page is ok.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -195,6 +212,9 @@ export default class PageComponent extends Component<PagePropTypes> {
 
     lazyResources.addScrollHandler();
 
+    if (this.props.inAssigned) {
```
TomWoodward, line 215: instead of short circuiting on `inAssigned`, you should check for the target attribute in the click handler. but actually i just tested and it seems like links with `target="blank"` on them work as expected already

src/app/content/components/Assigned.tsx
```diff
@@ -90,7 +90,7 @@ export default () => {
     <ErrorModal />
     <ErrorBoundary>
       <AssignedTopBar section={section} />
-      <Page>
+      <Page inAssigned={true} >
```
TomWoodward, line 93: i would pass props for what we actually want to do here, rather than leaking context and forcing more logic into the page component. maybe `<Page lockNavigation>` to indicate that we don't want to allow navigating to different pages in this view.

TomWoodward (changes requested):

src/app/content/components/Page/PageComponent.tsx
TomWoodward: well thats not good

src/app/content/components/Page/PageComponent.tsx
RoyEJohnson: Should checking for a target attribute be added to [here](https://github.com/openstax/rex-web/blob/main/src/app/content/components/Page/contentLinkHandler.ts#L121)?

TomWoodward: Handling of links with target blank seemed to work already when I tested it, are you having problems with it?

RoyEJohnson: I've been testing it by clicking the figure 1 link. When I added the `target="_blank"` attribute, it still opened in the same tab, navigating to the Rex page.
I see that the "Simplify an expression" link properly opens in another tab, but that aborts [here](https://github.com/openstax/rex-web/blob/main/src/app/content/components/Page/contentLinkHandler.ts#L134) whereas the figure links have a `reference` and so don't abort there.

TomWoodward: Ok so yea I think you were right about handling the target attribute here https://github.com/openstax/rex-web/pull/1919#discussion_r1176853470


But I think there is also a different bug here having to do with figure links on the same page, we can address that in a separate change if you want

RoyEJohnson: I've put in a fix for both of these issues. Clicking a figure link works, but awkwardly: it looks like the component re-mounts, so it scrolls to the top of the page and re-displays toasts, then scrolls down to the figure, which winds up partly obscured by the TopBar.

RoyEJohnson: Oh, but it broke a test, so maybe not a correct fix.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -47,6 +47,19 @@ export default class PageComponent extends Component<PagePropTypes> {
 
       transformContent(parsedContent, parsedContent.body, this.props.intl);
 
+      if (this.props.lockNavigation) {
```
TomWoodward, line 50: i see why this needs to be here, because in `transformContent` you don't have access to the props to perform the mutation based on the `lockNavigation` flag, but can you move the content of this if statement into a function in https://github.com/openstax/rex-web/blob/ecb62adf6d2bad690953b5b31a3665d5f5058e23/src/app/content/components/Page/contentDOMTransformations.ts and just call it from here? that way it reduces visible complexity in this file

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -47,6 +47,19 @@ export default class PageComponent extends Component<PagePropTypes> {
...
+        const links = parsedContent.body.querySelectorAll('a[href]');
+
+        for (const link of links) {
+          const {origin, pathname} = new URL(link.getAttribute('href'), assertWindow().location.href);
+          const loc = assertWindow().location;
```
TomWoodward, line 55: this code needs to work in pre-rendering, so you can't use window.location, you need to use the navigation state path. you already have access to it at `props.currentPath`

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -47,6 +47,19 @@ export default class PageComponent extends Component<PagePropTypes> {
...
+
+          if (origin !== loc.origin || pathname !== loc.pathname) {
```
TomWoodward, line 57: there is logic elsewhere (in recipes) that does (or is supposed to) add target blank to any link outside the content, so you don't have to handle that case here. (i think thats what you're looking at the origin for)

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -47,6 +47,19 @@ export default class PageComponent extends Component<PagePropTypes> {
 
       transformContent(parsedContent, parsedContent.body, this.props.intl);
 
+      if (this.props.lockNavigation) {
+        const links = parsedContent.body.querySelectorAll('a[href]');
+
+        for (const link of links) {
+          const {origin, pathname} = new URL(link.getAttribute('href'), assertWindow().location.href);
```
TomWoodward, line 54: i think you can use this utility here https://github.com/openstax/rex-web/blob/ecb62adf6d2bad690953b5b31a3665d5f5058e23/src/app/content/utils/urlUtils.ts#L104
```suggestion
          const pathname = fromRelativeUrl(link.getAttribute('href'), this.props.currentPath);
```

TomWoodward (changes requested):

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -37,6 +38,20 @@ export default class PageComponent extends Component<PagePropTypes> {
   private processing: Array<Promise<void>> = [];
   private componentDidUpdateCounter = 0;
 
+  private linksToOtherPagesOpenInNewTab(rootEl: HTMLElement) {
```
TomWoodward, line 41: ok you moved it to a function, but i would really like to get it out of this file, this file has a lot of complexity already. can you please move to the [contentDOMTransformations.ts](https://github.com/openstax/rex-web/blob/ecb62adf6d2bad690953b5b31a3665d5f5058e23/src/app/content/components/Page/contentDOMTransformations.ts) file? you will still need to call it directly because of the extra parameters (unless you feel like adding additional props to the `transformContent` function), but even just moving the function would be nice

TomWoodward (changes requested):

TomWoodward (approved):

TomWoodward:
@RoyEJohnson looks like we're still lacking some code coverage here

RoyEJohnson:
@TomWoodward Where do you see that? I had seen it in the "Files changed" tab, but it's not there anymore.

TomWoodward:
@RoyEJohnson i don't think this card was ever moved into review, but now its gotten old enough to have some conflicts so those have to be resolved. if its ready please make sure the issue is moved into the review column

TomWoodward (changes requested):
@RoyEJohnson i was trying to get a link for QA to evaluate this and noticed that its actually assigning target blank for figure links on the same page. isn't it supposed to not do that?

my test link https://rex-web-navigation-in-a-0hywoj.herokuapp.com/apps/rex/assigned/123456?book=13ac107a-f15f-49d2-97e8-60ab2e3b519c&section=d0a86917-6447-4b49-bed1-efb5c352cc77&section=065c3182-adac-48fc-a3ab-03d487982dd5

RoyEJohnson:
Figure links on the same page still use a full /books path rather than the /rex/assigned path. We talked about that a little; it's a different fix that needs done.

TomWoodward:
Right ok I remember talking about this, I think there is a thing that updates the urls that needs to be changed to use the current url rather than just the books one

I guess what we have here is better than what is on prod, but we need to make a new card for that issue

TomWoodward (approved):

PR #1922, opened 2023-04-25
Adjust toast top for Assigned

RoyEJohnson:
for: https://github.com/openstax/unified/issues/2288

### DIFF
--- a/src/app/content/components/Assigned.tsx
+++ b/src/app/content/components/Assigned.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { useSelector } from 'react-redux';
-import styled, {css} from 'styled-components/macro';
+import styled, { css } from 'styled-components/macro';
 import AccessibilityButtonsWrapper from '../../components/AccessibilityButtonsWrapper';
 import Button from '../../components/Button';
 import { useServices } from '../../context/Services';
@@ -37,6 +37,7 @@ const StyledButton = styled(Button)`
 
 // Override layout for Toast
 const assignedMobileTop = (props: any) => getMobileSearchFailureTop(props) - bookBannerMobileMiniHeight;
+// tslint:disable-next-line: variable-name
 const ToastOverride = styled(PageToasts)`
   top: ${topbarDesktopHeight}rem;
   left: 0;
--- a/src/app/content/components/Assigned.tsx
+++ b/src/app/content/components/Assigned.tsx
@@ -35,10 +35,12 @@ const StyledButton = styled(Button)`
   margin: 0 auto;
 `;
 
-// Override top for Toast
+// Override layout for Toast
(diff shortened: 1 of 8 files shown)


src/app/content/components/Assigned.tsx
```diff
@@ -90,7 +90,7 @@ export default () => {
     <ErrorModal />
     <ErrorBoundary>
       <AssignedTopBar section={section} />
-      <Page>
+      <Page toastTop={5.2}>
```
jivey, line 93: I think there's a toolbarHeight var somewhere that we could reference for this, or maybe we set one up so it's easy to adjust if the toolbar changes.

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -130,9 +131,16 @@ export default class PageComponent extends Component<PagePropTypes> {
 
   public render() {
     const pageIsReady = this.props.page && this.props.textSize !== null;
+    const PT = this.props.toastTop ?
+      styled(PageToasts)`
+        top:${this.props.toastTop}rem;
+      ` : PageToasts
+    ;
```
jivey, line 138: Usually creating a styled component within another component logs a console warning, you may want to move this outside of PageComponent.

jivey (changes requested):

RoyEJohnson:
Should be much better this time.

jivey (changes requested):
Looking good, looks like there's just some minor lint issues: https://github.com/openstax/rex-web/actions/runs/4823447051/jobs/8591956424?pr=1922#step:6:17

And this is up to you, but we may want to make the override always 100% width, at this specific width it seems to be too narrow:
image

once it goes wider, it looks fine, so maybe we should override whatever the first screenshot is doing to match this: 
image

RoyEJohnson:
I've fixed the lint complaints in files I changed, but there are some in other files I left.

src/app/content/components/Page/PageContent.tsx
```diff
@@ -56,10 +56,12 @@ export default styled(MainContent)`
     z-index: 1;
   }
 
+  // stylelint-disable selector-class-pattern
```
jivey, line 59:
```suggestion
  /* stylelint-disable selector-class-pattern */
```

src/app/content/components/Page/PageContent.tsx
```diff
@@ -56,10 +56,12 @@ export default styled(MainContent)`
...
   .MathJax_Display .highlight,
   .MathJax_Preview + .highlight {
     display: inline-block;
   }
+  // stylelint-enable selector-class-pattern
```
jivey, line 64:
```suggestion
  /* stylelint-enable selector-class-pattern */
```

jivey (changes requested):
At first I was worried about the linter errors but I saw from your other PR that deleting node_modules fixes it.

Looks good, just one small change to fix the stylelint, I guess it doesn't like the double slashes 🤔

jivey (approved):

PR #1923, opened 2023-04-26
Optimize Images in DOMTransformations

RoyEJohnson:
for: https://github.com/openstax/unified/issues/2136


TomWoodward:
this looks good so far. needs code coverage, that build test assertion needs to be updated with the new path. and you probably need to copy or move this https://github.com/openstax/rex-web/blob/main/src/test/fixtures/apps/archive/codeversion/resources/c7cc4c5e00a4bc07cb74f29cac2fb3c6cd6d53b5 to /src/test/fixtures/apps/image-cdn/v1/f=webp/apps/archive/codeversion/resources/c7cc4c5e00a4bc07cb74f29cac2fb3c6cd6d53b5

src/app/content/components/Content.browserspec.tsx
```diff
@@ -10,8 +10,8 @@ const TEST_CASES: { [testCase: string]: (target: Page) => Promise<void> } = {
   Desktop: setDesktopViewport, Mobile: setMobileViewport,
 };
 const EXPECTED_SCROLL_TOPS: { [testCase: string]: number[] } = {
-  Desktop: [242, 90, 122, 242, 365, 668, 761, 1263, 1607],
-  Mobile: [239, 66, 96, 239, 523, 1263, 1402, 1751, 2118],
+  Desktop: [242, 90, 122, 242, 365, 668, 761, 977, 1321],
+  Mobile: [239, 66, 96, 239, 523, 1263, 1402, 1618, 1985],
```
TomWoodward, line 14: don't update these

src/test/fixtures/apps/archive/codeversion/contents/testbook1-shortid:testpage12-shortid.json
```diff
@@ -2,6 +2,6 @@
...
... customers, clients, employees, and the surrounding community and environment</dd> </dl> <dl id=\"fs-idm407518992\"> <dt id=\"34873\">compliance</dt> <dd id=\"fs-idm405304016\">the extent to which a company conducts its business operations in accordance with applicable regulation and statutes</dd> </dl> <dl id=\"fs-idm403168464\"> <dt id=\"67088\">corporate culture</dt> <dd id=\"fs-idm401618864\">the shared beliefs, values, and behaviors that create the organizational context within which employees and managers interact <a data-type=\"cite\" href=\"\"><sup class=\"os-end-note-number\">20</sup></a> </dd> </dl> </div> <div id=\"table-of-links\"> <h3 class=\"os-title\" data-type=\"title\">Table of Links</h3> <ul> <li> <a href=\"#main-content\">Main Content</a> </li> <li> <a href=\"#fs-id1166227577868\">A note</a> </li> <li> <a href=\"#term104\">A term</a> </li> <li> <a href=\"#fs-id1166227682186\">A blockquote</a> </li> <li> <a href=\"#fs-id1166227571565\">A paragraph</a> </li> <li> <a href=\"#CNX_Chem_06_00_CrabNeb\">A figure</a> </li> <li> <a href=\"#fs-idp13638832\">A table</a> </li> <li> <a href=\"#fs-idm397777376\">A glossary term</a> </li> </ul> </div> </div> </body>\n </html>",
```
TomWoodward, line 5: the `src` in the content is a relative url, we transform it to rebase it, the previous fixture was correct

TomWoodward (changes requested):

src/test/fixtures/apps/archive/codeversion/contents/testbook1-shortid:testpage12-shortid.json
RoyEJohnson: The reason I don't get code coverage is that no tests have images with absolute paths beginning with /apps/archive. Do I need to create a new fixture or add a new image to an existing fixture? Or something else?

TomWoodward (approved):

TomWoodward:
is the test image too small after running through the image-cdn? that could... definitely present a problem if it happens with real content

RoyEJohnson:
The code coverage issue happens because there is no match [here](https://github.com/openstax/rex-web/blob/use-optimized-images/src/app/content/components/Page/contentDOMTransformations.ts#L74).
It does seem like the urls should be updated by the time the transformer is called, but for some reason, that match isn't happening. It's not an issue of anything after the image-cdn; it's not getting to that.

RoyEJohnson:
It looks like the only one with an image is testpage12, and that doesn't seem to be fully processed when it's used in a test.
I have copied the image tag from testpage12 to testpage1, which does get passed through `optimizeImages` in the testing of Page.spec.

PR #1931, opened 2023-05-22
FE Playwright test: assert note in content & MH page

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/2449

1.  When a highlight is broken into multiple pieces due to content styling, count only the unique highlight ids
2. Edit note in content page
3. Add waits for loading highlights in content page & MH modal
4. Assert edits made to note card in MH modal against content page notecard


playwright/src/fixtures/content.page.ts
```diff
@@ -285,6 +325,7 @@ class ContentPage {
       await this.page.mouse.down()
       await this.page.mouse.move(boundary.width - 20 + boundary.x, boundary.y + boundary.height - 10)
       await this.page.mouse.up()
+
```
Malar-Natarajan, line 328: @gregfitch this function `selectText` which I use to select text for highlighting sometimes drags beyond the range of the selector and selects extra text. (I have used only paragraph selectors, so at any point it should be selecting just one paragraph but occassionally it drags into the next para also).
Is there any way this can be prevented

gregfitch: This is generally due to the page layout, especially if the last line of text is less than half the width of the content block. I'm noticed browsers sometimes move to the next text block if you run beyond the closing tags 'location'. There's not much you can do to combat this as the block size changes with screen size (or browser), other than try to move fewer extra pixels during the mouse drag. This is one of the reasons why I used random text selects instead of full paragraphs most of the time, especially if it could grab parts that aren't in the highlightable bounds.

Malar-Natarajan: ok, thank you.

playwright/src/fixtures/content.page.ts
```diff
@@ -131,15 +133,19 @@ class ContentPage {
   async clickHighlight(highlight_id: string) {
     // Click on a highlight
     // param: highlight_id of the highlight to be clicked
-    this.highlightIdlocator = this.page.locator(`[data-highlight-id="${highlight_id}"][data-highlighted="true"]`)
-    await this.highlightIdlocator.click()
+    const highlightIdlocator = await this.page.$$(`[data-highlight-id="${highlight_id}"][data-highlighted="true"]`)
+
```
Malar-Natarajan, line 137: @gregfitch The `highlightIdlocator` defined in line 136 returns undefined. what am I doing wrong here.

gregfitch (approved):

PR #1933, opened 2023-05-30
Adjust token font size (Firefox only)

RoyEJohnson:
For: https://github.com/openstax/unified/issues/1652


RoyEJohnson:
I made a CSS adjustment that only Firefox supports.
I'm not sure that I chose the best place to put the change, and I'm not _sure_ there's no other content that would match the selector (though it seems unlikely).

Dantemss (approved):
I think the circles are the same size now but the letters are a different size, which will be impossible to fix without changing the font. Still, better than before IMO.

PR #1935, opened 2023-06-07
handle network failure for some archive/highlighter/search fetches with inline error message

jivey:
for: https://github.com/openstax/unified/issues/546

examples below, right now these take up the entire screen with a modal:
image


src/app/messages/pl/messages.json
```diff
@@ -107,7 +107,9 @@
   "i18n:notification:update:body": "Kliknij poniżej, by przejść do najnowszej wersji oprogramowania.",
   "i18n:notification:update:header": "Dostępna aktualizacja",
   "i18n:notification:update:reload": "Aktualizuj",
+  "i18n:notification:toast:archive:load-failure": "Nie możemy załadować tej strony. Sprawdź połączenie z internetem i spróbuj ponownie.",
   "i18n:notification:toast:search:highlight-not-found": "Nie odnajdujemy wyników dla tego zapytania. Spróbuj innego.",
+  "i18n:notification:toast:search:load-failure": "Nie możemy zakończyć wyszukiwania. Sprawdź połączenie z internetem i spróbuj ponownie.",
```
jivey, line 112: @ewaforman Would you mind taking a look at these translations for "We're unable to load this page" and "We're unable to complete your search request"? I'd be happy to make any changes you suggest!

src/app/messages/es/messages.json
```diff
@@ -107,7 +107,9 @@
   "i18n:notification:update:body": "Haga clic a continuación para utilizar la versión más reciente de este software.",
   "i18n:notification:update:header": "Actualizaciones disponibles",
   "i18n:notification:update:reload": "Actualización",
+  "i18n:notification:toast:archive:load-failure": "No podemos cargar esta página. Por favor, compruebe su conexión y vuelva a intentarlo.",
   "i18n:notification:toast:search:highlight-not-found": "No podemos encontrar resultados de esa búsqueda. Por favor, intente con otra.",
+  "i18n:notification:toast:search:load-failure": "No podemos completar su solicitud de búsqueda. Por favor, compruebe su conexión y vuelva a intentarlo.",
```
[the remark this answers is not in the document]
bethshook: looks good

Dantemss (approved):
I... think this is OK but I'd like @TomWoodward 's opinion on whether we should only catch certain errors or if we should try to preserve some error detail fields.

src/app/messages/pl/messages.json
ewaforman: @jivey Looks good.

TomWoodward (changes requested):
this is what i imagined for the highlights errors.

for the content loading errors we lack anything to put in the content pane, so i was thinking we'd put the error in the content area itself, not as a toast message. but the task wasn't very well defined and this looks like it'd work well so i'm fine with keeping what josiah did already as long as erika is happy with it

i would like to add some sentry reporting to these `catch` blocks for any case where the error isn't a `failed to fetch` network error. if its some kind of error code from the api or a malformed response or something i do want that in sentry

TomWoodward:
@jivey i know you have a few things going on, but don't forget about buttoning up this PR

jivey:
@erikayvonne Do you approve of these error messages appearing in these red toast messages? As Tom noted above, we could also show them in the content pane, but that would be a bit more work.
image

erikayvonne:
sorry I missed this @jivey - this looks good to me, thanks!

src/gateways/createArchiveLoader.spec.ts
```diff
@@ -246,6 +251,56 @@ describe('archiveLoader', () => {
...
+          expect(captureException).not.toHaveBeenCalled();
+          expect((error as ToastMesssageError).messageKey).toEqual(
+            'i18n:notification:toast:archive:load-failure'
+          );
+        } else {
+          expect(error).toBeTruthy();
+        }
+      });
+
+      xit('captures errors if something else went wrong with the fetch', async() => {
```
jivey, line 280: There is an issue with this test not showing that the Sentry call is working. It seems like the Sentry object on the implementation side is not being properly mocked (I've confirmed it is being called with manual testing.) This seems to be a setup/teardown bug, because it is passing in isolation.

jivey: I've narrowed this down to `resetModules()` in `afterEach`, removing it makes the new specs pass, but breaks the others.

TomWoodward: Does it work if you take stuff out of the imports at the top of the spec file and add them as requires in the before each?

jivey: That does work! In hindsight that makes sense, forcing the reload after it gets reset. Thanks!

TomWoodward: Yea if you don't you get mixed references to globals in the module cache. Usually this happens with react but in this case it might have been sentry

TomWoodward (approved):

PR #1937, opened 2023-06-08
Upgrade insecure packages

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2440


src/app/auth/init/establishState.ts
```diff
@@ -10,7 +11,7 @@ const initializer: Initializer = async({dispatch, userLoader}) => {
   const user = await userLoader.getCurrentUser();
 
   if (user) {
-    dispatch(receiveUser(formatUser(user)));
+    dispatch(receiveUser(formatUser(user as AccountsUser)));
```
jivey, line 14: This is might be a @TomWoodward question, but if we need to narrow this, it might make sense to replace REX's `AccountsUser` with `ApiUser` from `ts-utils`

jivey (approved):

RoyEJohnson:
@TomWoodward 
If I don't cast `user` [here](https://github.com/openstax/rex-web/blob/49a7643c0be298a42ff3d8837a69cf68e04b7ef4/src/app/auth/init/establishState.ts#L14), the complaint is:
```Argument of type 'User' is not assignable to parameter of type 'AccountsUser'.
  Type 'TokenUser' is missing the following properties from type 'AccountsUser': is_not_gdpr_location, contact_infos
  ```

PR #1940, opened 2023-06-20
Pre-fetch images during prerender to prime the image-cdn cache

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2434

Is it really as simple as this, or am I missing something?


jivey (approved):
Code change looks good, I'm not sure if more is needed to make this work, maybe check with @TomWoodward

TomWoodward (changes requested):
We only want to do this head request in prerendering, anything in the app code will also happen in end user sessions

Also, we should be including node-fetch I'm browser side code

RoyEJohnson:
Thanks, @TomWoodward, how the browser and prerender were separated was completely unclear, so this is extremely helpful.

RoyEJohnson:
I ran `prettier` on contentDOMTransformations, which is why there's so much reformatting there. I wasn't sure we wanted prefetch in `prerender/local`, but I figured it didn't hurt.

TomWoodward (changes requested):
can you unprettier it so i can tell what you actually changed please

the project linter will fail according to the project config when necessary

RoyEJohnson:
Unprettiered.

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -19,15 +20,17 @@ export function linksToOtherPagesOpenInNewTab(rootEl: HTMLElement, currentPath:
 // from https://github.com/openstax/webview/blob/f95b1d0696a70f0b61d83a85c173102e248354cd
 // .../src/scripts/modules/media/body/body.coffee#L123
 // We are passing Document because it is required to prerender.
-export const transformContent = (document: Document, rootEl: HTMLElement, intl: IntlShape) => {
+export const transformContent = (
+  document: Document, rootEl: HTMLElement, intl: IntlShape, services: AppServices & MiddlewareAPI
```
TomWoodward, line 24: just services, don't depend on the redux middleware here too

TomWoodward (approved):
this is fine, just commented on some extra stuff you're depending on that you don't need

PR #1943, opened 2023-06-28
Skip key terms results when selecting first search result

RoyEJohnson:
For: https://github.com/openstax/unified/issues/1761


Dantemss (approved):
My only comment is that `skipKeyTerms` not only filters out key terms but also returns the first result, so maybe the method name should indicate that. But it's a very minor thing.

RoyEJohnson:
I renamed it `firstNonKeyTermResult`

PR #1947, opened 2023-07-06
Add citation metadata

RoyEJohnson:
For: https://github.com/openstax/unified/issues/1938

### DIFF
--- a/src/app/content/components/Attribution.tsx
+++ b/src/app/content/components/Attribution.tsx
@@ -96,6 +96,44 @@ const AttributionDetails = styled(Details)`
 
   ${disablePrint}
 `;
+// date is initialized as UTC, conversion to local time can change the date.
+// this compensates
+function compensateForUTC(date: Date): void {
+  date.setMinutes(date.getMinutes() + date.getTimezoneOffset());
+}
+
+function getPublishDate(book: Book): Date | null {
+  if (!('publish_date' in book)) {
+    return null;
+  }
+  const date = new Date(book.publish_date);
+
+  compensateForUTC(date);
+  return date;
+}
+
+function getAuthors(book: Book) {
+  if (!('authors' in book)) {
+    return [];
+  }
+  const seniorAuthors = book.authors.filter((author) => author.value.senior_author);
+
+  return seniorAuthors.length > 0
+    ? seniorAuthors
+    : book.authors.slice(0, 2);
+}
+
+export function attributionValues(book: Book) {
+  const bookPublishDate = getPublishDate(book);
+  const authorsToDisplay = getAuthors(book);
+
+  return {
+    bookTitle: book.title,
+    copyrightHolder: 'OpenStax',
+    bookPublishDate,
+    authorsToDisplay,
+  };
+}
 
 // tslint:disable-next-line:variable-name
 const AttributionContent = htmlMessage('i18n:attribution:text', Content);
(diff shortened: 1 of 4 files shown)


src/app/content/components/Attribution.tsx
```diff
@@ -96,6 +96,44 @@ const AttributionDetails = styled(Details)`
...
+
+function getAuthors(book: Book) {
+  if (!('authors' in book)) {
+    return [];
+  }
+  const seniorAuthors = book.authors.filter((author) => author.value.senior_author);
+
+  return seniorAuthors.length > 0
+    ? seniorAuthors
+    : book.authors.slice(0, 2);
```
Dantemss, line 123: I know this is code that was already there so normally I wouldn't complain... too much. Except that now we're using it when making actual citations...

According to the MLA: `If a source has two authors, name both authors in your MLA in-text citation and Works Cited entry. If there are three or more authors, name only the first author, followed by et al.`

But we're not necessarily using MLA... who is responsible for our citation format anyway?

I won't hold back the PR because of this though... we can always make another card to fix.

Dantemss (approved):
I think it's good but see comments

src/app/content/hooks/receiveContent.ts
```diff
@@ -12,9 +12,29 @@ import { content as contentRoute } from '../routes';
...
+    'en-us',
+    {year: 'numeric',  month: 'short', day: 'numeric'}
+  );
+  const authors = authorsToDisplay.map((a) => ({
+    name: 'citation_author', content: a.value.name,
+  }));
+
+  return [
+    {name: 'citation_title', content: bookTitle},
+    {name: 'citation_publisher', content: copyrightHolder},
```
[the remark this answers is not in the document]
TomWoodward: the copyright holder is actually not always openstax, there are at least cases where its katalyst, TEA, or The Michelson 20MM Foundation  they're defined here https://github.com/openstax/rex-web/pull/1947/files/8ad26f33798ba6bbac50de73cbec8e6b48e19177#diff-653cd6db0a45d9a4e127f19042c6e957ce79eecb1725f4eb8b50c625ce4d0b65R153

src/app/content/hooks/receiveContent.ts
```diff
@@ -12,9 +12,24 @@ import { content as contentRoute } from '../routes';
 import * as select from '../selectors';
 import { getCanonicalUrlParams } from '../utils/canonicalUrl';
 import { createTitle, getPageDescription } from '../utils/seoUtils';
+import { attributionValues } from '../components/Attribution';
```
TomWoodward, line 15: i don't really like the redux code importing from the react code, can you move this logic to a utils file?

TomWoodward (changes requested):

src/app/content/hooks/receiveContent.ts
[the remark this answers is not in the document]
Dantemss: Hm in that case if the copyright holder is not always OpenStax, maybe the citation_publisher needs to be always "OpenStax".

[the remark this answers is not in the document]
TomWoodward: I think we might still need to say the publisher is katalyst at least, not sure about the other ones. I guess the katalyst case is hard coded to "openstax Poland" in the translation files for the attribution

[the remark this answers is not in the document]
RoyEJohnson: Do we need a separate lookup table, then? Since copyright holder and publisher aren't the same thing, and it's not always OpenStax?

src/app/content/components/utils/attributionValues.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+function localeToLanguage(locale: string): string {
+  const languageLookup: {[key: string]: string} = {
+    es: 'Spanish',
+    pl: 'Polish',
+    en: 'English',
+  };
+
+  return languageLookup[locale];
```
TomWoodward, line 56: i think you can replace this with either
```
new Intl.DisplayNames(["en"], { type: "language" }).of(locale);
```

or

```
new Intl.DisplayNames([locale], { type: "language" }).of(locale);
```

depending on if you want the result to be eg "Spanish" or "español"

TomWoodward (changes requested):
looking pretty good

src/app/content/components/utils/attributionValues.ts
RoyEJohnson: I thought there must be something of that sort available. Thx.

RoyEJohnson: Looks like it gets a @ts-expect-error until the Upgrade TypeScript PR goes through.
https://github.com/microsoft/TypeScript/issues/41338#issuecomment-1257926423

TomWoodward: please add some more comments around the expect-error with a link to that issue and mention the ts version its fixed in for future us

TomWoodward (approved):

TomWoodward (approved):
looks fine to me

src/app/content/components/utils/attributionValues.ts
```diff
@@ -0,0 +1,64 @@
...
+export function attributionValues(book: Book) {
+  const bookPublishDate = getPublishDate(book);
+  const authorsToDisplay = getAuthors(book);
+  const locale = book.language;
+
+  return {
+    bookTitle: book.title,
+    publisher: locale === 'pl' ? 'OpenStax Poland' : 'OpenStax',
+    // Per https://github.com/microsoft/TypeScript/issues/41338#issuecomment-1571465883
+    // DisplayNames is not defined by TypeScript until version v4.5
```
TomWoodward, line 58: the TS pr is merged, so you may be able to remove this

RoyEJohnson:
I have given up on the Zotero event, so it _won't work with Zotero_, but the tags are properly created.

PR #1951, opened 2023-07-10
add integration to report user and content info to ga4

TomWoodward:
for: openstax/unified#2492


Dantemss (approved):
I feel like the only way to make sure GA code is correct is to try it out. But it looks good to me.

PR #2002, opened 2023-07-19
Upgrade TypeScript to ^4 (4.9.5)

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2510

### DIFF
--- a/script/domVisitor.ts
+++ b/script/domVisitor.ts
@@ -86,7 +86,7 @@ async function visitPages(
         anyFailures = true;
         bar.interrupt(`- (${matches.length}) ${pageUrl}#${matches[0]}`);
       }
-    } catch (e) {
+    } catch (e: any) {
       anyFailures = true;
       bar.interrupt(`- (error loading) ${pageUrl}`);
       if (e.message) {
--- a/src/app/components/DotMenu.tsx
+++ b/src/app/components/DotMenu.tsx
@@ -48,8 +48,7 @@ export const DotMenuDropdownList = styled(DropdownList)`
 `;
 
 // tslint:disable-next-line:variable-name
-// @ts-ignore
-export const DotMenuDropdown = styled((props: DropdownProps) => <Dropdown toggle={<DotMenuToggle />} {...props} />)`
+export const DotMenuDropdown = styled((props: DropdownProps) => <Dropdown {...props} toggle={<DotMenuToggle />} />)`
   .focus-within ${DotMenuIcon} {
     color: ${theme.color.primary.gray.base};
   }
--- a/src/app/components/DotMenu.tsx
+++ b/src/app/components/DotMenu.tsx
@@ -48,6 +48,7 @@ export const DotMenuDropdownList = styled(DropdownList)`
 `;
 
 // tslint:disable-next-line:variable-name
+// @ts-ignore
 export const DotMenuDropdown = styled((props: DropdownProps) => <Dropdown toggle={<DotMenuToggle />} {...props} />)`
   .focus-within ${DotMenuIcon} {
     color: ${theme.color.primary.gray.base};
--- a/src/app/notifications/components/UpdatesAvailable.tsx
+++ b/src/app/notifications/components/UpdatesAvailable.tsx
@@ -31,7 +31,7 @@ const UpdatesAvailable = ({className}: {className?: string}) => {
   // if the readyPromise exists, wait for it before rendering
   // if it doesn't exist, we render anyway and in that case
   // onClick() calls activateSwAndReload(undefined) to reload without waiting
-  if (readyPromise && !sw) {
+  if (assertWindow().navigator.serviceWorker && !sw) {
     return null;
   }
 
--- a/src/app/utils.ts
+++ b/src/app/utils.ts
@@ -150,7 +150,7 @@ export const merge = <T1 extends {}, T2 extends {}>(thing1: T1, thing2: T2): T1 
   ...getCommonProperties(thing1, thing2).reduce((result, property) => ({
     ...result,
     ...(isPlainObject(thing1[property]) && isPlainObject(thing2[property])
-      ? {[property]: merge(thing1[property], thing2[property])}
+      ? {[property]: merge(thing1[property] as object, thing2[property] as object)}
       : (Array.isArray(thing1[property]) && Array.isArray(thing2[property]))
         ? {[property]: [...thing1[property] as unknown as [], ...thing2[property] as unknown as []]}
         : {}
--- a/src/app/utils.ts
+++ b/src/app/utils.ts
@@ -6,7 +6,6 @@ import Sentry from '../helpers/Sentry';
 import { receiveLoggedOut } from './auth/actions';
 import { recordError, showErrorDialog } from './errors/actions';
 import { notFound } from './errors/routes';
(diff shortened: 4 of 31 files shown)


src/app/utils.ts
```diff
@@ -150,7 +150,7 @@ export const merge = <T1 extends {}, T2 extends {}>(thing1: T1, thing2: T2): T1
   ...getCommonProperties(thing1, thing2).reduce((result, property) => ({
     ...result,
     ...(isPlainObject(thing1[property]) && isPlainObject(thing2[property])
-      ? {[property]: merge(thing1[property], thing2[property])}
+      ? {[property]: merge(thing1[property] as object, thing2[property] as object)}
```
RoyEJohnson, line 153: Complaint was:
```
Argument of type 'T1[keyof T1 & keyof T2]' is not assignable to parameter of type '{}'.
  Type 'T1[string] | T1[number] | T1[symbol]' is not assignable to type '{}'.
    Type 'T1[string]' is not assignable to type '{}'.ts(2345)
```

RoyEJohnson: Here's how ChatGPT suggested rewriting the function:
```
// Utility function to get common properties between two objects
const getCommonProperties = <T1 extends {}, T2 extends {}>(obj1: T1, obj2: T2): Array<keyof T1 & keyof T2> =>
  Object.keys(obj1).filter((key) => key in obj2) as Array<keyof T1 & keyof T2>;

// Merges two objects (thing1 and thing2) into a new object
export const merge = <T1 extends {}, T2 extends {}>(thing1: T1, thing2: T2): T1 & T2 => {
  const commonProperties = getCommonProperties(thing1, thing2);

  const mergedObject = commonProperties.reduce((result, property) => {
    const mergedValue = isPlainObject(thing1[property]) && isPlainObject(thing2[property])
      ? merge(thing1[property] as PlainObject, thing2[property] as PlainObject)
      : Array.isArray(thing1[property]) && Array.isArray(thing2[property])
        ? [...thing1[property], ...thing2[property]]
        : {};

    return { ...result, [property]: mergedValue };
  }, {});

  return { ...thing1, ...thing2, ...mergedObject };
};
```

src/app/components/DotMenu.tsx
```diff
@@ -48,8 +48,7 @@ export const DotMenuDropdownList = styled(DropdownList)`
 `;
 
 // tslint:disable-next-line:variable-name
-// @ts-ignore
-export const DotMenuDropdown = styled((props: DropdownProps) => <Dropdown toggle={<DotMenuToggle />} {...props} />)`
+export const DotMenuDropdown = styled((props: DropdownProps) => <Dropdown {...{...props, toggle: <DotMenuToggle />}} />)`
```
RoyEJohnson, line 51: TS indicated that `props` would always overwrite `toggle`, so I rewrote to ensure `toggle` has precedence.

Dantemss: I believe you should be able to just move the toggle property to before the props property. And maybe also change the type of props to `Omit<DropdownProps, 'toggle'>`

Sorry, I started looking into this because it shows up as assigned to me in the GitHub UI but eventually I saw it was actually Josiah so I'll let him do the rest :)

src/app/utils.ts
TomWoodward: the same function exists in ts-utils [here](https://github.com/openstax/typescript-service-template/blob/main/packages/utils/src/misc/merge.ts) which uses 4.6

ts-utils is now a dependency of rex because we're using the user loader, so you can just use the helper function from the library if thats easier

src/app/components/DotMenu.tsx
TomWoodward: yea you should just be able to spread props before the extra param
```suggestion
export const DotMenuDropdown = styled((props: DropdownProps) =>  <Dropdown {...props} toggle={<DotMenuToggle />} />)`
```

i believe this issue is happening because of a bug in the styled-component prop types where custom styled component props like this one wouldn't get strong type checks (so we were actually not providing a toggle prop even if its required by the types)

script/domVisitor.ts
```diff
@@ -86,7 +86,7 @@ async function visitPages(
         anyFailures = true;
         bar.interrupt(`- (${matches.length}) ${pageUrl}#${matches[0]}`);
       }
-    } catch (e) {
+    } catch (e: any) {
```
TomWoodward, line 89: in a perfect world all these errors would be typed as `unknown` and we'd have to use guards before accessing them... but you're maintaining the existing level of safety which is fine for this pr

src/app/content/highlights/hooks/loadMore.ts
```diff
@@ -56,7 +56,7 @@ export const hookBody: ActionHookBody<
       highlights = await loadMore(services, summaryPageSize);
     } catch (error) {
       services.dispatch(toggleSummaryHighlightsLoading(false));
-      throw ensureApplicationErrorType(error, new HighlightPopupLoadError({ destination: 'myHighlights' }));
+      throw ensureApplicationErrorType(error as Error, new HighlightPopupLoadError({ destination: 'myHighlights' }));
```
TomWoodward, line 59: you could change the `ensureApplicationErrorType` helper to accept `any` as its first argument and internally do an `instanceof Error` check, that'd be more correct and have less changes to all these places

src/app/notifications/components/UpdatesAvailable.tsx
```diff
@@ -31,7 +31,7 @@ const UpdatesAvailable = ({className}: {className?: string}) => {
   // if the readyPromise exists, wait for it before rendering
   // if it doesn't exist, we render anyway and in that case
   // onClick() calls activateSwAndReload(undefined) to reload without waiting
-  if (readyPromise && !sw) {
+  if (assertWindow().navigator.serviceWorker && !sw) {
```
TomWoodward, line 34: why did you change this? i mean i think i see how it does the same thing i'm just not sure why you would have to change it

TomWoodward (changes requested):

src/app/notifications/components/UpdatesAvailable.tsx
Dantemss: Hm yeah not entirely sure what's going on. From the code above, the type of `readyPromise` should be `Promise<whatever> | undefined`.

Dantemss: Maybe it's thinking `assertWindow().navigator.serviceWorker` is always defined which is probably not true for older browsers.

RoyEJohnson: That seems to be correct; `serviceWorker` is of type `ServiceWorkerContainer`, which is an interface with no `| undefined`. A little odd that it doesn't tell me `serviceWorker` is always defined in the `if`. Taking the test out of the `if` breaks a test (`'doesn\'t render if the serviceWorker or readyPromise change after render()`)

TomWoodward (approved):

src/app/notifications/components/UpdatesAvailable.tsx
Dantemss: Taking the test out would probably break IE and other old stuff: https://caniuse.com/serviceworkers

jivey (approved):

PR #2018, opened 2023-07-25
Added book-scan task for Concourse; Added cache and block analytics scripts in domVisitor

Dantemss:
For: https://github.com/openstax/unified/issues/2480

Related PR: https://github.com/openstax/unified-deployment/pull/335

The caching is very aggressive and not compliant with cache headers, it simply pretends every url can be cached instead.
quick-lru was already being used by something else in our bundle so I picked it


concourse/book-scan/task.yml
```diff
@@ -0,0 +1,20 @@
+---
+platform: linux
+
+image_resource:
+  type: docker-image
+  source:
+    password: ((dockerhub-password))
+    username: ((dockerhub-username))
+    repository: node
+    tag: 14.15.4-buster
```
Dantemss, line 10: I couldn't see an easy way to get the rex image from the build-rex job instead, but I'm open to suggestions.

Dantemss:
This might end up having to be a @TomWoodward review...

script/validate-modified-books.bash
```diff
@@ -9,11 +9,10 @@ while read -r row; do
   book_version=$(jq -r '.book_version' <<< "$row")
   is_new=$(jq -r '.is_new' <<< "$row")
 
-  node script/entry.js domVisitor errorsExist \
+  REACT_APP_UNLIMITED_CONTENT=true node script/entry.js domVisitor errorsExist \
     --bookId="$book_id" \
     --bookVersion="$book_version" \
     --rootUrl="$BASE_URL" \
-    --queryString="validateLinks" \
```
Dantemss, line 16: It seems to me that this queryString and REACT_APP_UNLIMITED_CONTENT have the same effect

TomWoodward:
The query string is a way of setting that config. Because the scan is
running against a remote server it depends on how the target environment is
configured. Heroku has unlimited content, sandbox environments don't

script/domVisitor.ts
```diff
@@ -12,6 +13,15 @@ import createOSWebLoader from '../src/gateways/createOSWebLoader';
...
 import progressBar from './utils/progressBar';
 
+// Blocking GTM prevents most analytics scripts from loading
+const blockedRequests = [
+  /^https?:\/\/(?:www\.)?googletagmanager\.com/,
+  /^https?:\/\/js\.pulseinsights\.com/,
+];
+
+// Most pages perform less than 10 requests when GTM is blocked
+const cacheMaxSize = 50;
```
jivey, line 23: Is this 50 globally, and is that enough for an entire book to not see misses?

jivey (commented):
This seems good to me, but it does deal with REX areas and concourse stuff I haven't dealt with, so might be good to get a second look from @TomWoodward

script/domVisitor.ts
Dantemss: It's 50 requests in the LRU cache. From my testing this seems enough to get a pretty good hit rate, mainly because blocking GTM greatly reduces the number of requests (to about 10 per page).

concourse/book-scan/script.bash
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
TomWoodward, line 1: i think you should use the docker image built for CI that has all this installed already

TomWoodward (changes requested):

concourse/book-scan/script.bash
Dantemss: Ah I see what you mean... I also probably need to figure out why the runs on concourse have so many errors...

Dantemss:
Need to upload a new CI image to dockerhub based on this branch and then update the branch

script/domVisitor.ts
```diff
@@ -76,7 +86,32 @@ async function visitPages(
...
-      await page.goto(`${rootUrl}${pageUrl}${appendQueryString}`);
+      // validateLinks should persist within the SPA so we allow it when clicking links
+      if (appendQueryString && appendQueryString !== '?validateLinks') {
+        await page.goto(`${rootUrl}${pageUrl}${appendQueryString}`);
+      } else {
+        // If there's no queryString, we can click the content links instead which are much faster
+        // Prefer using env variables instead of queryStrings when possible
+        const pageComponents = pageUrl.split('/');
+        const slug = pageComponents[pageComponents.length - 1];
+        const linkSelector = `a[href="${slug}"]`;
```
TomWoodward, line 97: but i don think the archive url will get written into the anchor, so you may need to loosen up this check, maybe `a[href^="${slug}"]` or whatever the starts with check is

TomWoodward (changes requested):

TomWoodward (approved):

PR #2032, opened 2023-09-05
Fix search-highlighted hashlinks

RoyEJohnson:
For: https://github.com/openstax/unified/issues/1715

### DIFF
--- a/src/app/content/components/Page/contentLinkHandler.ts
+++ b/src/app/content/components/Page/contentLinkHandler.ts
@@ -95,7 +95,7 @@ export const reduceReferences = (
   }
 };
 
-const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageReferenceMap | PageReferenceError) =>
+const isPathReferenceForBook = (pathname: string, book: Book) => (ref: PageReferenceMap | PageReferenceError) =>
   isPageReferenceError(ref)
     ? false
     : content.getUrl(ref.params) === pathname
@@ -104,9 +104,31 @@ const isPathRefernceForBook = (pathname: string, book: Book) => (ref: PageRefere
         || ('uuid' in ref.params.book && ref.params.book.uuid === book.id)
       );
 
-// tslint:disable-next-line: max-line-length
-export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => ContentLinkProp, services: AppServices & MiddlewareAPI) =>
-  async(e: MouseEvent) => {
+const doubleClick : {
+  timer: number | undefined;
+  lastHref: string
+} = {
+  timer: undefined,
+  lastHref: '',
+};
+function isDoubleClick(href: string) {
+  const result = doubleClick.timer && doubleClick.lastHref === href;
+
+  doubleClick.lastHref = href;
+  assertWindow().clearTimeout(doubleClick.timer);
+  doubleClick.timer = assertWindow().setTimeout(
+    () => doubleClick.lastHref = '',
+    500
+  );
+
+  return result;
+}
+
+export const contentLinkHandler = (
+  anchor: HTMLAnchorElement,
+  getProps: () => ContentLinkProp,
+  services: AppServices & MiddlewareAPI
+) => async(e: MouseEvent) => {
     const {
       references,
       navigate,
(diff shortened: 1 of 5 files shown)


Dantemss:
Sorry, I feel like I don't understand how this all works enough to know if this won't break something else, so I have to defer to Tom or Beth.

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -138,7 +138,7 @@ export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => Co
 
     e.preventDefault();
 
-    if (isHtmlElementWithHighlight(e.target)) {
+    if (isHtmlElementWithHighlight(e.target) && hash !== reference?.match) {
```
TomWoodward, line 141: i think this line was added here https://github.com/openstax/rex-web/pull/485/files#diff-3f2d87fbf043fd71538b475ed64199e922160643163bab450faa5f753dd9024aR73 as a way to not clear the highlight editing when clicking figure links. as you can see its supposed to allow clicks through when they're the focused highlight but i don't seem to be able to click any links inside highlights so i think something has gotten broken along the way.

TomWoodward (changes requested):

src/app/content/components/Page/contentLinkHandler.ts
RoyEJohnson: This change did make the click work on Figure 1.7 for me.
You can try it in the [deployment (with a query already entered)](https://rex-web-fix-search-high-e3k7v6.herokuapp.com/books/american-government-2e/pages/1-2-who-governs-elitism-pluralism-and-tradeoffs?query=figure%201.7&target=%7B%22type%22%3A%22search%22%2C%22index%22%3A0%7D#fs-id1171472068209)

bethshook: or do we just need another card

TomWoodward: no, i don't think the fix here is correct. it only addresses hash links, [other search highlights over other content links](https://rex-web-fix-search-high-e3k7v6.herokuapp.com/books/american-government-2e/pages/2-1-the-pre-revolutionary-period-and-the-roots-of-the-american-political-tradition?query=Thomas%20Jefferson%2C%20the%20Declaration%20of%20Independence%20officially&target=%7B%22type%22%3A%22search%22%2C%22index%22%3A0%7D#fs-id1166226347619) will continue to not work. we need a more generic solution for allowing clicks to go through when clicking in highlights.

i was saying that previously we tried to make it so clicking in a user highlight would first focus it, and then allow you to click the link with the second click. this appears to be broken now, and in any case would not have worked with search highlights because its specifically looking for the focused user highlight. 

maybe we can do a more dom-based check by looking for the focused class on the highlight

in any case i think there is more work to do on this issue to find a good solution, preferably one that works for both user and search highlgihts

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -138,7 +138,7 @@ export const contentLinkHandler = (anchor: HTMLAnchorElement, getProps: () => Co
 
     e.preventDefault();
 
-    if (isHtmlElementWithHighlight(e.target)) {
+    if (isHtmlElementWithHighlight(e.target) && !e.target.closest('a')) {
```
TomWoodward, line 141: we're in the link click handler, i don't think there'll ever _not_ be a closest anchor. this change makes it impossible to select a highlight that contains only an anchor.

i think it would work to change the conditional below `e.target.getAttribute('data-highlight-id') !==  focusedHighlight` to not rely on the `focusedHightlight` state and instead use the css, like `!e.target.matches('.focus')`, that way it'll match either a focused user highlight or a focused search highlight

TomWoodward (changes requested):

src/app/content/components/Page/contentLinkHandler.ts
RoyEJohnson: That does make the highlight clickable, but breaks [a test](https://github.com/openstax/rex-web/blob/main/src/app/content/components/Page/contentLinkHandler.spec.ts#L293).

TomWoodward: because the test is written to rely on the redux state [here](https://github.com/openstax/rex-web/blob/main/src/app/content/components/Page/contentLinkHandler.spec.ts#L288) instead of the class, it doesn't seem like this test is integrating the entire highlighter, so you may have to add the `focus` class manually to the element that it [creates](https://github.com/openstax/rex-web/blob/main/src/app/content/components/Page/contentLinkHandler.spec.ts#L256)

RoyEJohnson: That does make the test happy.

TomWoodward (approved):

PR #2040, opened 2023-09-21
Add Aria roles to navigable divs

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2600


Dantemss (approved):
I do worry that we'll have some css selector or something that fails because the main content is now a different tag. So this requires some careful regression testing.

PR #2047, opened 2023-09-27
Suppress highlight dialog in mobile

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2593

You might want to review one commit at a time. The first commit is the functional change.
The second and third commits are refactoring. The last is just reformatting with prettier.

### DIFF
--- a/src/app/content/highlights/components/Card.tsx
+++ b/src/app/content/highlights/components/Card.tsx
@@ -147,7 +147,6 @@ function Card(props: CardProps) {
   const [highlightRemoved, setHighlightRemoved] = React.useState<boolean>(
     false
   );
-  const isMobile = useMatchMobileMediumQuery();
   const computedProps = useComputedProps(props);
 
   if (
@@ -155,8 +154,7 @@ function Card(props: CardProps) {
     !props.page ||
     !props.book ||
     !locationFilterId ||
-    highlightRemoved ||
-    isMobile
+    highlightRemoved
   ) {
     return null;
   }
@@ -199,10 +197,14 @@ function NoteOrCard({
       });
     }
   }, [locationFilterId, props, setHighlightRemoved]);
-
+  const isMobile = useMatchMobileMediumQuery();
   const style = highlightStyles.find(
     search => props.data && search.label === props.data.color
   );
+
+  if (!annotation && isMobile) {
+    return null;
+  }
 
   return (
     <div onClick={focusCard} data-testid='card'>
--- a/src/app/content/highlights/components/Card.tsx
+++ b/src/app/content/highlights/components/Card.tsx
@@ -6,7 +6,7 @@ import React from 'react';
 import { connect, useSelector } from 'react-redux';
 import styled from 'styled-components';
 import { useServices } from '../../../context/Services';
-import { useFocusIn } from '../../../reactUtils';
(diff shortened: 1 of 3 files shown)


src/app/content/highlights/components/Card.tsx
```diff
@@ -101,25 +101,28 @@ const Card = (props: CardProps) => {
   const location = React.useMemo(() => {
     return props.page && getHighlightLocationFilterForPage(locationFilters, props.page);
   }, [locationFilters, props.page]);
+  const locationFilterId = location && stripIdVersion(location.id);
+  const { page, book } = props;
 
-  const isMobile = useMatchMobileMediumQuery();
+  const onRemove = React.useCallback(
+    () => {
+      if (props.data && page && locationFilterId) {
```
Dantemss: Is this here just to fix types? I'm wondering if it would cause the remove highlight button to not work at all if something is wrong with page or locationFilterId. Would that matter?

src/app/content/highlights/components/Card.tsx
```diff
@@ -54,94 +54,117 @@ export interface CardProps {
...
+return React.useMemo(
+    () => ({
+      focusCard,
+      editing,
+      setEditing,
+      annotation,
+      hasUnsavedHighlight,
+      element,
+    }),
+    [annotation, editing, focusCard, hasUnsavedHighlight]
```
Dantemss, line 126: Don't you need setEditing and element in the array?

src/app/content/highlights/components/Card.tsx
RoyEJohnson: I did not change the code of onRemove; I just put it in useCallback. The call to props.remove requires `page.id` and `locationFilterId`. I don't have a real understanding of how that works.

src/app/content/highlights/components/Card.tsx
RoyEJohnson: State set functions are known not to change, and refs are explicitly not change-tracked.

src/app/content/highlights/components/Card.tsx
Dantemss: Looking at it again, if page or locationFilterId are null or undefined you don't render anything... you could probably move the `onRemove` definition to below the if (...) { return null; } and not have to have this condition here but it's fine.

Dantemss (approved):

src/app/content/highlights/components/Card.tsx
RoyEJohnson: Hooks have to be called every time their caller is called. It's part of how React[ keeps track of them](https://legacy.reactjs.org/docs/hooks-rules.html#explanation). The way around it would be to put the part after the `return null` into a separate function (which would include the onRemove), which seemed like too much trouble to me.

RoyEJohnson: On second thought, I went ahead and broke that down.

src/app/content/highlights/components/Card.tsx
TomWoodward: roy is right, but you may see us listing dependencies in other places because the linter has special cases to not complain if you're using the result of `useRef` or `useState`, but if you're composing those things through a helper hook the linter can't keep up. its the same thing either way because they never update, but you end up needing to specify a bunch of extra dependencies sometimes.

PR #2048, opened 2023-10-02
Show delete modal before saving empty annotation

RoyEJohnson:
For: https://github.com/openstax/unified/issues/859


Dantemss (approved):
If you run into weirdness due to double clicks or the modal being bypassed somehow, I'd split updateAnnotation into 2 functions... but if not then what you have seems good to me.

PR #2051, opened 2023-10-04
Handle large numbers of highlights

RoyEJohnson:
For: https://github.com/openstax/unified/issues/887


src/app/content/highlights/components/EditCard.tsx
```diff
@@ -45,192 +49,415 @@ export interface EditCardProps {
...
 
-  const blurIfNotEditing = React.useCallback(() => {
-    if (!props.hasUnsavedHighlight && !editingAnnotation) {
-      props.onBlur();
+function LoginConfirmation({
+  data,
+  onBlur,
+}: Pick<EditCardProps, 'data' | 'onBlur'>) {
+  const loginLink = useSelector(selectAuth.loginLink);
+  const trackShowLogin = useCallbackFor(useAnalyticsEvent('showLogin'));
```
TomWoodward, line 89: instead of wrapping this in a ref utility, if `useAnalyticsEvent` is returning a new function ref every time you call it we should probably fix it internally to use useCallback

src/app/content/highlights/components/utils/useCallbackFor.ts
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+// Some functions get recreated but shouldn't be reacted to
+// This makes a non-changing callback
+export default function useCallbackFor<T extends Array<unknown>>(fn: (...args: T) => unknown) {
+    const ref = React.useRef(fn);
+    const cb = React.useCallback(
```
TomWoodward, line 7: unless i'm missing something, the presence of this useCallback hook isn't really doing anything. since you're passing `[]` for the dependencies, you might as well define this as 

```
export default function useInitialValue<T>(thing: T) {
  return React.useRef(thing).current;
}
```

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
```diff
@@ -138,16 +138,6 @@ exports[`EditCard matches snapshot when editing 1`] = `
...
-  <div
-    confirmLink="/accounts/login?r=/"
-    confirmMessage="i18n:highlighting:login:link"
-    data-analytics-label="login"
-    data-analytics-region="highlighting-login"
-    drawFocus={false}
-    message="i18n:highlighting:login:prompt"
-    mock-confirmation={true}
-    onCancel={[MockFunction]}
-  />
```
TomWoodward, line 150: why was login removed from this snap and not the others?

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
```diff
@@ -170,23 +160,6 @@ exports[`EditCard matches snapshot when focused 1`] = `
...
-    mock-note={true}
-    note=""
-    onChange={[Function]}
-    onFocus={[Function]}
-    textareaRef={
-      Object {
-        "current": null,
-      }
-    }
-  />
```
TomWoodward, line 189: why are these removed?

TomWoodward (changes requested):

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -45,192 +49,415 @@ export interface EditCardProps {
 
 // tslint:disable-next-line:variable-name
 const EditCard = React.forwardRef<HTMLElement, EditCardProps>((props, ref) => {
+  if (!props.isActive) {
```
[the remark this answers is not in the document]
RoyEJohnson: From what I can tell, everything worked as expected: highlights displayed and saved and deleted without a problem. All of their state is derived from props, apart from `confirmingDelete`, which I don't think should persist if they go to a different highlight.
This is the central change for the PR, by the way. Having all the editcards render without displaying is the source of the bogdown.

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -45,192 +49,415 @@ export interface EditCardProps {
...
-      props.onBlur();
+function LoginConfirmation({
+  data,
+  onBlur,
+}: Pick<EditCardProps, 'data' | 'onBlur'>) {
+  const loginLink = useSelector(selectAuth.loginLink);
+  const trackShowLogin = useCallbackFor(useAnalyticsEvent('showLogin'));
+
+  React.useEffect(() => {
+    if (!data) {
```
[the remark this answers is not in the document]
RoyEJohnson: This was a test from the original code, I just kept it in the new place. Maybe it only mattered for trackShowCreate?
```
React.useEffect(() => {
    if (props.data || !props.isActive) { return; }
    if (authenticated) {
      trackShowCreate();
    } else {
      trackShowLogin();
      ```

File: src/app/content/highlights/components/Confirmation.tsx [hunk 7]
```diff
@@ -55,7 +55,7 @@ const Confirmation = React.forwardRef<HTMLElement, Props>((
 ) => {
   const drawFocusRef = useDrawFocus();
   return <Overlay
-    ref={mergeRefs(ref, drawFocus ? drawFocusRef : null)}
+    ref={mergeRefs(ref, drawFocusRef)}
```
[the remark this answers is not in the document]
RoyEJohnson: Leaving this ternary in dropped code coverage. I'm not sure how, since I see it gets called with and without `drawFocus={false}` in the code. I've put it back in, and will see if I need to write a new test.
image

src/app/content/highlights/components/EditCard.tsx
RoyEJohnson: I think this is a good catch. I was not paying attention to what I wrapped there.

src/app/content/highlights/components/utils/useCallbackFor.ts
RoyEJohnson: Using an empty dependency list means define it once, not redefine it every time.

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
RoyEJohnson: This is the result of splitting the ActiveEditCard and the LoginConfirmation. You only get one or the other.

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
RoyEJohnson: This looks like the other case: not logged in, so you only get the login confirm part.

src/app/content/highlights/components/utils/useCallbackFor.ts
TomWoodward: yes, `return React.useRef(thing).current;` also only uses the first thing passed in and ignores updates. to update it every time you'd have to separately set `ref.current = thing` in each call. i think my suggestion in the earlier comment is functionally the same as yours but could be used in more ways because its not limited to functions

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
TomWoodward: oh i see, the test names don't make it very clear but i see in the `matches snapshot when focused` that its not setting a user

src/app/content/highlights/components/utils/useCallbackFor.ts
RoyEJohnson: If I return `ref.current` each time, it's still changing. I need a wrapper that doesn't change, and I really only needed it for function calls. I didn't want to change from calling a function to calling an object method or something.

src/app/content/highlights/components/EditCard.tsx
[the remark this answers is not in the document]
TomWoodward: well they're intended to all render at the same time, as long as there is space in the margin they show google docs style:

image

TomWoodward (changes requested):

src/app/content/highlights/components/utils/useCallbackFor.ts
TomWoodward: it does not change each time https://codesandbox.io/s/naughty-ellis-txftc6?file=/src/App.js

theres something up with the code sandbox it double renders initially and that causes the counter to start at two, but you can click the button and see that its rendering and not continuing to affect the ref or effect

RoyEJohnson: It looks like it's not needed after all. There's a quirk in lint that doesn't want `props.function` as a dependency and insists on `props` instead, but if I just assign the function to a variable, the dependency is ok. Everything works without being wrapped in `useCallbackFor`, so I've removed it.

src/app/content/highlights/components/Confirmation.tsx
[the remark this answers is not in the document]
RoyEJohnson: 💡  The EditCard mocks the Confirmation dialog, and the Confirmation spec only tests with defautl `drawFocus` of true.

src/app/content/highlights/components/__snapshots__/EditCard.spec.tsx.snap
```diff
@@ -219,22 +192,6 @@ exports[`EditCard matches snapshot with data 1`] = `
   data-analytics-region="edit-note"
   data-highlight-card={true}
 >
-  <div
```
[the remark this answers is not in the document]
RoyEJohnson: Added.

jivey (changes requested):
Sorry for just now looking at this PR, I saw the initial review from Tom and figured I should wait until those rounds were done. I pulled this down locally to test, but I'm still seeing a delay when adding a highlight to a page with a large amount of highlights:

https://github.com/openstax/rex-web/assets/34174/ceed5f6d-4816-461e-8774-5c013da46362

I'm also seeing warnings in the console:

```
Warning: LoginOrEdit: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop. (https://fb.me/react-special-props)
    in LoginOrEdit (at EditCard.tsx:54)
```

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `ForwardRef(_c)`.
    in LoginOrEdit (at EditCard.tsx:54)
```

src/app/content/highlights/components/Card.spec.tsx
```diff
@@ -181,7 +181,7 @@ describe('Card', () => {
       picker.props.onEdit();
     });
 
-    expect(() => component.root.findByType(EditCard)).not.toThrow();
+    expect(() => component.root.findAllByType(EditCard)).not.toThrow();
```
RoyEJohnson, line 184: Just occurred to me that this is probably not a valid test now.

jivey (approved):
I'm not sure if @TomWoodward has any other feedback, but this looks good to me and is testing well.

src/app/content/highlights/components/EditCard.tsx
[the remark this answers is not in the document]
TomWoodward: ok i'm beginning to understand the situation here. but now i think that you have two checks short-circuiting the edit card display, one here and one [here](https://github.com/openstax/rex-web/pull/2051/files#diff-019ddbca5476ddffda6217696eb10a2acd1b4a66c6e54843595a298e5c260af6R216), are both necessary?

TomWoodward (changes requested):

RoyEJohnson:
The `isActive` check that is external to EditCard just short-circuits to avoid unnecessary creation of a couple of callbacks. It's a micro-optimization that I'm fine with taking out.

TomWoodward (approved):

PR #2057, opened 2023-10-18
Restructure social media links for at

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2601
The last commit is just Prettier reformatting. The actual changes are all in the first commit.

### DIFF
--- a/src/app/components/Footer/index.tsx
+++ b/src/app/components/Footer/index.tsx
@@ -21,25 +21,25 @@ const Mission = htmlMessage('i18n:footer:copyright:mission-text', Styled.Mission
 const Copyrights = htmlMessage('i18n:footer:copyright:bottom-text', Styled.Copyrights);
 
 // tslint:disable-next-line:variable-name
-const BareMessage: React.SFC<{id: string}> = ({id}) => <FormattedMessage id={id}>
+const BareMessage: React.FunctionComponent<{id: string}> = ({id}) => <FormattedMessage id={id}>
   {(msg) => msg}
 </FormattedMessage>;
 
 // tslint:disable-next-line:variable-name
-const ColumnHeadingMessage: React.SFC<{id: string}> = ({id}) => <Styled.ColumnHeading>
+const ColumnHeadingMessage: React.FunctionComponent<{id: string}> = ({id}) => <Styled.ColumnHeading>
   <BareMessage id={id} />
 </Styled.ColumnHeading>;
 
 // tslint:disable-next-line:variable-name
-const FooterLinkMessage: React.SFC<{id: string, href: string, target?: string, rel?: string }> =
+const FooterLinkMessage: React.FunctionComponent<{id: string, href: string, target?: string, rel?: string }> =
   ({id, href, target, rel }) => <Styled.FooterLink href={href} target={target ? target : '_self'} rel={rel ? rel : ''}>
   <BareMessage id={id} />
 </Styled.FooterLink>;
 
 // tslint:disable-next-line:variable-name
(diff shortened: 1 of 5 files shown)


src/app/components/Footer/styled.tsx
```diff
@@ -86,7 +92,9 @@ export const boxed = css`
 `;
 
 const contentWrapperAndNavWidthBreakpoint =
-  '(max-width: ' + remsToEms(contentWrapperMaxWidth + verticalNavbarMaxWidth * 2) + 'em)';
+  '(max-width: ' +
+  remsToEms(contentWrapperMaxWidth + verticalNavbarMaxWidth * 2) +
+  'em)';
```
jivey, line 97: This one makes it harder to read to me, maybe use interpolation instead?

src/app/components/Footer/styled.tsx
```diff
@@ -95,13 +103,21 @@ export const FooterWrapper = styled.footer`
...
-        0rem, calc(${toolbarWidth}rem - (100vw - ${contentWrapperMaxWidth}rem) / 2), ${toolbarWidth}rem
-      );
-    }
-  ` : ``}
+  ${props =>
+    props.isVerticalNavOpen === false
+      ? css`
+          @media (min-width: ${theme.breakpoints
+              .desktopBreak}em) and ${contentWrapperAndNavWidthBreakpoint} {
+            padding-left: clamp(
```
jivey, line 111: The indentation here doesn't seem right, and to me, the original first line of the ternary reads much better without the extra line breaks and indentation.

src/app/components/Footer/index.tsx
```diff
@@ -16,71 +16,145 @@ export const supportCenterLink = 'https://help.openstax.org/s/';
...
 
 // tslint:disable-next-line:variable-name
-const SocialIconMessage: React.FunctionComponent<{id: string, href: string, Icon: React.ComponentType}> =
-  ({id, href, Icon}) => <Styled.SocialIcon aria-label={useIntl().formatMessage({id})} href={href}
-    target='_blank' rel='noopener'>
+const SocialIconMessage: React.FunctionComponent<{
+  id: string;
+  href: string;
+  Icon: React.ComponentType;
+}> = ({ id, href, Icon }) => (
```
jivey, line 64: I know they are doing slightly different things, but the newline rules for this and https://github.com/openstax/rex-web/commit/1f1c8efa0557e37c46c7e08e47d19bd6b924bde9?w=0#diff-43286799a0a570580be50717cd6ba2a7a59d87e249a707fad745e4222daeaa26R167 seem inconsistent

jivey (changes requested):
The icon changes look good, I'm not sure about the prettier changes though. I don't see a prettier file in root, but maybe we can adjust some settings. I left a few comments about some style things (that are subjective for sure.) I'm also wondering if we want to discuss a standard style as a team.

I use the built in formatter function that my LSP provides, and it mostly does what I want. That might be something the investigate for this, because as far as I understand, LSP formatters are more powerful?

src/app/components/Footer/styled.tsx
RoyEJohnson: This is sort of a nightmare of nesting.
There are some things I disagree with prettier about, but at least it is consistent, and IMO it does more good than harm. We get a lot of dense code.

RoyEJohnson:
I was clear whether you were requesting that the prettier changes all be backed out, or just that some issues be addressed. I addressed the first two issues you highlighted.

jivey (approved):

PR #2058, opened 2023-10-18
Persist search string when navigating internal links

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2244


src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -66,14 +95,25 @@ export const reduceReferences = (
...
+  const referenceMap: {
+    [key: string]: PageReferenceMap | PageReferenceError | undefined;
+  } = {};
 
-  for (const reference of references) { referenceMap[reference.match] = reference; }
+  for (const reference of references) {
+    referenceMap[reference.match] = reference;
+  }
 
-  const options = createNavigationOptions(systemQueryParams as SystemQueryParams);
```
Dantemss, line 73: These lines the linter is breaking down are really not that long...

src/app/content/components/Page/contentLinkHandler.ts
```diff
@@ -96,73 +136,98 @@ export const reduceReferences = (
...
-    }
-  };
+  if (hasUnsavedHighlight && !(await showConfirmation(services))) {
+    return;
+  }
+
+  const additionalQueryString = new URLSearchParams(
+    persistentQueryParams as Record<string, string>
+  ).toString();
+  const extendedSearchString = (search + '&' + additionalQueryString).substring(
```
Dantemss, line 204: You might have to do something when `additionalQueryString` is empty so there's not a `&` at the end?

src/app/content/components/Page/contentLinkHandler.spec.ts
```diff
@@ -152,6 +152,7 @@ describe('contentLinkHandler', () => {
 
     it('intercepts clicking content links with slug', async() => {
       const link = `/books/${book.slug}/pages/page-title`;
+      prop.currentPath = '/asdf?foo=bar';
```
Dantemss, line 155: Is this unrelated? This seems to be the only instance in the PR of something changing that has to do with currentPath.

Dantemss (changes requested):
Since there's already a test file, it may be nice to add a test to show what changed.

src/app/content/components/Page/contentLinkHandler.spec.ts
RoyEJohnson: I modified the test so that I could have one with a pre-existiing search string.

src/app/content/components/Page/contentLinkHandler.ts
RoyEJohnson: queryString.parse cleans this up

src/app/content/components/Page/contentLinkHandler.spec.ts
RoyEJohnson: I further modified the test to check the search string result.

Dantemss (approved):

PR #2059, opened 2023-10-19
 updates

staxbot:


theresamasciale:
I'm getting an error when I try to view the deployment. I checked on two different browsers and I get the same error.

Screenshot 2023-10-19 at 9 22 55 AM

theresamasciale:
@RoyEJohnson I'm wondering if this is because the bookpage isn't published yet (since we're waiting for the hotfix). I've made it a meeting point to discuss with disco at 10. Maybe they can redo the PR once the hotfix goes out and it's published? (if that is the issue here)

PR #2062, opened 2023-10-23
Fix aria roles per Lighthouse

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2628

### DIFF
--- a/src/app/components/NavBar/index.tsx
+++ b/src/app/components/NavBar/index.tsx
@@ -33,33 +33,50 @@ export const Dropdown: FunctionComponent<{user: User, currentPath: string}> = ({
     }
   };
 
-  return <OnScroll callback={blockScroll}>
-    <Styled.DropdownOverlay tabIndex='-1' ref={overlay} data-testid='nav-overlay'>
-      <a aria-hidden='true' tabIndex={-1} href='/'>
-        <Styled.OverlayLogo
-          src={openstaxLogo}
-          alt={useIntl().formatMessage({id: 'i18n:nav:logo:alt'})}
-        />
-      </a>
-      <div>
-        <FormattedMessage id='i18n:nav:hello:text' values={{name: user.firstName}}>
-          {(msg) => <Styled.OverlayHeading>{msg}</Styled.OverlayHeading>}
-        </FormattedMessage>
-        <Styled.DropdownList>
-          <li>
-            <FormattedMessage id='i18n:nav:profile:text'>
-              {(msg) => <a href='/accounts/profile' target='_blank'>{msg}</a>}
-            </FormattedMessage>
-          </li>
-          <li>
-            <FormattedMessage id='i18n:nav:logout:text'>
-              {(msg) => <a href={'/accounts/logout?r=' + currentPath}>{msg}</a>}
-            </FormattedMessage>
-          </li>
-        </Styled.DropdownList>
-      </div>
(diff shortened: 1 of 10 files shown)


src/app/content/components/BuyBook.tsx
```diff
@@ -47,13 +47,11 @@ const BuyPrintDisclosure = styled.p`
 // tslint:disable-next-line: variable-name
 const BuyBook = () => {
   const config = useSelector(buyPrintConfig);
-  const formatMessage = useIntl().formatMessage;
 
   if (!config) { return null; }
 
   return <BuyBookAlignment>
     <BuyBookLink
-      aria-label={formatMessage({id: 'i18n:toolbar:buy-book:aria-label:text'})}
```
RoyEJohnson, line 56: It doesn't need aria-label, since it has a text value.
Lighthouse didn't like the text value being different from the aria-label value.

src/app/content/components/LabsCall.tsx
```diff
@@ -160,7 +160,7 @@ const LabsCTA = () => {
     <LabsCallWrapper data-async-content>
       <LabsCallHeader>
         <Column noShrink>
-          <LabsLogo src={icon} role='img' alt='' />
```
RoyEJohnson, line 163: `The ARIA img role can be used to identify multiple elements inside page content that should be considered as a single image.`
The element is a single `` tag. It does not need the `img` role.

src/app/components/NavBar/index.tsx
```diff
@@ -33,33 +33,50 @@ export const Dropdown: FunctionComponent<{user: User, currentPath: string}> = ({
...
+            id='i18n:nav:hello:text'
+            values={{ name: user.firstName }}
+          >
+            {msg => <Styled.OverlayHeading>{msg}</Styled.OverlayHeading>}
+          </FormattedMessage>
+          <Styled.DropdownList id='dropdown-menu' role='menu'>
+            <li role='presentation'>
+              <FormattedMessage id='i18n:nav:profile:text'>
+                {msg => (
+                  <a href='/accounts/profile' target='_blank' role='menuitem'>
```
RoyEJohnson, line 60: The reformatted block may make it hard to find the changes. I added the role attribute here

src/app/components/NavBar/index.tsx
```diff
@@ -33,33 +33,50 @@ export const Dropdown: FunctionComponent<{user: User, currentPath: string}> = ({
...
+                    {msg}
+                  </a>
+                )}
+              </FormattedMessage>
+            </li>
+            <li role='presentation'>
+              <FormattedMessage id='i18n:nav:logout:text'>
+                {msg => (
+                  <a href={'/accounts/logout?r=' + currentPath} role='menuitem'>
+                    {msg}
```
RoyEJohnson, line 70: also added role attribute here (and the `presentation` role to the enclosing `li`)

src/app/content/components/BuyBook.tsx
jivey: I also like that the normal text is more descriptive to the screen reader as well. This appears to be the only thing that used it, can we delete the key from the messages files too?

jivey (changes requested):
Looking great! One tiny thing, I think we can delete `i18n:toolbar:buy-book:aria-label:text` from the messages files.

jivey (approved):

PR #2066, opened 2023-10-30
Label login nudge close button; force focus to stay on it

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2589


src/app/content/components/NudgeStudyTools/index.tsx
```diff
@@ -121,6 +128,44 @@ const NudgeStudyTools = () => {
...
+function CloseButtonHoldingFocus({
+  positions,
+  dismiss,
+}: {
+  positions: Exclude<ReturnType<typeof usePositions>, null>;
+  dismiss: () => void;
+}) {
+  const ref = React.useRef<HTMLElement>();
+
+  React.useEffect(
```
jivey, line 140: Suggestion: this seems like a helpful hook, it might be nice if this lived somewhere in utils that we could import in other overlays.

src/app/content/components/NudgeStudyTools/index.tsx
```diff
@@ -33,19 +33,38 @@ import {
...
+  if (!positions) { return null; }
+
+  return (
+    <NudgeStudyTools2 positions={positions} />
+  );
+}
+
 // tslint:disable-next-line: variable-name
-const NudgeStudyTools = () => {
+const NudgeStudyTools2 = ({
```
jivey, line 48: This seems harder to read, is splitting out this component necessary since we have the early return? Alternatively, maybe we adjust the var name, this makes me think it's a versioned component at first glance.

jivey (changes requested):
Looking good, just a thought about the readability of the split component.

src/app/content/components/NudgeStudyTools/index.tsx
RoyEJohnson: The reason for splitting is to allow the short-circuit to avoid all the hooks that are associated with the real component. Sometimes I name these pretesting components like MaybeNudgeStudyTools. In this case, I'm just moving the test up to the caller that already had another short-circuit defined.

src/app/content/components/NudgeStudyTools/index.tsx
RoyEJohnson: I didn't find another file where I thought it belonged, but I did separate it out as a hook within the file.

src/app/content/components/NudgeStudyTools/index.tsx
RoyEJohnson: You can't return before hooks or you get warnings about [conditional hooks](https://dev.to/lexlohr/conditional-hooks-1fpj).

jivey: Haha I never remember this until the linter yells at me about it, thanks for the reminder :)

jivey (approved):
Looks good, I don't know why the build is failing though, it seems unrelated.

RoyEJohnson:
Had to separate it back out (used "Maybe" for name) because the top level screens for presence of `window` which the mobile media detector requires.

jivey (approved):

PR #2067, opened 2023-11-01
Do not auto select first search result

RoyEJohnson:
For: https://github.com/openstax/unified/issues/1700
It may be helpful to look at the 2nd and 4th commits individually. The 2nd is where I made the functional change. The 4th is where I removed pieces of code that it _seemed_ were no longer involved, due to the functional change. I was unable to get the tests to cover them, but that could be due to my limited grasp of the mechanisms.


Dantemss (approved):
The changes more or less makes sense to me but I'm mostly trusting your code.

src/app/content/search/hooks/index.ts
```diff
@@ -63,24 +62,13 @@ export const receiveSearchHook: ActionHookBody<typeof receiveSearchResults> = (s
...
-    selectedResult
-    // We are clearing selected result when requesting a new search so in the theory this should never happen
-    && (isEqual(select.selectedResult(state), selectedResult)
-      // selectedResult bookId data is different than current book id
-      || book.id !== getIndexData(selectedResult.result.index).bookId)
-  ) {
-    return;
-  }
-
   if (selectedResult) {
```
RoyEJohnson, line 65: @TomWoodward I added you as a reviewer so you could sanity check some decisions I made, such as removing the code above and below.

PR #2068, opened 2023-11-02
Use polyfll for Intl-DisplayNames

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2650 Since it's only outdated browsers that need the polyfill, it didn't seem worth it to make the whole attribution chain async. Instead, if the polyfll is needed, it just lists the locale for the Language.

### DIFF
--- a/src/app/content/components/utils/attributionValues.spec.ts
+++ b/src/app/content/components/utils/attributionValues.spec.ts
@@ -9,8 +9,8 @@ jest.mock(
     })
 );
 
-describe('locale not found', () => {
-  it('loads english messages', () => {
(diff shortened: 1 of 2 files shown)


Dantemss (approved):
Seems acceptable to me but @erikayvonne should confirm.

Summary: Modern browsers will display `Language: English` etc in the attribution but older browsers will display `Language: en`. This is because the feature we were using to map the locale `en` to the display name `English` is missing in older browsers and was causing errors.

erikayvonne:
thanks @Dantemss - this is good to go on my end.

PR #2079, opened 2023-11-30
When scrolling element into view, compensate for images loading

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2673
For: https://github.com/openstax/unified/issues/2683

### DIFF
--- a/src/app/content/components/Page/PageContent.tsx
+++ b/src/app/content/components/Page/PageContent.tsx
@@ -150,6 +150,8 @@ export default styled(MainContent)`
 
   .os-figure,
   .os-figure:last-child {
+    table-layout: fixed;
+    width: 100%;
     margin-bottom: 5px; /* fix double scrollbar bug */
   }
 
--- a/src/app/domUtils.ts
+++ b/src/app/domUtils.ts
@@ -76,7 +76,7 @@ export const findFirstAncestorOrSelf = <T = HTMLElement>(
   predicate: ((e: any) => boolean) | ((e: any) => e is T)
 ): T | void => {
   if (isHtmlElement(node) && predicate(node)) {
-    return node;
+    return node as T;
   } else if (node.parentElement) {
     return findFirstAncestorOrSelf(node.parentElement, predicate);
   }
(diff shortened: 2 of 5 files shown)


src/app/domUtils.ts
```diff
@@ -97,7 +97,17 @@ const getScrollPadding = () => {
...
+
+  document?.addEventListener(
+    'load', scroll, true
+  );
+
+  setTimeout(
+    () => document?.removeEventListener('load', scroll, true),
+    3000
+  );
+  return scroll();
```
jivey, line 110: This looks like a double fire would occur under load conditions. Would it make sense to use a `document.readyState === 'complete'` check to split these? That would also let you skip the listeners entirely when the document is loaded.

src/app/content/components/Page/PageContent.tsx
```diff
@@ -150,6 +150,8 @@ export default styled(MainContent)`
 
   .os-figure,
   .os-figure:last-child {
+    table-layout: fixed;
```
jivey, line 153: If I'm reading the spec right, won't this fix the column width, which could change how some content is displayed? I'm wondering if it could cause overflow issues. I'm also wondering if this is for render performance to help page load?

jivey (changes requested):

src/app/domUtils.ts
RoyEJohnson: This is listening for images that lazy load, which happens as the page scrolls, throwing off the coordinates of the scroll target.
I can debounce the listener to avoid rapid multiple fire (in testing, I saw something like 97 image load events on a run), but I don't notice any janky behavior from the multiple calls.

src/app/content/components/Page/PageContent.tsx
RoyEJohnson: If you go to [this page](https://rex-web-production.herokuapp.com/books/workplace-software-skills@c795d5f/pages/preface?archive=https%3A%2F%2Fopenstax.org%2Fapps%2Farchive-preview%2F20231109.173216) and shift-scroll, you will see the main content area scroll horizontally.
When you scroll to the bottom of the page, the images there load, and the main content width resets.

src/app/domUtils.ts
RoyEJohnson: As an example, if you go [here](https://rex-web-production.herokuapp.com/books/microbiology/pages/8-3-cellular-respiration), search for "chemical reaction", and then select the result (in section 8.1) that begins "Overall," the target result will scroll a bit off the top of the screen.

jivey (approved):
Thanks for the explainers! They are also nice as a form of documentation for future us 📓

PR #2087, opened 2023-12-09
FE playwright test: validate ch dropdown count in MH

Malar-Natarajan:
for: https://app.zenhub.com/workspaces/openstax-unified-5b71aabe3815ff014b102258/issues/openstax/unified/2721

Assert chapter count in MH chapter dropdown filter


playwright/src/utilities/user.ts
```diff
@@ -62,12 +62,17 @@ async function accountsUserSignup(page: Page, url = '', student: Student = new S
   await page.fill('[placeholder="Last name"]', student.last)
   await page.fill('[placeholder="me@myemail.com"]', student.email)
   await page.fill('[placeholder="Password"]', student.password)
-  await page.check('#signup_terms_accepted')
-  await page.click('text=Continue')
```
Malar-Natarajan, line 66: @gregfitch the mobile tests (for iPhone 12) are not picking these existing locators. They dont fail for invisible/inactive locators; looks as if they just skip those lines and moves to the next and when they are not able to find the locator in the new page the test times out. If I change the TOS checkbox & Continue buttons to `page.evaluate` then it works. Now its doing the same for `Finish` button.
Have you encountered this issue before?

gregfitch: (Assuming you mean webkit/Safari) I haven't, but using `evaluate` there (_not really what it's meant for_) means that it waits for whatever Promise it is given as an argument. If I had to guess, I would say the webkit version is acting before the DOM element is 'ready' for the action or command (ie a race condition - it sees the element so it sends the action, but the action isn't actually completed). Using evaluate is probably giving the site extra time to get its ducks in a row before sending the action (kind of like making a script sleep for a moment). This ***shouldn't*** happen, but if it is...webkit has always been the difficult child in automated testing.

Malar-Natarajan: I am thinking to proceed with the `evaluate` and if I still see frequent webkit failures in the CI checks, disable webkit from CI.

playwright/src/fixtures/MHmodal.ts
```diff
@@ -33,10 +33,20 @@ class MHModal {
   async chapterDropdownCount() {
     // Total number of checkboxes in the chapter dropdown
     const Toc = new TOC(this.page)
-    const chapterDropdownCount = (await Toc.chapterCount()) + (await Toc.standalonePagesCount())
+    const chapterDropdownCount =
+      (await Toc.chapterCount()) +
+      (await Toc.standalonePagesCount()) +
+      (await Toc.unitIntroCount()) +
+      (await Toc.eobDropdownCount())
     return chapterDropdownCount
```
gregfitch, line 41: As you're not doing anything to adjust `chapterDropdownCount`, just return the value instead of assigning it to a constant and then returning the constant.

playwright/src/fixtures/MHmodal.ts
```diff
@@ -33,10 +33,20 @@ class MHModal {
...
   }
 
+  async chapterDropdownText(n: number) {
+    // Chapter name listed under the chapter dropdown
+    const chapterName = await this.checkBoxStatus.nth(n).textContent()
+    return chapterName
```
gregfitch, line 47: same as above

gregfitch (changes requested):

gregfitch (approved):

PR #2089, opened 2023-12-14
Trap focus in delete-highlight modal

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2698


bethshook (approved):

src/app/reactUtils.ts
```diff
@@ -18,6 +18,45 @@ export const useDrawFocus = <E extends HTMLElement = HTMLElement>() => {
...
+// Exported to allow testing
+export function createTrapTab(ref: React.MutableRefObject<HTMLElement | null>) {
+  return (event: KeyboardEvent) => {
+    if (event.key !== 'Tab') { return; }
+
+    const el = ref.current;
+    if (!el) { return; }
+
+    const focusableItemQuery =
+      'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
```
jivey, line 30: Should we skip `disabled` attributes here? I'm thinking disabled-but-visible inputs could be a concern.

src/app/reactUtils.spec.tsx
```diff
@@ -243,6 +244,91 @@ describe('useOnDOMEvent', () => {
...
 
+describe('useTrapTabNavigation', () => {
+  let window;
+  let addEventListener: jest.SpyInstance;
+
+  beforeEach(() => {
+    window = assertWindow();
+
+    addEventListener = jest.spyOn(window.document, 'addEventListener');
+    // removeEventListener = jest.spyOn(window.document, 'removeEventListener');
```
jivey, line 255: Seems like we can remove this

jivey (changes requested):
This is testing well and overall looks good to me, I have a question about skipping disabled elements in the selector.

jivey (approved):

PR #2091, opened 2023-12-15
Add margin to DisplayNote to account for side toolbar

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2640


jivey (approved):

jivey (changes requested):
I already approved but I realized one small tweak, it's probably good for us to use `verticalNavbarMaxWidth` like the toolbar so they are in sync if the toolbar width ever changes.

jivey (approved):

PR #2095, opened 2023-12-18
Test for mobile device rather than screen resolution

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2645


Dantemss (changes requested):
Tom says not to rely on the User Agent. Maybe some combination of screen size and `pointer`? https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer

TomWoodward (changes requested):
left a comment on the parent issue

PR #2097, opened 2024-01-02
Add aria-label and aria-expanded attributes to DotMenuToggle

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2720
For: https://openstax.atlassian.net/browse/DISCO-106


src/app/components/DotMenu.tsx
```diff
@@ -19,13 +19,22 @@ export const DotMenuIcon = styled(EllipsisV)`
...
-  </PlainButton>
-))`
+export const DotMenuToggle = styled(
+  React.forwardRef<HTMLDivElement>((props, ref) => {
+    const {isOpen} = props;
+
+    return (
+      <PlainButton {...{
+        'aria-label': 'Actions',
+        'aria-expanded': isOpen,
```
TomWoodward, line 29: you don't have to embed these in the `props` spread, you can define them as normal react props either before or after the `{...props}` depending on which way you want to override

TomWoodward (changes requested):

TomWoodward (approved):

jivey (approved):

PR #2099, opened 2024-01-03
Add default focus style to styled Checkbox

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2713

### DIFF
--- a/src/app/components/Checkbox.tsx
+++ b/src/app/components/Checkbox.tsx
@@ -48,6 +48,9 @@ const Checkbox = ({children, className, ...props}: React.PropsWithChildren<Props
   {children}
 </label>;
 
+// Browser default outline for focus items per
+// tslint:disable-next-line:max-line-length
+// https://stackoverflow.com/questions/20609485/what-is-the-default-style-of-the-blue-focus-outline-in-chrome/38498539#38498539
 const defaultFocusOutline = `
   outline: 0.1rem dotted #212121;
   outline: 0.5rem auto -webkit-focus-ring-color;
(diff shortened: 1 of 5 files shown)


src/app/components/Checkbox.tsx
```diff
@@ -48,6 +48,11 @@ const Checkbox = ({children, className, ...props}: React.PropsWithChildren<Props
   {children}
 </label>;
 
+const defaultFocusOutline = `
+outline: 0.1rem dotted #212121;
+outline: 0.5rem auto -webkit-focus-ring-color;
```
TomWoodward, line 53: indent the styles by one please

src/app/components/Checkbox.tsx
```diff
@@ -48,6 +48,11 @@ const Checkbox = ({children, className, ...props}: React.PropsWithChildren<Props
   {children}
 </label>;
 
+const defaultFocusOutline = `
+outline: 0.1rem dotted #212121;
```
TomWoodward, line 52: the dotted style is the firefox style, chrome has a blue gradient thing, is it possible to use the default focus styles here implicitly or do we have to use a custom style? or is that what you're doing by defining both the dotted and the webkit at the same time like this?

TomWoodward (changes requested):

src/app/components/Checkbox.tsx
RoyEJohnson: I depended on [this StackOverflow post](https://stackoverflow.com/a/38498539), which suggests that you have to spec it. `initial` does not work.

TomWoodward: ah ok... yea i don't see a simpler solution here since we're doing a hidden-checkbox situation. you may want to just throw that stack overflow link in a comment above the `defaultFocusOutline` style

TomWoodward (approved):

jivey (approved):

Malar-Natarajan:
@RoyEJohnson The CI check `Lint` is failing in this PR. Could you please resolve. Thanks!

PR #2100, opened 2024-01-04
Allow tabbing to vertical menu items while nudge is displayed

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2691
Added fix for: https://github.com/openstax/unified/issues/2703

### DIFF
--- a/src/app/content/components/NudgeStudyTools/index.tsx
+++ b/src/app/content/components/NudgeStudyTools/index.tsx
@@ -32,6 +32,7 @@ import {
   useIncrementPageCounter,
   usePositions,
 } from './utils';
+import { nudgeStudyToolsTargetId } from './constants';
 
 // tslint:disable-next-line: variable-name
 const NudgeStudyTools = ({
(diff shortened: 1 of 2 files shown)


TomWoodward (approved):

src/app/content/components/NudgeStudyTools/index.tsx
```diff
@@ -119,19 +120,40 @@ const NudgeStudyTools = ({
 
 function useTabNavigationInterceptor() {
   const ref = React.useRef<HTMLElement>();
+  const isMobile = useMatchMobileMediumQuery();
 
   React.useEffect(
     () => {
-      const abortTabbing = (event: Event) => {
+      if (isMobile) {
```
jivey, line 127: Does this need to exclude mobile? I'm thinking external keyboard, iPad narrow window view, etc.

RoyEJohnson: Looks like keyboard navigation does need to be supported for mobile.
https://bootcamp.uxdesign.cc/keyboard-accessibility-mobile-devices-688dd73bdb7b

RoyEJohnson: The bad news is that I don't get keydown events from the browser when in responsive mode. :(

jivey: I wonder if the xcode ios simulator would work for testing this 🤔

RoyEJohnson: It got merged, so maybe we'll want a new issue for the mobile.

PR #2101, opened 2024-01-04
Pad selected search-highlight so it is visibly taller than unselected

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2696


src/app/content/components/Page/PageContent.tsx
```diff
@@ -139,6 +139,7 @@ export default styled(MainContent)`
       &.focus {
         ${SELF_AND_CHILD_MATH_SELECTOR} {
           background-color: #ff9e4b;
+          padding: 0.2rem 0;
```
TomWoodward, line 142: we want this to apply to user highlights as well right? maybe we should move it up and add an `&.focus` [here](https://github.com/openstax/rex-web/pull/2101/files#diff-b3c6f238c5a4a53db657bc6f4ac089bc43d6735b106368460a1b0a2c0e2096ddR54)

TomWoodward (changes requested):

src/app/content/components/Page/PageContent.tsx
RoyEJohnson: I added it to the highlight.color.focus spec.

TomWoodward (approved):

jivey (approved):

PR #2102, opened 2024-01-05
Focus on TOC when TOC is opened

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2711
Also added an aria-label to HighlightsHelpInfo close button

### DIFF
--- a/src/app/content/components/TableOfContents/index.spec.tsx
+++ b/src/app/content/components/TableOfContents/index.spec.tsx
@@ -78,11 +78,17 @@ describe('TableOfContents', () => {
       store.dispatch(actions.closeToc());
     });
     expect(component.root.findByType(TableOfContents).props.isOpen).toBe(false);
+
+    const {root} = renderToDom(<TestContainer store={store}>
+      <ConnectedTableOfContents />
+    </TestContainer>);
+    const sb = root.querySelector('[data-testid="toc"]');
+
     renderer.act(() => {
       store.dispatch(actions.openToc());
+      sb?.dispatchEvent(new Event('transitionend'));
     });
     expect(component.root.findByType(TableOfContents).props.isOpen).toBe(true);
-    jest.runAllTimers();
   });
 
   it('resets toc on navigate', () => {
--- a/src/app/content/components/TableOfContents/index.spec.tsx
+++ b/src/app/content/components/TableOfContents/index.spec.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { unmountComponentAtNode } from 'react-dom';
+import { act as reactDomAct } from 'react-dom/test-utils';
 import renderer from 'react-test-renderer';
 import ConnectedTableOfContents, { TableOfContents } from '.';
 import createTestStore from '../../../../test/createTestStore';
@@ -7,13 +8,13 @@ import { book as archiveBook, page, shortPage } from '../../../../test/mocks/arc
 import { mockCmsBook } from '../../../../test/mocks/osWebLoader';
 import { renderToDom } from '../../../../test/reactutils';
 import TestContainer from '../../../../test/TestContainer';
+import * as reactUtils from '../../../reactUtils';
 import { AppState, Store } from '../../../types';
 import { assertWindow } from '../../../utils';
 import * as actions from '../../actions';
 import { initialState } from '../../reducer';
 import { formatBookData } from '../../utils';
 import * as domUtils from '../../utils/domUtils';
-import * as reactUtils from '../../../reactUtils';
 
 const book = formatBookData(archiveBook, mockCmsBook);
 
@@ -62,7 +63,6 @@ describe('TableOfContents', () => {
     expect(scrollSidebarSectionIntoView).toHaveBeenCalledTimes(2);
   });
 
-  jest.useFakeTimers();
   it('opens and closes', () => {
     jest.spyOn(reactUtils, 'useMatchMobileQuery')
       .mockReturnValue(true);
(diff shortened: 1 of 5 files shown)


TomWoodward (approved):

jivey (approved):
Looks good, I think the spec failure is flaky, I reran it and hopefully it passes 🤞

RoyEJohnson:
@jivey Malar remarked that the mobile TOC doesn't trap tab navigation, so I added that, which involved some refactoring, so I thought you should review the 2nd commit, which is what has changed since you approved.

jivey (commented):
@RoyEJohnson Can you describe the focus issue that means we need to resort to useTimeout? Maybe the animation is still running? If there's no other way that's fine, I'm just curious why.

RoyEJohnson:
@jivey I'm not sure, but I think the issue is that the children get a re-rendered after the effect runs, which loses the focus. I was able to move the move the effect to the SidebarBody (one level down) but it didn't help. I might be able to move it into the TocSection, where it shouldn't have that issue.

jivey:
@RoyEJohnson Yeah if moving it to a different component gets around it that seems better to me? I guess I worry about the timeout being brittle. Maybe @TomWoodward has other thoughts.

RoyEJohnson:
@jivey I'm going to say it's an issue with the animation. Though it works with any time value >= 50, I'm changing it to use `sidebarTransitionTime` for extra safety and also as a hint as to why it's there.

jivey:
@RoyEJohnson Nice, that seems like a nice way to handle it 👍

jivey (approved):

TomWoodward (changes requested):
the toc is never actually removed from the dom, its only visually hidden with css when closed, this is because the default state of the toc (when tocOpen=null) will either be open or closed depending on a pure css media query, this is so the toc state is correct before JS loads.

so all that said, i am concerned about that setTimeout, because the transition time isn't even a JS thing its a css thing. if the problem is that there is a re-render in the sidebar somewhere as it opens that is resetting the focus, i wonder if that render is really necessary? i feel like we need to understand this a bit better and see if we can figure it out without the timeout

RoyEJohnson:
Observations:
If the TOC is open initially (isTocOpen is null), focus can be set immediately with no need for timeout.
If the TOC is not already open (isTocOpen is false), focus seemingly cannot be set until the open transition completes.
If the TOC has been opened (isTocOpen is true) then clicking doesn't change state, so the effect does not fire.

By attaching a `transitionend` event listener to the sidebar, I can get the focus to work for the 2nd case. I'm not clear how important it is to have it move focus into the TOC when the TOC is already open and the control isn't really doing anything.

TomWoodward:
@RoyEJohnson in your first and third cases, we wouldn't want to set the focus anyway right? in the first case this is just a desktop page load and we don't want to fuss with the focus, and in the third case they'd be trying to close the toc.

i wonder if something about the transition is preventing the browser from allowing you to focus it? like maybe you can't focus it if the opacity is not 1? if you're able to handle it with an event listener that sounds great.

RoyEJohnson:
In the every case, I'm talking about if you click the button. So the page has just loaded, and you click the button, nothing happens. This is, in particular, for a [page that isn't found](https://rex-web-production.herokuapp.com/books/university-physics-volume-1/pages/1-5-esXtimates-and-fermi-calculations), where you have the "Try another page in the Contents" button.

TomWoodward:
well normally clicking the button on the left would close the ToC, the not found page is special that it has that open link even if the thing is open already. i'm not sure how that was ever intended to work. for it to trigger focus we'd have to change more about how that button works. maybe this should be a separate card for that.

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +24,185 @@ interface SidebarProps {
...
+        'ol > li a, old > li summary'
+      ) as HTMLElement;
+      const el = mRef.current;
+      const transitionListener = () => {
+        firstItemInToc?.focus();
+      };
+
+      if (props.isTocOpen) {
+        // This is primarily for code coverage
+        transitionListener();
```
TomWoodward, line 75: is this still necessary since you figured out the event thing?

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +24,185 @@ interface SidebarProps {
   page?: Page;
 }
 
+function TabTrapper({
```
TomWoodward, line 27: is there a reason for this to be a component and not just a hook?

TomWoodward (changes requested):

src/app/content/components/TableOfContents/index.tsx
RoyEJohnson: Yes, `transitionListener` otherwise doesn't get called because the `transitionEnd` event on the element doesn't happen in testing.

src/app/content/components/TableOfContents/index.tsx
RoyEJohnson: Conditional execution. [Check for mobile](https://github.com/openstax/rex-web/pull/2102/files#diff-26081ab52cd6866f337549a6aa3caad9d12eb13241e23c921b76e6f51c746b35R34) fails if window is undefined.

Malar-Natarajan:
@RoyEJohnson are there any pending changes to be done here. The merge button is blocked for this reason.

TomWoodward (approved):
i pushed a commit fixing the test for transitionend and coverage, looks good otherwise

src/app/content/components/TableOfContents/index.spec.tsx
```diff
@@ -71,13 +75,45 @@ describe('TableOfContents', () => {
...
+    });
+
+    expect(focusSpy).not.toHaveBeenCalled();
+
+    reactDomAct(() => {
+      store.dispatch(actions.openToc());
+    });
+    reactDomAct(() => {
+      sb?.dispatchEvent(new Event('transitionend'));
+    });
```
TomWoodward, line 112: the trick was that the event listener isn't added until the thing re-renders with the open prop, so you need to have the event dispatch in a separate act call after you update the open state

PR #2103, opened 2024-01-08
Ensure toasts are annnounced by screen readers

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2712
Load ToastContainerWrapper unconditionally, and mark as an alert region.


TomWoodward (changes requested):
the documentation says:

> Importantly, the element with the ARIA live attributes must be available when the page loads. Otherwise, many screen readers will not detect updates to the element. Additionally, the element must be empty on page load unless an immediate screen reader announcement is desired.

i'm concerned that the alerts won't work unless we add the alert container directly to the index.html template. did you investigate this?

Dantemss (commented):
I'll defer to Tom's review on this one. It may also be helpful to have QA look at this with a screenreader so we know how it behaves.

RoyEJohnson:
The [documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role) only warns that you can't create the alert when you need to use it; that it has to change to be noticed. The screenreader I'm running does encounter the alert and closes it after reading it.
Certainly we'd want QA to confirm that they are getting the desired behavior.

TomWoodward:
i just want to make sure it works, i'd certainly rather not deal with having only one alert container defined in the index.html that everything has to share access to

RoyEJohnson:
Is there a process for having QA test it with a screenreader before approving?

TomWoodward:
@Malar-Natarajan can test it in the heroku env during the normal review phase

TomWoodward (changes requested):

src/app/content/components/Page/PageToasts.tsx
```diff
@@ -61,10 +62,13 @@ export const ToastContainerWrapper = styled.div`
 const PageToasts = (props: ToastProps | {}) => {
   const toasts = useSelector(groupedToastNotifications).page;
   const mobileToolbarOpen = useSelector(mobileToolbarOpenSelector);
+  const ref = React.useRef<HTMLElement | null>(null);
```
TomWoodward, line 65: this is unused now

TomWoodward (changes requested):

TomWoodward (approved):
as long as the alert works, looks fine to me

PR #2104, opened 2024-01-08
Restrict images in figures to context text width

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2673

### DIFF
--- a/generic-styles/content-common.less
+++ b/generic-styles/content-common.less
@@ -75,7 +75,7 @@ figure {
 
   // do not wrap this in a .media so it applies to editing in Aloha
   img {
-    max-width: 100%;
+    width: 100%;
     margin: 0; //overrides top&bott margin on img, all img EXCEPT those in figure tags should have 1em top&bott margin
   }
 


TomWoodward (changes requested):
styling content should be done in the content styles (ideally the baked styles in cnx-recipes, but also for now in the rex content `.less` files) and why does the need an explicit width to make it work? shouldn't it be responsive?

RoyEJohnson:
The attribute width is something Chrome/lightouse wants for layout efficiency.
The issue being solved is a weird glitch: before the image is loaded, the attribute width is respected, causing the horizontal overflow. After the image loads, the image is a reasonable size and the overflow goes away.
I'll see if I can make the change in the .less file. There is also a change being made in recipes, I think.

TomWoodward:
if i change the `figure img` to be `width: 100%;` instead of `max-width: 100%` i think its behaving better, do you see that as well?

RoyEJohnson:
Yes, that works for me.

RoyEJohnson:
Does that mean it is a change that should be made on the Content side?

TomWoodward:
i tried testing it on a figure image that was more narrow and it still seemed to look right, but i would appreciate some more verification. and yes the change would be in both the rex `.less` files (for non-baked style books) and in cnx-recipes (for most books)

TomWoodward (approved):
there is a page later in the book that beth linked that has some smaller image figures and they all look fine, i think this is ok 

https://rex-web-another-shot-at-zq8b9b.herokuapp.com/books/workplace-software-skills@c795d5f/pages/1-1-computing-from-inception-to-today?archive=https%3A%2F%2Fopenstax.org%2Fapps%2Farchive-preview%2F20231109.173216

note that we also need to make the change in cnx-recipes before we can close the parent issue

RoyEJohnson:
cnx-recipes, not ce-styles? Is that why my attempt to do it in ce-styles didn't seem to work?

TomWoodward:
i thought the styles were here https://github.com/openstax/cnx-recipes/blob/main/styles/books/generic/webview.scss but they might have moved to a new repo at some point.

PR #2140, opened 2024-01-22
Hide EditCard from touch devices, not from narrow screens

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2645
Also fixes: https://github.com/openstax/unified/issues/2749


jivey (approved):
I didn't follow all of it, but it seems from the card notes that this behavior is testing correctly.

PR #2158, opened 2024-01-26
Use-mark-instead-of-span

RoyEJohnson:
### This is just a test branch. It should not be merged
For: https://github.com/openstax/unified/issues/2647
Also for: openstax/unified#2718


jivey (commented):
Do we know if there are disadvantages to this resolving to codeload? I usually use the full URL so I've never seen it before. Maybe it's fine, I couldn't find a lot of info about it, just this: https://stackoverflow.com/questions/60188254/how-is-codeload-github-com-different-to-api-github-com

RoyEJohnson:
I do not know. I notice that the pattern-library in osweb also resolves to codeload.
Are the tags created this way going to be the standard way we release Highlighter now? Will they be merged to main there at some point and "real" tags will be created from the main branch?

jivey:
That's a good @TomWoodward question, is the plan is to move to the releases to GitHub tags? Or should I push a new build to NPM? I think only REX and Tutor use highlighter, so it's probably easy to point things to the GitHub tag.

TomWoodward:
idk but other apps use the full github url in the package.json so maybe that is why its resolving this way?

jivey:
Yeah for consistency I'd recommend using the full url

RoyEJohnson:
Updated to use the full url.

jivey (approved):

RoyEJohnson:
https://github.com/openstax/rex-web/pull/2176 uses the same Highlighter branch.

PR #2159, opened 2024-01-26
Constrain tab navigation in toc

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2753


src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +24,171 @@ interface SidebarProps {
...
+  isTocOpen,
+}: {
+  mRef: MutableRefObject<HTMLElement>;
+  isTocOpen: boolean;
+}) {
+  const isPhone = useMatchMobileMediumQuery();
+  const isMobile = useMatchMobileQuery();
+
+  React.useEffect(() => {
+    const refIsSensible = mRef && 'current' in mRef && mRef.current;
```
jivey, line 38: Suggestion: you could move this directly into the if statement instead of creating an one off variable

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +24,171 @@ interface SidebarProps {
...
-  />
-);
+const SidebarBody = React.forwardRef<
+  HTMLElement,
+  React.ComponentProps<typeof SidebarPaneBody>
+>((props, ref) => {
+
+  return (
+    <React.Fragment>
+      {typeof window !== 'undefined' && (
```
jivey, line 72: I don't see window accessed, is this here for prerendering?

jivey (commented):
~I thought we added `createTrapTab` for stuff like this, maybe it doesn't work for the TOC?~

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +24,171 @@ interface SidebarProps {
...
+
+  React.useEffect(() => {
+    const refIsSensible = mRef && 'current' in mRef && mRef.current;
+
+    if (!refIsSensible) {
+      return;
+    }
+    const otherRegions =
+      document?.querySelectorAll(
+        '[data-testid="navbar"],[data-testid="bookbanner"]'
```
jivey, line 45: I'm not sure we want to use testids for functionality, I usually think of these as being free to change for testing purposes.

jivey (commented):
Forget my earlier comment about createTabTrap, I misread the diff and thought we were adding another one. 

Thinking out loud here... in regards to my other comment about the Menu not trapping tab, could we wrap the entire sidebar with a trap? Does this need to target TOC specifically? I wonder if the functionality could be simplified to just check for it being in that entire container or not on mobile/ipad.

RoyEJohnson:
The overlay that obscures the main content only appears when the TOC is open, so we only trap navigation then.
In addition to the TOC, the sidebar _and_ the title banner _and_ the top nav bar (but _not_ the jump to main content shortcuts) are in the tab cycle.

jivey (changes requested):
The tablet width + overlay displaying case you mentioned works, I was thinking of the phone width when the main menu / toolbar is open. On a phone, the focus escapes the toolbar (not the TOC.) It feels like we are really close to having this work nicely for both cases. 

Maybe I'm misunderstanding how the trap works or the HTML is structured, but wouldn't targeting `data-testid="toolbar"` and `data-testid="toc"` cycle through the visible elements, making it just work for the mobile case?

jivey (approved):

PR #2163, opened 2024-02-05
Add aria label to recurring chapter names

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2664

### DIFF
--- a/src/app/content/components/TableOfContents/index.tsx
+++ b/src/app/content/components/TableOfContents/index.tsx
@@ -7,8 +7,8 @@ import { closeMobileMenu, resetToc } from '../../actions';
 import { isArchiveTree } from '../../guards';
 import { linkContents } from '../../search/utils';
 import * as selectors from '../../selectors';
-import { ArchiveTree, Book, Page, State } from '../../types';
-import { archiveTreeContainsNode, getArchiveTreeSectionType } from '../../utils/archiveTreeUtils';
+import { ArchiveTree, Book, LinkedArchiveTreeSection, Page, State } from '../../types';
+import { archiveTreeContainsNode, getArchiveTreeSectionType, splitTitleParts } from '../../utils/archiveTreeUtils';
 import { expandCurrentChapter, scrollSidebarSectionIntoView, setSidebarHeight } from '../../utils/domUtils';
 import { stripIdVersion } from '../../utils/idUtils';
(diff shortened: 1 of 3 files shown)


jivey (changes requested):

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -23,28 +25,218 @@ interface SidebarProps {
   page?: Page;
 }
 
+function TabTrapper({
```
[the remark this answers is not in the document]
RoyEJohnson: This PR was based on another PR because it needed some of the same changes. That's merged now.

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -136,6 +136,22 @@ function TocNode({
...
 }
 
+function maybeAriaLabel(page: LinkedArchiveTreeNode) {
+  const [num, titleText] = splitTitleParts(page.title);
+
+  if (num) {
+    return {};
+  }
+
+  const [parentNum, parentTitleText] = splitTitleParts(page.parent!.title);
```
jivey, line 146: Can't parent still be undefined here? It seems like a guard could be added to the early return above and then you'd have a narrowed type

jivey (changes requested):
This is looking good just one minor type concern

src/app/content/components/TableOfContents/index.tsx
RoyEJohnson: The test data doesn't have any with null parents. The prospect of having to add to the test data made me look at likely TOC items like "Index", which has as its parent the book itself. I don't think there are any TOC entries that don't have a parent.

jivey: Does that mean the type should be changed? Maybe @TomWoodward can chime in, I don't know these types well enough to feel confident about asserting the data will always be there.

TomWoodward: i think in real life the only thing in the linked section that won't have a parent is the book node. its possible the types could be updated to make parent required

RoyEJohnson: Turns out the type I was using was a union, and it only needs to be the one where `parent` is required.

jivey: Ah good find

jivey (approved):

PR #2164, opened 2024-02-05
Use Alt+h to go between highlighted text and editor

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2766
https://openstax.atlassian.net/browse/DISCO-113
For: https://github.com/openstax/unified/issues/2767 (fix related focus issue)
https://openstax.atlassian.net/browse/DISCO-116


src/app/content/highlights/components/EditCard.tsx
```diff
@@ -266,6 +266,8 @@ function AnnotationEditor({
   const initializeColor = React.useCallback(() => {
     if (!props.highlight.getStyle()) {
       onColorChange(highlightStyles[0].label, true);
+      textarea.current?.blur();
+      textarea.current?.focus();
```
Dantemss, line 270: This on purpose or just for testing?

src/app/content/highlights/constants.ts
```diff
@@ -14,5 +14,6 @@ export const highlightIndicatorSizeForBlock = 1.2;
 export const highlightIndicatorSize = 0.9;
 
 export const highlightKeyCombination: KeyCombinationOptions = {
-  key: 'h',
+  key: '˙', // key isn't h when alt is pressed
```
Dantemss, line 17: Is this consistent for different keyboards? Otherwise we may have to use the `code` property instead or something.

src/app/content/keyboardShortcuts/components/__snapshots__/KeyboardShortcutsPopup.spec.tsx.snap
```diff
@@ -417,7 +417,7 @@ exports[`KeyboardShortcuts renders keyboard shortcuts modal if it is open 1`] =
               <span
                 className="c13"
               >
-                h
+                i18n:a11y:keyboard-shortcuts:keys:alt+h
```
jivey, line 420: This might need a translation key updated

jivey, line 420: This might need a translation key updated

jivey, line 420: This might need a translation key updated

jivey (changes requested):
Sorry for all the comment spam, kept clicking the wrong thing. Left a note about the translation keys.

src/app/content/highlights/components/EditCard.tsx
RoyEJohnson: This is on purpose. Focus wasn't sticking without it.

src/app/content/highlights/constants.ts
RoyEJohnson: It's standard according to [this](https://www.toptal.com/developers/keycode/table).

jivey: In my Windows tests, this is still interpreted as `h`:

```
{
 "key": "h",
 "keyCode": 72,
 "which": 72,
 "code": "KeyH",
 "location": 0,
 "altKey": true,
 "ctrlKey": false,
 "metaKey": false,
 "shiftKey": false,
 "repeat": false
}
```

I think we'll need to check the modifiers in addition to the key/code.

src/app/content/highlights/components/EditCard.tsx
jivey: Was the focus not sticking because of a re-render? This seems like a render or timing issue we should dig into

jivey (changes requested):
Left some comments outside of the review

src/app/content/highlights/components/EditCard.tsx
RoyEJohnson: It seems to be the same issue as [this](https://stackoverflow.com/questions/3685242/unable-to-type-in-textbox-even-with-focus-set): the focus is set, but keyboard input is ignored until you click on the field. Their solution was the same as mine: blur, then focus again (only they used a timeout, which doesn't seem to be necessary).
It only happens when the highlight is a new selection. If I disable the call to `onColorChange`, I am able to type.

src/app/content/highlights/constants.ts
RoyEJohnson: I was checking the altKey modifier anyway. I just have to use `code` instead of `key`.

src/app/content/highlights/components/EditCard.tsx
RoyEJohnson: This issue is the updating of the DOM that `onColorChange` does. In particular, if something is removed from the DOM, focus is lost (although it seems to be only half-lost in this case -- the focus outline remains, it just doesn't direct keyboard input there)

jivey (approved):

PR #2166, opened 2024-02-07
Add label and role to DisplayNote

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2760
My translations should be reviewed!


src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -92,30 +94,39 @@ const DisplayNote = React.forwardRef<HTMLElement, DisplayNoteProps>((
...
-    />}
-  </div>;
+  return (
+    <div
+      className={className}
+      ref={mergeRefs(ref, element)}
+      tabIndex={-1}
+      data-highlight-card
+      role='dialog'
+      aria-label={formatMessage({id: 'i18n:highlighter:display-note:label'})}
```
jivey, line 104: Should this reflect the state of the card? This will show "highlight editor" for a non-editable state, which feels off to me. Maybe we should have "Highlighted note" for this and put "Highlight editor" on EditCard?

src/app/messages/en/messages.json
```diff
@@ -59,6 +59,7 @@
   "i18n:footer:social:in:alt": "linkedin",
   "i18n:footer:social:rice-logo:alt": "Rice University logo",
   "i18n:footer:social:tw:alt": "twitter",
+  "i18n:highlighter:display-note:label": "highlight editor",
```
jivey, line 62: Totally nitpicky (sorry lol) but if we can capitalize that it will match the other highlighter messages

jivey (changes requested):
Left a thought about making the messaging work with the card state. 

Regarding the translations, I usually nicely tag bethshook and ewaforman and hope that they have time 😄 I've found that the online translators aren't always great at making technical jargon sound right.

src/app/content/highlights/components/DisplayNote.tsx
RoyEJohnson: Changed the message file and added one for the EditCard

RoyEJohnson:
@bethshook @ewaforman I've added you as reviewers so you could check/correct my translations in the messages files. Thanks.

bethshook:
Spanish looks good

src/app/content/highlights/components/EditCard.tsx
```diff
@@ -44,6 +44,7 @@ export interface EditCardProps {
   data?: HighlightData;
   className: string;
   shouldFocusCard: boolean;
+  role?: string;
```
jivey, line 47: This doesn't seem to be used?

jivey (changes requested):

src/app/content/highlights/components/EditCard.tsx
RoyEJohnson: You are correct. That got leftover from trying to figure out where role needed to be specified.

jivey (approved):

Malar-Natarajan:
@RoyEJohnson sorry if you were not in the loop but Ewa is no longer working with OpenStax or Rex projects. So I believe if we need her help we may need to route through Erika & Paulina

jivey:
@Malar-Natarajan @RoyEJohnson That's my bad. Maybe we should check with Erika about how we should handle these

PR #2171, opened 2024-02-08
Make colorpicker radiobuttons in a fieldset

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2759
For: https://openstax.atlassian.net/browse/DISCO-96

### DIFF
--- a/src/app/content/highlights/components/ColorPicker.tsx
+++ b/src/app/content/highlights/components/ColorPicker.tsx
@@ -6,7 +6,7 @@ import { match, not } from '../../../fpUtils';
 import { highlightStyles } from '../../constants';
 import { cardPadding } from '../constants';
 import ColorIndicator from './ColorIndicator';
-import type { HTMLDivElement, HTMLInputElement } from '@openstax/types/lib.dom';
+import { HTMLDivElement, HTMLInputElement } from '@openstax/types/lib.dom';
 
 interface SingleSelectProps {
   color?: HighlightColorEnum;
@@ -109,7 +109,13 @@ const ColorPicker = ({className, ...props}: Props) => {
       if (!ref.current || event.currentTarget !== event.target) {
         return;
       }
-      (ref.current.querySelector<HTMLInputElement>('input[checked]'))?.focus();
+      const firstChecked = ref.current.querySelector<HTMLInputElement>('input[checked]');
+
+      if (firstChecked) {
+        firstChecked.focus();
+      } else {
+        ref.current.querySelector<HTMLInputElement>('input')?.focus();
+      }
     },
     []
   );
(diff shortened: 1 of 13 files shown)


src/app/content/highlights/components/ColorPicker.tsx
```diff
@@ -62,33 +63,88 @@ const ColorButton = styled(({className, size, style, ...props}: ColorButtonProps
...
+      ref={ref as unknown as LegacyRef<HTMLElement>}
+      onKeyDown={handleKeyNavigation}
+      onFocus={initialFocus}
+    >
+      {highlightStyles.map((style) => <ColorButton key={style.label}
+        name={style.label}
+        checked={props.multiple ? props.selected.includes(style.label) : props.color === style.label}
+        style={style}
+        size={props.size}
+        tabindex={-1}
```
jivey, line 130: React wants this to be `tabIndex` which I never remember until I randomly notice it yelling at me in the console

src/app/content/highlights/components/ColorPicker.tsx
```diff
@@ -62,33 +63,88 @@ const ColorButton = styled(({className, size, style, ...props}: ColorButtonProps
...
-          : props.onChange(style.label)}
-    />)}
-  </div>;
+  return (
+    <div
+      className={className}
+      tabIndex={0}
+      role='radiogroup'
+      aria-label='colors'
+      ref={ref as unknown as LegacyRef<HTMLElement>}
```
jivey, line 121: Wouldn't initializing the ref to `null` make this unnecessary?

jivey (changes requested):

jivey:
Not a change request, just wondering to expand my React knowledge... do you think using callback refs would be a more "Reacty" way of approaching this? I was peeking at https://react.dev/learn/manipulating-the-dom-with-refs#how-to-manage-a-list-of-refs-using-a-ref-callback but I dunno if it has render tradeoffs and might be too complicated for this.

src/app/content/highlights/components/ColorPicker.tsx
RoyEJohnson: That actually does work. I thought the differing types were the problem. Thanks.

RoyEJohnson:
I think callback refs could be used to avoid making separate components where the only reason for the separate component was to allow it to have a ref. Since we already have a ColorButton, I don't see it offering an advantage here. My inclination is to make separate components, anyway, because it's more straightforward. There would probably be an efficiency advantage to callback refs if the list of sub-units was very long.

jivey (approved):
Looks good but you've got some lint and snapshot failures

PR #2175, opened 2024-02-12
Provide accessible label for textarea on EditCard

RoyEJohnson:
For: https://github.com/openstax/unified/issues/2708
https://openstax.atlassian.net/browse/DISCO-117


src/app/content/highlights/components/Note.tsx
```diff
@@ -49,16 +75,22 @@ const Note = ({onChange, onFocus, note, textareaRef}: Props) => {
...
-    value={note}
-    onFocus={onFocus}
-    maxLength={noteMaxLength}
-    onChange={(e: React.ChangeEvent<HTMLTextAreaElement>) => {
-      onChange(e.target.value);
-    }}
-    placeholder={useIntl().formatMessage({id: 'i18n:highlighting:card:placeholder'})}
-  />;
+  return (
+    <WrapperLabel for='note-textarea'>
```
jivey, line 79: Do we still need the `for` if the wrapper makes it implicit?

jivey (changes requested):
I was glancing at the associated card and saw a linked figma design, did we decide not to use that?

I saw some likely small bugs, in Firefox the label doesn't look right:
image
and in Chrome, if it's the first time showing the editor and you click the label text directly, it shows the textarea as focused, but it doesn't seem to be (it has no cursor, typing doesn't work)
image

RoyEJohnson:
The focus issue is addressed in a different PR. The layout should be more like the Figma now (I used blue instead of light blue, because light blue isn't visible enough). Firefox and Safari layout should look right.

jivey (approved):

PR #2180, opened 2024-02-21
DISCO-107 redirect book slugs to book details page

bethshook:
for: openstax/unified#2772


script/utils/prepareRedirects.ts
```diff
@@ -37,13 +39,34 @@ const prepareRedirects = async(
...
         from: pathname,
-        to: decodeURI(content.getUrl({ book: { slug: bookSlug }, page: { slug: page.slug } })) + (query || ''),
+        to,
       });
+
+      if (!pathname.endsWith('/')) {
+        redirects.push({
+          from: `${pathname}/`,
+          to,
+        });
```
TomWoodward, line 53: do this with ```{from: `${pathname}/`, to: pathname}``` and it'll do two redirects, one from the slash to the no slash and one from the no slash to the new location.

this way any previous redirects for standard pages that have slash-to-no-slash redirects will have the same slash-to-no-slash redirects if the pages are subsequently moved.

not that the way you have it would be broken, but it would be better for the redirects for particular urls to not change because they're allowed to be indefinitely cached and it seems like it could be a source of unexpected behavior in the future to alter them

src/app/content/hooks/locationChange/resolveContent.ts
```diff
@@ -147,6 +147,12 @@ const resolvePage = async(
 ) => {
   const {dispatch, getState} = services;
   const state = getState();
+
+  if (match.params.book && 'slug' in match.params.book && !match.params.page) {
+    // book slug provided but no page
+    await processBrowserRedirect(services);
```
TomWoodward, line 153: if you allow this to hit the page not found because it has the wrong slug conditional [below](https://github.com/openstax/rex-web/pull/2180/files#diff-3b681369c849d5f3e705f6e56269f0a1c93047c1049cd36b99f3e9335ae62db4R158) it'll get redirected through the existing logic

TomWoodward (changes requested):

src/app/content/hooks/locationChange/resolveContent.ts
bethshook: but the logic below requires a page id and we don't have that

TomWoodward (approved):

PR #2188, opened 2024-03-02
DISCO-100 Added lighthouse automated test and prerenderspec

Dantemss:
For: https://github.com/openstax/unified/issues/2229

Related PR: https://github.com/openstax/unified-deployment/pull/347


RoyEJohnson (approved):

Dantemss:
@TomWoodward asking for your review in case you feel like I should organize this code differently

Dantemss:
If this approach looks ok, we'll also need to give unified-deployment access to this bucket

script/lighthouse-top-pages.bash
```diff
@@ -0,0 +1,36 @@
...
+)
+
+lighthouse_base_report_path=/tmp/lighthouse-reports
+
+mkdir -p "$lighthouse_base_report_path"
+
+bucket_name=unified-web-lighthouse
+
+# Note: this probably fails due to pagination if there are > 1,000 prefixes
+last_prefix=$(aws s3api list-objects-v2 --bucket "$bucket_name" --delimiter / \
```
TomWoodward, line 21: could there be a "last-prefix.txt" file that you write to when you do the upload so that you don't have to scan the whole bucket?

TomWoodward (changes requested):
specifically adding the prerenderspec here as part of the CI test suite looks fine, but i think we should rework the way that the production check is integrated here.

i don't think you're getting anything by integrating the production check into the prerenderspec and i think its a bit confusing. we already have the `test:lighthouse:manual` command that takes a url and generates the report html, can we use that? or if you want to have your script boot puppeteer and calls the shared `checkLighthouse` function that would be ok too.

i also think that the script can be created as a concourse task in `rex-web/concourse` and put the report files in a task input/output, then leave the s3 download/uplaod to another pipeline step to keep this stuff kind of separated

Dantemss:
This should finally be ready for review once again. I tested it in the unified-test pipeline and it seems to work.

RoyEJohnson (approved):

src/test/browserutils.ts
```diff
@@ -115,13 +116,34 @@ export const h1Content = (target: puppeteer.Page) => target.evaluate(() => {
...
+    }
+  });
 
-  expect(lhr.categories.customAccessibility.score).toBeGreaterThanOrEqual(1);
-  expect(lhr.categories.accessibility.score).toBeGreaterThanOrEqual(1);
-  expect(lhr.categories.seo.score).toBeGreaterThanOrEqual(0.69);
-  expect(lhr.categories['best-practices'].score).toBeGreaterThanOrEqual(0.79);
+  if (typeof scores === 'string') {
+    return new Promise<void>((resolve) => writeFile(scores, JSON.stringify(result), () => resolve()));
+  }
```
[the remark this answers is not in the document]
Dantemss: Done

script/lighthouse-top-pages.bash
Dantemss: Done

script/lighthouse.ts
```diff
@@ -0,0 +1,42 @@
...
+    const filename = `${pageUrl.replace(/[^a-z0-9]+/gi, '-').replace(/^-+/i, '')}.json`;
+    const targets = mostRecentReportDir ? await new Promise<ScoreTargets | undefined>(
+      (resolve) => readFile(
+        `${mostRecentReportDir}/${filename}`,
+        { encoding: 'utf8' },
+        (err, data) => err ? resolve(undefined) : resolve(JSON.parse(data))
+      )
+    ) : undefined;
+
+    const result = await checkLighthouse(browser, pageUrl, targets);
```
TomWoodward, line 28: so checkLighthouse throws if you don't meet the targets, which means it won't write the new results file, so basically if there is any regression the job'll continuously fail until we fix it and reach the targets again, right?

TomWoodward (approved):

script/lighthouse.ts
Dantemss: It's been a while since I worked on this, but that sounds about right.

Dantemss:
Since there's nothing for Malar to test here and we already merged the unified-deployment PR, I'll merge this one.

Dantemss:
The PR is linked but the Jira task keeps failing. I'll try to take a look and figure out why.

Dantemss:
> > The PR is linked but the Jira task keeps failing. I'll try to take a look and figure out why.
> 
> make sure the jira ticket shows the issue as linked, and then try rerunning just that one check -- i think normally it only runs on push so if you're editing the title it doesn't recheck

The Jira card had been set to "Done" and that prevents jira-linked-action from matching.

PR #2189, opened 2024-03-06
Add alt-S to navigate between search regions

RoyEJohnson:
[DISCO-90]

[DISCO-90]: https://openstax.atlassian.net/browse/DISCO-90?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/components/Page/searchHighlightManager.ts
+++ b/src/app/content/components/Page/searchHighlightManager.ts
@@ -2,7 +2,6 @@ import Highlighter, { Highlight } from '@openstax/highlighter';
 import { HTMLElement } from '@openstax/types/lib.dom';
 import isEqual from 'lodash/fp/isEqual';
 import { IntlShape } from 'react-intl';
-import { scrollTo } from '../../../domUtils';
 import { AppState } from '../../../types';
 import { memoizeStateToProps } from '../../../utils';
 import * as selectSearch from '../../search/selectors';
@@ -75,10 +74,9 @@ const selectResult = (services: Services, previous: HighlightProp | null, curren
     allImagesLoaded(services.container).then(
       () => {
         const target = selectedElements[0] as HTMLElement;
-        const container = target.closest('[tabindex]') as HTMLElement;
+        const focusTarget: HTMLElement | null = target.querySelector('[tabindex="0"]');
 
-        scrollTo(target);
-        container?.focus();
+        focusTarget?.focus();
       }
     );
   }
--- a/src/app/content/components/Topbar/index.tsx
+++ b/src/app/content/components/Topbar/index.tsx
@@ -281,7 +281,7 @@ function Topbar(props: Props) {
       ].map((q) => document?.querySelector(q));
 
       // Determine which region we are in (if any)
-      const currentSectionIndex = targets.findIndex((el) => el?.contains(document?.activeElement ?? null));
+      const currentSectionIndex = targets.findIndex((el) => el?.contains(document?.activeElement!));
 
       // If not in any, go to search input
       if (currentSectionIndex < 0) {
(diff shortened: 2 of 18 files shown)


src/app/content/components/Page/searchHighlightManager.ts
```diff
@@ -75,10 +74,9 @@ const selectResult = (services: Services, previous: HighlightProp | null, curren
     allImagesLoaded(services.container).then(
       () => {
         const target = selectedElements[0] as HTMLElement;
-        const container = target.closest('[tabindex]') as HTMLElement;
+        const focusTarget: HTMLElement | null = target.querySelector('[tabindex="0"]');
 
-        scrollTo(target);
```
Dantemss, line 80: Is this `scrollTo()` just gone now? I noticed it was gone from the tests too.

RoyEJohnson: Yes. `focus` will scroll an element into view unless you [specify that it shouldn't](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#preventscroll).

src/app/content/components/Topbar/index.tsx
```diff
@@ -35,191 +38,327 @@ interface Props {
...
+        return;
+      }
+      const targets = [
+        '[class*="SearchInputWrapper"] input',
+        '[class*="SearchResultsBar"]',
+        'main',
+      ].map((q) => document?.querySelector(q));
+
+      // Determine which region we are in (if any)
+      const currentSectionIndex = targets.findIndex((el) => el?.contains(document?.activeElement!));
```
Dantemss, line 220: I'd do something like `document?.activeElement && el?.contains(document.activeElement)` because activeElement can be null and idk if contains likes nulls or undefineds.

src/app/content/components/Topbar/index.tsx
```diff
@@ -35,191 +38,327 @@ interface Props {
...
+    // Only in desktop layout
+    () => {
+      if (isMobile) {
+        return;
+      }
+      const targets = [
+        '[class*="SearchInputWrapper"] input',
+        '[class*="SearchResultsBar"]',
+        'main',
+      ].map((q) => document?.querySelector(q));
```
Dantemss, line 217: Could you make `targets` an `HTMLElement[]` here? So you don't have to keep casting it later.

src/app/content/components/Topbar/index.tsx
```diff
@@ -35,191 +38,327 @@ interface Props {
...
+    if (props.query !== query && props.query !== prevQuery.current) {
+      setQuery(props.query);
+    }
+    prevQuery.current = props.query;
+  }
 
-    const onSearchSubmit = (e: React.FormEvent) => {
+  const onSearchChange = React.useCallback(
+    (e: React.FormEvent<HTMLInputElement>) => {
+      setQuery((e.currentTarget as any).value);
```
Dantemss, line 271: Eh I get that the form will always have currentTarget, but I'd at least cast it as `{ value: string }` or whatever it is

Dantemss (approved):

src/app/content/components/Topbar/index.tsx
RoyEJohnson: `null` is ok for `contains`. `activeElement` _can_ be undefined, but there's no good way to make it so. In general, it will be set to `body` if you `blur` it from anything else.
I changed `document?activeElement!` to `document?activeElement ?? null` but could not write a test to hit the undefined branch.

PR #2201, opened 2024-03-21
[DISCO-165] Fix colorpicker focus

RoyEJohnson:
[DISCO-165]

[DISCO-165]: https://openstax.atlassian.net/browse/DISCO-165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/content/highlights/components/ColorPicker.tsx
```diff
@@ -104,22 +104,18 @@ const ColorPicker = ({className, ...props}: Props) => {
...
+    () => {
+      if (color && document?.activeElement === ref.current) {
+        ref.current?.querySelector<HTMLInputElement>(`input[name="${color}"]`)?.focus();
       }
     },
-    []
+    [color]
   );
 
+  React.useEffect(focusOnSelected, [focusOnSelected]);
```
Dantemss, line 117: I think this will steal focus whenever this component is rendered... is that an OK thing?

RoyEJohnson: It will only change focus if the container is the active element (already has the focus - this is the 2nd part of the `if` in line 110).
The rule for radio groups is that the container is keyboard-focusable, and when it receives focus, the selected radio item gets focused (or if none is selected, the first one, but we always have a selection).

Dantemss (approved):

src/app/content/highlights/components/ColorPicker.tsx
Dantemss: ah ok I missed that

PR #2203, opened 2024-03-25
Style section highlights headers as h2/h3

RoyEJohnson:
[DISCO-43]
I made the Popup Header an h1 so that we have the correct sequence of headers.
It made no sense to me that it was an h3 previously. It is a major, separate section. If we think it should be h3, then the section headers need to become h4 and h5.

[DISCO-43]: https://openstax.atlassian.net/browse/DISCO-43?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/components/SectionHighlights.tsx
+++ b/src/app/content/components/SectionHighlights.tsx
@@ -30,7 +30,7 @@ export const HighlightsChapterWrapper = styled.div`
 `;
 
 // tslint:disable-next-line:variable-name
-const HighlightsChapter = styled.div`
+const HighlightsChapter = styled.h2`
   ${h4Style}
   font-weight: bold;
   display: flex;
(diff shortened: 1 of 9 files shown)


jivey (commented):
This is a good change. I'm wondering about the awkwardness of the h3Style on the h1. Maybe we should update the heading style names to be more abstract like large/medium/small? Or add aliases?

jivey (approved):
Yeah we could totally make a case for either way... I think I mentally tend to separate the style from the markup (though hopefully they match in relative scale) because I'm used  making small tweaks in different areas anyway. But with REX I don't have a good sense of how these are used, so I think this is a good way to handle it. And in the future we might replace some of these with typography helpers from ui-components.

RoyEJohnson:
@jivey I added a commit here to pull the put-away X out of the popup header. Thought you might want to look it over. I had to change Header to be a div and then put the h1 tags explicitly around the heading text.

jivey (approved):
That's a nice idea for handling the h1 in Header 👍

PR #2204, opened 2024-03-26
Fix header and list-of-links structure in footer

RoyEJohnson:
[DISCO-44]

[DISCO-44]: https://openstax.atlassian.net/browse/DISCO-44?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/Footer/index.tsx
+++ b/src/app/components/Footer/index.tsx
@@ -176,7 +176,7 @@ const Footer = ({
     <Styled.InnerFooter>
       <Styled.FooterTop>
         <Styled.TopBoxed>
-          <Styled.Heading role='heading' aria-level={2}>
+          <Styled.Heading>
             <BareMessage id='i18n:footer:heading' />
           </Styled.Heading>
           <Mission />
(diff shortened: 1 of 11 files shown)


src/app/components/Footer/index.tsx
```diff
@@ -72,53 +72,67 @@ const SocialIconMessage: React.FunctionComponent<{
   </Styled.SocialIcon>
 );
 
+function LinkList({children}: React.PropsWithChildren<{}>) {
+  return (
+    <Styled.LinkListWrapper>
+      {React.Children.toArray(children).map((c) => <li>{c}</li>)}
```
Dantemss, line 78: I noticed this one doesn't set the key property, unlike the other PR

Dantemss (approved):

src/app/components/Footer/index.tsx
RoyEJohnson: Oh, yeah. No warning on the Rex system for that. The set of children won't be changing, but I pushed a use-index-as-key update.

Malar-Natarajan:
@RoyEJohnson Lint is failing in this PR even after a re-run. Could you please look into it.

PR #2207, opened 2024-03-28
Group checkboxes in dropdowns

RoyEJohnson:
[DISCO-46]

[DISCO-46]: https://openstax.atlassian.net/browse/DISCO-46?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/theme.ts
```diff
@@ -9,6 +9,18 @@ export interface ColorSet {
   darkest?: string;
 }
 
+export const hiddenButAccessible = `
```
jivey, line 12: Nitpicking but `visuallyHidden` could be a good name for this

jivey (approved):

src/app/theme.ts
RoyEJohnson: There's a `visuallyHidden` [in Dropdown](https://github.com/openstax/rex-web/blob/make-group-for-dropdown-checkboxes/src/app/components/Dropdown.tsx#L49) (and a contrasting `visuallyShown`). Since it uses height/width `0`, it's also hidden from screenreaders. Probably need to rename it to `fullyHidden` or something. I'm not entirely clear on what the point of it is, why it's not `display: none`.

Malar-Natarajan:
@RoyEJohnson Could you please resolve the conflicts on this PR

PR #2214, opened 2024-04-05
Add live region for navigation changes

RoyEJohnson:
[DISCO-187]

[DISCO-187]: https://openstax.atlassian.net/browse/DISCO-187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
Adding you as a reviewer, Tom, because I suspect there are decisions I made that are open to criticism.


src/app/components/PageTitleConfirmation.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const el = ref.current;
+
+    if (!el) { return; }
+    const listener = ((e: CustomEvent) => {
+      setDocTitle(e.detail);
+    }) as EventListener;
+
+    el.addEventListener('page-title', listener);
+    return () => el.removeEventListener('page-title', listener);
+  }, []);
```
Dantemss, line 28: Doesn't this need to re-trigger if the ref changes?

src/app/components/PageTitleConfirmation.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default function announcePageTitle(title: string) {
+  const el = document?.getElementById('page-title-confirmation');
+
+  el?.dispatchEvent(
+    new CustomEvent('page-title', {
+      detail: title,
+    })
+  );
+}
```
Dantemss, line 12: I feel like this function could be a callback that calls setDocTitle directly instead of dispatching the event and using the listener

Dantemss: You might need to create a `usePageTitleConfirmation` pattern for that though.

src/app/components/PageTitleConfirmation.tsx
RoyEJohnson: The ref won't change unless there's a fresh render, but I can move the `el` definition outside the useEffect just in case.

src/app/components/PageTitleConfirmation.tsx
RoyEJohnson: The only way I could see doing that was to use a Context, and I found this lighter and less coupled. There might be a more clever way to feed setDocTitle to the announce function so it could be called.

Dantemss (approved):
Well not sure about setting the fn property on the exported function, but I guess it's fine?

src/app/components/NavBar/index.tsx
```diff
@@ -134,6 +134,7 @@ const NavigationBar = ({user, loggedOut, currentPath}: NavigationBarProps) =>
       </a>
       {loggedOut && <LoggedOutState currentPath={currentPath} />}
       {user && <LoggedInState user={user} currentPath={currentPath} />}
+      <Styled.PageTitleConfirmation />
```
TomWoodward, line 137: does it matter where in the document you put this? it doesn't read it in order or anything right? maybe put it up the tree a bit in Layout or Content or something? NavBar seems a bit weird

src/app/components/NavBar/__snapshots__/index.spec.tsx.snap
```diff
@@ -853,6 +958,13 @@ exports[`content in browser when logged in matches snapshot 1`] = `
         </svg>
       </button>
     </div>
+    <div
+      aria-live="polite"
+      className="c13"
+      id="page-title-confirmation"
+    >
+      Loaded page ""
```
TomWoodward, line 966: does this read out on page load? would we prefer that it be empty at first instead of having "loaded page [silence]"?

src/app/components/PageTitleConfirmation.spec.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { renderToDom } from '../../test/reactutils';
+import { act } from 'react-dom/test-utils';
+import announcePageTitle, {
+  PageTitleConfirmation
+} from './PageTitleConfirmation';
+
+describe('On scroll', () => {
```
TomWoodward, line 8: `on scroll`?

src/app/components/PageTitleConfirmation.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+type FunctionWithProperty = {
+  (title: string): void;
+  fn?: (title: string) => void;
+};
+
+const announcePageTitle: FunctionWithProperty = (title: string) => {
```
TomWoodward, line 8: i think this component should work based on being connected to the redux state instead of having a separate callback like this

TomWoodward (changes requested):

src/app/components/NavBar/index.tsx
RoyEJohnson: Moved to Layout

src/app/components/NavBar/__snapshots__/index.spec.tsx.snap
RoyEJohnson: I have it skip the first non-blank value

src/app/components/PageTitleConfirmation.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    store.dispatch(setHead({...initialState.head, title: ''}));
+    const component = renderToDom(
+      <TestContainer store={store}>
+        <PageTitleConfirmation className='any' />
+      </TestContainer>
+    );
+
+    expect(component.node.textContent).toBe('');
+    store.dispatch(setHead({...initialState.head, title: 'initial'}));
```
TomWoodward, line 27: i believe its gonna want the dispatches to be wrapped in `act`

src/app/components/PageTitleConfirmation.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+import TestContainer from '../../test/TestContainer';
+import { setHead } from '../head/actions';
+
+describe('PageTitleConfirmation', () => {
+  it('skips blank, skips first non-blank, announces subsequent', async() => {
+    const state = {
+      ...initialState,
+    };
+
+    state.head.title = 'initial';
```
TomWoodward, line 16: i don't think its breaking anything in this case, but redux makes many assumptions about state data being immutable and you may get weird behavior by altering the data by reference like this

src/app/components/PageTitleConfirmation.tsx
```diff
@@ -0,0 +1,37 @@
...
+  }, [title]);
+
+  return (
+    <div id='page-title-confirmation' className={className} aria-live='polite'>
+      {message}
+    </div>
+  );
+}
+
+export default connect(({ head: { title } }: AppState) => ({ title }))(styled(
```
TomWoodward, line 33: there is a selector for the title https://github.com/openstax/rex-web/blob/main/src/app/head/selectors.ts#L19

using it will affect nobody's life but it follows the convention. when you have derived values in selectors it does make a difference because they're memoized based on the input state and will help reduce react re-renders

TomWoodward (changes requested):
this is looking fine to me, please confirm that those tests aren't complaining about the `act` call and i will approve.

src/app/components/PageTitleConfirmation.spec.tsx
RoyEJohnson: No complaints from the test.
```
% yarn jest PageTitleConfirmation.spec
yarn run v1.22.22
$ /Users/rej2/Documents/GitHub/rex-web/node_modules/.bin/jest PageTitleConfirmation.spec
 PASS  src/app/components/PageTitleConfirmation.spec.tsx (6.811 s)
  PageTitleConfirmation
    ✓ skips blank, skips first non-blank, announces subsequent (28 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        7.778 s
Ran all test suites matching /PageTitleConfirmation.spec/i.
✨  Done in 21.04s.
```

src/app/components/PageTitleConfirmation.tsx
RoyEJohnson: I added a auto-focus to main when its content changes, and also revised this to use the selector.

src/app/components/PageTitleConfirmation.spec.tsx
RoyEJohnson: No complaints from the test.
```
% yarn jest PageTitleConfirmation.spec
yarn run v1.22.22
$ /Users/rej2/Documents/GitHub/rex-web/node_modules/.bin/jest PageTitleConfirmation.spec
 PASS  src/app/components/PageTitleConfirmation.spec.tsx (6.811 s)
  PageTitleConfirmation
    ✓ skips blank, skips first non-blank, announces subsequent (28 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        7.778 s
Ran all test suites matching /PageTitleConfirmation.spec/i.
✨  Done in 21.04s.
```

TomWoodward:
@RoyEJohnson can you remove the focus on main thing and we'll get the announcement merged? i have concerns about when the content is changed as part of a navigation with a target, like clicking a search result or a link with a hash

TomWoodward (approved):

RoyEJohnson:
I made a [separate subtask issue](https://openstax.atlassian.net/browse/DISCO-278) for setting focus on the main area.

PR #2220, opened 2024-04-11
add chapter name and book title to attribution link

jivey:
https://openstax.atlassian.net/browse/DISCO-84


RoyEJohnson (approved):

jivey:
@RoyEJohnson I missed a couple of other citation links that needed this aria label as well, if you don't mind taking another look

src/app/messages/en/attributionText.ts
```diff
@@ -28,7 +28,7 @@ export const defaultText = `
     If you are redistributing all or part of this book in a digital format,
     then you must include on every digital page view the following attribution:
     <blockquote>
-      Access for free at <a target="_blank" href="https://openstax.org{introPageUrl}">https://openstax.org{introPageUrl}</a>
+      Access for free at <a target="_blank" href="https://openstax.org{introPageUrl}" aria-label="{introPageTitle}">https://openstax.org{introPageUrl}</a>
```
RoyEJohnson, line 31: The quotes are superfluous, aren't they? Should just be `aria-lable={introPageTitle}` ?
And the interpolation in the href? That works? I'd write
```
href={`https://openstax.org${introPageUrl}`}
```

jivey: This string gets formatted with the useIntl stuff, so it's like how we interpolate in the messages.json file

PR #2221, opened 2024-04-12
Focus on main when content changes

RoyEJohnson:
[DISCO-198]

[DISCO-198]: https://openstax.atlassian.net/browse/DISCO-198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/components/Page/scrollToTopOrHashManager.ts
+++ b/src/app/content/components/Page/scrollToTopOrHashManager.ts
@@ -30,16 +30,8 @@ const scrollToTargetOrTop = (container: HTMLElement | null, hash: string, previo
   if (getScrollTarget(container, hash)) {
     scrollToTarget(container, hash);
   } else if (previous) {
-    scrollToTop();
+    assertWindow().scrollTo(0, 0);
   }
-};
-
-const scrollToTop = () => {
-  const window = assertWindow();
-  const document = window.document;
-
-  document.querySelector('main')?.focus();
-  window.scrollTo(0, 0);
 };
 
 const getScrollTarget = (container: HTMLElement | null, hash: string): HTMLElement | null => {
--- a/src/app/content/components/Page/scrollToTopOrHashManager.ts
+++ b/src/app/content/components/Page/scrollToTopOrHashManager.ts
@@ -2,7 +2,7 @@ import { HTMLElement } from '@openstax/types/lib.dom';
 import { scrollTo } from '../../../domUtils';
 import * as selectNavigation from '../../../navigation/selectors';
 import { AppState } from '../../../types';
-import { assertWindow, memoizeStateToProps, resetTabIndex } from '../../../utils';
+import { assertWindow, memoizeStateToProps } from '../../../utils';
 import { isSearchScrollTarget } from '../../search/guards';
 import { selectedResult } from '../../search/selectors';
 import * as select from '../../selectors';
@@ -26,17 +26,19 @@ const scrollToTarget = (container: HTMLElement | null, hash: string) => {
   }
 };
 
-const scrollToTargetOrTop = (container: HTMLElement | null, hash: string) => {
+const scrollToTargetOrTop = (container: HTMLElement | null, hash: string, previous: boolean) => {
   if (getScrollTarget(container, hash)) {
     scrollToTarget(container, hash);
-  } else {
+  } else if (previous) {
     scrollToTop();
   }
 };
 
 const scrollToTop = () => {
   const window = assertWindow();
-  resetTabIndex(window.document);
+  const document = window.document;
+
+  document.querySelector('main')?.focus();
   window.scrollTo(0, 0);
 };
 
@@ -59,7 +61,7 @@ const scrollToTopOrHashManager = (
     return;
   }
   if (previous?.page !== current.page) {
-    scrollToTargetOrTop(container, current.hash);
+    scrollToTargetOrTop(container, current.hash, Boolean(previous));
   } else if (previous?.hash !== current.hash) {
     scrollToTarget(container, current.hash);
   }
--- a/src/app/content/components/__snapshots__/Content.spec.tsx.snap
+++ b/src/app/content/components/__snapshots__/Content.spec.tsx.snap
@@ -712,6 +712,10 @@ Array [
   overflow: visible;
 }
 
+.c79:focus-visible {
+  outline: none;
+}
+
 .c79 .highlight {
   position: relative;
   z-index: 1;
@@ -5514,6 +5518,10 @@ Array [
 
 .c79 {
   overflow: visible;
+}
+
+.c79:focus-visible {
+  outline: none;
 }
 
 .c79 .highlight {
--- a/src/app/content/components/__snapshots__/Content.spec.tsx.snap
+++ b/src/app/content/components/__snapshots__/Content.spec.tsx.snap
@@ -3299,7 +3299,6 @@ li[aria-label="Current Page"] .c61 {
             >
               <input
                 aria-label="Search this book"
-                autoFocus={true}
                 className="c35"
                 data-testid="mobile-search-input"
                 onChange={[Function]}
@@ -7895,7 +7894,6 @@ li[aria-label="Current Page"] .c61 {
             >
               <input
                 aria-label="Search this book"
-                autoFocus={true}
                 className="c35"
                 data-testid="mobile-search-input"
                 onChange={[Function]}
--- a/src/app/content/__snapshots__/routes.spec.tsx.snap
+++ b/src/app/content/__snapshots__/routes.spec.tsx.snap
@@ -279,6 +279,10 @@ Array [
   overflow: visible;
 }
 
+.c4:focus-visible {
+  outline: none;
+}
+
 .c4 .highlight {
   position: relative;
   z-index: 1;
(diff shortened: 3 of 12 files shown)


src/app/components/MainContent.tsx
```diff
@@ -31,22 +31,36 @@ const ContentStyles = styled(({ textSize, ...props }) => <DynamicContentStyles {
...
-        book={book}
+  ({book, children, className, ...props}, ref) => {
+    const initialLoad = React.useRef(true);
+
+    React.useEffect(
+      () => {
+        if (initialLoad.current) {
+          initialLoad.current = false;
+        } else {
+          (ref as React.MutableRefObject<HTMLDivElement>).current?.focus();
```
Dantemss, line 42: Just curious why the typecast is needed

Dantemss (approved):
I feel like in principle it should work

src/app/components/MainContent.tsx
RoyEJohnson: This ref comes in as a union of this type and a function that doesn't have `current`.
I'm open to the idea that I should [handle the function-ref case](https://stackoverflow.com/a/76739143), but this was my inclination.

RoyEJohnson: I have updated the code using the linked approach.

Malar-Natarajan:
@RoyEJohnson The CI check `Test` is failing. Could you please have a look.

src/app/components/MainContent.tsx
```diff
@@ -31,22 +31,40 @@ const ContentStyles = styled(({ textSize, ...props }) => <DynamicContentStyles {
...
+    const initialLoad = React.useRef(true);
+    const ref = React.useRef<HTMLDivElement>(null);
+
+    React.useImperativeHandle(forwardedRef, () => ref.current as HTMLDivElement);
+
+    React.useEffect(
+      () => {
+        if (initialLoad.current) {
+          initialLoad.current = false;
+          document?.getElementById('root')?.querySelector('a')?.focus();
```
TomWoodward, line 44: the task is to focus the main container right? why are you focusing the first anchor here?

TomWoodward (changes requested):
i don't know about coupling this change to a re-render of the MainContent children, aren't there situations where children will change but still be the same page? seems liek before it was related to an explicit navigation event, was that not working for some reason?

src/app/components/MainContent.tsx
RoyEJohnson: Because the first load of the page should focus on the Skip to content link. Only subsequent changes should focus on changed content. I guess this is the function that `resetTabIndex` was doing previously, but it did it on every content change. Setting focus to document.body might be more clear (or, equivalently, calling blur on activeElement)

RoyEJohnson:
@Malar-Natarajan Got it figured out.

Malar-Natarajan:
@RoyEJohnson looks like there are changes requested. The merge button is disabled.

src/app/components/MainContent.tsx
RoyEJohnson: Maybe we just want this always to happen where previously we had it resetting the focus.
Intuitively, that sounds likely. `scrollToTopOrHashManager` is called by PageComponent on DidMount and DidUpdate. They are differentiated by whether there is a `previous`.

TomWoodward (changes requested):
i'm not totally clear on what your changes are trying to accomplish, is the idea to stop `resetTabIndex` from being called when navigating to new pages? do we need `resetTabIndex` to exist at all if you've added another place that focuses the `main` element? would it be more correct to update `resetTabIndex` to focus the `main` element instead?

RoyEJohnson:
@TomWoodward
`resetTabIndex` sets the focus so that it's ready to go to the Skip to content on the first tab press (or something like that). It was previously set to run every time `scrollToTop` was called, which happened basically on every content update.
My first stab at this took it out entirely, but we still want it for the initial load.
For other content updates, we need the container of the updated content to get focus. And I just realized I don't have that happening yet.

RoyEJohnson:
I think I've got this down to the most straightforward way of doing it.

src/app/content/components/Page/scrollToTopOrHashManager.ts
```diff
@@ -26,17 +26,21 @@ const scrollToTarget = (container: HTMLElement | null, hash: string) => {
...
 };
 
-const scrollToTop = () => {
+const scrollToTop = (previous: boolean) => {
   const window = assertWindow();
-  resetTabIndex(window.document);
+  if (previous) {
+    document?.querySelector('main')?.focus();
+  } else {
+    resetTabIndex(window.document);
```
TomWoodward, line 42: does this need to happen on initial load? doesn't the focus start at the top by default?

TomWoodward (changes requested):

src/app/content/components/Page/scrollToTopOrHashManager.ts
RoyEJohnson: You are correct, it works without it.

RoyEJohnson: But when I take it out, the lighthouse test fails. Put it back in, it passes.

RoyEJohnson: I think the best I can do is inline the resetTabIndex sequence. It seems that the Lighthouse test does not set up with the focus ready to move to the Skip-to link.

TomWoodward: which test are we talking about?

RoyEJohnson: The [a11y Lighthouse](https://github.com/openstax/rex-web/actions/runs/9776275545/job/26988439259?pr=2221#step:6:830) check

RoyEJohnson: The [failure](https://github.com/openstax/rex-web/actions/runs/9779824805/job/26999974157#step:6:770) indicates the focus isn't ready to tab to Skip to content.

TomWoodward:
@RoyEJohnson the test failure on the `isSkipToContentSelected` test appears to be a legitimate one, the search input in the toolbar is grabbing focus, i can reproduce this on production. its the `mobile-search-input`, which is shown even on medium screen sizes, its the same breakpoint for when the toc is default closed

the lighthouse error (i added some debug logging for the failing lighthouse audit) is some kind of race condition with the osano consent banner. this is interesting because the `jest-puppeteer.config.json` (which is being used for this test) attempts to block osano loading.  

@RoyEJohnson can you take over making the mobile search input not grab focus unless it is the hidden one that you have to click the button to show

@Dantemss can you look into why osano is being loaded in the puppeteer tests?

TomWoodward (changes requested):
that recent merge should fix the lighthouse error, just the mobile input focus needs to be fixed for the overall focus management to work

src/app/content/components/__snapshots__/Content.spec.tsx.snap
```diff
@@ -3272,7 +3272,6 @@ li[aria-label="Current Page"] .c60 {
             >
               <input
                 aria-label="Search this book"
-                autoFocus={true}
```
TomWoodward, line 3275: but we do want this to become focused if you're on the extra small mobile view where the input is hidden initially and you need to press a button to open it

TomWoodward (changes requested):

src/app/content/components/__snapshots__/Content.spec.tsx.snap
RoyEJohnson: For the record, that does not happen in Production. You press the button to open the search input and you still have to navigate to it.

RoyEJohnson: I have added an Effect to the input to grab focus when the mobile toolbar opens.

TomWoodward (approved):

Malar-Natarajan:
@RoyEJohnson could you please resolve conflicts in this PR

RoyEJohnson:
@Malar-Natarajan Resolved.

TomWoodward (changes requested):
would it have been simpler to call `this.container.focus()` from PageComponents componentDidUpdate function? there is already a bunch of stuff there that happens in reaction to the page content being updated, it would be a logical place to add this handler.

also doesn't it already scroll to the top of the page? there is a whole section of that logic that handles scrolling to the target anchor or top of the page

RoyEJohnson:
It does look like setting focus in ComponentDidUpdate simplifies things.

RoyEJohnson:
The build test error appears to be swapping the position of the `class` and `tabindex` attributes. 🤷

TomWoodward (approved):
logic is looking cleaner now, although now that i'm looking at where you changed the code in scrollToTargetOrTop i'm wondering if it would be best to focus the target instead of the whole content when linking to a specific target

RoyEJohnson:
Some scroll-to targets are not focusable. Footnotes, for example, are simply divs. They have id's so you can scroll to them, but no tabindex, so they don't focus.

TomWoodward:
i wonder what the intended screen reader functionality is there.... surely if its scrolling to a target on the page its not supposed to start reading from the top of the content

RoyEJohnson:
It would be reasonable for screenreaders to recognize that content has scrolled and start reading visible content.

RoyEJohnson:
I don't think I made any net changes to what's actually here, but the build test works now. All I did was restore some old versions of files to see what caused the test to fail. Then I got rid of those changes and squashed a few commits.

TomWoodward:
Somehow I feel like the screen reader is not going to do the right thing here with the target scrolling, but I'll approve what you have working so far

TomWoodward (approved):

PR #2224, opened 2024-04-19
Link accessible stylesheet for Osano

RoyEJohnson:
[DISCO-238]

[DISCO-238]: https://openstax.atlassian.net/browse/DISCO-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -10,7 +10,7 @@
     -->
     <link rel="manifest" href="/dist/assets/manifest.json">
 
-
+    <link rel="stylesheet" href="https://ram.openstax.org/osano/accessible.css">
     <!--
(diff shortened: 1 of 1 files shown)


jivey (changes requested):
The code change itself looks good but I'm wondering about potential confusion regarding where osano style tweaks (accessible or otherwise) should live. `index.html` already has inline osano styles, maybe these should live there? Or if we need these to be in RAM, maybe those inline styles should move there too and we call it like `osano.css`?

jivey (approved):

PR #2235, opened 2024-05-13
Update ABL scripts to use CORGI

TylerZeroMaster:
part of openstax/ce#2182

This new script effectively performs the same steps as the old transformation script:
- Fetch ABL from https://corgi.ce.openstax.org/api/abl/ 
   - Fetches entries intended for REX where code_version <= `REACT_APP_ARCHIVE`
- Select the newest version of each book
- Select versions that are not in config.books.json

`REACT_APP_ARCHIVE` is read from `src/config.archive-url.json`

The tests show the input and output formats (the output format is the same as before: `{ <uuid>: <commit_sha>, ... }`)


script/validate-abl-import/test.bash
```diff
@@ -27,8 +27,10 @@ for input_file in "$my_dir"/*.input.json; do
     # Actual test goes here
     # -------------------------
     json=$(cat "$input_file")
-    node "$my_dir/../entry.js" transform-approved-books-data --data "$json" > "$output_file"
+    configured_books="{}"
```
TomWoodward, line 30: shouldn't there be tests that exercise the logic of diffing the books against the existing config?

TomWoodward (changes requested):

script/validate-abl-import/test.bash
TylerZeroMaster: Thanks for the quick review! Yes, that completely slipped my mind. I added some additional test data for this. The smaller of the new data sets demonstrates what occurs when new versions are present, the other uses some real data to demonstrate that there are no new entries when all ABL versions are already present.

TomWoodward (approved):

TomWoodward:
i think we should wait to merge this until the current batch of updates is done.

@TylerZeroMaster is the new system being used already? are we good to merge this whenever?

TylerZeroMaster:
> i think we should wait to merge this until the current batch of updates is done.
> 
> @TylerZeroMaster is the new system being used already? are we good to merge this whenever?

@TomWoodward The new system is in production, but it is not being used yet: there is another PR in progress to update Enki/webhosting pipeline to use the new system as well. We can wait to merge that Enki PR alongside this one.

Additionally, if the ABL on Github is updated, the version on CORGI will need to be updated to match. I have a script that will "fast-forward" the CORGI ABL data to match what is present on Github, but I have been running it manually.

Overall, the changes can be merged anytime, but I would appreciate a little bit of warning ahead of time.

script/transform-approved-books-data.bash
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -e
+
+data="$1"
```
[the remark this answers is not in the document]
TylerZeroMaster: That is a good point. I originally used inputs like this to separate fetching from transformation and the errors that could occur in each process. I updated the transformation script to read input from 2 files and do some additional checks on the input.

TomWoodward:
@TylerZeroMaster ok, we will shoot to merge this maybe next week after the current batch of updates is done. 

also the "ABL Schema Tests' check is failing with some jq compile error

RoyEJohnson (approved):

TylerZeroMaster:
> @TylerZeroMaster ok, we will shoot to merge this maybe next week after the current batch of updates is done.
> 
> also the "ABL Schema Tests' check is failing with some jq compile error

Ahh, I think that was because I was using jq version 1.7 which has some changes that must not be backwards compatible. It has been fixed. Let me know when we are ready to merge this change.

PR #2238, opened 2024-05-14
Redirect search param

bethshook:
https://openstax.atlassian.net/browse/DISCO-363


src/app/content/utils/processBrowserRedirect.ts
```diff
@@ -2,21 +2,28 @@ import { History } from 'history';
...
     .then((res) => res.json())
     .catch(() => []);
 
   for (const {from, to} of redirects) {
     if (from === services.history.location.pathname) {
-      if (matchForRoute(external, services.router.findRoute(to))) {
-        window.location.href = window.location.origin + to;
-      }
-      services.history.replace(to);
+      const {pathname, search, hash} = new URL(to, base);
```
TomWoodward, line 23: i would add the url parsing within `findRouteMatch` if the given location is a string. the option for it to be a string is something we added in your previous redirects change so that should be safe

src/app/content/utils/processBrowserRedirect.ts
```diff
@@ -2,21 +2,28 @@ import { History } from 'history';
...
+      const match = services.router.findRoute({pathname, search, hash, state: {}});
+      services.history.replace(to); // hack to show blank page instead of flashing rex when target is external
```
TomWoodward, line 25: i believe if you delete [this return statement](https://github.com/openstax/rex-web/blob/main/src/app/navigation/middleware.ts#L27C7-L27C13) then it'll route to the error route screen, which renders a loader, that might be a better solution than this to prevent displaying rex

TomWoodward (changes requested):

src/app/navigation/utils/index.ts
```diff
@@ -61,8 +62,9 @@ export const findRouteMatch = (routes: AnyRoute[], location: Location | string):
       const keys: Key[] = [];
       const re = pathToRegexp(path, keys, {end: true});
       const match = re.exec(typeof location === 'string' ? location : location.pathname);
+      const search = typeof location.search === 'string' ? location.search : '';
       if (match) {
-        return formatRouteMatch(route, (typeof location !== 'string' && location.state) ?? {}, keys, match);
+        return formatRouteMatch(route, (typeof location !== 'string' && location.state) ?? {}, keys, match, search);
```
bethshook, line 67: @TomWoodward without the changes to this file, the `?message=retired` redirects will never hit `matchSearch` because there is no search param on the match

bethshook: (this is re your suggestion to incorporate your changes to matchSearch and remove my change to formatRouteMatch)

src/app/content/utils/processBrowserRedirect.ts
bethshook: the `?message=retired` redirects I'm trying to handle don't hit that return statement because they don't match a notFound or external content route

src/app/navigation/utils/index.ts
TomWoodward: well in that case the target is still a rex page, i guess the issue is you don't want to flash the previous book before switching to the new book....

can we like.... ignore that a little bit for now? maybe a better solution would be to rearrange the resolveContent hook to not dispatch receiveBook if there is a redirect

bethshook: I can remove the `services.history.replace(to)` from processBrowserRedirect if that's what you mean - I don't think the flash is a big deal. could rearrange resolveContent but that will take more time.

TomWoodward: i think just remove the replace and call it good for now, unless you've already done more

bethshook: it's removed

TomWoodward (approved):

RoyEJohnson (approved):

PR #2244, opened 2024-05-24
Fix focus after go-to-top

RoyEJohnson:
[DISCO-308]

[DISCO-308]: https://openstax.atlassian.net/browse/DISCO-308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/highlights/components/HighlightsPopUp.tsx
+++ b/src/app/content/highlights/components/HighlightsPopUp.tsx
@@ -142,7 +142,7 @@ const HighlightsPopUp = ({ closeMyHighlights, ...props }: Props) => {
           <CloseIcon colorSchema={props.bookTheme}/>
         </CloseIconWrapper>
       </Header>
-      {props.user ? <ShowMyHighlights /> : <LoginForHighlights />}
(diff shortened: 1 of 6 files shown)


Dantemss (approved):

TomWoodward (commented):
Where does the existing refElement point to? That's the scroll container right? Is that not a good thing to focus when clicking the button?

RoyEJohnson:
The focus gets set to the Modal container, which was [created to be focusable](https://github.com/openstax/rex-web/pull/2244/files#diff-27b81276801e32883304efcdc5c389792d19980a49c1141cc02ba70495b31dfcR43) before this PR.
But maybe I'm misunderstanding what you're asking, @TomWoodward ?

TomWoodward:
No that seems fine, I was just wondering if the same ref could be used for both focusing and scrolling, but it sounds like the way you have it set up makes sense

PR #2249, opened 2024-05-31
Ensure lazy loaded images load when visible initially

RoyEJohnson:
[DISCO-285]

Lazy loaded items don't always notice that they're visible. There was a check already for when the component updated, but that doesn't fire on initial load, so it had to be added to `ComponentDidMount` as well.

[DISCO-285]: https://openstax.atlassian.net/browse/DISCO-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (approved):
Interesting, I'm guessing this is from of a fixed "bug" in Chrome at some point? Seems like an good change to me, though you may want to update the PR title.

PR #2250, opened 2024-06-03
DISCO-326 Fix prerendering tags

Dantemss:
My guess is they started actually copying the stack tags into the fleet resource


RoyEJohnson (approved):
It's doing variable substitution without variable indicators now?

PR #2253, opened 2024-06-10
Update assignable launch

TomWoodward:
https://openstax.atlassian.net/browse/DISCO-14

adds token parsing for assignable integration and textSize support from the assignable nav. integrates with changes from https://github.com/openstax/assignments/compare/main...text-resizer?expand=1


src/app/content/components/AssignedTopBar.tsx
```diff
@@ -39,21 +42,60 @@ const StyledSectionTitle = styled.h2`
   `)}
 `;
 
+declare const window: any;
```
jivey, line 45: I don't notice any custom properties so it seems like you wouldn't need it this.... maybe prerendering related?

jivey (approved):
Looks good, just left a few questions to better my understanding.

src/app/content/launchToken.ts
```diff
@@ -0,0 +1,44 @@
...
+  }).join(''));
+
+  const token = JSON.parse(jsonPayload);
+  return 'sub' in token ? JSON.parse(token.sub) : undefined;
+};
+
+export const pullToken = (window: Window) => {
+  const searchParams = new URLSearchParams(window.location.search);
+
+  const launchToken = searchParams.get('t') ?? undefined;
```
RoyEJohnson, line 21: The `??` branch seems pointless.

RoyEJohnson (approved):

src/app/content/components/AssignedTopBar.spec.tsx
```diff
@@ -11,8 +11,17 @@ import { MiddlewareAPI, Store } from '../../types';
 import { setTextSize } from '../actions';
 import { LinkedArchiveTreeSection } from '../types';
 import { AssignedTopBar } from './AssignedTopBar';
+import { assertDocument, assertWindow } from '../../utils/browser-assertions';
+
+jest.mock('react', () => {
+  const react = (jest as any).requireActual('react');
+  return { ...react, useEffect: react.useLayoutEffect };
```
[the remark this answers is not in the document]
TomWoodward: its possible that newer versions of react or the testing library don't have this problem, but there are a few places in the rex code that do this. the react issue is here https://github.com/facebook/react/issues/14050#issuecomment-438173736

src/app/content/components/AssignedTopBar.tsx
TomWoodward: looks like an artifact of an earlier age

src/app/content/launchToken.ts
TomWoodward: you're right, the conditional below catches it either way

PR #2258, opened 2024-06-17
improve contrast on button focus outline

bethshook:
https://openstax.atlassian.net/browse/DISCO-216


jivey (approved):

src/app/components/Button.tsx
```diff
@@ -82,6 +82,11 @@ const Button = styled(ButtonHoc)`
     border: none;
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
+
+    &:focus {
+      outline: solid ${theme.color.white};
+      box-shadow: inset 0 0 0 3px ${theme.color.black};
```
jivey, line 88: Might want to make this rem?

RoyEJohnson: There's a `defaultFocusOutline` style in src/app/theme.ts that you might consider. It tries to make the outline the same as the browser default. Of course, black and white will give you maximum contrast.

RoyEJohnson (approved):

src/app/components/Button.tsx
bethshook: that's good to know but this change is actually in response to a browser default - one of them is too close to the orange of this button

PR #2269, opened 2024-07-02
DISCO-290 Set Rex heading levels for Assignable

Dantemss:
I can see like 10 different ways to do this so maybe it's best if you told me if you like this or not


TomWoodward (changes requested):
overall this is fine, just some comments on the actual mutation

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -42,6 +46,40 @@ function removeDocumentTitle(rootEl: HTMLElement) {
   ].join(',')).forEach((el) => el.remove());
 }
 
+// set the top heading's level to topHeadingLevel and adjust other headings accordingly
+function changeHeadingLevels(document: Document, rootEl: HTMLElement, topHeadingLevel: number) {
+  const headingLevels = [ 1, 2, 3, 4, 5, 6 ];
+  const currentTopHeading = headingLevels.find((level) => rootEl.querySelectorAll(`h${level}`)?.length);
+
+  if (!currentTopHeading || topHeadingLevel <= currentTopHeading) {
```
TomWoodward, line 54: if desired topHeading level is 2, and the currentTopHeading is 3, we don't want to skip we want to bump them all up.

if they are equal we can return

TomWoodward (changes requested):
would be nice to see a spec, but looks generally good to me

src/app/content/components/Page/contentDOMTransformations.ts
bethshook: ah ok, was confused on this point

TomWoodward (approved):
looks good but hold off for a sec while we verify if the top level should be 2 or 3

PR #2270, opened 2024-07-03
Content warning

RoyEJohnson:
[DISCO-341]

[DISCO-341]: https://openstax.atlassian.net/browse/DISCO-341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/content/components/Page/PageComponent.tsx
```diff
@@ -151,6 +152,7 @@ export default class PageComponent extends Component<PagePropTypes> {
       <this.highlightManager.CardList />
       <PT />
       <RedoPadding>
+        <ContentWarning book={this.props.book} />
```
TomWoodward, line 155: can you move the usage of this to Content (src/app/content/components
/Content.tsx) and use a redux selector to pull the book in within the component

the PageComponent can be delicate and i don't want to add stuff to it, and moving it to Content will ensure the prompt doesn't show up for assignable users

TomWoodward (changes requested):

TomWoodward (changes requested):
the most recent failure is that the fixture files need to be updated. the fixtures are in `src/test/fixtures/apps/cms/api/v2/pages` and are named by url, so the file name needs to be updated to include the new query parameter, and the body of the file updated to include the expected value for the field

src/app/content/components/cookie.ts
```diff
@@ -0,0 +1,36 @@
+function tenYearsFromNow() {
```
TomWoodward, line 1: there is a cookie npm library that should handle all this stuff in this file

src/app/content/components/ContentWarning.tsx
```diff
@@ -0,0 +1,74 @@
...
+function checkCookie(id: string) {
+  return cookieKey(id) in cookie.hash;
+}
+
+function cookieKey(id: string) {
+  return `content-warning-${id}`;
+}
+
+function inAWeek() {
+  return new Date(Date.now() + 7 * 24 * 3600 * 1000);
```
TomWoodward, line 73: the spec was 28 days

TomWoodward (changes requested):

TomWoodward (approved):

jivey (approved):

PR #2268, opened 2024-07-01
update prerender to include a content manifest csv output

TomWoodward:
https://openstax.atlassian.net/browse/DISCO-23

adds a csv output of all book toc node info for use by reporting platforms to associate uuids with slugs with titles (for querying and output formatting in reporting dashboards)

also moved sitemap index output to queue generation to avoid loading all the books twice

PR #2326, opened 2024-08-12
fix build

TomWoodward:
https://openstax.atlassian.net/browse/DISCO-467

the problem seems to be because in https://github.com/openstax/rex-web/pull/2268/files i moved the `renderAndSaveSitemapIndex` call out of a worker and into the scheduling thread, because all the required book data is already loaded. but because the scheduling thread is not running in a worker it didn't have the same environment variables loaded (also missing the default value assigned in `fleet.ts`). i moved some of the config definitions to a new file so they could share their defaults, i think that'll fix it.

### DIFF
--- a/script/prerender/fileUtils.ts
+++ b/script/prerender/fileUtils.ts
@@ -7,7 +7,7 @@ import once from 'lodash/once';
 import path from 'path';
 import createCache, { Cache } from '../../src/helpers/createCache';
 import { directoryExists, readFile, writeFile } from '../../src/helpers/fileUtils';
-import { BUCKET_REGION, PUBLIC_URL, BUCKET_NAME } from "./constants";
+import { BUCKET_REGION, PUBLIC_URL, BUCKET_NAME } from './constants';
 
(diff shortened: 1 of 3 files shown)


script/prerender/fileUtils.ts
```diff
@@ -51,7 +51,7 @@ export const createDiskCache = <K extends string, V>(prefix: string): Cache<K, V
 
 // Generates a release path for a file without a leading /, used when uploading the release to S3
 function prefixReleasePath(filepath: string) {
-  let basePath = assertDefined(process.env.PUBLIC_URL, 'PUBLIC_URL environment variable not set');
+  let basePath = `${PUBLIC_URL}`; // clone string to not leak the potential slice on next line.
```
Dantemss, line 54: Should be fine without since slice() doesn't modify the original string.

Dantemss (approved):
Seems good to me. Let's give it a try.

script/prerender/fileUtils.ts
TomWoodward: ah right

PR #2328, opened 2024-08-14
update pipeline version

bethshook:
https://openstax.atlassian.net/browse/DISCO-469


jivey (approved):
I think there were some issues with this pipeline (that maybe are resolved?) The code change looks good though.

PR #2332, opened 2024-08-21
Make highlight Actions menu items (Edit and Delete) buttons

RoyEJohnson:
[DISCO-307]
For clarity, changed `ol` to `menu`
Also simplified return value of DropdownItem.

[DISCO-307]: https://openstax.atlassian.net/browse/DISCO-307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/Dropdown.spec.tsx
+++ b/src/app/components/Dropdown.spec.tsx
@@ -107,6 +107,7 @@ describe('Dropdown', () => {
     const useOnEscSpy = jest.spyOn(reactUtils, 'useOnEsc');
 
     const focus = jest.fn();
+    const focus2 = jest.fn();
     const addEventListener = jest.fn();
     const removeEventListener = jest.fn();
     const createNodeMock = () => ({focus, addEventListener, removeEventListener});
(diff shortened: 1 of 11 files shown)


RoyEJohnson:
Lint had a problem with the "Stop containers" phase.
image

jivey (changes requested):
The markup changes are great. I need some clarification regarding the card, shouldn't the first menu item get focused? 

There's also a very minor style regression in the MH modal 
image

RoyEJohnson:
As I read [the comment](https://openstax.atlassian.net/browse/DISCO-307?focusedCommentId=12169), the focus concern is that the "dialog" (the DisplayNote) doesn't get focus when it appears. That is by design; it's what alt-H is for.
In the Disclosure Pattern, the focus does not automatically move from the control button. The user can open and close it repeatedly. They tab to the first revealed item. I'm not finding that to be a hard and fast rule, but it's what I recall LA communicated to us before.
I'll figure out the style regression.

RoyEJohnson:
1st: using `ol` tag. 2nd: using `menu` tag. 3rd: what's in dev
image
image
image

RoyEJohnson:
I'm going to have it move focus to the first item.

RoyEJohnson:
@jivey Focusing on the first item can result in the hover-highlight competing with the focus-highlight, so I had to add code to make focus follow the mouse within the menu.

jivey (changes requested):
The styles look good, but with the hover change, it can get into a weird state where one item is highlighted by the mouse and the other the keyboard:
image

I'm not sure how [react-aria-component's Menu](https://react-spectrum.adobe.com/react-spectrum/Menu.html#menu) handles it  internally, but that seems like a good interaction goal if we can make it work like that.

RoyEJohnson:
I resolved the focus/hover issue by styling focus and not hover.

jivey (approved):
Looks good and feels good to use, thanks for going through those review rounds with me.

PR #2336, opened 2024-08-23
fix: outdated `caniuse-lite`

### LINKED ISSUE #2335: `Browserslist: caniuse-lite is outdated.`

hamirmahal:
fixes #2335.


TomWoodward (changes requested):
@hamirmahal thank you for the PR! I appreciate it, but unfortunately something about this change seems to be breaking the node/babel-register config that is used for the static site rendering part of the app, so it doesn't seem like this is going to be as simple as it looks.

TomWoodward:
closing due to inactivity. the check did pass again after reverting the update. looks like more work would be needed to make this change.

PR #2350, opened 2024-09-17
Do not set shouldFocusCard on new highlight

RoyEJohnson:
[CORE-514]

[CORE-514]: https://openstax.atlassian.net/browse/CORE-514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -50,9 +50,9 @@ const Wrapper = ({highlights, className, container, highlighter}: WrapperProps)
     if (element.current.contains(activeElement)) {
       focusedHighlight.focus();
     } else {
-      setShouldFocusCard(true);
+      setShouldFocusCard(focusedId !== undefined);
```
Dantemss, line 53: I'm trying to reason about this change but I'm getting stuck... it seems you're fixing the case when `focusedId === undefined`. But would it even get to this point? `focusedHighlight` should not find anything in that case unless `highlight.id` is also undefined. Is this something that happens?

RoyEJohnson: This is something that happens due to timing issues in Firefox. In some cases, deleting a highlight followed by highlighting some text caused the focus to jump to the highlight editor, causing the color to be selected automatically. This change stopped that behavior.

Dantemss (approved):

PR #2364, opened 2024-10-21
fix crash when creating initials for user without first and last name

jivey:
https://openstax.atlassian.net/browse/CORE-587

Assignable students do not have first and last names associated, so REX crashes trying to make initials for the dropdown menu. Suggestions for alternatives to "AU" (Anonymous User? Assignable User?) are welcome.

### DIFF
--- a/src/app/components/NavBar/__snapshots__/index.spec.tsx.snap
+++ b/src/app/components/NavBar/__snapshots__/index.spec.tsx.snap
@@ -180,32 +180,8 @@ exports[`content in browser assignable user renders 1`] = `
   box-shadow: 0 0 0.2rem 0.2rem rgba(0,0,0,0.3);
 }
 
-.c6 {
-  display: -webkit-box;
-  display: -webkit-flex;
-  display: -ms-flexbox;
-  display: flex;
-  -webkit-box-pack: center;
-  -webkit-justify-content: center;
-  -ms-flex-pack: center;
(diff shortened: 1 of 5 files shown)


bethshook:
that's a tough one. my only other ideas are `ME` and `OS`

bethshook (approved):

RoyEJohnson:
A question mark would be unambiguous.

RoyEJohnson (approved):

TomWoodward:
if you made them spaces it would show an empty circle, i feel like even that would be less confusing than `AU`, i'm not sure what people are gonna think when they see that. only other thing i can think of is change the thing entirely to show like a profile icon (maybe only if there are no initials?)

src/app/components/NavBar/__snapshots__/index.spec.tsx.snap
```diff
@@ -1135,6 +1135,31 @@ exports[`content in browser matches snapshot for logged out 1`] = `
...
+  height: 6rem;
+  max-width: 128rem;
+  margin: 0 auto;
+}
+
+.c2 {
+  display: block;
+  width: auto;
+  height: 3.5rem;
+}
```
jivey, line 1161: Something with the snapshots is duplicating styles, there's multiple copies of the same rules in this spec...

TomWoodward (approved):

PR #2366, opened 2024-10-24
Text size

TomWoodward:
matching text size name change and token format change from this pr https://github.com/openstax/assignments/pull/199


jivey (changes requested):
Changes look good but we probably want some tests for the token change, or did the existing ones not break?

TomWoodward:
the existing checks didn't break but i should check the coverage i'm sure i have to add something

TomWoodward:
i tried to fix all the warning output, locally i think there might be one more async setState but its only popping up intermittently and i'm not sure from where. in the github action outptut i also see some weird babel error, i'm gonna call it good enough for now. test coverage is back up to 100% and i added the threshold to the jest config

RoyEJohnson (approved):
Looks reasonable to me.

src/app/content/hooks/storeTextSize.spec.ts
```diff
@@ -81,6 +81,13 @@ describe('loadStoredTextSize', () => {
   });
 
   it('loads the value from launchToken', async() => {
+    helpers.launchToken = {tokenString: '', tokenData: {text_size: 2}};
```
bethshook, line 84: how is this different from the test below?

src/app/content/search/components/SearchResultsSidebar/__snapshots__/index.spec.tsx.snap
```diff
@@ -872,7 +872,7 @@ exports[`SearchResultsSidebar matches snapshot with related key terms 1`] = `
         aria-current={true}
         className="c15 c16"
         data-testid="related-key-term-result"
-        href="books/book-slug-1/pages/test-page-1?query=cool%20search&target=%7B%22index%22%3A0%2C%22type%22%3A%22search%22%7D#test-pair-page1"
```
bethshook, line 875: I'm not clear on why the query is removed

TomWoodward: It isn't set in the test, it was leaking from another one and now it isn't

src/app/content/highlights/components/EditCard.spec.tsx
```diff
@@ -20,9 +20,9 @@ import * as onClickOutsideModule from './utils/onClickOutside';
 import { MAIN_CONTENT_ID } from '../../../context/constants';
 import { renderToDom } from '../../../../test/reactutils';
 
-jest.mock('./ColorPicker', () => (props: any) => <div mock-color-picker {...props} />);
-jest.mock('./Note', () => (props: any) => <div mock-note {...props} ref={props.textareaRef} />);
-jest.mock('./Confirmation', () => (props: any) => <div mock-confirmation {...props} />);
+jest.mock('./ColorPicker', () => (props: any) => <div mock-color-picker='true' data-props={props} />);
+jest.mock('./Note', () => (props: any) => <div mock-note='true' data-props={props} ref={props.textareaRef} />);
+jest.mock('./Confirmation', () => (props: any) => <div mock-confirmation='true' data-props={props} />);
```
jivey, line 25: It makes sense that these shouldn't get spread onto the div, but I wonder if they should get added at all, they don't seem to be used?

src/app/content/__snapshots__/routes.spec.tsx.snap
```diff
@@ -778,7 +778,7 @@ Array [
...
...turpis. Donec mollis ac massa ac elementum. Nulla maximus purus sed orci efficitur, ut porta mauris elementum.</p><p id=\\"p9\\">Vestibulum blandit enim quis sapien ultrices pretium. Cras dapibus ornare fringilla. Duis ut laoreet dolor. Proin ullamcorper molestie dui eu vestibulum. Nullam convallis mollis mauris nec accumsan. Proin egestas enim non nisl suscipit, ut feugiat felis pharetra. In faucibus fringilla ipsum vel sollicitudin. Maecenas blandit augue odio, ac mattis justo feugiat sit amet. Etiam nec molestie eros. Sed gravida velit libero, at pretium diam venenatis sit amet. Donec tellus leo, faucibus eu sapien nec, mollis efficitur lorem. Lorem ipsum dolor sit amet, consectetur adipiscing elit.</p></section><dl id=\\"test-pair-page1\\"><dt>test term</dt><dd>test definition with <math>math element</math></dd></dl></div><img alt=\\"A photo is shown of the Crab Nebula.\\" data-media-type=\\"image/jpeg\\" id=\\"89491\\" src=\\"/apps/image-cdn/v1/f=webp/apps/archive/codeversion/resources/c7cc4c5e00a4bc07cb74f29cac2fb3c6cd6d53b5\\"><a href=\\"../../../books/book-slug-1/pages/test-page-1#hash\\" target=\\"_blank\\">hashlink</a><a href=\\".\\" target=\\"_blank\\">Current link</a>",
```
jivey, line 781: Seems like we changed the document-title from an h3 to an h2 but the snapshot didn't get updated? Just making sure.

jivey (approved):
Left some very minor questions but seems good to me

bethshook (approved):

src/app/content/highlights/components/EditCard.spec.tsx
TomWoodward: they're there so that they show up in the snapshot, but react was complaining about the non-divvy attributes

src/app/content/__snapshots__/routes.spec.tsx.snap
TomWoodward: yea this was the assignable view heading stuff, the tests were never updated

PR #2383, opened 2025-01-08
Update eslint settings and fix errors

RoyEJohnson:
[CORE-674]
In addition to fixing uses of `any`, there were a few non-null assertions that I changed to casts. If that seems pointless, I could disable the rule and change the casts back to assertions.

[CORE-674]: https://openstax.atlassian.net/browse/CORE-674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,26 @@
 {
-  "extends": ["react-app"],
+  "extends": [
+    "react-app",
+    "react-app/jest",
+    "eslint:recommended",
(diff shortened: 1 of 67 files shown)


TomWoodward (approved):
looks fine to me as long as the types are all still working right

PR #2388, opened 2025-01-19
ci: only run `Jira` job on pull requests

### LINKED ISSUE #2387: `CI / Jira` has failed on the last several pushes to `main`
`CI / Jira` has failed on the last several pushes to `main`.

This creates a lot of noise. If the check isn't supposed to run on pushes to `main`, we can skip it.

hamirmahal:
Closes #2387


.github/workflows/ci.yml
```diff
@@ -10,6 +10,7 @@ jobs:
   linked_issue:
     name: Jira
     runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
```
hamirmahal, line 13: If this job isn't supposed to run on pushes to `main`, we can skip it.

This will reduce noise and allow contributors and maintainers to focus on failures that are actually problematic.

PR #2433, opened 2025-02-19
Core 750: proper custom controls

jomcarvajal:
Card: https://openstax.atlassian.net/browse/CORE-750

Screenshot 2025-02-19 at 1 57 00 PM
Screenshot 2025-02-19 at 1 58 20 PM
Screenshot 2025-02-19 at 1 56 13 PM


TomWoodward (changes requested):
i think this needs more work reorganizing the markup.

first, the summary element states `A <summary> element may only be used as the first child of a <details> element.` also, in the [tree navigation](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) example linked in the report, the sub-tree list is not inside the anchor.

unless you found some docs saying that its ok for the subtree to be inside the anchor, i think you need to move it outside and link it with an id.

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -147,28 +172,24 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
...
-  }
+interface DetailsComponentProps {open: boolean; }
+class DetailsComponent extends React.Component<DetailsComponentProps> {
   public render() {
-    const {open, defaultOpen: _, ...props} = this.props;
-    const {defaultOpen} = this.state;
+    const {open, ...props} = this.props;
 
-    return <Details {...props} open={open || defaultOpen} />;
+    return <DetailsTree role='treeitem' {...props} open={open} />;
```
TomWoodward, line 187: should this have an href? maybe just `href="#"`? i think it might be required

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -150,51 +156,106 @@ function maybeAriaLabel(page: LinkedArchiveTreeSection) {
...
+  book: Book | undefined;
+  page: Page | undefined;
+  activeSection: React.RefObject<HTMLElement>;
+  onNavigate: () => void;
+}) {
+  const sectionType = getArchiveTreeSectionType(item);
+  const [isOpen, setOpen] = React.useState<boolean>(shouldBeOpen(page, item));
+  const toggleOpen = (event: React.MouseEvent<HTMLDivElement, MouseEvent> | React.KeyboardEvent<HTMLDivElement>) => {
+    if ('key' in event) {
+      if (event.key !== 'Enter' && event.key !== ' ') {
```
TomWoodward, line 179: i believe this needs a `preventDefault` and i think you may be required to add support for the additional keyboard operations defined https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/#kbd_label

you may actually be closer to achieving the disclosure pattern https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/ which i think would also be an acceptable solution here, it just requires different roles

TomWoodward (changes requested):

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -150,51 +156,106 @@ function maybeAriaLabel(page: LinkedArchiveTreeSection) {
...
   onNavigate,
+  role,
 }: {
+  id?: string;
   book: Book | undefined;
   page: Page | undefined;
   section: ArchiveTree;
   activeSection: React.RefObject<HTMLElement>;
   onNavigate: () => void;
+  role: 'tree' | 'group';
```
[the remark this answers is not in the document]
jomcarvajal: I tried with different books but I never received a role = 'tree' when matched book id and section id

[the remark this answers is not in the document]
TomWoodward: the one in the tree probably has a trailing `@`. there is a utility somewhere called i think `stripIdVersion` that you can call on the tree node id and then they'll match

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -82,28 +83,18 @@ export const NavItem = styled(NavItemComponent)`
...
 
-  ${/* suppress errors from https://github.com/stylelint/stylelint/issues/3391 */ css`
-    :hover ${SummaryTitle},
-    :focus ${SummaryTitle} {
+  a[role='treeitem'] {
+    :focus,
+    :hover {
+      outline: none;
       ${activeState}
-    }
```
RoyEJohnson, line 97: You need this to close the hover

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -147,28 +137,64 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
...
 // tslint:disable-next-line:variable-name
 export const NavDetails = styled(DetailsComponent)`
+  ${labelStyle}
+  display: flex;
   overflow: visible;
+  list-style: none;
+  cursor: pointer;
+  text-decoration: none;
+
+  :focus,
```
RoyEJohnson, line 176: Lint will complain about "Expected selector .... to come before selector ...."
You can silence that by inserting a comment before this line:
`/* stylelint-disable no-descending-specificity */`

RoyEJohnson: There are a lot of lint errors you'll have to fix. Some of them can be done automatically.
First, run `yarn lint` to see where the problems are. Then fix individual files with (for example)
`yarn tslint --fix src/app/content/components/TableOfContents/*.tsx`
That will catch things like wrong quote type and trailing whitespace and missing semicolon.

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -69,7 +69,7 @@ const SidebarBody = React.forwardRef<
   React.useEffect(
     () => {
       const firstItemInToc = mRef?.current?.querySelector(
-        'ol > li a, old > li summary'
+        'ol > li a, old > li div:first-child'
```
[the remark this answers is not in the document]
RoyEJohnson: Old is not a tag; it has to be `ol`.

src/__snapshots__/index.prerenderspec.ts.snap
```diff
@@ -1,15 +1,86 @@
...
-  \\"books\\": {
-    \\"testbook1-uuid\\": {
-      \\"defaultVersion\\": \\"1.0\\",
-      \\"dynamicStyles\\": true
-    }
-  },
-  \\"code\\": \\"test\\",
-  \\"id\\": \\"test\\"
-}"
+"<!DOCTYPE html>
```
RoyEJohnson, line 4: This snapshot (part of test:build) is failing. The result it is getting is a significantly larger webpage that is bundled rather than formatted for reading.
It's the response.text from fetching rex/release.json. Is this expected?

jomcarvajal: As far as I know I didn't modify nothing related with this. Can you help me to understand how to fix it ?

RoyEJohnson: Ask about it in the #engineering channel. It's beyond me.

jomcarvajal: It seems like the failed snapshot jumps in actual main branch. That means I update this snap by mistake and then github jumps with an error. So there is some kind of incongruity between local and remote branch. I just copy/paste snapshot from main and it works.

RoyEJohnson: Maybe it was spurious. It seems to be ok with the latest CI run.

src/app/content/components/ContentLink.tsx
```diff
@@ -95,14 +96,30 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
         navigate(navigationMatch, options);
       }
     }}
+    onKeyDown={(e) => onKeyDown && onKeyDown(e, async() => {
```
RoyEJohnson, line 99: You can actually write this execute-if-defined as `onKeyDown?.(e, async() => {`
An similarly for the `onClick` below.

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const searchTreeItem = (filteredTreeItems: Element[], currentItemIndex: number, key: string) => {
+    /*
+      According Keyboard Support for Navigation Tree, the search starts with treeItems
+      next to current treeItem and if there is no result, it will look for previous treeItems
+    */
+    const firstSearchOptions = filteredTreeItems.slice(currentItemIndex + 1, filteredTreeItems.length);
+    const secondSearchOptions = filteredTreeItems.slice(0, currentItemIndex);
+    let result;
+    for (const treeItem of secondSearchOptions) {
```
RoyEJohnson, line 71: I'm pretty sure you can concat the search options together and do a single `find`.
```
  const searchOptions = filteredTreeItems.slice(currentItemIndex + 1, filteredTreeItems.length)
    .concat(filteredTreeItems.slice(0, currentItemIndex));
  const result = searchOptions.find((treeItem) =>
    treeItem.textContent?.trim().toLowerCase().startsWith(key.toLowerCase())
  );
```

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const onKeyDownNavGroupSupport = ({
+    event,
+    item,
+    isOpen,
+    onSelect,
+  }: KeyboardSupportProps) => {
+    event.preventDefault();
+
+    const filteredTreeItems = getTreeItems();
```
RoyEJohnson, line 94: Extract these three lines to a little helper function:
```
function getCurrentItemIndex(item: KeyboardSupportProps['item']) {
  const filteredTreeItems = getTreeItems();
  const currentItem = filteredTreeItems.find((treeitem) => treeitem.id === item.id);
  const currentItemIndex = currentItem ? filteredTreeItems.indexOf(currentItem) : -1;

  return { filteredTreeItems, currentItemIndex };
}
```

RoyEJohnson (commented):
I've made some notes. It's looking good.

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,183 @@
...
+
+  const focusLastTreeItem = (filteredTreeItems: Element[]) => {
+    (document?.querySelector(`[id="${filteredTreeItems[filteredTreeItems.length - 1]?.id}"]`) as HTMLElement)?.focus();
+  };
+
+  const focusOnTab = (filteredTreeItems: Element[], shiftKey: boolean) => {
+    const pageElements = Array.from(
+      document?.querySelectorAll('input, button, select, textarea, a[href]') as ArrayLike<Element>
+    );
+    let index: number;
```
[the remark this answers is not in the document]
jomcarvajal: I don't see it possible. Doing multiple tests and taking in count the active element in that point should be inside the tree structure, we always gonna find only index between 0 and total of interactive elements

playwright/src/fixtures/toc.ts
```diff
@@ -33,7 +33,7 @@ class TOC {
 
   async unitIntroCount() {
     // Total number of unit introduction pages in the book
-    const unitIntroPageLocator = this.page.locator('//li[@data-type="unit"]/details/ol[1]/li[1][@data-type="page"]')
+    const unitIntroPageLocator = this.page.locator('//li[@data-type="unit"]/a/ol[1]/li[1][@data-type="page"]')
```
TomWoodward, line 36: isn't this no longer true? if its true i would be worried about putting lists inside an anchor, but i'm pretty sure you updated that already

src/app/components/Details.tsx
```diff
@@ -50,3 +50,17 @@ export const Details = styled.details`
     }
   `}
 `;
+
+// Other components than ToC use Details, so we need to style them separately
+// tslint:disable-next-line:variable-name
+export const DetailsTree = styled.a`
```
TomWoodward, line 56: you've left these elements named the way they were before but they no longer really make sense with those names. this component no longer acts like a `details` component because the tree that it expands is not inside of it, its now more of a `CollapseToggle` or something like that.

src/app/content/components/ContentLink.tsx
```diff
@@ -95,14 +96,30 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
...
+
+      if (hasUnsavedHighlight && !await showConfirmation(services)) {
+        return;
+      }
+
+      if (onClick) {
+        onClick();
+      }
+
+      navigate(navigationMatch, options);
```
TomWoodward, line 113: onclick has the navigate inside of it, you don't have to call it again

src/app/content/components/ContentLink.tsx
```diff
@@ -95,14 +96,30 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
         navigate(navigationMatch, options);
       }
     }}
+    onKeyDown={(e) => onKeyDown?.(e, async() => {
+      /*
+        All this logic has to be inside onKeyDown === Enter || Space
+      */
+      e.preventDefault();
+
+      if (hasUnsavedHighlight && !await showConfirmation(services)) {
```
TomWoodward, line 105: this is duplicated from the onclick, i would extract a handler used in both callbacks

src/app/content/components/ContentLink.tsx
```diff
@@ -95,14 +96,30 @@ export const ContentLink = (props: React.PropsWithChildren<Props>) => {
         navigate(navigationMatch, options);
       }
     }}
+    onKeyDown={(e) => onKeyDown?.(e, async() => {
+      /*
+        All this logic has to be inside onKeyDown === Enter || Space
```
TomWoodward, line 101: should you be checking the key pressed here? i don't see that filtering happening anywhere

src/app/content/components/TableOfContents/index.tsx
```diff
@@ -150,51 +162,131 @@ function maybeAriaLabel(page: LinkedArchiveTreeSection) {
...
+      event: e,
+      item,
+      isOpen,
+      onSelect: toggleOpen,
+    });
+  };
+
+  return (
+    <Styled.NavItem key={item.id} sectionType={sectionType}>
+      <TocNode
```
TomWoodward, line 202: as i mentioned in the `details` comment, a better name for this at this point might be `TocSectionToggle`

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -147,28 +138,65 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
...
+
+  ::before {
+    display: none;
+  }
+
+  ::-moz-list-bullet {
+    list-style-type: none;
+  }
+
+  ::-webkit-details-marker {
```
TomWoodward, line 192: its not a details element, do you need this?

src/app/content/components/TableOfContents/styled/index.tsx
```diff
@@ -147,28 +138,65 @@ export const NavOl = styled.ol<{section: ArchiveTree}>`
...
+  :hover {
+    outline: none;
+    ${activeState}
+  }
+
+  ::before {
+    display: none;
+  }
+
+  ::-moz-list-bullet {
```
TomWoodward, line 188: pretty sure styled-components will compile in the vendor prefixes for you

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
TomWoodward, line 1: file name doesn't make sense, nothing in this file is a hook

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+export interface KeyboardSupportProps {
+  event: React.KeyboardEvent<HTMLAnchorElement>;
+  item: LinkedArchiveTreeNode;
+  isOpen?: boolean;
+  onSelect: () => void;
+}
+
+const getTreeItems = () => {
+  const treeItems = Array.from(document?.querySelectorAll('a[role="treeitem"]') as ArrayLike<HTMLElement>);
```
TomWoodward, line 13: use assertDocument() so that it throws if the logic is accidentally included in the nodejs render. you can pass the type to the `querySelectorAll` generic to avoid the cast

TomWoodward, line 13: this will behave strangely if there is more than one tree on the page, you should probably select relative to the currently focused link

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const treeItemResults = treeItems.filter((el) => {
+    const style = window?.getComputedStyle(el);
+    return (style?.display !== 'none');
```
TomWoodward, line 16: is there something you can add to the css selector like `[open]` to do this filter?

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+    treeItem.textContent?.trim().toLowerCase().startsWith(key.toLowerCase())
+  );
+  if (result) document?.querySelector<HTMLElement>(`[id="${result?.id}"]`)?.focus();
+};
+
+function getCurrentItemIndex(item: KeyboardSupportProps['item']) {
+  const filteredTreeItems = getTreeItems();
+  const currentItem = filteredTreeItems.find((treeitem) => treeitem.id === item.id);
+  /* istanbul ignore next */
+  const currentItemIndex = currentItem ? filteredTreeItems.indexOf(currentItem) : -1;
```
TomWoodward, line 78: could you just use findIndex here?

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+  return { filteredTreeItems, currentItemIndex };
+}
+
+export const onKeyDownNavGroupSupport = ({
```
TomWoodward, line 83:
```suggestion
export const treeNavSubtreeOnKeyDown = ({
```
might make it more clear what this function is for

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+    case 'Tab':
+      focusOnTab(filteredTreeItems, event.shiftKey);
+      break;
+    default:
+      searchTreeItem(filteredTreeItems, currentItemIndex, event.key);
+      break;
+  }
+};
+
+export const onKeyDownNavItemSupport = ({
```
TomWoodward, line 133:
```suggestion
export const treeNavItemOnKeyDown = ({
```

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
```diff
@@ -0,0 +1,171 @@
...
+  document?.querySelector<HTMLElement>(`[id="${filteredTreeItems[filteredTreeItems.length - 1]?.id}"]`)?.focus();
+};
+
+const focusOnTab = (filteredTreeItems: Element[], shiftKey: boolean) => {
+  const pageElements = Array.from(
+    document?.querySelectorAll('input, button, select, textarea, a[href]') as ArrayLike<HTMLElement>
+  );
+  let index = pageElements.indexOf(document?.activeElement as HTMLElement);
+  if (shiftKey) {
+    focusFirstTreeItem(filteredTreeItems);
```
TomWoodward, line 52: why are you changing the treeitem focus and then focusing something else right after?

TomWoodward (changes requested):

src/app/content/components/ContentLink.tsx
jomcarvajal: Removing `navigate(navigationMatch, options)` produce that page doesn't update the content. It means maybe this navigation is important. The onClick() call doesn't bring an invoke of navigation.

src/app/content/components/ContentLink.tsx
jomcarvajal: Yeah I was sure that will work but that was not the case. I have to preventDefault behavior of keys because Tree navigation is very special. Removing preventDefault and 'enter' as one of my cases of keyboardSupport  just do nothing. I'll explore more about it

src/app/content/components/TableOfContents/keyboardSupport.hook.tsx
TomWoodward: You need to have different css selectors for the content links and the toggles, then you can do `a.content-link,a.toggle[open]` (or whatever the right selector would be

TomWoodward: Oh nevermind, I'm misinterpreting what you're trying to do. I think I see now you're trying to filter out the ones that are hidden.... I'll think about it some more

jomcarvajal:
> I went back and reviewed the [w3c page](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) on the tree pattern, and in particular the markup structure and event handlers don't seem to be lining up with what I see in the pr. Notably I didn't think our toggles are supposed to be links because they didn't go anywhere. It also seems like the JavaScript example they have there might solve some of the issues I've been noting in your event handlers.
> 
> We might consider using the [react aria](https://react-spectrum.adobe.com/react-aria/Tree.html) library that we use in other places, they have a tree view component and I think their library is generally pretty extensible in terms of display styling

Hmmm I see... it is weird because when I was reviewing w3c page, the structure of using <a /> as toggles is right there and that is why I kept using it. Probably I'll start a new PR from zero then.

jomcarvajal:
@TomWoodward It is true that we are using links that don't go anywhere, but this was done this way following the W3C example, which clearly shows that role=treeitem elements should be links in this pattern.

When you mention that "the markup structure and event handlers don't seem to be lining up with what I see in the PR," could you clarify what you mean by that? I'm asking because the context of TableOfContents differs from the example provided. We have a dynamic tree in a context that involves handling a specific structure for reading data, such as book, page, section, among others.

Regarding your comment about using react-aria-components, I get the impression that you are suggesting I should remove my code entirely and use the library instead. Is that correct? Also, I understand that this change would come from ui-components, which implies a more general implementation.

On another note, when using the Tree from react-aria-components, I’ve noticed that its keyboard interaction implementation is somewhat limited. Additionally, the way styles are managed is different from what we currently have, which could lead to a greater effort to avoid breaking any existing functionality.

Given that there are many doubts and concerns, and since I don’t fully understand your vision on this, I would prefer if we could meet and review the code together. This way, we can clear up any misunderstandings.

TomWoodward:
@jomcarvajal 

> It is true that we are using links that don't go anywhere, but this was done this way following the W3C example, which clearly shows that role=treeitem elements should be links in this pattern.

look again, those links go places, in the example they provide there are index pages for each subtree that we don't have. in the react-aria implementation it looks like they actually don't have role=treeitem on the folder nodes that can't be selected

> When you mention that "the markup structure and event handlers don't seem to be lining up with what I see in the PR," could you clarify what you mean by that? I'm asking because the context of TableOfContents differs from the example provided. We have a dynamic tree in a context that involves handling a specific structure for reading data, such as book, page, section, among others.

they have example javascript for handling keyboard events which looks simpler than yours, handles fewer events, overrides less default functionality (the tab key for example) 

> Regarding your comment about using react-aria-components, I get the impression that you are suggesting I should remove my code entirely and use the library instead. Is that correct? Also, I understand that this change would come from ui-components, which implies a more general implementation.

my suggestion to use react-aria-components is because we seem to be struggling to get this right, and that library is designed to solve this problem for us. i'm not sure if the treeview component that i've been referencing is a perfect fit, maybe they have something else, but it seemed like it might be worth looking into. i like using react-aria-components because the whole point of it is to follow the w3c standards by default so we don't have to worry about it. not saying we have to go that route, but i think its worth considering.

i'm not suggesting that you do it through the ui-components library though (unless that makes sense for some other reason), you could also just implement the react-aria dependency right in rex.

PR #2447, opened 2025-03-11
Set overflow-wrap so long URLs break as needed

RoyEJohnson:
[CORE-805]

[CORE-805]: https://openstax.atlassian.net/browse/CORE-805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/content.browserspec.ts
+++ b/src/app/content/content.browserspec.ts
@@ -68,15 +68,12 @@ describe('content', () => {
 
     // scroll down and make sure it worked
     await scrollDown();
-    await scrollTocDown(20);
     expect(await getScrollTop()).not.toBe(0);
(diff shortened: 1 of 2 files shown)


Dantemss (approved):
You might need to fix or remove the scroll tests. And no idea what needs to be manually tested here. But the change looks good.

RoyEJohnson:
The test (as per the card) was to open the Citation/Attribution section at the bottom of a page. It contains some long links, which were not being wrapped, so on narrow screens (phones), they would become horizontally-scrollable. This made them wrap.
Before (staging) and after (local)
image
image

PR #2451, opened 2025-03-26
Set unique aria-label for search result link

RoyEJohnson:
[CORE-799]

### DIFF
--- a/src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
+++ b/src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
@@ -1,4 +1,5 @@
 import { SearchResultHit } from '@openstax/open-search-client';
+import { HTMLDivElement } from '@openstax/types/lib.dom';
 import isEqual from 'lodash/fp/isEqual';
 import { OutputParams } from 'query-string';
 import React from 'react';
@@ -55,8 +56,7 @@ function useKeyTermPair({
   return pair;
 }
 
-// tslint:disable-next-line: variable-name
-const OneSearchResultHit = ({
+function HighlightPreview({
   activeSectionRef,
   book,
   hit,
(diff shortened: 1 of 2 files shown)


src/app/content/search/components/SearchResultsSidebar/__snapshots__/index.spec.tsx.snap
```diff
@@ -870,6 +870,7 @@ exports[`SearchResultsSidebar matches snapshot with related key terms 1`] = `
       </h3>
       <a
         aria-current={true}
+        ariaLabel="undefined in test-page-1"
```
TomWoodward, line 873: i think you have to put your ariaLabel in state for it to not show this on the first render, because the ref will be empty until it renders the first time

src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
```diff
@@ -65,47 +65,65 @@ const OneSearchResultHit = ({
...
+  const isSelected = isEqual(selectedResult, thisResult);
+  const target: SearchScrollTarget = {
+    elementId: thisResult.result.source.elementId,
+    index,
+    type: 'search',
+  };
+  const divRef = React.useRef<HTMLDivElement>(null);
+  const excerptText = divRef.current?.textContent;
+  const page = getPage(hit);
+  const ariaLabel = `${excerptText} in ${page.slug}`;
```
TomWoodward, line 84: do we also need to do like `result number ${index+1} in ${page.slug}: ${excerptText}` or do you think this is unique enough?

TomWoodward (changes requested):

src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
RoyEJohnson: We'll do that, then we know.

src/app/content/search/components/SearchResultsSidebar/__snapshots__/index.spec.tsx.snap
RoyEJohnson: Right. Have to do the set in a useEffect.

TomWoodward (approved):

Dantemss (approved):

PR #2454, opened 2025-03-31
add padding bottom to avoid overlap with banners

jomcarvajal:
[CORE-802]

The issue more than resize text is related with overlap text in all resolutions. Instead of only focus on small resolutions, I added padding according resolutions. The padding is set by the number of toast messages in queu and a constant value taken from an aprox height of those toastBanners

https://github.com/user-attachments/assets/cbe9c380-0079-4ea8-a22a-c9b307e662c6

[CORE-802]: https://openstax.atlassian.net/browse/CORE-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/content/components/Page/PageToasts.tsx
```diff
@@ -63,7 +78,12 @@ const PageToasts = (props: ToastProps | {}) => {
   const mobileToolbarOpen = useSelector(mobileToolbarOpenSelector);
 
   return (
-    <ToastContainerWrapper role='alertdialog' {...props} mobileToolbarOpen={mobileToolbarOpen}>
+    <ToastContainerWrapper
+      role='alertdialog'
```
RoyEJohnson, line 82: I think that with the role "alertdialog" it does not matter that the content overlaps.
It was previously "alert". The difference is that "alertdialog" is something that is supposed to be dismissed -- like these. That role change was [a change I made](https://github.com/openstax/rex-web/pull/2448) recently.

jomcarvajal: We should cancel the card CORE-802 ?

RoyEJohnson: I think it should be reviewed along with CORE-804 (the card my PR was attached to). I think they're effectively the same.
I don't think it's _wrong_ to make room for the boxes, but I think if we want to do that, we might just want to have them appear in a reserved space rather than absolute-positioned.

PR #2457, opened 2025-04-03
remove broken redirect from polish marketing

bethshook:
https://openstax.atlassian.net/browse/CORE-865


Dantemss (approved):
OK I see the issue... in the new book only `1-wprowadzenie` exists as a page (all 3 exist as unit slugs, which maybe is fine). I guess even in the old book there were 2 pages with that name...

Dantemss:
Merging this to get these builds started

PR #2458, opened 2025-04-07
adjust aria-live for error toasts

bethshook:
https://openstax.atlassian.net/browse/CORE-746


bethshook:
@RoyEJohnson since you are tagged on this ancient pr I wonder if you could take a look at my tests and see if you have any idea for getting them to pass now that I've implemented this timeout for displaying error modals. I've tried a couple of approaches including `useFakeTimers()` but no dice. current approach is pretending the timeout isn't there but even that is resulting in a failed spec.

RoyEJohnson:
For Page.spec, could you mock out the toast stuff entirely?

RoyEJohnson (approved):
🎉

PR #2459, opened 2025-04-14
Remove `TextResizerMenu` from tab order

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-798

Remove the menu box from the focus order.

Relevant labels are:
"i18n:toolbar:textresizer:button:decrease:aria-label": "Decrease text size"
"i18n:toolbar:textresizer:button:increase:aria-label": "Increase text size"
"i18n:toolbar:textresizer:popup:heading": "Text size"

I think the labels for the buttons are clear enough. The range input, however, uses `i18n:toolbar:textresizer:popup:heading`, or "Text size", which could be too generic or not descriptive enough.

Example:

https://github.com/user-attachments/assets/5668f88e-0144-40f8-905c-b150eb9c38d2


src/app/content/components/Topbar/TextResizer.tsx
```diff
@@ -57,7 +57,7 @@ export const TextResizer = (props: TextResizerProps) => {
       dataAnalyticsLabel='Change text size'
       {...props}
     >
-      <Styled.TextResizerMenu tabIndex={0} bookTheme={props.bookTheme} textSize={props.textSize}>
+      <Styled.TextResizerMenu tabIndex={-1} bookTheme={props.bookTheme} textSize={props.textSize}>
```
RoyEJohnson, line 60: Does it need a `tabIndex` setting at all? It's a div, which normally will not receive focus. Does this div receive focus programmatically?

TylerZeroMaster: You're right. For some reason I thought 0 was the default, but that would mean that all non-interactive elements would be in the tab list, so I am not sure what I was thinking.

RoyEJohnson (approved):

PR #2470, opened 2025-05-05
Image preview modal for mobile breakpoint 

P-Gill97:
https://openstax.atlassian.net/browse/CORE-652


src/app/content/components/Page/MediaModalManager.tsx
TomWoodward, line 1: this whole MediaModalManger setup is pretty convoluted, can you look at the highlightManager and try to share some of the patterns?

don't need the singleton class. stylistically, would be more consistent if it was a function and not a class. PageComponent can construct the manager and store the manager reference

PageComponent should orchestrate managers but there shouldn't be any business logic in there, move all the event listeners into the manager, you can bind your listener to the page component element instead of each individual image so that you don't have to re-apply the listeners when the content changes

src/app/content/components/Page/PageComponent.tsx
```diff
@@ -149,6 +147,7 @@ export default class PageComponent extends Component<PagePropTypes> {
     return <MinPageHeight>
       <this.highlightManager.CardList />
       <PT />
+      <MediaModalPortal />
```
TomWoodward, line 150: if you look above, the CardList is coming from the manager so that you don't need all the wonky binding logic. if you prefer, you could leave the component separate and pass the manager into it as a prop here, but either of those approaches would be more clear than the singleton binding its doing now

TomWoodward (changes requested):

src/app/content/components/Page/contentDOMTransformations.ts
```diff
@@ -249,3 +251,11 @@ function moveFootnotes(document: Document, rootEl: HTMLElement, intl: IntlShape)
     sup.appendChild(link);
   }
 }
+
+function enhanceImagesForAccessibility(rootEl: HTMLElement) {
+  rootEl.querySelectorAll('img').forEach((img) => {
+    img.setAttribute('tabindex', '0');
+    img.setAttribute('role', 'button');
+    img.setAttribute('aria-label', 'Open media preview');
```
RoyEJohnson, line 259: This will potentially have a bunch of buttons with the same label. If the images reliably have alt text, you can work that into the label (the label will be announced _instead of_ the alt text).

src/app/content/components/Page/MediaModal.spec.tsx
```diff
@@ -0,0 +1,67 @@
+import renderer, { act } from 'react-test-renderer';
+import MediaModal from './MediaModal';
+import React from 'react';
+
+describe('MediaModal', () => {
+  const childContent = <div>Test Content</div>;
```
RoyEJohnson, line 6: You could define the entire Component to be rendered so that each test just calls
`renderer.create(<Component />)` or `renderer.create(<Component isOpen={false} />)`

src/app/content/components/Page/MediaModal.tsx
```diff
@@ -0,0 +1,95 @@
...
+// tslint:disable-next-line:variable-name
+const ModalWrapper = styled.div`
+  position: fixed;
+  inset: 0;
+  overflow: hidden;
+  z-index: ${theme.zIndex.highlightSummaryPopup + 1};
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  pointer-events: none;
```
RoyEJohnson, line 54: Why `pointer-events: none`?

src/app/content/components/Page/contentDOMTransformations.ts
P-Gill97: Yeah, you're right. It would be confusing. 

I think something like this would be better. 
```
const alt = img.getAttribute('alt');
const label = alt ? `Open preview of ${alt}` : 'Open media preview';
img.setAttribute('aria-label', label);

```

src/app/content/components/Page/mediaModalManager.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { ReactNode, useEffect } from 'react';
+import { createPortal } from 'react-dom';
+import MediaModal from './MediaModal';
+import { HTMLElement, MouseEvent, KeyboardEvent, HTMLImageElement } from '@openstax/types/lib.dom';
+
+
+export function createMediaModalManager(container: HTMLElement | null) {
```
RoyEJohnson, line 7: This would be more readable if you could extract out the definitions for subunits like MediaModalPortal and create them with a single call (e.g., MediaModalPortal = createMediaModalPortal(mount)).
`handleMediaInteraction` and the listener functions could be defined in a hook.
The keydown functionality could be another hook.

TomWoodward (changes requested):
most of these are nitpicks but i'm really kinda worried that we're doing something bad by marking an image as a button, can we do some more research on acceptable approaches here or maybe ask level access about it? 

i'm a little concerned that we're supposed to be adding a transparent button overlay over the image that would appear as a separate button/tab-order element to a screen reader or keyboard user, but that would be annoying because to do it cleanly we'd probably need an image wrapper element

RoyEJohnson:
I think maybe the accessible option is a small, visible expand-icon button overlayed on the corner of the image.

src/app/content/components/Page/mediaModalManager.tsx
```diff
@@ -0,0 +1,94 @@
...
+      return () => { setModalContent = null; };
+    }, []);
+    if (typeof document === 'undefined') return null;
+
+    return isOpen ? createPortal(
+      <MediaModal isOpen={isOpen} onClose={() => setIsOpen(false)}>
+        {modalContent}
+      </MediaModal>,
+      document.body
+    ) : null;
```
TomWoodward, line 33: the modal can exist and not be open. you can simplify this to 
```suggestion
    setModalContent = (content) => {
      setContent(content);
      setIsOpen(true);
    };

    if (typeof document === 'undefined') return null;

    return createPortal(
      <MediaModal isOpen={isOpen} onClose={() => setIsOpen(false)}>
        {modalContent}
      </MediaModal>,
      document.body
    );
```

and it'll still work i think

src/app/content/components/Page/mediaModalManager.tsx
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  const mount = (newContainer: HTMLElement) => {
+    // detach from previous if different container
+    if (container && container !== newContainer) {
+      detachListeners();
+    }
+
+    container = newContainer;
+    attachListeners();
```
TomWoodward, line 80: but if the container is the same container (which i think will happen when navigating between pages), you're attaching the listeners again each time mount is called right? aren't you getting duplicate listeners?

TomWoodward (changes requested):

src/app/content/components/Page/mediaModalManager.tsx
P-Gill97: The listeners won't be duplicated according to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener). The browser dedupes listeners by type and listener reference. If the container is the same container which means it will have the same reference. It should not duplicate the listeners. So I believe this is safe. 
```
  const mount = (newContainer: HTMLElement) => {
    if (container !== newContainer) {
      detach();
      container = newContainer;
    }
    attach();
  };
```

RoyEJohnson (commented):
My concerns are addressed; I'll leave the approval for Tom.

TomWoodward (approved):

src/app/content/components/Page/mediaModalManager.tsx
P-Gill97: That's strange, we are both looking at this file right? (line 79) Maybe it's a cache issue?
 
https://github.com/openstax/rex-web/blob/image-preview/src/app/content/components/Page/mediaModalManager.tsx

TomWoodward (approved):
my github ui seemed to be messed up and i was seeing the old code, sorry about that, not sure what happened

TomWoodward:
i have an interest in merging this in case we want to use it for the math too @P-Gill97

P-Gill97:
Ok i'll merge it.

PR #2472, opened 2025-05-09
Core 829 add login gate to books with content warning

RoyEJohnson:
[CORE-829]
The card says don't work on this before summer.
The change is pretty simple. The test is a mess.
I wound up splitting test cases into separate files because it refused to re-render.

[CORE-829]: https://openstax.atlassian.net/browse/CORE-829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/NavBar/index.tsx
+++ b/src/app/components/NavBar/index.tsx
@@ -151,6 +151,13 @@ const LoggedOutState: FunctionComponent<{currentPath: string}> = ({currentPath})
   </Styled.Link>}
 </FormattedMessage>;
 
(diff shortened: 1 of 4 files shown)


Dantemss (approved):
Seems simple enough but check with Tom and/or Erika if the reason the card said to wait was because they wanted to wait to release. Also uh I got things to try to finish before I travel so I'll trust you to check that the change works in Heroku.

PR #2473, opened 2025-05-09
CORE-854: Replace blockquote tag around attribution text

RoyEJohnson:
[CORE-854]
I replaced it with paragraph tags.

[CORE-854]: https://openstax.atlassian.net/browse/CORE-854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dantemss (commented):
Won't this change the actual styles? Maybe we should use aria-role instead?

Dantemss (approved):
Oh I see, it seems the blockquote tag was not even needed here. The card just says to remove but if you think it works better with `<p>`, I defer to you.

PR #2475, opened 2025-05-14
CORE-902 content portal routing

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-902

Note: The cookieyes button does not seem to appear when running locally (I suspect it's added by javascript that doesn't run locally, but I am not 100% sure).

I needed to change some stuff about content routing in order for the portal to stay alive through content links. Maybe there's a better way, but I added `portalName` to `Params` and have it propagate through `ContentLink`. Here's a more detailed explanation:

1. `ContentLink` creates a url using `ContentRoute` `getUrl` inside `getBookPageUrlAndParams` [here](https://github.com/openstax/rex-web/blob/9294afd1a0f92e6ec3e0841a18f6acf7c2b143e1/src/app/content/components/ContentLink.tsx#L65).
2. The above url returns a non-portal url if the portal name is not included in `Params` (because `getUrl` returns urls based on matching params).
3. Content link href is transformed to a relative path [here](https://github.com/openstax/rex-web/blob/9294afd1a0f92e6ec3e0841a18f6acf7c2b143e1/src/app/content/components/ContentLink.tsx#L67).
4. The above function uses common parts of the url to perform the transformation. If `getBookPageUrlAndParams` returns a url that does not have a portal, then the relative url that is ultimately used in the content link will not have a portal either.

I did not make this change at a deeper level because having portal name in params elsewhere would interfere with things like fetching/receiving content.

Aside: I am not a huge fan of using regex for the navigation selector. I wondered if there was a way to use the content routes to extract the portal name instead, or if `portalName` should be added to the redux state. It seemed pretty closely linked to the pathname, which is why I decided to do it this way. Any suggestions would be welcome.

## Desktop
Screenshot 2025-05-14 at 10 29 45 AM

## Mobile
Screenshot 2025-05-14 at 10 29 59 AM

## Contact Dialog
Screenshot 2025-05-14 at 10 30 18 AM

## Content Links
Screenshot 2025-05-14 at 10 32 35 AM


RoyEJohnson (approved):
I appreciate the description of what you were doing.
I left some comments for your consideration, but it looks fine to me.

src/app/navigation/selectors.ts
```diff
@@ -14,6 +14,11 @@ export const pathname = createSelector(
   (state) => state.pathname
 );
 
+export const portalName = createSelector(
+  pathname,
+  (state) => state.match(/\/portal\/([^/]+)\//)?.[1]
```
TomWoodward, line 19: `navigation.match.params.portalName` will have the portal name if there is one

TomWoodward (changes requested):
looks good to me

TomWoodward (changes requested):

src/app/components/Footer/index.tsx
```diff
@@ -293,7 +303,7 @@ const PortalColumn2 = () => {
         <FooterLinkMessage href='/tos' id='i18n:footer:column3:terms' />
         <FooterLinkMessage href='/privacy-policy' id='i18n:footer:column3:privacy-policy' />
       </LinkList>
-      <ContactDialog contactFormParams={contactFormParams} />
+      <Dialog contactFormParams={contactFormParams} />
```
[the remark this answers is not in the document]
TylerZeroMaster: Okay, but this pattern is used all over the place in os-webview. I literally copied this hook from os-webview. I didn't see this comment before. I will make this change and update the tests.

src/app/components/Footer/index.tsx
```diff
@@ -209,4 +213,148 @@ const Footer = ({
...
+}
+
+export function useContactDialog() {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const open = React.useCallback(() => setIsOpen(true), [setIsOpen]);
+  const close = React.useCallback(() => setIsOpen(false), [setIsOpen]);
+
+  React.useEffect(
+    () => {
+      if (typeof window === 'undefined' || !window.parent || window.parent === window) {
```
TomWoodward, line 263: why are you checking `!window.parent || window.parent === window` here? this looks like a check to see if the rex page is embedded, thats not what we want right?

src/app/content/components/ContentLink.tsx
```diff
@@ -31,6 +31,7 @@ interface Props {
   handleClick?: () => void; // this one gets called instead of navigation
   navigate: typeof push;
   currentPath: string;
+  navigationParams: unknown;
```
TomWoodward, line 34: why unknown?

src/app/guards.ts
```diff
@@ -47,3 +47,7 @@ export const isHtmlElementWithHighlight = (thing: unknown): thing is dom.HTMLEle
 
 export const isKeyOf = <O extends {[key: string]: unknown}>(obj: O, thing: unknown): thing is keyof O =>
   typeof thing === 'string' && thing in obj;
+
+export const isPortaled = (thing: unknown): thing is {portalName: string} => (
+  typeof thing === 'object' && thing != null && 'portalName' in thing
```
TomWoodward, line 52: technically this'll still match `{portalName: undefined}` or `{portalName: ''}`, you might want to throw a `&& thing.portalName` on the end. but also wouldn't it be just as easy to check `params?.portalName` inline?

TomWoodward (changes requested):

src/app/components/Footer/index.tsx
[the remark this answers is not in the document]
RoyEJohnson: That's true; I was learning React as I ported osweb to use it, so there are some bad patterns in the codebase.

TomWoodward (approved):

PR #2481, opened 2025-05-28
CORE-799: Fix aria-label attribute

RoyEJohnson:
[CORE-799]


RoyEJohnson:
I noticed that there was potentially HTML in both the chapter title and the highlight text, so both are processed to be text now.
image

Dantemss (approved):
Interesting hack to extract the text from html tags

PR #2482, opened 2025-05-28
CORE-1000 - Fix optimized images

TylerZeroMaster:
https://openstax.atlassian.net/browse/CORE-1000

Summary: `optimizeImages` transforms `src` to a different url; `makeResourcesLazy` removes `src`. `optimizeImages` never found the `src` since `makeResourcesLazy` ran first. Reordering these calls fixes the problem. 

I added an image to the short page so that the updated `src` would be included in snapshot testing. It has a width and a height because that is required for the lazy loading manager to find the image. I also changed the query selector slightly. It was missing a closing bracket. That was in no way related to the problem, but it did add another layer to check, so it seemed worth fixing.


Dantemss (approved):

TylerZeroMaster:
@Dantemss I may have had that happen once as well. I have not been able to recreate it, but I recall seeing one broken image. I didn't look at the attributes, so I am not 100% sure it still had `data-lazy-src`, but your problem sounds the same. A refresh fixed it. 

My guess, based on `data-lazy-src` remaining, would be that it could be related to the way intersections are detected. This uses `getBoundingClientRect` and compares that to window dimensions on each scroll event. I don't see an immediate problem, but I wonder if switching to [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) or [the loading attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#lazy), supported by [most browsers](https://caniuse.com/loading-lazy-attr), would be worth exploring.

I think this would be a problem on production already since it uses the same lazy loading method.

Dantemss (approved):
Seems to work, the images load as I scroll

PR #2483, opened 2025-06-02
remove logic that show all cards when blur or navigate with keyboard

jomcarvajal:
[CORE-806]

Remove SHOW_ALL option for highlights when onBlur. Cards should stay away when press Esc and it can return pressing Enter.

[CORE-806]: https://openstax.atlassian.net/browse/CORE-806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Dantemss (approved):
Looks good to me. Send it to level access to re-review I guess.

PR #2486, opened 2025-06-12
CORE-1026: Quoted terms are matched exactly in search

RoyEJohnson:
[CORE-1026]

[CORE-1026]: https://openstax.atlassian.net/browse/CORE-1026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/components/TableOfContents/index.tsx
+++ b/src/app/content/components/TableOfContents/index.tsx
@@ -265,13 +265,12 @@ function TocSection({
         const active = page && stripIdVersion(item.id) === page.id;
 
         return (
-          <>
+          <React.Fragment key={item.id}>
             {isArchiveTree(item)
               ?
               <Styled.StyledTreeItem
                 section={section}
                 id={item.id}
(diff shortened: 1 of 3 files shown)


TomWoodward (changes requested):
@RoyEJohnson i'm not totally understanding the changes you made here, but i had understood that this issue would be a backend search change to remove the fuzzy matching logic for quoted portions of the search, i don't think that it can be done via the UI, maybe i'm misunderstanding or we need to check with dante though

Dantemss:
I suppose what this is doing is it does the fuzzy search and then filters the results in the frontend, which is inefficient but works and is an easy solution. I'd be inclined to approve (if it tests fine), we can always improve it later.

RoyEJohnson:
Dante's description is accurate.

TomWoodward (approved):

PR #2496, opened 2025-07-09
selector change in highlight test

omehes:
Tiny one line change - had to replace a selector for a page object

Note. The one highlight test will fail until the new behaviour is fixed.

### DIFF
--- a/playwright/e2e/ui/pages/home.py
+++ b/playwright/e2e/ui/pages/home.py
@@ -533,7 +533,7 @@ class HomeRex:
 
     @pytest.mark.asyncio
     async def close_unsuccessful_search_result_sidebar(self):
-        await self.page.locator("div.styled__SearchResultsBar-as3fh1-4.gldLiQ").get_by_role("button").click()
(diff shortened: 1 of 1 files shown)


Dantemss (changes requested):
It's not good to use these auto-generated classes from styled-components, as they can change anytime for whatever reason. Try to see if there are any other classes around the element that you can use, like a class on the parent. If not, maybe we can add one.

Dantemss:
I mean... you can add the class :) search for CloseIconWrapper and add a data-testid or className (there's an example elsewhere in Rex that already has a data-testid, make sure you found the one that goes to the search results sidebar).

Or try something like `div[data-testid="search-results-sidebar"] button`, if the button will accept the click as well (and if it's the only button in the search results sidebar).

omehes:
@Dantemss here is the pr with the updated selector which is related to https://github.com/openstax/rex-web/pull/2534, where I added test id to the element in question (as per our discussion)

jivey (approved):

Dantemss (approved):

PR #2540, opened 2025-07-24
add test to clean book page of possible highlights

omehes:


omehes:
closed to investigate some more aditions...

PR #2541, opened 2025-07-28
adjusted highlight selectors and tests

omehes:

added testid to highlight box elements
simplified selectors in highlights tests


RoyEJohnson (approved):
You will need to run
`yarn test:unit -u`
to update the snapshot that is causing the test failure. Other than that, good.

PR #2558, opened 2025-08-13
Core 20 highlight edit navigation

RoyEJohnson:
[CORE-20]
[CORE-459]
The latter one describes using tab to get to the dialog, but we took a different approach (we want to be able to tab through the highlights without the detours, and use a different key to go into the editor).

[CORE-20]: https://openstax.atlassian.net/browse/CORE-20?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-459]: https://openstax.atlassian.net/browse/CORE-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (commented):
Changes seem good, but after using it, I'm worried the hidden highlight popup won't be discoverable enough. But we can get some feedback using the deployed env.

RoyEJohnson:
@jivey I have modified the highlight editor to show instructions (only) when a highlight is focused, to make the editor more discoverable. The tests are not fixed yet, but you can see whether you think it's better.

jivey (approved):
I see the help popup when unauthenticated, it seems a bit narrow though:
image

Otherwise I think this is good, and the keyboard interactions make a lot of sense.

PR #2559, opened 2025-08-15
rex qa tests selector improvements - 2

omehes:


jivey (changes requested):

omehes:
@jivey I added some test-ids. Some playwright tests fail because I changed the selector and I think we have to deploy to staging for it to take effect...

jivey (approved):

PR #2565, opened 2025-08-25
added highlight deletion for algebra and trig book

omehes:


TomWoodward (approved):
note that you're only finding highlights in paragraphs and you're not asserting anything (like if there are no highlights to begin with most of the test won't run, not sure if you're actually trying to test anything here or just clean up the page)

code looks fine though

PR #2585, opened 2025-10-27
CORE-829 Add login gate

RoyEJohnson:
[CORE-829]
I implemented this so that the login message replaces the page content, rather than as a dialog.

[CORE-829]: https://openstax.atlassian.net/browse/CORE-829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/NavBar/index.tsx
+++ b/src/app/components/NavBar/index.tsx
@@ -151,6 +151,13 @@ const LoggedOutState: FunctionComponent<{currentPath: string}> = ({currentPath})
   </Styled.Link>}
 </FormattedMessage>;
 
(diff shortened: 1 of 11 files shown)


jivey (commented):
Code changes seem good, I am wondering about the general design. The login button looks a little strange because of the block style:

image

I'm also wondering if we are okay with search still working... granted the results don't show very much, but making sure that's not an issue.

jivey (approved):
I didn't check out what the build failure is but everything else looks good.

PR #2605, opened 2025-11-11
CORE-1278: search term glitch error 

jomcarvajal:
[CORE-1278]

[CORE-1278]: https://openstax.atlassian.net/browse/CORE-1278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/helpers/rangy.ts
```diff
@@ -11,17 +11,53 @@ if (!rangy.initialized) {
...
+      range2.commonAncestorContainer;
+    while (containerA) {
+      let current: typeof containerB | null = containerB;
+      while (current) {
+        if (containerA === current) return true;
+        current = current.parentNode;
+      }
+      containerA = containerA.parentNode;
+    }
+    return false;
```
Dantemss: How would they not share an ancestor? Would they need to be in different DOMs, or detached? Otherwise the body etc should be a common ancestor right?

When I tried the search in my browser, I only saw the error in `findText`, which is already handled by the try/catch you added.

Dantemss (approved):

PR #2614, opened 2025-11-26
CORE-1159: REX - Highlighted text and box issues inside Show/Hide Solution 

jomcarvajal:
[CORE-1159]

[CORE-1159]: https://openstax.atlassian.net/browse/CORE-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/highlights/components/CardWrapper.tsx
+++ b/src/app/content/highlights/components/CardWrapper.tsx
@@ -253,8 +253,30 @@ const Wrapper = ({highlights, className, container, highlighter, dispatch}: Wrap
   </div>;
 };
 
+function extractTextFromHTML(html: string): string {
+  const tmp = document?.createElement('div');
+  if (tmp) {
+    tmp.innerHTML = html;
+    return tmp.textContent || '';
+  }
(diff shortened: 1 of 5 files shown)


src/app/content/highlights/components/CardWrapper.tsx
```diff
@@ -264,8 +264,30 @@ const Wrapper = ({highlights, className, container, highlighter, dispatch}: Wrap
   </div>;
 };
 
+function extractTextFromHTML(html: string): string {
```
RoyEJohnson: We already have a [stripHTML utility](https://github.com/openstax/rex-web/blob/main/src/app/utils.ts#L155) that should work for this.

RoyEJohnson (changes requested):
I like the tests.
One comment on a redundant function, and one question.

RoyEJohnson (approved):

PR #2619, opened 2025-12-03
CORE-1217: (FIX) handleGlobalMouseUp event

jomcarvajal:
[CORE-1217]

Previous logic was causing issues when focusing highlights

[CORE-1217]: https://openstax.atlassian.net/browse/CORE-1217?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/highlights/components/CardWrapper.spec.tsx
+++ b/src/app/content/highlights/components/CardWrapper.spec.tsx
@@ -728,4 +728,95 @@ describe('CardWrapper', () => {
       expect(component.toJSON()).not.toBeNull();
     });
   });
+
+  describe('handleGlobalMouseUp', () => {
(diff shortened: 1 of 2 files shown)


jivey (approved):
This resolves the snapping issue I was seeing on main, so good to get merged, but I don't believe either PR fixed the original concern. Locally and in the deployed env, I still don't see the highlight popup when releasing outside of the container until after I get it to show up once (by releasing inside the container):

https://github.com/user-attachments/assets/d97fa56a-ad84-48cd-a188-e74ca984bbe6

PR #2626, opened 2025-12-08
Core-1399: Keep highlight editor open when highlight is created

RoyEJohnson:
[CORE-1399]
[CORE-1408]

[CORE-1399]: https://openstax.atlassian.net/browse/CORE-1399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CORE-1408]: https://openstax.atlassian.net/browse/CORE-1408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/highlights/components/Note.tsx
+++ b/src/app/content/highlights/components/Note.tsx
@@ -4,6 +4,7 @@ import { useIntl } from 'react-intl';
 import styled from 'styled-components/macro';
 import { textStyle } from '../../../components/Typography/base';
 import theme from '../../../theme';
+import { KeyboardEvent } from '@openstax/types/lib.dom';
 import { cardPadding, cardWidth } from '../constants';
 
 interface Props {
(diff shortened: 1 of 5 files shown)


src/app/content/highlights/components/Note.tsx
```diff
@@ -47,6 +48,12 @@ const SimpleLabel = styled.label`
   padding-left: 0.2rem;
 `;
 
+// exported for test coverage reasons
+export function escapeHandler(onElement: HTMLTextAreaElement | null, shouldDo: boolean) {
+  if (shouldDo) {
```
Dantemss: it's unclear to me why this boolean is passed into the function, the check could just live outside, in escCb

Dantemss (approved):
Dismissing with esc works in Heroku but the blank highlight stays (it stays highlighted). Is this intended?

src/app/content/highlights/components/Note.tsx
RoyEJohnson: The boolean was the reason I needed to separate/export the function to get test coverage. The condition is always false in testing due to how refs are handled.

PR #2628, opened 2025-12-11
Fix book scan errors and flaky tests

Dantemss:
https://openstax.atlassian.net/browse/CORE-347

- Updated puppeteer (to the same version that lighthouse expects/uses - lighthouse cannot be updated due to the node version)
- Fixed book scan failure due to changes in the ToC structure.
- Fixed book scan failure due to unicode characters in book slugs (Spanish Physics).
- Retry network failures in book scan one time.
- Close and reopen book scan tab every 30 pages to prevent browser crashes.
- Fixed flaky test due to img with empty src attribute.
- Fixed some flaky tests missing act() calls.
- Allow some slack in scroll tests (OS X and GH actions seem to get consistently different results).


RoyEJohnson (approved):
I think some of the test stuff you had to clean up was from me, so sorry about that!
This looks good to me.

PR #2630, opened 2025-12-12
CORE-447: JAWS firefox - Chapter dropdown in MH modal closes only on 2 Esc keystrokes

jomcarvajal:
[CORE-447]

[CORE-447]: https://openstax.atlassian.net/browse/CORE-447?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
I'm afraid this approach is trading one accessibility problem for more.
The card notes that the Color menu works, and a difference I see is that the Color menu has a fieldset (with an invisible legend) wrapping the controls, while the Chapter menu doesn't. That may be the fix.

RoyEJohnson (approved):

PR #2639, opened 2025-12-31
CORE-1094: Rex: Highlight box remains at the top of the page when scroll-highlighting text

jomcarvajal:
[CORE-1094]

[CORE-1094]: https://openstax.atlassian.net/browse/CORE-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/app/content/highlights/components/cardUtils.ts
```diff
@@ -148,6 +148,7 @@ export const getSelectionDirection = (selection: Selection): 'forward' | 'backwa
     selection.focusNode
   );
 
+  // tslint:disable-next-line:no-bitwise
```
RoyEJohnson: While you were working on this, tslint got removed from the stack. Eslint is the only linter. I added a commit to add the no-bitwise rule and changed the comment accordingly.

src/typings/global.d.ts
```diff
@@ -8,6 +8,7 @@ import PromiseCollector from '../helpers/PromiseCollector';
 
 declare global {
 
+  type DomSelection = Selection;
```
RoyEJohnson: Rather than adding types here, I've imported the types where they are used in another commit.

RoyEJohnson (changes requested):
I have rebased and added a couple of small commits (explained in my comments).
I'm finding that when I extend my selection forward, the instruction box moves down, but when I extend upward (by shift-clicking before the start of the selection), it does not move.
If I select by dragging so that the page auto-scrolls while dragging, the box winds up a little below the frame (since the selected text goes right to the bottom of the viewport).
Do you think you can force the box into the viewport when the selection changes?

RoyEJohnson:
This me in Chrome. It's apparently only when the screen is too narrow to position the box outside the text.

https://github.com/user-attachments/assets/935d028f-da7a-4ada-82b6-3dd3836b86fb

RoyEJohnson (changes requested):
Just one little thing.

RoyEJohnson (approved):

PR #2645, opened 2026-01-09
add PlatformWrapper to hide platform-specific content in assignable view

jivey:
https://openstax.atlassian.net/browse/CORE-1428

We already have this assignable-specific wrapper, so this feels like the right level of abstraction for now. Elements with data-platform-hidden="assignable" get hidden with css when rendered in the assignable iframe context, and this can easily be extended in the future to use a query param if needed.


RoyEJohnson (approved):
I can't see the other PR, but this one seems harmless at worst.

PR #2647, opened 2026-01-12
CORE-1402: Update Rice logo in footer

jomcarvajal:
[CORE-1402]

Before: 
Screenshot 2026-01-09 at 12 10 51 PM

After:
Screenshot 2026-01-09 at 12 10 41 PM

[CORE-1402]: https://openstax.atlassian.net/browse/CORE-1402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (approved):
I'm not the authority on this, but it seems like the logo is a little undersized. It has some built-in padding, so some of its height is invisible. It's currently at 4rem, and I think it would look better at 4.7rem (unless you have instructions on the sizing already).
Other than that, good to go.

PR #2658, opened 2026-01-16
Core-1500: make TOC heading an h2

RoyEJohnson:
[CORE-1500]

I had Claude replace the snapshots with specific tests.

[CORE-1500]: https://openstax.atlassian.net/browse/CORE-1500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (approved):
This seems okay, but maybe we should add a `.toBe('h2')` if the specs should concern themselves with asserting the heading level.

PR #2670, opened 2026-01-22
add tab trap to chapter and color filter dropdowns

jivey:
https://openstax.atlassian.net/browse/CORE-1498


src/app/reactUtils.ts
```diff
@@ -38,7 +38,8 @@ export function createTrapTab(...elements: HTMLElement[]) {
   .map(
     (container) => {
       const contents = Array.from(container.querySelectorAll<HTMLElement>(focusableItemQuery))
-        .filter((el) => !isHidden(el));
+        // keep focusable form elements that might be visually hidden
```
RoyEJohnson: isHidden checks for width and height of zero. Accessibility tools treat those as if they are not there (they require at least a pixel of height and width). Were you doing this to catch the hidden-but-accessible items, or for another reason?

RoyEJohnson (approved):

PR #2716, opened 2026-01-28
CORE-1514: Ensure all interactive functionality is operable with the keyboard

jomcarvajal:
[CORE-1514]

Translation should probably be confirmed by our contact at Katalyst

[CORE-1514]: https://openstax.atlassian.net/browse/CORE-1514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (changes requested):
Looks good but I think we should keep it in the hidden accessibility links too. And I'm not sure who is available to check the translations these days, you may want to ask in the engineering channel.

jivey (changes requested):

jivey (approved):

PR #2730, opened 2026-02-03
CORE-1548: Fix key term aria-label to include visible text at start

OpenStaxClaude:
## Summary

This PR fixes the key term search result aria-labels to comply with both WCAG 2.5.3 (Label in Name) and WCAG 2.4.4 (Link Purpose).

Fixes: [CORE-1548](https://openstax.atlassian.net/browse/CORE-1548)
Related: [CORE-799](https://openstax.atlassian.net/browse/CORE-799)

## The Problem

We had two conflicting WCAG requirements:
- **WCAG 2.4.4 (Level A)**: Links must have unique, meaningful names (addressed in CORE-799)
- **WCAG 2.5.3 (Level A)**: Accessible name must START with the visible label text (this issue)

**Current aria-label**: `"Result 1 in Key Terms: rotational analog of linear momentum..."`  
**Visible text**: `"Angular momentum..."`

The visible text wasn't at the start of the aria-label, preventing speech input users from saying "click Angular momentum" to activate the link.

## The Solution

Reorder the aria-label to put the key term FIRST, then add the disambiguating context:

**New aria-label format**: `"[Key Term]: [definition] (Result [N] in Key Terms)"`  
**Example**: `"Angular momentum: rotational analog of linear momentum... (Result 1 in Key Terms)"`

## Why This Works

✅ **WCAG 2.5.3 Compliance**: The visible text "Angular momentum" now appears at the START of the aria-label  
✅ **WCAG 2.4.4 Compliance**: The "(Result 1 in Key Terms)" suffix still provides unique context  
✅ **Better UX**: Screen reader users hear the key term first, then definition, then location context

## Changes Made

- Modified `uniqueSearchLabel` function in `SearchResultHits.tsx` to handle key terms specially
- Updated call site to pass `isKeyTerm` flag and `keyTermName` when available
- Regular search results maintain their existing format (unchanged)

## Testing

The test snapshots will need to be updated to reflect the new aria-label format for key term results. CI will handle running tests and we can update snapshots as needed.

**Manual Testing Steps**:
1. Search for "angular momentum" on a physics textbook
2. Inspect key term results in Chrome DevTools
3. Verify aria-label starts with the key term name
4. Test with screen reader to verify announcement is clear
5. Verify multiple results have unique aria-labels

[CORE-1548]: https://openstax.atlassian.net/browse/CORE-1548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
+++ b/src/app/content/search/components/SearchResultsSidebar/SearchResultHits.tsx
@@ -56,7 +56,18 @@ function useKeyTermPair({
   return pair;
 }
 
-function uniqueSearchLabel(index: number, title: string, highlight: string) {
+function uniqueSearchLabel(
+  index: number,
+  title: string,
+  highlight: string,
+  isKeyTerm: boolean,
+  keyTermName?: string
+) {
+  if (isKeyTerm && keyTermName) {
+    // For key terms: "[Term]: [definition] (Result N in Key Terms)"
+    return `${stripHtml(keyTermName)}: ${stripHtml(highlight)} (Result ${index + 1} in ${stripHtml(title)})`;
+  }
+  // For regular search results: keep existing format
   return `Result ${index + 1} in ${stripHtml(title)}: ${stripHtml(highlight)}`;
 }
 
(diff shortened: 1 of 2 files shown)


RoyEJohnson (changes requested):
I've pushed a little update.
1. It seems all extra text should be at the end of the aria-label, regardless of which type of result it is
2. The extra text is the same for either, so make a variable for it
3. the addition of pair?.term to the HighlightPreview callback required updating the dependency list

Make sure that is all properly reasoned.

RoyEJohnson (changes requested):
I've pushed a little update.

1. It seems all extra text should be at the end of the aria-label, regardless of which type of result it is
2. The extra text is the same for either, so make a variable for it
3. the addition of pair?.term to the HighlightPreview callback required updating the dependency list

Make sure that is all properly reasoned.

RoyEJohnson (approved):

jivey (approved):

PR #2760, opened 2026-02-05
CORE-1532: Add math equation section into keyboard shortcut modal

jomcarvajal:
[CORE-1532]

Screenshot 2026-02-23 at 12 18 25 PM

[CORE-1532]: https://openstax.atlassian.net/browse/CORE-1532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (changes requested):
We can remove "equations are now focusable", and I think Control + Enter doesn't work in Chrome, strangely enough. So we probably want to specify Safari, Firefox, Edge. We can also remove NVDA from "(Windows, NVDA)" and "(macOS, NVDA)" - those shortcuts are specific to mathjax not NVDA, and NVDA only runs on windows. Can you also send this to Erika and Victoria for design input? I wonder if they have ideas for grouping the alternative context menu shortcuts. And maybe they have some ideas for showing how to enable Speech in there - it's not a shortcut, but it's required to navigate successfully with NVDA, so it feels like we should include it in some way.

jivey (approved):

PR #2761, opened 2026-02-06
prevent activating highlight card when activating dot menu

jivey:
https://openstax.atlassian.net/browse/CORE-1584
This is a quick fix for activating the menu, it only partially addresses that ticket. The movement on activation would still be good to fix.


src/app/content/highlights/components/DisplayNote.tsx
```diff
@@ -96,18 +110,20 @@ const DisplayNote = React.forwardRef<HTMLElement, DisplayNoteProps>((
...
-        <DropdownList>
-          <DropdownItem message='i18n:highlighting:dropdown:edit' onClick={onEdit} />
-          <DropdownItem
-            message='i18n:highlighting:dropdown:delete'
-            data-testid='delete'
-            onClick={() => setConfirmingDelete(true)}
-          />
-        </DropdownList>
-      </Dropdown>
+      <span ref={dropdownRef}>
```
Dantemss: Can't we pass the ref to the Dropdown instead? I fear adding a span might cause other issues?

RoyEJohnson (commented):
I'm actually working on the DisplayCard/EditCard positioning problem right now as well.
https://github.com/openstax/rex-web/pull/2758#pullrequestreview-3759571455

Dantemss (approved):

PR #2762, opened 2026-02-09
add label to highlight note textarea

jivey:
https://openstax.atlassian.net/browse/CORE-1565


RoyEJohnson (changes requested):
See Copilot's remarks. There are two comments, but it's one fix: make the label associated only with the textarea. It currently encloses the buttons as well.

RoyEJohnson (approved):

PR #2763, opened 2026-02-09
media modals: remove tabindex from images and tab trap

jivey:
https://openstax.atlassian.net/browse/CORE-1590
https://openstax.atlassian.net/browse/CORE-1546


RoyEJohnson (approved):
I adjusted the failing test. I was a little surprised it came out -1 rather than null or something, but I'm guessing that's an implicit default for image?
In any case, the behavior is as expected.

RoyEJohnson (changes requested):
I have to concede that Copilot makes a good point about testing the tab-trapping.

src/app/reactUtils/focusUtils.ts
```diff
@@ -43,7 +43,7 @@ export const focusableItemQuery = ([
   'select',
   'textarea',
   '[href]',
-  '[tabindex]:not([tabindex=\'-1\']',
+  '[tabindex]:not([tabindex=\'-1\'])',
```
jivey: Pre-existing but copilot rightly pointed out the missing trailing paren

RoyEJohnson (approved):

PR #2780, opened 2026-02-18
fix structure of figure tag

jivey:
https://openstax.atlassian.net/browse/CORE-1591


RoyEJohnson (approved):
This does match what I described (figcaption goes inside figure tag).

RoyEJohnson (approved):

PR #2784, opened 2026-02-19
creating one time login to rex and necessary adjustments to tests

omehes:


jivey (changes requested):
Copilot flagged a couple of reasonable concerns here, especially about the test execution order. I left a couple of comments on those copilot reviews, though for some reason copilot placed them several lines above the actual lines it was referencing.

jivey (approved):

PR #2790, opened 2026-02-24
CORE-1536: Return focus to editCard after cancel

jomcarvajal:
[CORE-1536]

[CORE-1536]: https://openstax.atlassian.net/browse/CORE-1536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jivey (changes requested):
Fix looks good, but can we add an assertion to an existing test in the EditCard spec (assuming there is one for canceling, maybe add a new one if not?) I'd love for any future changes to catch the focus not going back to the textarea.

jivey (approved):

PR #2800, opened 2026-02-26
adds new test for unlogged highlights option

omehes:


Dantemss (approved):
The copilot review says a test description should be changed. The rest is probably fine.

PR #2806, opened 2026-03-03
CORE-1589: Add aria-label to ToC TreeItem components for screenreader accessibility

OpenStaxClaude:
## Summary

This PR addresses CORE-1589 by adding `aria-label` attributes to Table of Contents TreeItem components to improve screenreader accessibility.

## Changes

- Added `aria-label` with `, link` suffix to leaf TreeItem components (sections that are links)
- Added `aria-label` with `, section` suffix to expandable section TreeItem components (sections with subsections)
- Added unit tests to verify aria-label props are correctly set on TreeItem elements

## Problem

The Table of Contents uses React Aria's Tree component, which implements the ARIA treegrid pattern. In this pattern:
- TreeItem renders as `role="row"` (receives keyboard focus during arrow navigation)
- TreeItemContent renders as `role="gridcell"`
- The actual `<a>` link is nested inside the gridcell

When users navigate with arrow keys, focus moves to the row element, not the link. Screenreaders announce the row's text content but do NOT announce that it's a link. This violates WCAG 4.1.2 (Name, Role, Value) because screenreader users cannot determine that these controls navigate to new pages.

## Solution

By adding `aria-label` attributes to the TreeItem components that include descriptive text like "[section title], link" or "[section title], section", screenreader users will now hear:
- For clickable items: "[section title], link"
- For expandable sections: "[section title], section"

This works within the existing React Aria treegrid pattern without requiring library customization.

## Testing

- Added unit tests to verify aria-label is correctly set on TreeItem elements
- Manual testing with screenreaders will be performed to verify announcements

## Related Issues

- Jira: [CORE-1589](https://openstax.atlassian.net/browse/CORE-1589)
- Related: CORE-1566 (adds `aria-current="page"` for current page indication)

[CORE-1589]: https://openstax.atlassian.net/browse/CORE-1589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
In the new test, both "toBeDefined" checks are failing.

RoyEJohnson (changes requested):
In the new test, both "toBeDefined" checks are failing.

RoyEJohnson (changes requested):
The merge conflict was an unfortunate goof I didn't catch, but it was not the reason the `toBeDefined` tests were failing. They are still failing. The treeItems variables are coming up with zero items.

RoyEJohnson (changes requested):
I have examined the node.props in the problem tests, and none of them includes 'aria-label'. Is there something that might keep the aria-label parameter from being received by the TreeItem component?
Also, Copilot has recommendations about using i18n translation tables, which would be good. Can you create the entries in the various messages.json files and use them?

RoyEJohnson (changes requested):
Still puzzling over the failing tests. While we're here, it looks like TocSection could stand to be refactored. The map function computes active, which is only used for TocLeaf, while stripped Title is only used for StyledTreeItem. Break that up for some clarity.
Also take a look a the possibility that isArchiveTree is returning negative in the test?

RoyEJohnson (changes requested):
TreeItemProps is apparently a namespace rather than a type. That raised a lint error.
I tried defining (rather than importing) TreeItemProps as `Parameters<typeof TreeItem>` but the test still failed.
It looks like there is some test interference happening. I commented out the tests before the aria-label tests, and then they passed. Look for unrestored mocks or other potential test leaks.

OpenStaxClaude:
Found and fixed the test interference issue!

## Root Cause
The aria-label tests were failing due to **test leakage** from earlier tests in the same describe block. Specifically:
- The `'expands and scrolls to current chapter'` test creates spies on `domUtils` 
- The `'opens and closes'` test creates spies on `mediaUtils`
- These spies were **never restored** after the tests completed

The main `describe('TableOfContents')` block had a `beforeEach` but **no `afterEach`**, so mocks persisted across tests and interfered with subsequent tests.

## The Fix
Added an `afterEach(() => { jest.restoreAllMocks(); })` block right after the `beforeEach` in the main describe block.

This ensures:
1. All spies/mocks are cleaned up after each test
2. Tests run in isolation without interference
3. The aria-label tests can properly render and find components with aria-label props

## Why This Matters
As you discovered, when you commented out earlier tests, the aria-label tests passed. This confirmed the issue was test interference rather than a problem with the aria-label implementation itself.

The tests should now pass consistently! ✅

RoyEJohnson (changes requested):
The new tests work now (hoorah!) but the next test group (expandParentOfCurrentPage) fails.  This is the error for all three tests. It is likely depending on some test leakage that got patched.
```
    [React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.
```

RoyEJohnson (commented):
The same set of tests are now failing with this
```
    TypeError: Expected "book_uuid" to be a string

      71 |     const path = assertDefined(findPathForParams(parsedParams, contentPaths), 'Invalid parameters for content path');
      72 |
    > 73 |     return pathToRegexp.compile(path)(parsedParams);
```

RoyEJohnson (changes requested):
The same set of tests are now failing with this
```
    TypeError: Expected "book_uuid" to be a string

      71 |     const path = assertDefined(findPathForParams(parsedParams, contentPaths), 'Invalid parameters for content path');
      72 |
    > 73 |     return pathToRegexp.compile(path)(parsedParams);
```

RoyEJohnson (commented):
Address Copilot's latest review.
Also need a test to cover the case where `isTocOpen && isMobile` is true and addEventListener is called in TableOfContents/index.tsx.

RoyEJohnson (changes requested):
Address Copilot's latest review.
Also need a test to cover the case where isTocOpen && isMobile is true and addEventListener is called in TableOfContents/index.tsx.

RoyEJohnson (changes requested):
Address Copilot's new comments.

RoyEJohnson (changes requested):
Please address Copilot's latest comments.

RoyEJohnson (changes requested):
Please address Copilot's new comments.

RoyEJohnson (changes requested):
Copilot has some more feedback to address.

RoyEJohnson (approved):

jivey (approved):

PR #2809, opened 2026-03-04
adds tests for nursing books and additional small refinements

omehes:


jivey (approved):
Seems alright. The nth selectors are fragile and ideally would use data-testids. Also suggested removing the commented code.

PR #2810, opened 2026-03-05
Refactor TOCControlButton to reuse TOCControl component

OpenStaxClaude:
## Summary

This PR refactors `TOCControlButton` to eliminate code duplication by reusing the `TOCControl` component instead of reimplementing the same rendering logic.

## Changes

- Updated `InnerProps` interface to accept optional `aria-expanded` and `aria-controls` props
- Enhanced `TOCControl` component to forward ARIA attributes to the underlying `OpenButton`
- Created new `withMobileResponsiveTocControl` HOC that wraps `TOCControl` with:
  - Redux connection via `tocConnector`
  - Mobile detection logic using `useMatchMobileQuery`
  - Proper state determination (closed by default on mobile)
  - Additional ARIA attributes (`aria-expanded`, `aria-controls`)
- Replaced `TOCControlButton` implementation to use the new HOC
- Removed unused imports from `index.tsx` (`TocIcon`, `useMatchMobileQuery`, `tocConnector`, `MiddleProps`, `OpenButton`)

## Benefits

- **Less code duplication**: The TOC button rendering logic now exists in only one place
- **Consistent rendering**: Both TOC controls use the same underlying component
- **Easier maintenance**: Changes to the TOC button appearance only need to happen in one place
- **Better separation of concerns**: Mobile logic is encapsulated in the HOC

## Files Changed

- `src/app/content/components/SidebarControl/types.ts` - Added optional ARIA props to `InnerProps`
- `src/app/content/components/SidebarControl/TOCControl.tsx` - Enhanced component and added new HOC
- `src/app/content/components/SidebarControl/index.tsx` - Simplified to use the new HOC

## Testing

The refactoring preserves all existing behavior:
- ✅ TOC button renders with correct icon and text
- ✅ Mobile-responsive behavior (TOC closed by default on mobile)
- ✅ Proper ARIA attributes (`aria-label`, `aria-expanded`, `aria-controls`)
- ✅ `data-testid='toc-button'` preserved for testing
- ✅ Existing usage of `StyledOpenTOCControl` unaffected

## Related Issue

Closes [CORE-1713](https://openstax.atlassian.net/browse/CORE-1713)

This refactoring was identified during work on CORE-1537 and separated into its own PR to keep changes focused.

[CORE-1713]: https://openstax.atlassian.net/browse/CORE-1713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/content/components/SidebarControl/TOCControl.tsx
+++ b/src/app/content/components/SidebarControl/TOCControl.tsx
@@ -49,7 +49,7 @@ export function lockTocControlState(isOpen: boolean, Control: React.ComponentTyp
 
 export function withMobileResponsiveTocControl(Control: React.ComponentType<InnerProps>) {
   return tocConnector(({open, close, ...props}: MiddleProps) => {
-    const isMobile = useMatchMobileQuery();
+    const isMobile = typeof window !== 'undefined' && useMatchMobileQuery();
     const isOpen = props.isOpen === null ? !isMobile : props.isOpen;
 
     return <Control
--- a/src/app/content/components/SidebarControl/TOCControl.tsx
+++ b/src/app/content/components/SidebarControl/TOCControl.tsx
@@ -49,6 +49,12 @@ export function lockTocControlState(isOpen: boolean, Control: React.ComponentTyp
 
 export function withMobileResponsiveTocControl(Control: React.ComponentType<InnerProps>) {
   return tocConnector(({open, close, ...props}: MiddleProps) => {
+    // Note: This conditional hook call violates React's Rules of Hooks, but is necessary
+    // because useMatchMobileQuery() calls assertWindow() which throws during SSR.
+    // Since window never changes from defined to undefined during runtime, the hook
+    // order remains consistent after the initial render. This pattern existed in the
+    // original TOCControlButton implementation.
+    // eslint-disable-next-line react-hooks/rules-of-hooks
     const isMobile = typeof window !== 'undefined' && useMatchMobileQuery();
     const isOpen = props.isOpen === null ? !isMobile : props.isOpen;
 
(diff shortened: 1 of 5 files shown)


RoyEJohnson (changes requested):
Address Copilot's comments.
Also, for functions that are new in this PR, use `function` notation rather than arrow function const, unless there's a reason to do the latter.

RoyEJohnson (changes requested):
Address Copilot's comments.
Also, for functions that are new in this PR, use function notation rather than arrow function const, unless there's a reason to do the latter.

RoyEJohnson (approved):

Dantemss (approved):

RoyEJohnson (changes requested):
I moved the window guard into mediaQueryUtils in my latest push.
Copilot has some complaints about it, but the main problem is that the new test doesn't successfully restore window, so subsequent tests all fail.
The goal is to make useMatchMediaQuery handle the case where window is undefined, and to properly test that case. Figure out the right way to make that work.

RoyEJohnson (changes requested):
The new test is failing, as are some subsequent tests.
```
  ● useMatchMobileQuery › returns false if window is not defined

    The error below may be caused by using the wrong test environment, see https://jestjs.io/docs/en/configuration#testenvironment-string.
    Consider using the "jsdom" test environment.
    
    ReferenceError: window is not defined

      1039 |       delete (globalThis as any).window;
      1040 |
    > 1041 |       component = renderer.create(<Component />);
           |                            ^
      1042 |
      1043 |       expect(() => component.root.findByProps({ 'data-test-id': 'mobile-resolution' })).toThrow();
      1044 |       expect(() => component.root.findByProps({ 'data-test-id': 'desktop-resolution' })).not.toThrow();
```

RoyEJohnson (changes requested):
The test is clever, but the null branch of line 21 does not get tested because window's type is not actually undefined.

RoyEJohnson (approved):

PR #2826, opened 2026-03-25
adding test to compare 2 pdf files

omehes:


RoyEJohnson (approved):
If you've considered Copilot's suggestions, itlooks ok to me.

PR #2833, opened 2026-03-27
Phase 2.3: Form Elements Migration

OpenStaxClaude:
[CORE-1697]
## Summary

Migrates form element components from styled-components to plain CSS, following the established patterns from Phase 1.1-2.1.

### Components Migrated

1. **TextArea** (Note component) - Textarea for highlight notes
2. **SimpleLabel** (Note component) - Label for the textarea
3. **AnswerInput** (Practice Questions) - Visually hidden radio inputs for answer selection

### Changes

- Created `Note.css` with styles for TextArea and SimpleLabel
- Updated `Note.tsx` to use plain HTML elements with CSS variables
- Updated `Answer/styled.tsx` to re-export the standard `hiddenButAccessibleClass` utility as `answerInputClass`
- Updated `Answer/index.tsx` to use plain `<input>` element with className
- Removed styled-components from these form elements

Visible change: There was a dynamic bit of padding in the highlighter note textarea when the text was empty. It was there to make room for when the field label was rendered inside the textarea, which we don't do anymore. It caused the textarea size to shift slightly when going from empty to not or vice-versa. Now it doesn't.

**Note:** Originally created `Answer.css`, but during code review it was identified that the codebase already has a standard `.hidden-but-accessible` utility class. The Answer.css file was removed in favor of using the existing utility for better consistency and screen reader support.

### Migration Approach

These components are **internal** (not exported) and not used in component selectors, so we used a simple **in-place migration** without the full hybrid approach. No legacy exports needed.

### Testing Notes

This is a **refactoring-only change** with no visual differences. All styling, layout, and behavior should remain identical:

- Highlight notes textarea should look and behave the same
- Practice question radio inputs should remain visually hidden but accessible
- All interactive behaviors (auto-resize, ESC handling, keyboard navigation) should work identically

### Manual Testing

1. **Highlights Note Textarea**
   - Open a textbook in REX
   - Select text to create a highlight
   - Click in the note textarea
   - Verify: correct width, gray border, blue label, auto-resize, ESC key behavior, character limit

2. **Practice Questions Answer Inputs**
   - Open a textbook and click Practice
   - Start a practice session
   - Verify: inputs visually hidden, custom indicator shows selection, keyboard navigation works, screen reader accessible

### Automated Testing

CI will run all existing tests to ensure no regressions.

## Related

- Jira ticket: [CORE-1697](https://openstax.atlassian.net/browse/CORE-1697)
- Follows patterns from Phase 1.1-2.1 and learnings from Phase 2.2

[CORE-1697]: https://openstax.atlassian.net/browse/CORE-1697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
Address Copilot's comments.

RoyEJohnson (changes requested):
The build test is failing due to css loading order. I think you need to put ContentPane.css in the global loader as we did with some other css files.
```
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.

chunk 432 [mini-css-extract-plugin]
Conflicting order. Following module has been added:
 * css ./node_modules/css-loader/dist/cjs.js??ref--5-oneOf-4-1!./node_modules/postcss-loader/src??postcss!./src/app/content/components/ContentPane.css
despite it was not able to fulfill desired ordering with these modules:
 * css ./node_modules/css-loader/dist/cjs.js??ref--5-oneOf-4-1!./node_modules/postcss-loader/src??postcss!./src/app/content/highlights/components/Note.css
   - couldn't fulfill desired order of chunk group(s) route-assigned
   - while fulfilling desired order of chunk group(s) route-content

```

RoyEJohnson (changes requested):
Address Copilot's latest comments

RoyEJohnson (changes requested):
The decision to set `border:none` in an earlier phase has come back to bite: the button in EditCard to create a new highlight is not a PlainButton and so relied on the default button decoration.
Options I see are:
1. Make a different base button class with border: none and apply it to every button except this one. This seems like an extensive bit of complication.
2. Change the global styling to exclude a class, and add that class to this button. This seems well-targeted to me.
3. Make this button a PlainButton so we don't have any default buttons. This would be a visual change in the PR but might be a good idea for stylistic consistency.

I am inclined to #2, but could go #3. What do you think?

RoyEJohnson (changes requested):
Go with the `:not([class*="button"])` approach.

RoyEJohnson (changes requested):
We need a test that covers the else-branch of the `shouldDo` test in Note.tsx line 20

RoyEJohnson (changes requested):
Consider Copilot's comments. I think as far as the button classes go, they will re-introduce the specificity problem we were working around: we want to be able to specify borders on some plain-buttons.

RoyEJohnson (changes requested):
button-primary, when disabled, has a box shadow. It should not.

RoyEJohnson (changes requested):
Address Copilot's latest comments

RoyEJohnson (changes requested):
SkipAndSubmitButtons should have a margin-left of 0.1rem due to the QuestionNavigation Wrapper spec, but that is not being applied to them.

RoyEJohnson (changes requested):
Address Copilot's comments

RoyEJohnson (changes requested):
The AllOrNone buttons have a line-height of 25 where it should be `normal`.

jivey (approved):

RoyEJohnson (approved):

PR #2842, opened 2026-04-02
adds another docx comparison test - converts to pdf and compares

omehes:

### DIFF
--- a/e2e_tests/docx-tools/test_compare_unzipped_docxs.py
+++ b/e2e_tests/docx-tools/test_compare_unzipped_docxs.py
@@ -22,7 +22,7 @@ Latest update on March 30th, 2026
 """
 
 HOME_DIR = os.path.expanduser("~")
-BASE_FROM_DIR = f"{HOME_DIR}/Downloads/2pdfs"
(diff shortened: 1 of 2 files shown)


e2e_tests/docx-tools/test_compare_unzipped_docxs.py
```diff
@@ -22,7 +22,7 @@
 """
 
 HOME_DIR = os.path.expanduser("~")
-BASE_FROM_DIR = f"{HOME_DIR}/Downloads/2pdfs"
+BASE_FROM_DIR = f"{HOME_DIR}/Downloads/corgi_input_files"
```
jivey: Will this break the workflow here? https://github.com/openstax/rex-web/blob/952c342a2fad708457848738ad11981d7435fe66/e2e_tests/pdf-tools/test_compare_pdf_files.py#L86

jivey (commented):
This seems more of a local script to manually compare, not a test to be executed in the rest of the test suite. Should this live here, if it doesn't use pytest? Maybe add some test assertions if it should.

jivey (approved):

PR #2843, opened 2026-04-03
adds new test for he/bookstore page

omehes:


TomWoodward (approved):
this is ok to merge, but you should consider:

you're putting osweb tests in the rex repo. the rex repo only serves the reading pages, so they're a little out of place here. i understand that its probably easier to have just the one test suite for all the automated tests, but its still a little weird

this page is manged by the CMS, and all the assertions that you have are likely to become broken as the team optimizes the content of the page to communicate what it needs to communicate. these style of tests for CMS pages are going to be very brittle and need frequent updating

PR #2845, opened 2026-04-06
adds new osweb tests for kinetic and k12 pages

omehes:


TomWoodward (approved):
same as the other cms pr, this is ok to merge, but keep in mind:
- this is a weird home for cms tests
- be careful making automated tests for cms content, the purpose of the cms is to make the content easy to change, these tests will need to be updated if the content of these pages is updated, so you're adding work to a process that is supposed to be streamlined

omehes:
osweb tests were moved to os-webview repo. closing this branch.

PR #2915, opened 2026-04-09
adds test to compare release.json and abl

omehes:


TomWoodward (changes requested):
release.json is literally generated from the ABL, and it will drift when the ABL is updated until the next release, i don't think we need this test

omehes:
As per Tom's review, closing as not needed.

PR #2916, opened 2026-04-09
adds test to search through all unzipped docx files

omehes:


TomWoodward (changes requested):
i'm interpreting that this test is unzipping files in you local downloads folder and asserting that they have docx files in them, and i'm assuming that there are probably other tests that run before this that trigger the download?

i'm not sure this test is really providing  value, i would focus on testing things that are related to the logic/function of the CMS, not the content itself. these zip files are uploaded by the content management team. if the zip files were somehow dynamically generated it'd be worth it, but since they're manually added this seems like a very odd testing pattern for something that is really out of the scope of the app.

omehes:
@TomWoodward This is not a test file but rather a tool for me to be able to search through docx files for a string or anything in the content of the book.

PR #2837, opened 2026-03-30
Phase 2.4: Dropdown System Migration

OpenStaxClaude:
## Summary

Migrates dropdown components from styled-components to global CSS files, continuing Phase 2 of the styled-components migration initiative.

**Related Jira Ticket**: [CORE-1698](https://openstax.atlassian.net/browse/CORE-1698)

### New Files
- **src/app/components/Dropdown.css** - Global CSS file containing all dropdown styles

### Modified Files
- **src/app/components/Dropdown.tsx** - Refactored to use plain React with CSS classes
- **src/app/index.tsx** - Added Dropdown.css import

### Components Migrated
- DropdownToggle - Cursor styling
- DropdownMenu - Positioning, shadow, border styles  
- DropdownFocusWrapper - Overflow control
- DropdownList - Menu item styles and focus states
- DotMenu - Verified previous partial migration is complete

## Key Features Preserved

✅ **fadeIn animation** (100ms ease-out) for dropdown appearance  
✅ **Complex :focus-within selectors** for tab-transparent dropdown variant  
✅ **Component selector compatibility** via styled() wrappers  
✅ **Keyboard navigation** and tab trapping  
✅ **Controlled/uncontrolled state** handling  
✅ **Safari focus handling** (callOrRefocus function)

## Migration Patterns Followed

Following patterns from `PLAIN_CSS_MIGRATION_LEARNINGS.md`:

✅ Uses `classnames` package for className composition  
✅ Wrapped components with `styled()` for backward compatibility  
✅ Used `React.forwardRef` for components that receive refs  
✅ Filtered transient props (starting with $) before spreading to DOM  
✅ Set CSS variables before `...style` spread  
✅ Imported global CSS from `src/app/index.tsx`

## Backward Compatibility

Maintained full backward compatibility:

- **DropdownToggle**: Wrapped with `styled()` - used in component selectors
- **DropdownList**: Wrapped with `styled()` - used in component selectors in highlights and filters

All existing consumers will continue to work without modification.

### Files Using Dropdown Components
- src/app/developer/components/Books.tsx
- src/app/content/highlights/components/DisplayNote.tsx
- src/app/content/highlights/components/SummaryPopup/ContextMenu.tsx
- src/app/content/components/popUp/Filters.tsx
- src/app/content/practiceQuestions/components/Filters.tsx
- src/app/content/studyGuides/components/Filters.tsx

### Test Plan

Manual testers should verify:

1. **Visual verification**
   - Dropdown menus appear correctly when clicked
[...]

PR #2920, opened 2026-04-13
Fix sticky positioning regression for BookBanner and Topbar

OpenStaxClaude:
## Summary

Fixes a regression in sticky positioning behavior for BookBanner and Topbar on mobile, introduced in PR #2837 (Phase 2.4: Dropdown System Migration).

**Issue:** When scrolling down, the BookBanner and Topbar scroll out of view as expected, but when scrolling back up, they incorrectly snap back to their sticky position instead of scrolling smoothly.

**Root Cause:** The Dropdown component wrapper had `position: relative` applied via styled-components, which created a containing block that interfered with `position: sticky` behavior in ancestor elements.

**Solution:** Moved `position: relative` from the styled-components wrapper to the CSS file with appropriate class selectors (`.dropdown-hidden` and `.dropdown-transparent`). This maintains dropdown menu positioning functionality while allowing parent elements to maintain their sticky positioning behavior.

## Changes

- **src/app/components/Dropdown.tsx**
  - Removed `position: relative` from the styled-components wrapper
  - Added `dropdown-hidden` className to TabHiddenDropDown root element

- **src/app/components/Dropdown.css**
  - Added `position: relative` and `overflow: visible` to `.dropdown-hidden` and `.dropdown-transparent` selectors

## Testing

Manual testing should verify:
1. BookBanner and Topbar stick correctly on mobile when scrolling
2. Dropdown menus still position correctly relative to their triggers
3. All dropdown functionality (open/close, keyboard navigation) still works

## Related

- Fixes regression from: #2837
- Related Jira ticket: [CORE-1698](https://openstax.atlassian.net/browse/CORE-1698)

[CORE-1698]: https://openstax.atlassian.net/browse/CORE-1698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/app/components/Dropdown.css
+++ b/src/app/components/Dropdown.css
@@ -1,5 +1,12 @@
 /* Dropdown component styles */
 
+/* Base dropdown container styles */
+.dropdown-hidden,
+.dropdown-transparent {
+  position: relative; /* needed for absolute positioning of dropdown menu */
+  overflow: visible;
+}
+
 /* Keyframe animation for dropdown fade-in */
 @keyframes dropdown-fade-in {
   0% {
--- a/src/app/components/Dropdown.tsx
+++ b/src/app/components/Dropdown.tsx
@@ -102,7 +102,7 @@ const TabHiddenDropDown = React.forwardRef<HTMLElement, TabHiddenProps>((
     if (toggleElement.current) { toggleElement.current.focus(); }
   });
 
-  return <div className={className} ref={mergeRefs(ref, container)}>
+  return <div className={classNames('dropdown-hidden', className)} ref={mergeRefs(ref, container)}>
     <DropdownToggle
       ref={toggleElement}
       component={toggle}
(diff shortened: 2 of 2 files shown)


Dantemss (approved):

RoyEJohnson (changes requested):
This issue is proving difficult to chase down. The working version is about two months old, so there are a lot of possible changes to look at, and it does not work for me when I run it locally. But it works on staging.
One thing I have found is that BarWrapper takes an `up` parameter, and it does not seem to be getting set the same way in my local current install as it does in the old, working version. Can you analyze how it is supposed to work?
The SHA of the working version is c1a635f5ac23080f7a6f4d0c82e2320ce68b4fe6, if you want to compare.

RoyEJohnson:
This appears to have only happened on my local installation.

PR #3114, opened 2026-07-29
finance book slug redirects

bethshook:
https://openstax.atlassian.net/browse/CORE-2607

Generated using the retire-books script with some modifications. Claude checked for completeness.


bethshook:
@TomWoodward this won't cover the bare book URLs that were in the missing paths list (`books/principles-of-finance-2e` and `books/principles-of-finance-2e`). I'm not sure what to do about those.

TomWoodward:
> @TomWoodward this won't cover the bare book URLs that were in the missing paths list (`books/principles-of-finance-2e` and `books/principles-of-finance-2e`). I'm not sure what to do about those.

i don't know where we're creating those but they go to the cms details page. does the script to remove a book from rex not add a redirect there? maybe we added these redirects after the last time we actually had to remove a book and the remove book process doesn't actually work anymore

TomWoodward (approved):

PR #3123, opened 2026-08-04
CORE-2611 Changed prerendering to vCPU scaling, give each task at least 2 vCPU (1 per thread)

Dantemss:
https://openstax.atlassian.net/browse/CORE-2611

The last prerendering fix showed that we're not as memory constrained as I though. This new setup based on vCPUs instead of memory should give us more uniform prerender times. Also remove some max memory constraints that wouldn't be hit in practice with vCPU scaling.


script/prerender/fleet.ts
```diff
@@ -60,10 +60,10 @@ import { RELEASE_ID, WORK_REGION, BUCKET_NAME, BUCKET_REGION, PUBLIC_URL } from
 const MAX_CONCURRENT_BOOKS = 5;
 
 // Number of concurrent prerender tasks to run
-const DESIRED_TASK_COUNT = 32;
+const DESIRED_TASK_COUNT = 16;
 
-// Total fleet memory capacity in MiB (must equal DESIRED_TASK_COUNT * 4096)
-const TOTAL_MEMORY_MIB = DESIRED_TASK_COUNT * 4096;
+// Total fleet vCPU capacity (must equal DESIRED_TASK_COUNT * 2)
+const TOTAL_VCPU_COUNT = DESIRED_TASK_COUNT * 2;
```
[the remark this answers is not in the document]
Dantemss: Changed the constraint

script/prerender/cfn.yml
```diff
@@ -42,11 +42,11 @@ Parameters:
...
 
-  TotalMemoryMiB:
-    Description: Total fleet memory capacity in MiB (must equal desired task count * 4096)
+  TotalVCpuCount:
+    Description: Total fleet vCPU capacity (must equal desired task count * 2)
     Type: Number
-    MinValue: 4096
-    ConstraintDescription: Must equal desired task count * 4096
+    MinValue: 2
+    ConstraintDescription: Must equal desired task count * 2
```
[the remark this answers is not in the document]
Dantemss: It's fine...

Dantemss:
I tested and it worked but... when the instance got killed due to low spot capacity the tasks were not moved to the new instance. I have to figure out why.

Dantemss:
I believe we're now using the instances efficiently. I saw a cluster running all 16 tasks in my tests and the instances were properly packed with tasks.

bethshook:
I was tagged on this initially but maybe @TomWoodward wants to review

script/prerender/cfn.yml
```diff
@@ -109,7 +110,7 @@ Resources:
               awslogs-region: !Ref AWS::Region
               awslogs-stream-prefix: workers
               mode: non-blocking
-          MemoryReservation: 4096
+          MemoryReservation: 3686
```
Dantemss, line 113: Apparently there's some memory overhead on every instance so this has to be slightly lower than the value we aim for (4GB).

script/prerender/fleet.ts
Dantemss, line 66: 32 vCPU seems to make the prerendering finish in a timely manner, about 25 minutes when everything goes well.

I decided I wanted 16 tasks (containers) with 2 vCPU each, and each task spawns 2 threads so they can use both vCPU. This is mostly based on wanting things to run on at least 1 full Intel core. Maybe 3 or 4 threads could work better, but determining that would require some actual benchmarking.

script/prerender/cfn.yml
```diff
@@ -123,6 +124,13 @@ Resources:
       EnableECSManagedTags: true
       EnableExecuteCommand: true
       LaunchType: EC2
+      PlacementStrategies:
+        - Field: attribute:ecs.availability-zone
+          Type: spread
+        - Field: cpu
+          Type: binpack
+        - Field: memory
+          Type: binpack
```
Dantemss, line 133: In theory this is supposed to try to fill the existing instances as much as possible

TomWoodward (approved):
Image

PR #3124, opened 2026-08-07
Element/selector updates after deployments (staging, prod)

omehes:

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,9 +1,19 @@
 name: Playwright
 
 on:
-  push:
-    branches:
-    - main  # Trigger QA tests only when code is merged into main
+  pull_request:
+    paths:
+      - 'e2e_tests/**'
+  workflow_dispatch:
+    inputs:
+      base_url:
+        description: 'Base URL to test against'
+        required: false
+        default: 'https://staging.openstax.org'
+        type: choice
+        options:
+          - 'https://staging.openstax.org'
+          - 'https://openstax.org'
 
 defaults:
   run:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -37,4 +37,4 @@ jobs:
     - env:
         TEST_USER_EMAIL: ${{ secrets.TEST_USER_EMAIL }}
         TEST_USER_PASSWORD: ${{ secrets.TEST_USER_PASSWORD }}
-      run: pytest --tb=long --showlocals --base-url="${BASE_URL:-https://staging.openstax.org}" --rex_user $TEST_USER_EMAIL --rex_password $TEST_USER_PASSWORD -vvv e2e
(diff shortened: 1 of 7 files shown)


e2e_tests/e2e/ui/pages/home.py
```diff
@@ -249,11 +249,13 @@ async def click_highlights_option_green_colour(self):
 
     @property
     def highlights_option_text_colour_check_purple(self):
-        return self.page.locator('div[color="purple"]').get_attribute("color")
-
+        return self.page.locator(".highlight-content-wrapper").filter(
+    has_text="notepurple"
+)
```
jivey: There's some missing indentation on the changes in this file

.github/workflows/playwright.yml
```diff
@@ -1,9 +1,19 @@
 name: Playwright
 
 on:
-  push:
-    branches:
-    - main  # Trigger QA tests only when code is merged into main
+  pull_request:
+    paths:
+      - 'e2e_tests/**'
```
jivey, line 6: I'm not sure we want to change this, this means every push to a branch that touches these will trigger a run. However the workflow dispatch idea seems good - would that be sufficient for testing?

e2e_tests/e2e/ui/pages/home.py
```diff
@@ -249,11 +249,13 @@ async def click_highlights_option_green_colour(self):
...
     @property
     def highlights_option_text_colour_check_green(self):
-        return self.page.locator('div[color="green"]').get_attribute("color")
+        return self.page.locator(".highlight-content-wrapper").filter(
+    has_text="notegreen")
```
jivey, line 258: Probably not a blocker but just noting that this feels very brittle. The old markup had the much more stable `color` attribute - maybe we can add that back, like `data-highlight-color`, if you need to target specific colors in the tests.

jivey (changes requested):
The only real blockers are the indentation and the pull request path - worried about that last one because I thought we'd switched to main push only to prevent them from accidentally eating the action budget from multiple pushes.

.github/workflows/playwright.yml
omehes: Dante and Claude came up with this and we agreed that this is good. It only runs my playwright tests when the actual tests are being changed.

e2e_tests/e2e/ui/pages/home.py
omehes: yeah, it was good but it's gone now, so I had to change the selector

e2e_tests/e2e/ui/pages/home.py
omehes: Done. Thx

.github/workflows/playwright.yml
jivey: I'll defer to Dante on that, but yeah just make sure multiple pushes in a row don't trigger new runs without cancelling the old ones.

jivey (approved):

PR #3137, opened 2026-09-08
CORE-2737: Use the published ui-components CSS colour audit in REX

OpenStaxClaude:
Second half of converging the two CSS colour audits, on top of [CORE-2736](https://github.com/openstax/ui-components/pull/148).

Jira: https://openstax.atlassian.net/browse/CORE-2737

Based on `CORE-2731-global-css-theme-tokens` (#3133). Review that first; the diff here is only the deduplication.

### Stacking

This needs two parents, and they are independent of each other — a diamond rather than a chain:

- **#3133 (CORE-2731)** for code: this PR deletes the engine copy that one adds.
- **#3138 (CORE-2853, node 22)** for environment: the ui-components bump will not `yarn install` on node 14.

2731 and 3138 share only `package.json`, in unrelated parts of it, and each is green without the other. So the suggestion is to land **#3138 on `main` on its own**, then rebase 2731 onto `main` — it picks up node 22 for free — and leave this PR on 2731 where it already sits. That gets both parents with nothing re-targeted. Stacking 3138 *on* 2731 would instead hold an already-green EOL-runtime upgrade behind a CSS refactor still in review.

Node is now the **only** thing standing between this and a green CI — see "What this needs" at the bottom.

## What this does

`src/test/cssColors.ts` — the REX copy of the parsing engine — is **deleted**. `NAMED_COLORS`, `stripNoise`, `declarations`, `takesColor`, `describeColor`, `findColors`, `stylesheetColors`, `colorKey` and `opaqueKey` now come from `@openstax/ui-components/theme/cssColors`.

The base branch split the audit in two while this PR sat in review (#3142 took the engine to `main`, `src/test/themeColors.ts` kept the theme-aware half), which turned out to be exactly the boundary this needed. So the change is now a deletion plus four import lines:

| file | change |
|---|---|
| `src/test/cssColors.ts` | deleted — 515 lines, all engine |
| `src/test/cssColors.spec.ts` | deleted — its cases live with the engine now |
| `src/test/themeColors.ts` | imports the engine instead of `./cssColors` |
| `src/test/themeColors.spec.ts` | same, plus a note saying where the engine's cases went |
| `src/test/cssMediaQueries.ts` | takes `stripNoise` from the package |
| `src/app/theme.spec.ts` | takes `describeColor` and `stripNoise` from the package |

Nothing REX-specific moved: `themeColorIndex` over REX's nested theme, `stylesheetFiles` with its two exclusions, `KNOWN_OFF_PALETTE`, `occurrence`, `colorViolations`, `tokenChoices` and the breakpoint proximity check all stay where the base put them.

Keeping a second copy of the engine's assertions in REX would have drifted the way the second copy of the code did, so `cssColors.spec.ts` goes with `cssColors.ts`. `themeColors.spec.ts` carries a note saying so, otherwise the next reader sees a spec that covers two functions out of an audit and cannot tell whether the rest is tested elsewhere or not at all.

## The load-bearing check

`theme.baseline.json` regenerates **byte-identically** against the imported engine — same md5 (`4be40476…`), empty `git diff`. That is what shows the swap is behaviour-preserving rather than merely compiling.

It was not a foregone conclusion: the published engine had since added `device-cmyk()` to `COLOR_FUNCTIONS` and `scrollbar` to `COLOR_SHORTHANDS`, put a dotall flag on the `rgba()` grammar, and switched `colorKey`/`opaqueKey` from raw channels to hex. None of it moves a baseline line — REX's stylesheets contain none of those constructs, and the key format change is invisible because `themeColorIndex` and its lookups share the function.

Verified locally on node 22 with `--ignore-engines`, at tag `test-core-2736-3`:

- `test:unit` — **243/243 suites, 2013/2013 tests, 233/233 snapshots**, coverage 100% on all four metrics
- `lint:typescript` (`eslint src && tsc`) — exit 0
- `CI=true yarn build` — exit 0 at the **default** heap
- `generate:theme-baseline` — byte-identical output

## The packaging blocker is fixed

The `"module": "index.js"` regression written up in the earlier revision of this description **is resolved in `test-core-2736-3`**, which this branch now points at (thanks @RoyEJohnson for cutting it). The field reads `./dist/esm/index.js` again, so webpack 4 takes the ESM entry and tree-shakes.

Measured on this branch just now:

| build | chunk 519 |
|---|---|
| `main` today | 2.8 MB |
| tag `-2` (broken `module`, CJS fallback) | 6.21 MB — over workbox's 5 MB precache limit, fatal under `CI=true` |
| **tag `-3`, this branch** | **2.4 MB** (507.65 KB gzipped), no precache warning, builds at the default heap |

Smaller than `main`, which is the acceptance criterion "the production bundle does not grow" met rather than merely survived.

That also retires item 3 of the old description — the `captureReactException` interaction only mattered under the Sentry-7 fallback, which Roy ruled out in favour of the node upgrade.

## What this needs

1. **CORE-2853 / #3138.** `@sentry/react@10.73.0` declares `engines: {node: ">=18"}`, REX pins 14.15.4, and every CI job here still dies in `yarn install` before anything compiles — that is precisely the failure on the current run. Nothing in this PR can fix it.
2. The screenshot suite, once the node upgrade lands and CI can get past install.

[CORE-2736]: https://openstax.atlassian.net/browse/CORE-2736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
Try building using tag test-core-2736-1

OpenStaxClaude:
Tried it. **The deduplication is green against `test-core-2736-1`** — but the tag's `dist` is half-built, so I had to complete it locally to get there.

## What passes

`test-core-2736-1` is `733c905` (the head of #148, so the same engine source I tested against last time) plus one `Add dist` commit. Against it:

- **`theme.baseline.json` regenerates byte-identically** — same md5 `4be4047…`, empty `git diff`. That was the load-bearing criterion.
- `tsc --noEmit` clean, `eslint src` clean.
- Full unit suite **242/242 suites, 1942/1942 tests**, coverage **100/100/100/100**.

I've also **regenerated the 5 snapshots** in this push. I held them back last time only because the tag was undecided; it now pins the component source, so they're the migration's real output rather than churn. All of it is the styled-components→plain-CSS move — `ManageCookiesLink` renders an inline `<style>`, `ToastContainer` carries `className="toast-container"` instead of a generated `.c0`.

On the `ToastContainer` one I flagged last time: the toast **is** still styled. The declarations that were in the `.c0` block (`position: fixed`, `right: 2rem`, `display: grid`, centred, `gap: 1vh`) are now in the package's `ToastContainer.css`, which the built module imports itself. Still worth a screenshot-suite look, since jest mocks `*.css` and a snapshot can't show it.

## The one thing to fix on the tag

It ships `dist/esm` but **no `dist/cjs`, and no CSS in either half**:

| | `.js` | `.d.ts` | `.css` |
|---|---|---|---|
| `dist/esm` as tagged | 76 | 76 | **0** |
| `dist/cjs` as tagged | **0** | **0** | **0** |
| what `scripts/build.bash` produces | 76 + 76 | 76 + 76 | 23 + 23 |

`/dist` is gitignored, so `Add dist` is a force-add — this one caught only the `tsc` output of the esm pass, missing the cjs pass and the `rsync` that copies the CSS in. (`1.23.2` has both halves plus CSS, so it's an omission in the hand-made commit, not a change in the build.)

It bites twice. `main` is `./dist/cjs/index.js` and the `exports` `require` condition is `./dist/cjs/*.js`, so anything reaching the package through `require` — the baseline generator and jest — resolves to nothing:

```
Error: Cannot find module '.../@openstax/ui-components/dist/cjs/theme/cssColors.js'
    at createEsmNotFoundErr (node:internal/modules/cjs/loader:1554:15)
```

And `dist/esm/components/ToastContainer.js` does `import './ToastContainer.css'` against a file that isn't in the tarball, so webpack can't resolve that either.

`scripts/build.bash` already does the right thing — the tag just needs all of its output:

```bash
npm run dist        # builds esm + cjs, rsyncs css into both
git add -f dist     # not just dist/esm
```

I got the results above by building the missing half locally from the tag's own source, so **the `dist/cjs` `moduleNameMapper` line is verified correct rather than provisional** — nothing in this PR's config changes when a complete tag exists, only the tag name. I have push access and am happy to cut it myself if that's easier, but didn't want to make tags in your repo uninvited.

## Still blocked on node 14, separately

Unchanged from last round: the tag depends on `@sentry/react@^10.60.0` → `10.73.0`, whose `engines` is `{"node": ">=18"}`, against REX's `.nvmrc` of `v14.15.4`. CI dies in `yarn install` before anything compiles, which is why the checks on this push are red regardless of the above. Locally it needs `--ignore-engines`, and that flag isn't a real fix — REX prerenders in node, so Sentry 10's code would actually run under node 14, not just sit next to it.

So the deduplication is finished and verified; what's left are two things that aren't about CSS colours. Still think this is grounds for splitting the bump into its own ticket behind the node 18 upgrade — holding the dedup behind that leaves two copies of the engine in `main` for however long it takes. Your call, though; happy to park it if you'd rather land them together.

OpenStaxClaude:
Follow-up: **the production build passes**, and it answers the `ToastContainer` question I'd left open.

`yarn build` exits 0 against the tag (with the missing `dist` half completed locally, as above):

```
Compiled with warnings.

/static/js/519.ce1ad062.chunk.js is 6.2 MB, and won't be precached.
```

That's the only warning, and it's the pre-existing chunk-size one — bundle sizes came out within a few hundred bytes of the previous build, so the bump doesn't move them. Note REX's `Test (build)` job runs with `CI=true`, which escalates warnings to errors; that's orthogonal to this PR but will matter whenever the build job does get to run.

More usefully, **the toast really is still styled, confirmed at the bundle level rather than inferred.** The emitted CSS contains:

```css
.toast-container{z-index:var(--ox-z-index-toasts);z-index:var(--toast-container-z-index,var(--ox-z-index-toasts));display:grid;justify-items:center;justify-content:center;grid-gap:1vh;gap:1vh}
.toast-container:not(.toast-container-inline){position:fixed;right:2rem}
```

So webpack does pull `ToastContainer.css` in through `sideEffects: ["**/*.css"]` with no import on REX's side, exactly as the package intends — same declarations the generated `.c0` block used to carry. A screenshot run would still be the real visual check, but this is a good deal stronger than a snapshot could give.

Two caveats on how I got there, so the result isn't over-read:

- It needed `NODE_OPTIONS=--openssl-legacy-provider` and a raised heap here, because I'm on node 24 rather than REX's 14 — environment, not the tag.
- It builds because I completed the tag's `dist` locally. **With `test-core-2736-1` exactly as published it cannot build**, since `dist/esm/components/ToastContainer.js` imports a `./ToastContainer.css` that isn't in the tarball. That's the same `git add -f dist` gap as above — it costs the CSS as well as the whole cjs half.

RoyEJohnson (changes requested):
This already stacked on CORE-2731. It also needs to be stacked on https://github.com/openstax/rex-web/pull/3138. I guess that means stack 3138 on 2731. Does that seem right to you?

OpenStaxClaude:
Half right, I think — the dependency is real, but I'd not put 3138 on top of 2731.

**The shape isn't a chain, it's a diamond.** 3137 genuinely needs both:

- **2731** for code — 3137 deletes the engine copy that 2731 adds, so it can't exist without it.
- **3138** for environment — the ui-components bump won't `yarn install` on node 14.

But 2731 and 3138 need nothing from each other. I checked: the only file both touch is `package.json`, and in unrelated parts of it (3138 does `.nvmrc`/`engines`/`@types/node`; 2731 does the `generate:theme-*` scripts). 2731 is green on node 14, and 3138 is green without any of the CSS work.

**So stacking 3138 on 2731 costs something for nothing.** It would mean an EOL-runtime upgrade that's already green and independent can't merge until a CSS refactor still in review iteration lands first — and 3138 eats a rebase every time 2731 moves. If we do have to linearise, I'd invert it (`main → 3138 → 2731 → 3137`) so the thing with no dependencies sits lowest and can merge first.

**What I'd actually suggest: don't re-parent anything.** 3138 is independent and green — land it on `main` on its own. Then rebase 2731 onto `main`, where it picks up node 22 for free, and 3137 stays exactly where it is, on 2731. That gets 3137 both parents with zero re-targeting, and each rebase is downhill.

The one thing to flag either way: **stacking won't turn 3137 green.** It's necessary but not sufficient — 3137 also needs the ui-components `module` fix from item 2 in the description. I just re-checked and `"module": "index.js"` is still there both on ui-components `main` and on `test-core-2736-2`, with no PR open for it. Until that's fixed webpack 4 falls back to `main`, bundles the CommonJS build, and chunk 519 comes out at 5.4 MB — over workbox's 5 MB precache limit, which is a hard failure under `CI=true`. So if the goal is to see this PR green, the packaging fix is the binding constraint rather than the stacking.

Happy to do whichever you prefer — say the word and I'll rebase, including your original ordering if you'd rather keep it as one stack for reviewing. I just didn't want to rewrite 3133's or 3138's branches under you while they're both mid-review.

OpenStaxClaude:
Heads-up: the base branch moved under this one. #3133 has been split in two at Roy's request, so `src/test/cssColors.ts` is now the parser alone (in #3142, against `main`) and everything theme-aware — `themeColorIndex`, `KNOWN_OFF_PALETTE`, `stylesheetFiles`, `occurrence`, `colorViolations` — lives in `src/test/themeColors.ts` on this branch's base.

That is the boundary this PR needed anyway: swapping in the published engine is now "delete `src/test/cssColors.ts` and re-point `themeColors.ts` and `cssColors.spec.ts` at `@openstax/ui-components`", with no theme-aware code to disentangle first. Expect a conflict on the next base merge; the resolution is to keep the base's split and drop this branch's edits to the REX-specific half.

Still a draft and still blocked by CORE-2736 either way.

RoyEJohnson (changes requested):
Resolve conflicts

OpenStaxClaude:
Conflicts resolved in 497aa28e — `mergeable` is now `MERGEABLE`.

**How it was resolved.** The base branch split the audit in two after this PR was cut (#3142 took the engine to `main`; `src/test/themeColors.ts` kept the theme-aware half), so the three conflicted files were three copies of the same disagreement: this branch had edited `cssColors.ts` as one file, the base had already cut it in two. I kept the base's split and dropped this branch's edits to the REX-specific half, which is what the swap wanted anyway. What is left of this PR is now a deletion plus four import lines:

- `src/test/cssColors.ts` deleted — 515 lines, all engine
- `src/test/cssColors.spec.ts` deleted — its cases live with the engine in ui-components
- `themeColors.ts`, `themeColors.spec.ts`, `cssMediaQueries.ts` and `src/app/theme.spec.ts` re-pointed at `@openstax/ui-components/theme/cssColors`

Your breakpoint audit comes through untouched: `cssMediaQueries` only ever needed `stripNoise`, which the package exports, and it keeps its own copy of the `<number>` grammar for the reason its comment gives — that grammar is an internal of the engine, not part of its published surface.

**Two things also landed in your favour since the last review.**

`test-core-2736-3` — the tag you pushed in b882ad92 — carries the `"module": "./dist/esm/index.js"` fix. That was the hard blocker in the old description, so I measured it on this branch rather than take it on faith. `CI=true yarn build` now completes at the **default** heap with chunk 519 at **2.4 MB** (507.65 KB gzipped) and no workbox precache warning — against 6.21 MB on tag `-2`, which was fatal under `CI=true`, and 2.8 MB on `main` today. The bundle is smaller than `main`, so the "does not grow" criterion is met rather than merely survived. I have rewritten the description accordingly and dropped the `captureReactException` section, which only mattered under the Sentry-7 fallback you ruled out.

Everything else is verified on node 22 at that tag: **243/243 suites, 2013/2013 tests, 233/233 snapshots**, coverage 100% on all four metrics; `eslint src && tsc` exit 0; and `theme.baseline.json` still regenerates byte-identically (same md5, empty `git diff`), which is the check that shows the engine swap changed no behaviour.

**What is still red, and why this stays a draft.** The CI run on this push fails in `Setup`, all four jobs, with:

```
error @sentry/react@10.73.0: The engine "node" is incompatible with this module.
Expected version ">=18". Got "14.15.4"
```

That is CORE-2853 / #3138, and nothing in this PR can move it — `yarn install` never gets far enough to compile a line. #3138 is still open against `main`. Land it there and this goes green on the next base merge; I will take it out of draft and run the screenshot suite then.

PR #3138, opened 2026-09-09
CORE-2853: upgrade node 14 -> 22

OpenStaxClaude:
Jira: [CORE-2853](https://openstax.atlassian.net/browse/CORE-2853)

Upgrades REX from node 14.15.4 (EOL April 2023) to node 22 LTS. Roy asked for
the `--openssl-legacy-provider` route and node 22 on the ticket.

**All CI checks are green.** Details and the two non-obvious bits are below.

### Changes

- `.nvmrc`: `v14.15.4` → `v22.23.2`. This stays the single source of truth —
  `ci.yml` reads it via `node-version-file` in every job and the Dockerfile
  `COPY`s it and runs `nvm install`, so both follow automatically.
- `package.json` `engines.node`: `14.x` → `22.x`. This is also what Heroku
  review apps read to pick their runtime. The old value made `yarn` refuse
  outright on a newer node, including inside nested `yarn` invocations —
  jest-puppeteer's `server.command` shells out to `yarn server`/`yarn start`,
  so the browser suites failed to start before a single test ran.
- `@types/node`: `^14.0.0` → `^22.0.0`, so `tsc` checks against the standard
  library we actually run. Two consequent type fixes: `useTimeout` and
  `useDebouncedWindowSize` held their timer ids in `React.useRef<number>`,
  which no longer accepts what `setTimeout` returns; they now use
  `ReturnType<typeof setTimeout>`, correct under either lib. Those were the
  only two `tsc` errors the bump produced.
- `NODE_OPTIONS=--openssl-legacy-provider` in `build:js` and
  `script/start.bash`. react-scripts 4.0.3 brings webpack 4.44.2, which calls
  `crypto.createHash('md4')`; node 17+ ships OpenSSL 3, where md4 is not in the
  default provider, so the build dies with `ERR_OSSL_EVP_UNSUPPORTED`. Set in
  the two places that run webpack rather than exported globally, so it is
  obvious what it is for. Both appends preserve any inherited `NODE_OPTIONS`.
- `README.md`: a short section naming the flag and why, so it is removable
  rather than folklore. Also drops the stale "14.x for rex-web" line from
  `PLAIN_CSS_MIGRATION_LEARNINGS.md` so no node version is written down twice.
- `ci:test:unit` no longer fails when the Codecov upload fails — see below.

### The one judgement call: the codecov script

The first CI run on this branch failed `Test (unit)` — but all 240 suites and
1920 tests passed. The failure was the upload afterwards. The `codecov` npm
package (3.8.3, deprecated by Codecov in 2021) throws where it reads the
upload response:

```
TypeError: result.split is not a function
    at node_modules/codecov/lib/codecov.js:212:25
    at node_modules/teeny-request/build/src/index.js:217:17
```

Not caused by this upgrade. The identical error, same file, same line, is in
main's most recent run —
[33783899064](https://github.com/openstax/rex-web/actions/runs/33783899064),
2dd8fae2, on node 14. What differs is the consequence: node 14 reported it as
`UnhandledPromiseRejectionWarning` and exited 0, so the job went green with a
warning nobody reads. Node 15 made unhandled rejections fatal by default, so on
22 the same latent bug takes the whole job down.

Coverage has therefore not been reaching Codecov for some time, and the
README's coverage badge is stale. Rather than let a dead third-party uploader
gate the test suite, the upload is now non-fatal and prints a line saying it
failed. That restores the pre-upgrade outcome without pretending it works.
[CORE-2855](https://openstax.atlassian.net/browse/CORE-2855) tracks fixing it
properly — it needs a `CODECOV_TOKEN` decision, or a decision to drop Codecov,
since `test:unit` already enforces 100% coverage via jest thresholds.

Happy to take the other call if you'd rather the upload stay fatal — that means
CORE-2855 lands first and this waits on it.

### Verification

CI, all green:

| Job | |
| --- | --- |
| `Setup` | ✅ 6m40s — includes the Heroku review app deploying |
| `Test (unit)` | ✅ 240/240 suites, 1920/1920 tests |
| `Test (build)` | ✅ `build:clean` → `test:sourcemap` → `prerender:local` → `prerenderspec` → `browserspec` |
| `Lint` | ✅ eslint + tsc + stylelint + shellcheck |
| `Scan` | ✅ but vacuous here — `modified_books` is empty on this PR, so it only resolved the URL |
| `ABL Schema Tests` | ✅ |
| `Jira` | ✅ |

Locally on node 22.23.2: `yarn install` with no `--ignore-engines`, `yarn lint`,
`yarn test:unit` (240/240, 1920/1920, 100% on all four coverage metrics),
`yarn build`, and `REACT_APP_ENV=test build:clean` + `prerender:local`
(12 pages) all clean.

Review app: [deployed](https://rex-web-core-2853-node--tjlicx.herokuapp.com/)
and serving. Loaded a book page in a real browser — renders, heading
"1.1 Physics: An Introduction", ~25k characters of content, no application JS
errors (only a third-party CookieYes site-registration warning). Note the AC
asked for "a prerendered page loads from it", which a review app can't do:
`heroku-postbuild` runs `build:clean` only, and the Procfile serves with
`--fallback404`, so review apps serve the SPA shell and never prerendered HTML.
The prerender path is covered by `Test (build)` instead.

**Screenshot suite — the AC's real question is answered, but not the way it was
framed.** The suite cannot pass, on any node: `toMatchImageSnapshot` is
registered in `src/test/setup.ts`, which belongs to the *unit* jest config,
while the screenshot specs run under `--config jest-puppeteer.config.json`
whose only setup file never calls `expect.extend`. So all 39 tests fail with
`toMatchImageSnapshot is not a function` — and it isn't in `ci.yml`, which is
why the rot went unnoticed.
[CORE-2857](https://openstax.atlassian.net/browse/CORE-2857) has the details.

That doesn't leave rendering unverified, though. Running in the
`openstax/rex-web:CI-2023-08-10` container against a `REACT_APP_ENV=test`
build, with a throwaway two-line `expect.extend` to bring the suite to life,
the real image comparisons run — and the results are **identical on both
runtimes**:

| | node 14.15.4 (main) | node 22.23.2 (this branch) |
| --- | --- | --- |
| passed | 17 | 17 |
| failed | 22 | 22 |

`comm` on the two sorted failure lists returns nothing in either direction —
same 21 test names, same everything. The 22 failures are all the desktop and
print variants and are pre-existing drift. So node 22 demonstrably does not
change rendering, which is what that criterion was actually asking.

### Deliberately not in this PR

- The dependency pins that exist only because of node 14 — `@aws-sdk/*
  <3.567.0`, `glob <9`, `lighthouse ^9`, `stylelint <14`, and
  `@openstax/ui-components #1.21.1` (the immediate driver, via
  `@sentry/react@^10.60.0` needing node >= 18). Lifting them here would make
  the node change unreviewable and confuse a bisect.
- react-scripts 5 / webpack 5, which retires the flag:
  [CORE-2854](https://openstax.atlassian.net/browse/CORE-2854).
- The CI image rebuild the ticket asked for. The Dockerfile cannot be built at
  all right now, for a reason unrelated to node: its base is `debian:bullseye`,
  which is EOL, and the `bullseye-security` Release file expired around 7 Sep,
  so the very first `apt-get update` exits 100. Date-triggered and present on
  main. [CORE-2856](https://openstax.atlassian.net/browse/CORE-2856), which
  also matters for the `release` target the prerender fleet runs from. It
  doesn't block this PR: every job runs `setup-node` with `node-version-file`,
  installing node 22 over whatever the image ships, and CI is green that way. I
  checked the node-22 part of the image separately — nvm 0.35.3, the version
  the Dockerfile installs, resolves and installs `v22.23.2` from `.nvmrc`
  fine, and the currently published image is buster-based with glibc 2.28,
  exactly node 22's floor, which is why node 22 runs on it at all.

### Still needs someone with the environment

The concourse prerender fleet, which is where the `@aws-sdk` pins live. It runs
node from a Dockerfile-built image, so CORE-2856 is a prerequisite for
exercising it.

[CORE-2853]: https://openstax.atlassian.net/browse/CORE-2853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CORE-2855]: https://openstax.atlassian.net/browse/CORE-2855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
Address Copilot's comment

RoyEJohnson (changes requested):
See previous comments.

Dantemss:
@RoyEJohnson I think for this we should do a prerendering test run. Grab a release in https://github.com/openstax/unified/issues (has to be one of the newest 4 and one that we don't actually plan to use), set it to the last commit SHA in this PR and then lock it.

Dantemss:
If it fails, updates to unified-deployment may be needed

PR #3139, opened 2026-09-09
CORE-2855: drop codecov, code climate, greenkeeper and pyup

OpenStaxClaude:
Jira: [CORE-2855](https://openstax.atlassian.net/browse/CORE-2855)

Started as "drop Codecov rather than fix the upload"; grew, on review, into retiring every dead code-quality/dependency service still wired into this repo. Four of them, one commit each.

## Why deletion instead of a fix

The `codecov` npm package pinned at 3.8.3 was deprecated by Codecov in 2021, and its upload has been throwing `TypeError: result.split is not a function` for some time — nothing has landed on Codecov since. Node 14 surfaced that as an `UnhandledPromiseRejectionWarning` and exited 0, so the `Test (unit)` job stayed green and the failure was invisible. From node 15 on, unhandled rejections are fatal, which is how #3138 found it.

Fixing it properly means swapping in `codecov/codecov-action`, and v4+ wants a `CODECOV_TOKEN` that this repo does not have. That is not worth it, because `test:unit` already runs `jest --coverage` against a `coverageThreshold` of 100% on branches, functions, lines and statements — locally and in CI. A dip in coverage already fails the build; Codecov was reporting on a gate that is enforced harder upstream of it.

Pulling that thread found three more services in the same state — a config file and/or a badge, and nothing in CI feeding them.

## Changes

**Codecov** (80f6b59c)

- remove the `codecov` devDependency (and its now-orphaned transitive deps from `yarn.lock`)
- remove the `codecov` npm script
- delete `codecov.yml`
- `ci:test:unit` is now `CI=true npm run-script test:unit` — no trailing upload step, nothing swallowed, a genuine failure fails CI

**Code Climate** (c3c25f08)

- delete `.codeclimate.yml`
- drop the `Maintainability` and `Test Coverage` badges from the README

There is no `cc-test-reporter` upload step anywhere in CI (`.github/workflows/`, `concourse/`, `jenkins/`, `Makefile`), so the `Test Coverage` badge has been rendering a number nothing produces, and the config only ever excluded `*spec.ts(x)` from an analysis nobody reads.

**Greenkeeper and pyup** (7dccbe7b)

- delete `greenkeeper.json`
- drop the `Updates` (pyup) and `Greenkeeper` badges from the README

Greenkeeper shut down in 2020, and its successor was never set up here. pyup has no config file and no CI step — just a badge. Dependency bumps actually arrive via Dependabot (#3134 for the `e2e_tests` python side, #3135 for JS), which covers what both of these were nominally for.

That was the last badge in the README, so it now opens on the title.

## Verification

`grep -ri` for `codecov`, `codeclimate`, `greenkeeper` and `pyup` across the repo returns nothing outside coincidental base64 substrings in `yarn.lock` integrity hashes. No build, CI or lint step referenced any of the deleted files, so every removal here is inert — the only behavioural change in the whole PR is that `ci:test:unit` no longer ends in a broken upload.

## Notes

- **Repo settings, not code.** If the **Code Climate GitHub app** is still installed on `openstax/rex-web` it will keep analysing pushes and may keep posting a status regardless of the deleted config. Same for any lingering Codecov/pyup app authorisation. Uninstalling those is an org/repo settings action that a commit cannot do — flagging it rather than leaving it implied.
- **Overlaps #3138.** That PR changes the same `ci:test:unit` line to make the upload non-fatal as a stopgap. Whichever merges second needs a trivial conflict resolution; if this one lands first, #3138 should drop its stopgap hunk, since there is no longer an upload to guard. Nothing else in this PR touches anything #3138 changes.

[CORE-2855]: https://openstax.atlassian.net/browse/CORE-2855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RoyEJohnson (changes requested):
@OpenStaxClaude Go ahead and remove the code climate baggage in this PR, as well.

OpenStaxClaude:
Done in c3c25f08 — `.codeclimate.yml` deleted and both Code Climate badges (`Maintainability`, `Test Coverage`) pulled from the README.

It was in the same shape as codecov: no `cc-test-reporter` upload step anywhere in CI, so the `Test Coverage` badge has been rendering a number nothing generates, and the config only ever excluded `*spec.ts(x)` from an analysis nobody reads. Nothing in the build referenced either file, so the removal is inert.

Two things a commit here cannot reach, flagging rather than guessing:

- If the **Code Climate GitHub app** is still installed on `openstax/rex-web`, it will keep analyzing on push and posting status — that needs uninstalling from the repo/org settings.
- I left the `pyup` and `Greenkeeper` badges alone. Both look just as dead (Greenkeeper shut down in 2020), but they are neither codecov nor Code Climate, so I did not widen the ticket. Happy to sweep them here if you would rather have one cleanup than two — say the word.

PR description updated to match. Marking ready for review once checks report.

RoyEJohnson (changes requested):
Do the pyup and Greenkeeper baggage as well.

OpenStaxClaude:
Done in 7dccbe7b — `greenkeeper.json` deleted, and the `Updates` (pyup) and `Greenkeeper` badges pulled from the README. That was the last badge in the file, so it now opens on the title.

Both were dead in the same way as the other two: Greenkeeper shut down in 2020 and its successor was never set up here; pyup has no config file and no CI step, just a badge. What actually bumps dependencies on this repo is Dependabot — #3134 for the `e2e_tests` python side, #3135 for JS — which covers what both of these nominally did.

So the branch is now four commits, one service each:

| commit | service | removed |
| --- | --- | --- |
| 80f6b59c | Codecov | `codecov.yml`, devDependency, npm script, `yarn.lock` entries |
| c3c25f08 | Code Climate | `.codeclimate.yml`, 2 badges |
| 7dccbe7b | Greenkeeper + pyup | `greenkeeper.json`, 2 badges |

`grep -ri` for all four names across the repo now returns nothing outside coincidental base64 substrings in `yarn.lock` integrity hashes, and no build, CI or lint step referenced any of the deleted files. The only behavioural change in the whole PR remains `ci:test:unit` no longer ending in a broken upload.

One thing still outside what a commit can reach: if the **Code Climate GitHub app** (or a lingering Codecov/pyup authorization) is still installed on the repo, it will keep running on push regardless of the deleted configs — that needs uninstalling from repo/org settings.

Title and description updated to match the wider scope. Watching checks now; will take it out of draft when they are green.

jivey (approved):
Seems reasonable to me

RoyEJohnson (approved):

